PR #4, opened 2018-07-24
Pull image and accept registry out

### LINKED ISSUE #2: Enhancement: Option to pull image if it does not exist on Docker host
Currently, the first example will fail if your Docker host does not have the `nginx` image pulled. Maybe a way to indicate the image should be pulled if it doesn't exist in the host could be useful?

gianarb:
Other than pulling an image to be sure that it exists this PR introduced
a mechanism to download images from a private repository.

Fixed #2 

@sirlatrom what do you think?

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -16,6 +16,7 @@ type RequestContainer struct {
 	Env          map[string]string
 	ExportedPort []string
 	Cmd          string
+	RegistryCred string
 }
 
 // Container is the struct used to represent a single container.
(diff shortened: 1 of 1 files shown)


sirlatrom:
@gianarb Sorry for not getting back to you. This looks like a fine way to do it. Perhaps looking into https://github.com/docker/cli/blob/6686ada6a4cf4e872af4d9bf3f61c387be934d87/cli/config/credentials/credentials.go#L12 could be an idea if it's a valid use case to use the client's config/auth store from their Docker CLI (if available).

gianarb:
Way better @sirlatrom do you have a chance to work on it?

PR #29, opened 2018-11-23
Implemented a create-only function

### LINKED ISSUE #20: Allow manual start of container
The java implementation actually requires users of the library to start the container on their own or use a test framework integration.

The go lib should probably also provide a way to start the container on request rather than on creation of the testcontainer setup.

gianarb:
As requested here #20 this PR implements a flow to only create
containers. The container has a new `Start` function to run it when you
wish.

@mraerino

```go
c, err := CreateContainer(ctx, "nginx", input)

....
....

err := c.Run(ctx)

.....
.....

```

Fixed #20

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -26,7 +26,8 @@ type RequestContainer struct {
 // Container is the struct used to represent a single container.
 type Container struct {
 	// Container ID from Docker
-	ID string
+	ID      string
+	started bool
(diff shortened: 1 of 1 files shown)


mraerino:
`Run` should actually also implement the waiting strategies. So you might want to move that to `Run` from `RunContainer`.

gianarb:
Reverted I will open a new one soon.

PR #30, opened 2018-11-23
Refactor docker handling to create client once for a context

mraerino:
This is a suggestion for improved handling of the docker client.
It seems reasonable to reuse the client.


gianarb:
Thanks

PR #32, opened 2018-11-23
Implemented function to create-only container

gianarb:
We need the ability to create a container without running it.
This PR introduced this feature with the method `CreateContainer`.
It works like the `RunContainer` but without executing the container
cmd.

You can run it as you wish with the function `container.Run(ctx)`.

The `Run(ctx)` function doesn't need a wait strategy. If you need one
you need to set it up by yourself programmatically.

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -29,6 +29,8 @@ type Container struct {
 	ID string
 	// Cache to retrieve container infromation without re-fetching them from dockerd
 	raw *types.ContainerJSON
+
+	cli *client.Client
 }
 
 // LivenessCheckPorts returns the exposed ports for the container.
@@ -56,11 +58,7 @@ func inspectContainer(ctx context.Context, c *Container) (*types.ContainerJSON, 
 	if c.raw != nil {
 		return c.raw, nil
 	}
-	cli, err := client.NewEnvClient()
-	if err != nil {
-		return nil, err
-	}
-	inspect, err := cli.ContainerInspect(ctx, c.ID)
+	inspect, err := c.cli.ContainerInspect(ctx, c.ID)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


gianarb:
@mraerino  this represents what I have in my mind I think. I don't think we should support the wait strategy in the `Run` function. If you need one you can call it after the `Run` func.

mraerino:
Are you sure there should not be sugar to have a blocking Run call if a waitingStrategy is supplied when creating the container?
In Java this works like this: https://www.testcontainers.org/usage/options.html#startup-timeout

gianarb:
I can see the flavor @mraerino now that you showed me that but I am not sure how we can reach that API in Go.

Maybe we should wrap the ExportPort around a more complicated object, not like today that it is just a string `80/tcp` but something more like:

```go
ExportPort{
    ID: `80/tcp",
    WaitFor: []WaitStrategy // This will be associated on the port `80/tcp`
}
```
I think this is related to #33 and @MeneDev can give us an idea.

mraerino:
Another way would be to set the port on the `WaitStrategy` struct.
This would allow having wait strategies that know which ports to connect on (e.g. postgres)

gianarb:
Yep, I probably like more the idea to have a reference of the port in the wait strategy as suggested #33 and as yo usaid @mraerino

mraerino:
The http strategy would then look like this:

```golang
type httpWaitStrategy struct {
	// all WaitStrategies should have a startupTimeout to avoid waiting infinitely
	startupTimeout time.Duration

	// httpWaitStrategy has additional properties
	port		  uint16
	path              string
	statusCodeMatcher func(status int) bool
	useTLS            bool
}
```

gianarb:
Yes I think more than how it is implemented internally we should spent some time around how to make the interface easy and general enough.

PR #34, opened 2018-11-23
BIG CLEANUP AND REFACTORING

gianarb:
Hello everyone.
I started this project few months ago but right now using it I
discovered that something needs to change.

So right now I redesigned the function, deprecating the old one.
The new core functions have the same name used from the java
implementation:

* `GetContainerIpAddress` returns the ip of the docker gateway. I decided to go
with this implementation because it should be the most "reusable"
approach right now. We will need to work and port a lot of the logics
the java project learn during all these years.
* `GetMappedPort` returns the port opened in the host.

Combining this two functions together you can reach a particular port in
your container. These two functions should be used as part of the Wait
interface.

For example now it works with docker-in-docker, docker in linux, and I
think it will also work in docker for mac if somebody can try.

There are two way to create a container, the signature is the same:

* `CreateContainer` returns a container that it never started. So you
cna run the function `Run` by yourself programmatically where you need
it in your test. #29
* `RunContainer` create and run the container.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,18 +32,22 @@ func TestNginxLatestReturn(t *testing.T) {
 		t.Error(err)
 	}
 	defer nginxC.Terminate(ctx, t)
-	ip, err := nginxC.GetIPAddress(ctx)
+	ip, err := nginxC.GetContainerIpAddress(ctx)
 	if err != nil {
 		t.Error(err)
 	}
-	resp, err := http.Get(fmt.Sprintf("http://%s", ip))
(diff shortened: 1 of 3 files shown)


mraerino:
@gianarb this looks really good!
I agree on the decision to have the same method names that the Java impl has.

I've also prepared some refactoring, but this will happily work with this one :+1:

mraerino:
@gianarb I actually just drafted #35 which largely affects what you are doing here.
I will send a PR for my proposal, so you can consider it better.

PR #36, opened 2018-11-25
Cleanup & Implement Interfaces Proposal and Reaper Sidecar

### LINKED ISSUE #24: Integrate with ryuk to guarantee container termination
The java implementation is using a sidecar container to cleanup containers after the test run, even if the test suite is suddenly interrupted. The implementation of that sidecar is available here: https://github.com/testcontainers/moby-ryuk

The things that have to be done to support this:
- Start a goroutine when the container is started and inside that:
- Connect to the sidecar via TCP
- Send filter expression to mark container to clean up
- Close connection once container is terminated (indicated via channel)

### LINKED ISSUE #33: Current wait strategy are weird.
Hello @MeneDev I looked again at the WaitStrategy we have right now, but I don't think they have a lot of sense. I think we should rewrite them to wait on a single port, and probably the waitingFor parameter should be an array of waiting strategies. What do you think?

### LINKED ISSUE #35: Proposal for the public API
*This has been implemented in PR #36*

In the past days I have been thinking about the API of this library. It is important becuase it determines the developer/user experience of this library. It is the first thing people see when using the library.

## Basics

I propose to follow these principles
- Follow Golang conventions for structure and naming
- Hide implementation detail
- Follow learnings of the java implementation, but not necessarily naming schemes
- Provide high-level abstractions, but allow access to lower-level concepts

### Naming

> Go doesn't provide automatic support for getters and setters. There's nothing wrong with providing getters and setters yourself, and it's often appropriate to do so, but it's neither idiomatic nor necessary to put Get into the getter's name. If you have a field called owner (lower case, unexported), the getter method should be called Owner (upper case, exported), not GetOwner.
(https://golang.org/doc/effective_go.html#Getters)

### `Context`

> Programs that use Contexts should follow these rules to keep interfaces consistent across packages and enable static analysis tools to check context propagation:
> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx
(https://golang.org/pkg/context/)

### Types

For the sake of hiding implementation detail, typing might not be the same as in the Docker or Kubernetes clients.
Ports might just be uint16 since that is the appropriately sized number type.

This might mean that a lot of the docker config needs to be reimplemented by this library.
For now it might be best to expose certain important features and allow access to others through a `RawConfig` property.

## `Container` interface

```golang
type Container interface {
	Endpoint(context.Context) (string, error)             // get ip:port string for the first exposed port
	PortEndpoint(context.Context, uint16) (string, error) // get ip:port string for the given exposed port
	IPAddress(context.Context) (string, error)            // get IP address where the container port is exposed
	MappedPort(context.Context, uint16) (uint16, error)   // get externally mapped port for a container port
	Ports(context.Context) (map[uint16]uint16, error)     // get all exposed ports
	Start(context.Context) error                          // start the container
	StartWithParams(context.Context, interface{}) error   // start the container with special parameters
	Terminate(context.Context) error                      // terminate the container
}
```

## `ContainerProvider` interface

```golang
type ContainerRequest struct {
	Image        string
	Env          map[string]string
	ExposedPort  []uint16
	Cmd          string
	Labels       map[string]string
	RawConfig    interface{}
	RegistryCred string
	WaitingFor   wait.WaitStrategy
}

// ContainerProvider allows the creation of containers on an arbitrary system
type ContainerProvider interface {
	CreateContainer(context.Context, ContainerRequest) (Container, error)
}
```

## Constructor functions

```golang
type ProviderType int

const (
	ProviderDocker ProviderType = iota // Docker is default = 0
)

type GenericContainerRequest struct {
	ContainerRequest              // embedded request for provider
	Started          bool         // whether to auto-start the container
	ProviderType     ProviderType // which provider to use, Docker if empty
}

func GenericContainer(ctx context.Context, req GenericContainerRequest) (Container, error) {
	// create provider and container
}
```

☝️ the name for this function is taken from the java implementation and allows for more types such as database and message queue containers

mraerino:
This is a collection of refactors and features which are hard to PR separately because the features need the structure from the refactor. If desirable, I can still split it up into PRs that are merged sequentially.

*Note for Reviewers: Consider going through it commit by commit, not by looking at changed files*

### Refactor wait package to fulfill Go conventions (d4f6280)

This is mostly a cleanup of the `wait` package where I went through what my IDE marked based on the output of `go fmt` and `go vet`.

@MeneDev If you have anything to remark on this one, I'm happy to take input.

### Use docker daemon host to determine container host (c760f11)

Since the current strategy for determining the container address did not work on macOS, I looked how the Java implementation solved this.
You can see it here: [DockerClientConfigUtils.java](https://github.com/testcontainers/testcontainers-java/blob/f113979ad5207464490cc4f19f488bb27795802c/core/src/main/java/org/testcontainers/dockerclient/DockerClientConfigUtils.java#L48)
For now this strategy will serve people well enough. I provided the option to specify this through an env variable called `TC_HOST`.

### Refactor Docker logic to satisfy new interfaces (bed7da0)

Fixes #35

This is the implementation of my proposal. Subject to discussion of course.

### Remove container automatically after stopping (a0414b1)

This is simply passing a parameter to the docker daemon to auto-remove the container upon stopping.

### Refactor waiting strategies to target a single port (9e45426)

Fixes #33

Targeting a single port makes the logic more simple. It will still need a solution if multiple strategies are needed. But this can be a separate feature.

### Enable bind mounts and labels as container params (c2e913b)

This brings some features for the container config.
Needed for the Reaper implementation.

### Use reaper sidecar container to clean up resources (fa93a6d)

Fixes #24

This is really similar to the Java implementation, I think.
Nothing special.
The labels used for cleanup are: `org.testcontainers.golang=true` and `org.testcontainers.golang.sessionId=<sessionId>`

(@gianarb this is based on your refactor commit)


Mohammed90:
> to fulfill Go conventions

Including the conventional [Go Code Review Comments?](https://github.com/golang/go/wiki/CodeReviewComments) If so, then `GetContainerIpAddress` should be `GetContainerIPAddress`, as per [this guideline](https://github.com/golang/go/wiki/CodeReviewComments#initialisms).

mraerino:
@Mohammed90 yes, I know. `GetContainerIpAddress()` has been replaced by `Host()` in d323eb4

README.md
```diff
@@ -32,18 +32,22 @@ func TestNginxLatestReturn(t *testing.T) {
 		t.Error(err)
 	}
 	defer nginxC.Terminate(ctx, t)
-	ip, err := nginxC.GetIPAddress(ctx)
+	ip, err := nginxC.GetContainerIpAddress(ctx)
```
Mohammed90: Then, I guess this should be:

```suggestion
	ip, err := nginxC.Host(ctx)
```

README.md
```diff
@@ -32,18 +32,22 @@ func TestNginxLatestReturn(t *testing.T) {
...
+	}
+	resp, err := http.Get(fmt.Sprintf("http://%s:%d", ip, port))
 	if resp.StatusCode != http.StatusOK {
 		t.Errorf("Expected status code %d. Got %d.", http.StatusOK, resp.StatusCode)
 	}
 }
 ```
 This is a simple example, you can create one container in my case using the
-`nginx` image. You can get its IP `ip, err := nginxC.GetIPAddress(ctx)` and you
+`nginx` image. You can get its IP `ip, err := nginxC.GetContainerIpAddress(ctx)` and you
```
Mohammed90:
```suggestion
`nginx` image. You can get its IP `ip, err := nginxC.Host(ctx)` and you
```

README.md
```diff
@@ -65,17 +69,20 @@ func TestRedisPing(t testing.T) {
         t.Error(err)
     }
     defer redisC.Terminate(ctx, t)
-    redisIP, err := redisC.GetIPAddress(ctx)
-    if err != nil {
-        t.Error(err)
-    }
-
+	ip, err := redisC.GetContainerIpAddress(ctx)
```
Mohammed90: And this should be the only one left out:

```suggestion
	ip, err := redisC.GetContainerIpAddress(ctx)
```

mraerino:
@Mohammed90 Oh, i totally forgot the README. Let me get that quickly!

MeneDev:
No remarks regarding `Refactor wait package to fulfill Go conventions`, thanks for your improvements!

gianarb:
This ia great work but I get 

```
$ go mod tidy
$ go test ./...
# github.com/testcontainers/testcontainer-go [github.com/testcontainers/testcontainer-go.test]
./docker.go:310:32: p.client.DaemonHost undefined (type *client.Client has no field or method DaemonHost)
FAIL    github.com/testcontainers/testcontainer-go [build failed]
?       github.com/testcontainers/testcontainer-go/wait [no test files]
```

mraerino:
@gianarb `go mod tidy` seems to remove the docker client library from the go.mod file. Can you try `git reset --hard` and then just `go mod download`?

The problem with the go docker client:
The latest version which is versioned correctly is v0.13.1. That one is really old and is missing the `DaemonHost` feature.
I upgraded it by pointing it to a newer commit, because pointing it to a tag like `v17.05.0-ce` is useless.

gianarb:
Great @mraerino all of them are working! This is ready to go!

PR #38, opened 2018-12-18
Retrieve gateway address if inside container

pburakov:
Addresses todo at [docker.go#L319](https://github.com/testcontainers/testcontainer-go/blob/15ebe77a7f97d7229efdce775f5d9cff309e86e8/docker.go#L319). 

We were using this lib for testing custom go services at Spotify and needed to implement this when running integration tests in a container, so I thought it would be helpful. Mirrors [this](https://github.com/testcontainers/testcontainers-java/blob/3ad8d80e2484864e554744a4800a81f6b7982168/core/src/main/java/org/testcontainers/dockerclient/DockerClientConfigUtils.java#L15) Java implementation.

Would love to hear your comments and suggestions for a comprehensive testing strategy for this.

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"net/url"
 	"os"
+	"os/exec"
 	"strings"
 
 	"github.com/docker/docker/api/types"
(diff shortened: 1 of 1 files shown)


gianarb:
Perfect, it works! Thanks a lot and I am very happy to see this library used at Spotify!

PR #42, opened 2019-01-08
BC BREAk move to testcontainers-go

gianarb:
The right name for the package should be testcontainers-go not
testcontainer-go. The `s` is important for consistency with the other
libraries across languages.

This is BC break and to fix this you need to change the import in your
applications.

### DIFF
--- a/container.go
+++ b/container.go
@@ -1,11 +1,11 @@
-package testcontainer
+package testcontainers
 
 import (
 	"context"
 
 	"github.com/docker/go-connections/nat"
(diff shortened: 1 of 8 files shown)


bsideup:
AFAIK if you rename the Github repo, it will redirect, and the package will work just fine.

Docker did it once (from `docker/docker` to `moby/moby`)

gianarb:
Nice, I am never sure about how this works will all the different dependency managers that golang has. Anyway is better to do it now than later

PR #50, opened 2019-01-31
Fixes to Travis configuration

bmuschko:
- Fixes broken link to Travis CI job via badge
- Avoids running the default install script (`go get -t -v ./...`) which is unnecessary when using Go modules


gianarb:
Thanks a lot!

PR #53, opened 2019-03-09
Exposed containerID with the function GetContainerID

### LINKED ISSUE #52: Please expose ContainerID via Container interface
Hello,

I have a use case in which I want to use the **container-ID** in order to interact with the created container from another process.
Therefore, would it be possible to also expose the **container-ID** via the Container interface, as returned by e.g. the `testcontainers.GenericContainer()` function?

Thank you very much.

gianarb:
Close #52

Now the Container exposes a method called `GetContainerID`. It exposes
the container id provided by the provider (docker atm).

@m4x-h4xx how does it sounds? Thanks

### DIFF
--- a/container.go
+++ b/container.go
@@ -25,6 +25,7 @@ type ContainerProvider interface {
 
 // Container allows getting info about and controlling a single container instance
 type Container interface {
+	GetContainerID() string                                         // get the container id from the provider
 	Endpoint(context.Context, string) (string, error)               // get proto://ip:port string for the first exposed port
 	PortEndpoint(context.Context, nat.Port, string) (string, error) // get proto://ip:port string for the given exposed port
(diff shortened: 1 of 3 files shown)


m4x-h4xx:
Hi @gianarb ,
that is just wonderful. Thank you very much for your quick response!

Best regards
Max

container.go
```diff
@@ -25,6 +25,7 @@ type ContainerProvider interface {
 
 // Container allows getting info about and controlling a single container instance
 type Container interface {
+	GetContainerID(context.Context) string                          // get the container id from the provider
```
bsideup: This method can't be async (since the id is static and retrieving at container's startup), is it really necessary to pass the context as an argument?

gianarb: I have removed it, it should not be a requirement

Issue #54, opened 2019-03-24
Wait for specific console output


alde:
It'd be great if it was possible to wait for specific log output, such as 

`MySQL init process done. Ready for start up.`

The port is available far in advance of this, causing me to have to add a `time.Sleep(30*time.Second)` in my test only to make sure it will have started.

gianarb:
Thanks @alde good issue. This is in some way related to #19 .

This issue will be nice to have and it will involve a pre-feature that exposes an API to get/follow logs.

gianarb:
As suggested via Slack https://github.com/testcontainers/testcontainers-java/blob/master/core/src/main/java/org/testcontainers/containers/wait/strategy/LogMessageWaitStrategy.java

PR #55, opened 2019-04-08
Implement a Log wait strategy

alde:
This makes it possible to wait for specific log output.

For example it's useful for MySQL test containers, which has the
port and host available long before it's actually ready to
accept connections.

example usage:

```go
WaitingFor: wait.ForLog("port: 3306  MySQL Community Server - GPL"),
```

Solves #54

### DIFF
--- a/container.go
+++ b/container.go
@@ -2,6 +2,7 @@ package testcontainers
 
 import (
 	"context"
+	"io"
 
 	"github.com/docker/go-connections/nat"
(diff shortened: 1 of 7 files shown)


gianarb:
Wow @alde I have nothing to add about this PR! Clear and simple! Thanks a lot!

alde:
Thanks :) 👍

Issue #19, opened 2018-11-22
Implement database containers


mraerino:
The java implementation of testcontainers provides useful preconfigured setups for certain databases.
They usually predefine an image and setup the special waiting strategy for that database.

This should be straight-forward for Go as well since there are many database clients around to do this.

I think I'll get started with a `postgres` setup.

gianarb:
Yep, this looks like a great idea. Can you write down some code here to help me understand how you imagine interacting with the postgres utils? probably just  a fake test is enough

gianarb:
Should we start packaging this kind of solutions in a subpackage? Maybe a directory called `canned`?

fbiville:
I'd be happy to write a Neo4j integration.

gianarb:
I am gonna this issue for now! As you can read in our documentation Golang is not really flexible when it comes to managing multiple modules in the same project and dependency management is still a pain.

https://golang.testcontainers.org/modules/

We decided to stay focused on the core library for now. If you decide to develop your module open a PR and link it in our documentation. We are happy to share it with the community but not to maintain it directly at this stage ❤️

mdelapenya:
We started this PR, which could be the seed for modules: https://github.com/testcontainers/testcontainers-go/pull/618

It creates examples to be consumed by the website, but turning them into modules would be a matter of adjusting the templates.

mdelapenya:
Given #618 has been merged, we created https://github.com/testcontainers/testcontainers-go/issues/636, where you can find how to contribute new modules :)

PR #59, opened 2019-04-20
Mysql canned container

gianarb:
Reference #19

I bootstrapped a mysql canned container.

```go
ctx := context.Background()
c, err := MySQLContainer(ctx, MySQLContainerRequest{
        RootPassword: "root",
        Database:     "hello",
})
if err != nil {
        t.Fatal(err.Error())
}
defer c.Container.Terminate(ctx)
sqlC, err := c.GetDriver("root", "root", "hello")
if err != nil {
        t.Fatal(err.Error())
}
_, err = sqlC.ExecContext(ctx, "CREATE TABLE example ( id integer, data varchar(32) )")
if err != nil {
        t.Fatal(err.Error())
}
```


gianarb:
@mraerino @alde what do you think?

canned/mysql.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+type mysqlContainer struct {
+	Container testcontainers.Container
+	db        *sql.DB
+	req       MySQLContainerRequest
+	ctx       context.Context
+}
+
+func (c *mysqlContainer) GetDriver(username, password, database string) (*sql.DB, error) {
```
mraerino: Consider taking this from the request and fallbacking to the defaults of the image

canned/mysql.go
```diff
@@ -0,0 +1,95 @@
...
+	return db, nil
+}
+
+func MySQLContainer(ctx context.Context, req MySQLContainerRequest) (*mysqlContainer, error) {
+	provider, err := req.ProviderType.GetProvider()
+	if err != nil {
+		return nil, err
+	}
+
+	req.Image = "mysql:8.0"
```
mraerino: Consider making the version configurable from the request and having a default in a package constant.

mraerino: Maybe also make the image configurable so somebody can choose a custom image that is compatible with the options of the default image

canned/mysql.go
```diff
@@ -0,0 +1,95 @@
...
+	req.ExposedPorts = []string{"3306/tcp"}
+	req.Env = map[string]string{}
+	req.Started = true
```
mraerino: All these options basicly overwrite options that may be set by a user. Try checking if this is empty in the `ContainerRequest`

canned/mysql.go
```diff
@@ -0,0 +1,95 @@
...
+		req.Env["MYSQL_USERNAME"] = req.Username
+	}
+	if req.Password != "" {
+		req.Env["MYSQL_PASSWORD"] = req.Password
+	}
+	if req.Database != "" {
+		req.Env["MYSQL_DATABASE"] = req.Database
+	}
+
+	req.WaitingFor = wait.ForLog("port: 3306  MySQL Community Server - GPL")
```
mraerino: The java implementation actually tries connecting via a mysql client to verify that the database is ready.

canned/mysql.go
```diff
@@ -0,0 +1,95 @@
...
+
+	c, err := provider.CreateContainer(ctx, req.ContainerRequest)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to create container")
+	}
+	mysqlC := &mysqlContainer{
+		Container: c,
+	}
+	mysqlC.req = req
+	mysqlC.ctx = ctx
```
mraerino: This is not a common pattern since a context is usually valid for a single function call. Consider e.g. `context.WithTimeout()`

canned/mysql.go
```diff
@@ -0,0 +1,95 @@
...
+
+	if req.Started {
+		if err := c.Start(ctx); err != nil {
+			return mysqlC, errors.Wrap(err, "failed to start container")
+		}
+	}
```
mraerino: is this functionality not implemented in the testcontainer-go core?

alde:
Sorry, I've been away on vacation with limited internet access.

Love the direction, and I agree with most of @mraerino's points :)

canned/mysql/mysql.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/pkg/errors"
+	testcontainers "github.com/testcontainers/testcontainers-go"
+)
+
+const (
+	DefaultImageVersion = "mysql:8.0"
+)
+
+type MySQLContainerRequest struct {
+	testcontainers.GenericContainerRequest
```
gianarb: I think we should not accept a request here, we should build it based on what we know from the image the other params we get from the user. What do you think?

PR #66, opened 2019-05-17
Change isReaper with SkipReaper

mauromedda:
Change the name and the visibility of the skip Reaper flag included
in the ContainerRequest structure. Now is possible disable the Reaper
creation from during the container creation request.


gianarb:
Thanks! Sorry if I was a bit late. for a bit of time I though this was a BC, but it is not

Issue #73, opened 2019-07-17
Is it possible to attach created containers to a specific docker network?


mdelapenya:
The use case would be creating a dedicated network for testing/local dev, so it's possible to attach created containers to it.

At this moment, the NetworkingConfig is passed as `nil`.

If you consider implementing it, I'd be glad to contribute it :)

mdelapenya:
Meanwhile I get feedback from this, I continued attaching containers to the default `bridge` Docker network.

While it could be enough, I consider it be very convenient if the API exposed the `inspectContainer` method, so any client code could use whatever information they need, instead of using just the existing`Host(), Name(), Ports()`, etc.

Wdyt?

gianarb:
I think we can do that, it doesn't need to be part of the Container interface because I would like to keep that interface easy to implement eventually with containerd or some other provider.

mdelapenya:
Merged, thanks!

![image](https://media.giphy.com/media/D0WOL0ogZIoG4/giphy.gif)

PR #76, opened 2019-07-25
(#73) Support attaching containers to different networks than bridge

mdelapenya:
## What does this PR do?
This PR adds support for client code to be able to configure networks a container is connected to, which nowadays is impossible as the default network configuration when invoking the Docker client to create and run a container is passed as `nil`.

For that reason, a Network request object is needed, alongside the methods to create/remove/get a network from the Docker client. Those methods have been defined at a new NetworkProvider interface.

As a container could have multiple aliases when connecting to a network, we are also adding that support: when adding a network it's possible to set the aliases for the container in that network.

## Why is this PR important?
Client code would eventually like to connect a container run by this library to a specific set of networks, and this library currently disallows it always using the default Docker bridge network.

## Testing
A unit test that verifies that a container created with one network is connected to it.

## Other considerations
Any feedback about naming and Go-conventions is super well received! 😃 

Thanks!

### DIFF
--- a/container.go
+++ b/container.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"io"
 
-	"github.com/docker/docker/api/types"
 	"github.com/docker/go-connections/nat"
 	"github.com/pkg/errors"
 	"github.com/testcontainers/testcontainers-go/wait"
@@ -21,11 +20,8 @@ type DeprecatedContainer interface {
 
 // ContainerProvider allows the creation of containers on an arbitrary system
 type ContainerProvider interface {
-	CreateContainer(context.Context, ContainerRequest) (Container, error)      // create a container without starting it
-	CreateNetwork(context.Context, NetworkRequest) (Network, error)            // create a network
-	GetNetwork(context.Context, NetworkRequest) (types.NetworkResource, error) // get a network
-	RemoveNetwork(context.Context, NetworkRequest) error                       // remove a network
-	RunContainer(context.Context, ContainerRequest) (Container, error)         // create a container and start it
+	CreateContainer(context.Context, ContainerRequest) (Container, error) // create a container without starting it
+	RunContainer(context.Context, ContainerRequest) (Container, error)    // create a container and start it
 }
 
 // Container allows getting info about and controlling a single container instance
@@ -70,7 +66,7 @@ const (
 )
 
 // GetProvider provides the provider implementation for a certain type
-func (t ProviderType) GetProvider() (ContainerProvider, error) {
+func (t ProviderType) GetProvider() (GenericProvider, error) {
 	switch t {
 	case ProviderDocker:
 		provider, err := NewDockerProvider()
--- a/container.go
+++ b/container.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"io"
 
+	"github.com/docker/docker/api/types"
 	"github.com/docker/go-connections/nat"
 	"github.com/pkg/errors"
 	"github.com/testcontainers/testcontainers-go/wait"
@@ -20,8 +21,11 @@ type DeprecatedContainer interface {
 
 // ContainerProvider allows the creation of containers on an arbitrary system
 type ContainerProvider interface {
-	CreateContainer(context.Context, ContainerRequest) (Container, error) // create a container without starting it
-	RunContainer(context.Context, ContainerRequest) (Container, error)    // create a container and start it
+	CreateContainer(context.Context, ContainerRequest) (Container, error)      // create a container without starting it
+	CreateNetwork(context.Context, NetworkRequest) (Network, error)            // create a network
+	GetNetwork(context.Context, NetworkRequest) (types.NetworkResource, error) // get a network
+	RemoveNetwork(context.Context, NetworkRequest) error                       // remove a network
+	RunContainer(context.Context, ContainerRequest) (Container, error)         // create a container and start it
 }
 
 // Container allows getting info about and controlling a single container instance
@@ -37,6 +41,7 @@ type Container interface {
 	Terminate(context.Context) error                                // terminate the container
 	Logs(context.Context) (io.ReadCloser, error)                    // Get logs of the container
 	Name(context.Context) (string, error)                           // get container name
+	Networks(context.Context) ([]string, error)                     // get container networks
 }
 
 // ContainerRequest represents the parameters used to get a running container
@@ -49,8 +54,9 @@ type ContainerRequest struct {
 	BindMounts   map[string]string
 	RegistryCred string
 	WaitingFor   wait.Strategy
-	Name         string // for specifying container name
-	Privileged   bool   // for starting privileged container
+	Name         string   // for specifying container name
+	Privileged   bool     // for starting privileged container
+	Networks     []string // for specifying network names
 
 	SkipReaper bool // indicates whether we skip setting up a reaper for this
 }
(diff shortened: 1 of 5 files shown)


container.go
```diff
@@ -20,8 +21,11 @@ type DeprecatedContainer interface {
 
 // ContainerProvider allows the creation of containers on an arbitrary system
 type ContainerProvider interface {
-	CreateContainer(context.Context, ContainerRequest) (Container, error) // create a container without starting it
-	RunContainer(context.Context, ContainerRequest) (Container, error)    // create a container and start it
+	CreateContainer(context.Context, ContainerRequest) (Container, error)      // create a container without starting it
+	CreateNetwork(context.Context, NetworkRequest) (Network, error)            // create a network
```
bsideup: I would be better to treat networks as a domain object on their own and not mix with the containers. 

You can check Java's implementation for the inspiration:
https://www.testcontainers.org/features/networking/#advanced-networking

container.go
```diff
@@ -49,8 +54,9 @@ type ContainerRequest struct {
 	BindMounts   map[string]string
 	RegistryCred string
 	WaitingFor   wait.Strategy
-	Name         string // for specifying container name
-	Privileged   bool   // for starting privileged container
+	Name         string   // for specifying container name
+	Privileged   bool     // for starting privileged container
+	Networks     []string // for specifying network names
```
bsideup: I would suggest adding the network aliases too

docker.go
```diff
@@ -182,6 +183,39 @@ func (c *DockerContainer) Name(ctx context.Context) (string, error) {
 	return inspect.Name, nil
 }
 
+// Networks gets the names of the networks the container is attached to.
+func (c *DockerContainer) Networks(ctx context.Context) ([]string, error) {
+	inspect, err := c.inspectContainer(ctx)
```
bsideup: FYI in Java, we inspect it once after the start and reuse the object, unless user asks for the current state

docker.go
```diff
@@ -359,6 +409,51 @@ func (p *DockerProvider) daemonHost() (string, error) {
 	return p.hostCache, nil
 }
 
+// CreateNetwork returns the object representing a new network identified by its name
+func (p *DockerProvider) CreateNetwork(ctx context.Context, req NetworkRequest) (Network, error) {
```
bsideup: make sure you delete networks after the tests

container.go
mdelapenya: Yeah, I created it in the ContainerProvider because it seemed to me like Docker was the provider, and in that sense it would also be the network provider.

I can rework it to have a NetworkProvider 👍

mdelapenya: You're right: I will try to mimic the Java implementation as much as possible, thanks!

docker.go
mdelapenya: Addressed here: https://github.com/testcontainers/testcontainers-go/pull/76/commits/41baa6d602d6d8985acb216aa8468f88f07e4cc5

container.go
mdelapenya: Addresses here: https://github.com/testcontainers/testcontainers-go/pull/76/commits/f85faef1f07e90b0f410423570782cb794669a55

I used a `map[string][]string`, so that the map indexes each network alongside an array for the aliases for the container in that network

container.go
mdelapenya: Addresses here: https://github.com/testcontainers/testcontainers-go/pull/76/commits/1e5969298fbba8915301b38e3ef3f7616a62559d

mdelapenya:
Hey @bsideup, I think I addressed your suggestions, so this is ready to review.

Thanks!

docker.go
mdelapenya: Here I tried to use same pattern as in `.Name(ctx)` and `.Ports(ctx)`: inspect the container and retrieve the desired data

docker.go
bsideup: you only addressed it in test, but the library should cleanup containers/networks/everything on exit

docker.go
bsideup: I see. @gianarb perhaps you should consider doing a small refactoring, inspecting containers for each getter is an expensive operation sometimes

container.go
bsideup: Aliases are set "per container", not "per network", there is no need to have a Map for them.

See
https://github.com/testcontainers/testcontainers-java/blob/c094a034f733215bfd88d41538089699caab1616/core/src/main/java/org/testcontainers/containers/GenericContainer.java#L596

mdelapenya: In this PR I'm setting the network aliases `per container`, so a container in the network A could have a few aliases, and in the network B could have other aliases.

docker.go
mdelapenya: I've used the Reaper approach to delete networks, as the `ryuk` container supports it

mdelapenya:
Hi @bsideup any feedback about the removal part?

Thanks!

docker.go
gianarb: btw this is what `inspectContainer` does already https://github.com/testcontainers/testcontainers-go/blob/22d6b6a6e5d8bc2a0fcad730929783ad88a3c5d2/docker.go#L155 it caches the object

mdelapenya: Great then! And what about https://github.com/testcontainers/testcontainers-go/issues/73#issuecomment-512865607? I could elaborate a new PR exposing the inspectContainer method as public. Wdyt, makes sense?

mdelapenya:
Hi @gianarb, any feedback about the review/merge of this PR?

I'd like to continue sending code for the inspect method I mentioned, if you like the idea of course.

mdelapenya:
Hey @bsideup @gianarb Any guess about this one?  🙏

gianarb:
I am on holiday 🌴. I will be back next week!

Thanks for your patient

mdelapenya:
image

docker.go
gianarb: @mdelapenya I would like to avoid that for now. The idea is to keep the interface as small as possible in order to able to implement the Container interface for other runtimes such as containerd.  If you expose the inspectContainer publicly it means that we need to wrap the docker response around abstraction and I would like to avoid that for now.

docker_test.go
```diff
@@ -15,6 +15,74 @@ import (
 	"github.com/testcontainers/testcontainers-go/wait"
 )
 
+func TestContainerAttachedToNewNetwork(t *testing.T) {
+	networkName := "new-network"
```
gianarb: Hello,
I tried your branch and one of the network created in this test didn't got removed:

```
 $ docker network ls
NETWORK ID          NAME                DRIVER              SCOPE
6050f659c458        bridge              bridge              local
adbb6c9dfe3f        host                host                local
c40a6b07ec52        new-network         bridge              local
dccf217218e7        none                null                local
ff00e6906c6e        potato_net-test     bridge              local
```
Do you have an idea?

gianarb:
I left a comment but in general, the API is clear and I can merge it. Just need to double check my comment @mdelapenya . Thanks a lot and sorry I didn't have much time in August to work on testcontainers

mdelapenya:
No worries, I'll take a look as soon as possible.

Thanks!

docker_test.go
mdelapenya: Shouldn't this line remove it? https://github.com/testcontainers/testcontainers-go/pull/76/files/22d6b6a6e5d8bc2a0fcad730929783ad88a3c5d2#diff-98b81bf788c107c9f5f395060629d5ecR48

mdelapenya: Fixed! I sent this before your fix edb1f310022e48268867572431fb2a467c85c0b5, so I was following same pattern. I pushed-force one commit after rebasing with master

Thank you!

gianarb:
Thanks a lot @mdelapenya !

mdelapenya:
image

PR #85, opened 2019-08-16
Implemented stop/remove/state/image methods for containers, list containers method, added an ability to reuse existent container

azakharenko:
Reusing of containers can be used in testing (when you just run tests and you don't need to launch new container every time). I made it for my own purposes, but maybe someone else will find it useful.

### DIFF
--- a/container.go
+++ b/container.go
@@ -6,9 +6,9 @@ import (
 
 	"github.com/docker/docker/api/types"
 
-	"github.com/azakharenko/testcontainers-go/wait"
 	"github.com/docker/go-connections/nat"
 	"github.com/pkg/errors"
+	"github.com/testcontainers/testcontainers-go/wait"
 )
 
 // DeprecatedContainer shows methods that were supported before, but are now deprecated
--- a/container.go
+++ b/container.go
@@ -4,9 +4,11 @@ import (
 	"context"
 	"io"
 
+	"github.com/docker/docker/api/types"
+
+	"github.com/azakharenko/testcontainers-go/wait"
 	"github.com/docker/go-connections/nat"
 	"github.com/pkg/errors"
-	"github.com/testcontainers/testcontainers-go/wait"
 )
 
 // DeprecatedContainer shows methods that were supported before, but are now deprecated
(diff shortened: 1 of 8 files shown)


container.go
```diff
@@ -35,8 +40,14 @@ type Container interface {
 	SessionID() string                                              // get session id
 	Start(context.Context) error                                    // start the container
 	Terminate(context.Context) error                                // terminate the container
+	Stop(context.Context) error                                     // stop the container
+	Remove(context.Context, bool) error                             // remove the container
 	Logs(context.Context) (io.ReadCloser, error)                    // Get logs of the container
 	Name(context.Context) (string, error)                           // get container name
+	IsRunning(ctx context.Context) (bool, error)                    // is state of container 'running'
+	State(ctx context.Context) (*types.ContainerState, error)       // state of container
```
gianarb: This is a great work. I like it thanks! I am trying to keep the interface as agnostic as possible and the fact that it returns the `*types.ContainerState` tight the interface to Docker and I do not like it for now because the idea is to be able to implement it for other runtimes as well. So for now can you remove the `State` function? The rest sounds good to me!

container.go
```diff
@@ -35,8 +40,14 @@ type Container interface {
...
+	Image(context.Context) (string, error)                          // get container image
+	ResetCache(context.Context)                                     // reset internal testcontainers-go cache
```
gianarb: `ResetCache` should be a private function, it does not need to be in the Container interface. We can implement it only for the docker provider.

Thanks

container.go
```diff
@@ -51,6 +62,8 @@ type ContainerRequest struct {
 	WaitingFor   wait.Strategy
 	Name         string // for specifying container name
 	Privileged   bool   // for starting privileged container
+	Entrypoint   []string
+	DontRemove   bool
```
gianarb: There is another PR that has the same implementation, can you remove this part as well? Thanks

gianarb:
This is too old and big as PR. Can't merge it.
If you have time to work it in small PR it will we great.
my 2c, open an issue first to discuss how to approach all of this.

Thanks a lot!

PR #87, opened 2019-09-02
Using a slice of strings ([]string) for Docker Cmd.

### LINKED ISSUE #86: Define Cmd as Slice of Strings
**Describe the bug**
Not a bug, a suggestion.
When creating a `ContainerRequest`, we give the `Cmd` attribute a `string` type.  Then we [parse it to make it a slice of strings](https://github.com/testcontainers/testcontainers-go/blob/master/docker.go#L252)

Would it be a better idea to define the [Cmd attribute in the ContainerRequest struct](https://github.com/testcontainers/testcontainers-go/blob/master/container.go#L47) as a slice of strings (`[]string`)?

The implicit parsing (splitting) is not clear to a user of the project, I think it would be good to be explicit here.

**To Reproduce**
Intentionally leaving blank, not a bug, a suggested change.

**Expected behavior**
```
type ContainerRequest struct {
	Image        string
	Env          map[string]string
	ExposedPorts []string // allow specifying protocol info
	Cmd          []string
	Labels       map[string]string
	RegistryCred string
	WaitingFor   wait.Strategy
	Name         string // for specifying container name
	Privileged   bool   // for starting privileged container

	SkipReaper bool // indicates whether we skip setting up a reaper for this
}
```

ClaytonNorthey92:
Resolves #86

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,3 +61,17 @@ can use it to make a GET: `resp, err := http.Get(fmt.Sprintf("http://%s", ip))`
 To clean your environment you can defer the container termination `defer
 nginxC.Terminate(ctx, t)`. `t` is `*testing.T` and it is used to notify is the
 `defer` failed marking the test as failed.
+
+## Sending a CMD to a Container
+
+If you would like to send a CMD (command) to a container, you can pass it in to the container request via the `Cmd` field...
+
+```go
+req := ContainerRequest{
+	Image: "alpine",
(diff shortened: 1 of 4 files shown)


gianarb:
@ClaytonNorthey92 can you rebase please?
Thanks a lot

ClaytonNorthey92:
@gianarb just rebased! thanks!

Issue #93, opened 2019-09-05
Internal state of the DockerContainer is not reset after container.Terminate(ctx)


mdelapenya:
**Describe the bug**
I'm working on exposing the inspectContainer method, and I realised that the Terminate method does not reset the internal state of the DockerContainer instance: raw and uuid

**To Reproduce**
I'm sending a PR for this. First commit will be a unit test

**Expected behavior**
After a container is terminated, its internal state (uuid and raw) is reset

**docker info**
output of the command:

```
Client:
 Debug Mode: false

Server:
 Containers: 3
  Running: 3
  Paused: 0
  Stopped: 0
 Images: 306
 Server Version: 19.03.1
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 894b81a4b802e4eb2a91d1ce216b8817763c29fb
 runc version: 425e105d5a03fabd737a126ad93d62a9eeede87f
 init version: fec3683
 Security Options:
  seccomp
   Profile: default
 Kernel Version: 4.9.184-linuxkit
 Operating System: Docker Desktop
 OSType: linux
 Architecture: x86_64
 CPUs: 4
 Total Memory: 7.787GiB
 Name: docker-desktop
 ID: BMI5:JZNA:56SZ:5U24:RYWQ:OJQB:AXJA:LQP6:ZGVF:VGNO:M63Z:6W7X
 Docker Root Dir: /var/lib/docker
 Debug Mode: true
  File Descriptors: 55
  Goroutines: 65
  System Time: 2019-09-05T07:25:03.9382341Z
  EventsListeners: 2
 HTTP Proxy: gateway.docker.internal:3128
 HTTPS Proxy: gateway.docker.internal:3129
 Registry: https://index.docker.io/v1/
 Labels:
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false
 Product License: Community Engine
```

mdelapenya:
Merged, thanks!

[...]

PR #94, opened 2019-09-05
(#93) Reset internal state after container.Terminate(), and expose InspectContainer

mdelapenya:
## What does this PR do?
The original purpose of this PR was to expose the inspectContainer method on the DockerContainer model. In that process, I found that there is a local cache (raw) that allowed to invoke the public methods that internally use `inspectContainer` (Host, Ports...) even after terminating the container, because that local cache was not removed.

For that reason this PR adds that ability to remove the local cache (raw and sessionID) when terminating a container.

## Why is this PR important?
Nowadays the API allows to invoke another container method (Host, Ports...) after the container is terminated, and the internal cached state would satisfy the request even when the container is not existing anymore at Docker client level. We are removing the cached values so that we receive the response directly from the container provider.

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -153,6 +153,12 @@ func (c *DockerContainer) Terminate(ctx context.Context) error {
 		RemoveVolumes: true,
 		Force:         true,
 	})
+
+	if err == nil {
+		c.sessionID = uuid.UUID{}
+		c.raw = nil
+	}
+
 	return err
 }
 
--- a/docker_test.go
+++ b/docker_test.go
@@ -178,6 +178,40 @@ func TestContainerStartsWithTheReaper(t *testing.T) {
 	}
 }
 
+func TestContainerTerminationResetsState(t *testing.T) {
+	ctx := context.Background()
+	client, err := client.NewEnvClient()
+	if err != nil {
+		t.Fatal(err)
+	}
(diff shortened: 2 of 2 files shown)


container.go
```diff
@@ -30,6 +31,7 @@ type Container interface {
 	Endpoint(context.Context, string) (string, error)               // get proto://ip:port string for the first exposed port
 	PortEndpoint(context.Context, nat.Port, string) (string, error) // get proto://ip:port string for the given exposed port
 	Host(context.Context) (string, error)                           // get host where the container port is exposed
+	InspectContainer(context.Context) (*types.ContainerJSON, error) // get the JSON representing the container
```
gianarb: This can not be merged. At the moment I would like to keep the Container interface provider agnostic. Because I would like to keep the door open to eventually have a containerd provider.

exposing the inspect response will prevent this to happen.
The rest of the PR is fine, if you can remove this I will merge.

mdelapenya: Sure, I'll rework this to remove the inspect part

Thanks!!

mdelapenya: Hey @gianarb I removed the expose part, keeping the internal cache removal.

Hope it's good to go!

gianarb:
Thanks a lot

PR #97, opened 2019-09-11
(#25) Create very basic implementation of Local Docker Compose

### LINKED ISSUE #25: Docker compose support
The java implementation supports spinning up docker compose setups to test larger setups as a black box.

This is interesting for this library especially since many DevOps people are using Golang for their services anyways.

mdelapenya:
## What is this PR doing?
This is the very beginning of an initiative to port the Docker-Compose approach from the Java implementation to Golang.

It simply creates the [Local Docker Compose approach](https://github.com/mdelapenya/testcontainers-java/blob/4df20c2d1d090d8f2d3129ec24b0988faecbe896/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java#L642) (no socat container at this moment), which forces developers to have the docker-compose binary in PATH.

Following Java's approach, a LocalDockerCompose (LDC) object needs to be created with an array of paths to docker-compose.yml files, and an identifier (maybe this second field could be internally created).

It's possible to configure the LDC with a command (i.e. "up -d"), which will be executed after the method Invoke is called.

We have added a convenient `Down()` method, to be executed after tests run to destroy resources created by docker-compose.

Besides that, we are parsing Docker Compose YAML file to capture service names, and put into an array, so that it's possible to perform operations over the services in the compose file, accessing the `compose.Services` field.

## Why is it important?
This will allow developers to configure more elaborated setups for testing, like an Elasticsearch + Kibana, or Metricbeat + MySQL, as an example, being those services coordinated and scheduled by Docker Compose under the hood, with no need to orchestrate them on the programatic side.

On the other hand, using declarative way of describing services is more familiar to developers (🤞), so the entry barrier could be lower.

## How to test it?
I added a few unit tests, which verify that there are no errors after executing the compose, checking that environment variables are properly propagated from compose to the containers, and destroying it in a deferred manner.

## Related Issues
Closes #25


mdelapenya:
Hey @gianarb, any feedback with this one?

mdelapenya:
Closing for travis rebuild

gianarb:
Sweet! The direction is right! I am trying it just now!

gianarb:
I am looking at how testcontainers-java does that.

https://www.testcontainers.org/modules/docker_compose/
```
Note that it is not necessary to define ports to be exposed in the YAML file; this would inhibit reuse/inclusion of the file in other contexts.

Instead, Testcontainers will spin up a small 'ambassador' container, which will proxy between the Compose-managed containers and ports that are accessible to your tests. This is done using a separate, minimal container that runs socat as a TCP proxy.
```

mdelapenya:
Yeah, this first proposal is trying to address the easy one: the local docker-compose approach, which is totally decoupled from the one with the minimal socat container.

As decoupled they are, I'd review them separately. Wdyt?

gianarb:
Something I do not see from the examples is around how do you think we can get URLs to access containers deployed with docker-compose? Let's say we have a MySQL and we should like to get the service endpoint for it?
Thanks!

mdelapenya:
Ok, let me check out Java implementation for that. I agree that there should be a manner to access the container from the compose

Thanks!

mdelapenya:
Hey @gianarb I don't think the local compose for Java supports accessing a service. I looked through the code and asked in their repo (https://github.com/testcontainers/testcontainers-java/issues/1928) and did not find a manner to access the service 🤷‍♂

mdelapenya:
I think this PR is ready for review

mdelapenya:
Tests are passing locally, closing and reopening again

mdelapenya:
@gianarb not sure why, but Travis fails consistently with:

```
--- FAIL: TestLocalDockerComposeWithMultipleComposeFiles (2.35s)
    compose_test.go:114: {<nil> read |0: file already closed read |0: file already closed}
FAIL
FAIL	github.com/testcontainers/testcontainers-go	129.580s
```
although tests are passing locally. Can you double check please?

mdelapenya:
Reopening and checking Travis 🤞

mdelapenya:
@gianarb Any feedback about this one?

utility.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"math/rand"
+	"time"
+)
+
+var alfabet = []rune("123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz")
+
+// RandomString returns a random string
+func RandomString(length int) string {
```
ClaytonNorthey92: would it be better to generate a UUID? rather than write our own `RandomString` function here? if it is just being used as an identifier, we just need to guarantee its uniqueness, no?

compose.go
```diff
@@ -0,0 +1,213 @@
...
+type DockerCompose interface {
+	Down() ExecError
+	Invoke() ExecError
+	WithCommand([]string) DockerCompose
+	WithEnv(map[string]string) DockerCompose
+}
+
+// LocalDockerCompose represents a Docker Compose execution using local binary
+// docker-compose or docker-compose.exe, depending on the underlying platform
+type LocalDockerCompose struct {
```
ClaytonNorthey92: would it be possible to add a usage example to the README? I like that I can read through the tests for examples, but I think our README is lacking features in general

utility.go
mdelapenya: Totally agree, and I'm using it only for tests, which makes no sense to maintain new code.

Thanks for the feedback!

mdelapenya:
> I know it is a bit of work, but would it be possible to add a test that has more than one service in docker-compose and test the interaction between them?
> 
> for example, if we could use nginx or envoy to reverse proxy a request, for example, I might want to write to redis, and read that value from redis, so I might want a docker compose that I could test functionality that
> 
> write <key/value> --> nginx/envoy --> redis
> read <key/value> --> nginx/envoy --> redis
> 
> I think this would be a common use case that would be a good use case for docker-compose to perform, thoughts?

Wouldn't it mean testing that docker-compose works? I agree in adding a new test with a combination of services that interact with each other (i.e. a service that populates a database), but testing the network and the dns discovery would be a responsibility for docker guys imho, which I'm  sure it's something they are already doing and it's working 😃

ClaytonNorthey92:
> > I know it is a bit of work, but would it be possible to add a test that has more than one service in docker-compose and test the interaction between them?
> > for example, if we could use nginx or envoy to reverse proxy a request, for example, I might want to write to redis, and read that value from redis, so I might want a docker compose that I could test functionality that
> > write <key/value> --> nginx/envoy --> redis
> > read <key/value> --> nginx/envoy --> redis
> > I think this would be a common use case that would be a good use case for docker-compose to perform, thoughts?
> 
> Wouldn't it mean testing that docker-compose works? I agree in adding a new test with a combination of services that interact with each other (i.e. a service that populates a database), but testing the network and the dns discovery would be a responsibility for docker guys imho, which I'm sure it's something they are already doing and it's working 😃

ah yes, you're right! good point!

mdelapenya:
🤞 @gianarb is there anything I can do to unblock this? I'm using my fork in an internal project instead of using the upstream, and it's working quite good.

Thanks!

compose.go
```diff
@@ -0,0 +1,213 @@
...
+
+// WithEnv assigns the environment
+func (dc *LocalDockerCompose) WithEnv(env map[string]string) DockerCompose {
+	dc.Env = env
+	return dc
+}
+
+// ExecError is super struct that holds any information about an execution error, so the client code
+// can handle the result
+type ExecError struct {
```
ClaytonNorthey92: hey @mdelapenya sorry to add another suggestion so late 😬 but would it be a good idea to add the command (`cmd`) that failed to this struct?  I think that would be useful for debugging.  I know this is not trivial since the `Cmd` struct looks somewhat large.  Just and idea, wondering what you think?

mdelapenya: Sounds reasonable to me :wink:

ClaytonNorthey92: awesome thank you!

mdelapenya: Hey, I added info about the process: dirContext, binary name and arguments. I preferred not to add environment to avoid disclosing sensible information.

Please take a look :)

ClaytonNorthey92: looks good thank you! in the CI/CD build it looks like the cmd is blank (an empty slice)...could that be what is causing the builds to fail?

mdelapenya: Yeah, but not sure how the CMD was lost. I'll try to debug it locally and see what happens, because that test was passing from the beginning

gianarb:
Hello, I do not have a lot to add here. I appreciate your work @mdelapenya! It is enormous, and the fact that you use it already is significant. You should secure the CI pass obviously, other than that @ClaytonNorthey92 made a good set of suggestions, the one I have left relates to docs and example. Do you mind to add code blocks, very carefully on `NewLocalDockerCompose` so the godoc will improve. If you can add a basic `Example` even from a Test, you already wrote, it would be great.

mdelapenya:
Sorry for the push-f my bad. I have an alias for rebasing old pulls and I'm used to using it instead of a merge commit. 🤦‍♂

mdelapenya:
Well, I'm able to reproduce locally travis errors using a local Travis worker (https://stackoverflow.com/questions/21053657/how-to-run-travis-ci-locally). I'll try to debug the build and check what happens here

mdelapenya:
I found the issue comes from the checks to verify environment variables are present/absent. Debugging...

mdelapenya:
Hey @gianarb, I'm really glad to say that I found the issue with the failing tests: it was related to a container name not found.

Now I can say it works and tests pass 👏 

I'd like you and the community (cc/ @ClaytonNorthey92) to receive feedback about this implementation, to see if it can be merged 🤞 

Thanks for the patience!

gianarb:
My plan is to look at this one this week @mdelapenya !

mdelapenya:
> My plan is to look at this one this week @mdelapenya !

That's great! I'm available for a call (slack/zoom/hangout) if you prefer

mdelapenya:
> > My plan is to look at this one this week @mdelapenya !
> 
> That's great! I'm available for a call (slack/zoom/hangout) if you prefer

Hey @gianarb, I'd like to see progress in this review, as it's open for a long while. Let me know if you need any clarification.

Thanks!

mdelapenya:
Hey @gianarb, I'd like to unblock this. Wdyt having a peer-review in a call?

gianarb:
Sure can you rebase in the meantime?!

mdelapenya:
Yes! Let me do it sometime today. There are things that I don't like (although they work) from this PR, like error handling and docker-compose logs, but I'm super happy to help/contribute

mdelapenya:
Hey @gianarb, I rebased the branch and updated the PR

PR #98, opened 2019-09-28
canned postgresql

giorgioazzinnaro:
Ciao @gianarb @alde @mraerino 

I took inspiration from #59 also to address #19 
I tried to implement some of the suggestions by @mraerino in #59 ..

Let me know if I should change anything else!

### DIFF
--- a/canned/postgresql.go
+++ b/canned/postgresql.go
@@ -22,6 +22,8 @@ const (
 	postgresPort       = "5432/tcp"
 )
 
+// PostgreSQLContainerRequest completes GenericContainerRequest
+// with PostgreSQL specific parameters
 type PostgreSQLContainerRequest struct {
 	testcontainers.GenericContainerRequest
 	User     string
@@ -29,7 +31,8 @@ type PostgreSQLContainerRequest struct {
 	Database string
 }
 
-type postgresqlContainer struct {
+// PostgreSQLContainer should always be created via NewPostgreSQLContainer
+type PostgreSQLContainer struct {
 	Container testcontainers.Container
 	db        *sql.DB
 	req       PostgreSQLContainerRequest
@@ -37,7 +40,7 @@ type postgresqlContainer struct {
 
 // GetDriver returns a sql.DB connecting to the previously started Postgres DB.
 // All the parameters are taken from the previous PostgreSQLContainerRequest
-func (c *postgresqlContainer) GetDriver(ctx context.Context) (*sql.DB, error) {
+func (c *PostgreSQLContainer) GetDriver(ctx context.Context) (*sql.DB, error) {
 
 	host, err := c.Container.Host(ctx)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


canned/postgresql.go
```diff
@@ -0,0 +1,147 @@
...
+		if err := c.Start(ctx); err != nil {
+			return postgresC, errors.Wrap(err, "failed to start container")
+		}
+
+		db, err := postgresC.GetDriver(ctx)
+		if err != nil {
+			return nil, err
+		}
+
+		for {
```
alde: I guess this is because the postgres container "bounces" and logs the "database system is ready to accept connections" twice?

In the java testcontainer that's handled by having the Wait code have a state and wait for it twice.

I think I prefer this, as it keeps the Waits simple

giorgioazzinnaro: Yes, I don't understand why the Postgres container is doing that, but basically the process gets restarted, so you'd have to wait for the second `"database system is ready to accept connections"`.

Because of that though I realised, the wait strategy I put there is basically pointless? :thinking: 

I have another proposal to make the same wait process into a proper `SQLStrategy` wait strategy.

wait/sql.go
```diff
@@ -0,0 +1,56 @@
...
+	if err != nil {
+		return errors.Wrap(err, "could not retrieve the SQL connector from the provided function")
+	}
+
+	db := sql.OpenDB(conn)
+
+	for {
+		_, err := db.ExecContext(ctx, "SELECT 1")
+		if err != nil {
+			time.Sleep(500 * time.Millisecond)
```
alde: Maybe there should be a max wait time, just in case?

giorgioazzinnaro: Whoops... you're right. I completely forgot.
I did the same as in `wait/log.go` now. Also added the fluent builders as in LogStrategy for consistency

gianarb:
Sweet! Do you mind to add a godoc example as well?! Thanks a lot

giorgioazzinnaro:
Yeah, I think that'd be useful! Will do over the weekend.

As for the wait strategy, do you guys think we should keep the SQL client or go back to LogStrategy with multiple occurrences?

gordonbondon:
@giorgioazzinnaro I've created a PR to your branch to add support for custom dockerfiles giorgioazzinnaro/testcontainers-go#1

giorgioazzinnaro:
I merged #105 #98 and #102 in the master of my fork to see how the godoc turns out, so here it is:
https://godoc.org/github.com/giorgioazzinnaro/testcontainers-go/canned

Any other suggestions on how to proceed with this?

nikolayk812:
It would be great to have this feature being merged and released

DarrienG:
This looks pretty good. What's blocking it from being merged? Glad to help if I can.

phss:
Hey folks, just wanted to check again how's the progress with this PR? If there's anything I can do to help, let me know!

codyaray:
Looks like a file named `wait/sql.go` has been merged to upstream, but doesn't match the expectations here.
```
./postgres_container.go:115:30: not enough arguments in call to wait.ForSQL
	have (func(context.Context, wait.StrategyTarget, <T>) (driver.Connector, error), map[string]interface {})
	want (nat.Port, string, func(nat.Port) string)
./postgres_container.go:136:93: undefined: wait.SQLVariables
```

giorgioazzinnaro:
I guess we need to include the maintainers again in the discussion, I think we were still stuck at understanding whether it's sensible to throw all the possible _canned_ containers in the root module, or rather create a go module for each of them (so that pulling the main module won't bring along a dependency on Redis, Postgres, MySQL, etc ... clients). Since we last discussed this the support for Go modules has greatly improved 

@gianarb @mdelapenya what's your opinion on this?

mdelapenya:
We started this PR, which could be the seed for modules: https://github.com/testcontainers/testcontainers-go/pull/618

It creates examples to be consumed by the website, but turning them into modules would be a matter of adjusting the templates.

mdelapenya:
Given #618 has been merged, we created https://github.com/testcontainers/testcontainers-go/issues/636, where you can find how to contribute new modules :)

PR #109, opened 2019-10-29
Retry on any http failure

gavinbunney:
When using this for a variety of tests, generally it's better to retry on _any_ http failure raised rather than just a subset. If the container has issues bootstrapping, this makes the library handle those cases better where the listener may be "active" but still in an unexpected state


gianarb:
Thanks a lot

PR #111, opened 2019-11-04
feature/container-ip

gavinbunney:
Expose the IP address of the container to allow clients to query the network directly

### DIFF
--- a/container.go
+++ b/container.go
@@ -38,6 +38,7 @@ type Container interface {
 	Logs(context.Context) (io.ReadCloser, error)                    // Get logs of the container
 	Name(context.Context) (string, error)                           // get container name
 	Networks(context.Context) ([]string, error)                     // get container networks
+	ContainerIP(context.Context) (string, error)                    // get container ip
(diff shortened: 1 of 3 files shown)


gianarb:
I am trying to think about a way to avoid to expose another function, in favor of something a bit more magic. Do we know how the java-lib does this?

gavinbunney:
One approach might be to expose the `inspectContainer` as a package exported function? Would mean it would give everyone access to everything about the container rather than adding new functions to expose different parts of the struct

cars10w:
Hi all,
any news on this topic?  I am running into the situation in a CI environment which is mentioned in the [testcontainers-java docs](https://www.testcontainers.org/features/networking/#getting-the-container-ip-address). Solvable by using IP addresses of the containers themselves and not of the Host.

As far as I can see they don't have any special magic for all those API endpoints.

Thanks and best regards!

gianarb:
Rebased https://github.com/testcontainers/testcontainers-go/pull/179

PR #113, opened 2019-11-14
#110 feature: Context as io.Reader

### LINKED ISSUE #110: Allow Passing in Context as io.ReadCloser
**Description**
As of now, when we allow a user to build and image from a Dockerfile, we only allow them to pass in the build context as a string, which is meant to be the path to the build context on their filesystem.  Then [we archive that directory before we send it to the Docker API, so it becomes a type that implements ReadCloser](https://github.com/testcontainers/testcontainers-go/blob/master/docker.go#L278)

I would like to allow the ability to use a [ReadCloser](https://golang.org/pkg/io/#ReadCloser) as the context in `FromDockerfile`.  This would allow consumers of our API to dynamically set files within the archive before sending it to the Docker API, and would be less implicit than what we do now.  However, I would like to also keep our current functionality because I like the abstraction.

**To Reproduce**
Not a bug. No reproduction steps.

**Expected behavior**
I am not sure what we want the API to look like, maybe:
```go
type FromDockerfile struct {
    Context string
    ContextArchive io.ReadCloser
    Dockerfile string
}
```
if the consumer sets both `Context` and `ContextArchive` we could return an error, specifying that you should only choose one or the other.

**Additional context**
this [conversation](https://testcontainers.slack.com/archives/CBWBNR76G/p1572263731000400) started this idea

ClaytonNorthey92:
resolves #110

Added a ContextArchive attribute to the FromDockerfile struct, this
allows a user to pass in an io.Reader as the Docker build context.
A user will now be allowed to create a dynamic build in code and send it
to the build.

Added test cases for both scenarios (using a local filepath and an
archive).  Added example to README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,8 @@ nginxC.Terminate(ctx, t)`. `t` is `*testing.T` and it is used to notify is the
 
 Testcontainers-go gives you the ability to build and image and run a container from a Dockerfile.
 
-You can do so by specifiying a `Context` and optionally a `Dockerfile` (defaults to "Dockerfile") like so:
+You can do so by specifiying a `Context` (the filepath to the build context on your local filesystem) 
+and optionally a `Dockerfile` (defaults to "Dockerfile") like so:
 
 ```
 req := ContainerRequest{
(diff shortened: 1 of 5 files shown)


gianarb:
Thanks a lot!

PR #115, opened 2019-11-20
Add more info about failing test on Travis

mdelapenya:
## What does this PR do?
It adds test info to Travis logs when running Go tests.

## Why is it important?
Travis logs are not good to debug/trace what happens when a test fails. Here we find a manner to check what is the status of each test method, with a summary at the end of the log (see screenshot)

Screenshot 2019-11-20 at 14 28 06

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,10 +10,13 @@ services:
 env:
   - GO111MODULE=on
 
+before_script:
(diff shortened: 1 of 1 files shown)


gianarb:
I have never used this candy before! Thanks

PR #116, opened 2019-11-21
Support volume mounts

uportalis:
### Context
I wanted to run an [ETCD](https://etcd.io/) cluster using testcontainers-go.
In order to simulate ETCD node crash and recovery, I needed my data to be kept after a terminate of the container so that the new container can use it.
To do this, I used a bind mount so that the data stays on my host.
However, at start-up, ETCD does some Fsync on the data files which, I don't know why, is not working on my environment (Docker for Windows 10).

### Solution
The alternative I found is to use a Docker volume and mount it into the ETCD container.

### DIFF
--- a/container.go
+++ b/container.go
@@ -67,7 +67,7 @@ type ContainerRequest struct {
 	Cmd            []string
 	Labels         map[string]string
 	BindMounts     map[string]string
-	VolumeMounts     map[string]string
+	VolumeMounts   map[string]string
 	RegistryCred   string
 	WaitingFor     wait.Strategy
 	Name           string              // for specifying container name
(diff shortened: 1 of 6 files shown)


container.go
```diff
@@ -67,6 +67,7 @@ type ContainerRequest struct {
 	Cmd            []string
 	Labels         map[string]string
 	BindMounts     map[string]string
+	VolumeMounts     map[string]string
```
mdelapenya: Source Format (SF) Your editor should update formatting on save, could you double check it does? 😉

uportalis: Correct. I forgot to format it.

uportalis:
Regarding tests, I was wondering how to test it correctly as I could not find an existing test case for bind mount that would serve as example to test volume mount.
But you are right, this is something that should be tested.

uportalis:
Test case (read on bind mount, write on volume mount) added.

gianarb:
Thanks a lot

PR #117, opened 2019-11-21
#98 canned postgres

nikolayk812:
- no postgres client exposed to API
- less code

### DIFF
--- a/canned/postgres/postgres.go
+++ b/canned/postgres/postgres.go
@@ -1,0 +1,107 @@
+package postgres
+
+import (
+	"context"
+	"fmt"
+	"github.com/docker/go-connections/nat"
+	"github.com/pkg/errors"
+	"github.com/testcontainers/testcontainers-go"
+	"github.com/testcontainers/testcontainers-go/wait"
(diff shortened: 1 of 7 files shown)


nikolayk812:
Defintely inspired by @giorgioazzinnaro

canned/postgres/postgres.go
```diff
@@ -0,0 +1,115 @@
...
+	"fmt"
+	"github.com/docker/go-connections/nat"
+	"github.com/pkg/errors"
+	"github.com/testcontainers/testcontainers-go"
+	"github.com/testcontainers/testcontainers-go/wait"
+)
+
+const (
+	image             = "postgres"
+	tag               = "9.6.15"
```
borgotta: Don't you think that it will be better if we use some fresher versions by default? Like 12.1-alpine, for example.

DerThorsten:
Will this be merged ? is there anything where I could help to get this merged?

Streppel:
I second @DerThorsten . I can help too if needed.

PR #119, opened 2019-11-27
allow alternative reaper docker registry

makasprzak:
In our project we are obliged to pull docker images through a corporate docker registry mirror. The problem is that the image for the resource reaper (ryuk) is right now hardcoded. Therefore I'm proposing this little patch to work this around. Please see if you can make any use of it. Thanks


ClaytonNorthey92:
instead of only allowing the specifying of the registry, should we just allow the user to specify the entire reaper image?  it looks like that is [what the java project does](https://www.testcontainers.org/features/configuration/#customizing-ryuk-resource-reaper)

makasprzak:
Fair point @ClaytonNorthey92 , I applied the change

ClaytonNorthey92:
thank you @makasprzak !

gianarb:
Thanks for coding this and for the reviews!! YOU are amazing!

PR #120, opened 2019-12-03
feat: allow viewing of logs from stopped container

### LINKED ISSUE #118: Feature: Allow Ability to View Logs from Failed-to-Start Container
**Please Note**
~~I am not 100% sure that this is possible, but it would be useful for debugging.~~
based on this thread, this is possible https://testcontainers.slack.com/archives/CBWBNR76G/p1574767579003000

**Describe the bug**
When starting a container, you have the option to "wait" for something to occur to determine the container has "started".  For example you can [wait.ForLog](https://github.com/testcontainers/testcontainers-go/blob/master/wait/log.go#L66).  A problem is if the container doesn't "start" because what you were waiting for didn't occur, you may want to print the logs to see why it didn't occur (maybe there was an error, maybe you spelled something wrong in your `wait.ForLog` 😬 ).  The currently functionality is: if the container fails to start, the `io.ReadCloser` that the `.Logs` function returns is `nil`.

I would like to propose ~~, if possible,~~ that we allow the reading of logs from a container that failed to start due to a context timeout.  The Java port just added functionality for [checking if the reason for failure was due to context timeout](https://github.com/testcontainers/testcontainers-java/pull/1837)

I vote we:
* check the failure to start reason
  * if it is due to a context timeout, we include the logs from that container

**To Reproduce**
```go
ctx := context.Background()

req := ContainerRequest{
	Image: "alpine",

	// my container request will log out something unexpected, but I
	// may still want to know what it logged
	Cmd:        []string{"echo", "this is unexpected!"},
	WaitingFor: wait.ForLog("I expect this!").WithStartupTimeout(10 * time.Second),
}
	ContainerRequest: req,
	Started:          true,
})

// if the container failed to start because of a timeout waiting for the expected logs,
// there may still be value in whatever was in the logs of the failed container
if err != nil && err.Error() == "failed to start container: context deadline exceeded" {

	l, logErr := c.Logs(ctx)
	if logErr != nil {
		t.Error(err)
	}

	logBuf := new([]byte)
	// this will error due to the l being nil due to the container failing to start
	_, logErr = l.Read(*logBuf)
	if logErr != nil {
		t.Error(err)
	}

	t.Errorf("error starting container because the deadline exceeded, but here are the useful logs: %s", logBuf)
}

if err != nil {
	t.Error(err)
}
```

**Expected behavior**
I would expect the error `"error starting container because the deadline exceeded, but here are the useful logs: this is unexpected!"`

**Additional Context**
https://testcontainers.slack.com/archives/CBWBNR76G/p1574767579003000

ClaytonNorthey92:
setting any container besides a reaper to not auto-remove when creating

fixes #118 

Now when creating containers, allowing a user to set the `AutoRemove` parameter in the container request, if this is done, then the container will be removed upon stopping.  This is by default set to `false` for user-created containers, but when running a reaper container, this will set it to `true` (reaper cleans up containers, but nothing cleans up reaper so we have to make sure it auto-removes)

### DIFF
--- a/container.go
+++ b/container.go
@@ -76,6 +76,7 @@ type ContainerRequest struct {
 	NetworkAliases map[string][]string // for specifying network aliases
 	SkipReaper     bool                // indicates whether we skip setting up a reaper for this
 	ReaperImage    string              // alternative reaper image
+	AutoRemove     bool                // if set to true, the container will be removed from the host when stopped
 }
 
 // ProviderType is an enum for the possible providers
--- a/container_test.go
+++ b/container_test.go
@@ -5,6 +5,8 @@ import (
 	"bytes"
 	"context"
 	"io"
+	"io/ioutil"
+	"strings"
 	"testing"
 	"time"
 
@@ -128,7 +130,7 @@ func Test_BuildImageWithContexts(t *testing.T) {
 					Contents string
 				}{
 					{
-						Name:     "Dockerfile",
+						Name: "Dockerfile",
 						Contents: `FROM alpine
 						CMD ["echo", "this is from the archive"]`,
 					},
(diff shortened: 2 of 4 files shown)


ClaytonNorthey92:
I am contemplating adding this to the README or not...the only time a user would ever want to change this is if they are skipping the reaper container, I know we provide that option, but does that happen often?

gianarb:
I think we can leave it out for now. At some point we will have to figure out about how to setup a proper doc I presume!

Issue #121, opened 2019-12-17
Feature: waiting strategy for SQL based database containers


erdemtoraman:
This waiting strategy shouldn't depend on any specific driver but only the std library sql package. It should block until the database container is ready to serve.

PR #122, opened 2019-12-17
Adding a wait for sql strategy

erdemtoraman:
#121

A simplified sql db waiting strategy without adding any new dependency to the library. An example usage for a postgres container:

```go

	var env = map[string]string{
		"POSTGRES_PASSWORD": "password",
		"POSTGRES_USER":     "postgres",
		"POSTGRES_DB":       "service",
	}

	var port = "5432/tcp"

	dbURL := func(port nat.Port) string {
		return fmt.Sprintf("postgres://postgres:password@localhost:%s/service?sslmode=disable", port.Port())
	}

	req := testcontainers.GenericContainerRequest{
			Image:        "postgres:latest",
			ExposedPorts: []string{port},
			Cmd:          []string{"postgres", "-c", "fsync=off"},
			Env:          env,
			WaitingFor:   wait.ForSQL(nat.Port(port), "postgres", dbURL).Timeout(time.Second*5),
		},
		Started: true,
	}
	container, err := testcontainers.GenericContainer(context.Background(), req)

	if err != nil {
		log.Fatal("start ", err)
	}
	mappedPort, err := container.MappedPort(context.Background(), nat.Port(port))
	if err != nil {
		log.Fatal(err)
	}
	defer container.Terminate(context.Background())

	log.Println("postgres container ready and running at port: ",mappedPort)
	time.Sleep(time.Minute)
	time.Sleep(time.Minute)
```

### DIFF
--- a/wait/sql.go
+++ b/wait/sql.go
@@ -1,0 +1,63 @@
+package wait
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+	"github.com/docker/go-connections/nat"
+	"time"
+)
+
+func ForSQL(port nat.Port, driver string, url func(nat.Port) string) *waitForSql {
+	return &waitForSql{
+		Port: port,
+		URL: url,
+		Driver: driver,
+	}
+}
+
+type waitForSql struct {
+	URL            func(port nat.Port) string
+	Driver         string
+	Port           nat.Port
+	startupTimeout time.Duration
+}
+
+func (w *waitForSql) Timeout(duration time.Duration) *waitForSql {
+	w.startupTimeout = duration
+	return w
+}
+
+func (w *waitForSql) WaitUntilReady(ctx context.Context, target StrategyTarget) (err error) {
+	if w.startupTimeout == 0 {
+		w.startupTimeout = time.Second*10
+	}
+	ctx, cancel := context.WithTimeout(ctx, w.startupTimeout)
+	defer cancel()
+
+	ticker := time.NewTicker(time.Millisecond * 100)
+	defer ticker.Stop()
+
+	port, err := target.MappedPort(ctx, w.Port)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


wait/sql.go
```diff
@@ -0,0 +1,66 @@
...
+import (
+	"context"
+	"database/sql"
+	"fmt"
+	"github.com/docker/go-connections/nat"
+	"time"
+)
+
+//ForSQL constructs a new waitForSql strategy for the given driver
+func ForSQL(port nat.Port, driver string, url func(nat.Port) string) *waitForSql {
```
mdelapenya: What do you think about renaming this to `ForDB` and `waitForDB`? At the end of the day, we are waiting for the database. Using SQL for it seems an implementation detail.

erdemtoraman: that's a good point but I assumed it'd be more consistent to point out "sql" because it can only handle sql based databases, if we just call it `forDB` the client of this function might not immediately understand it'd not work for say mongodb.

ClaytonNorthey92:
could we add tests for this? maybe at least one for every SQL database type we support?

could we also add some documentation for this in the README?

wait/sql.go
mdelapenya: I agree in the sense using SQL distinguish from any other storage type. My only concern was related to my mind directly connecting `waitForSQL` to `wait for the execution of an SQL sentence`... which on the other hand is true as we use `SELECT 1`.

So I'm good with this one :)

nikolayk812:
@gianarb what do you think about this PR?

gianarb:
I would lke to get it merged! @ClaytonNorthey92 @mdelapenya thumbs up and I will click the button!

gianarb:
Thanks a lot @erdemtoraman !!
I created an issue to get an Example func for this feature https://github.com/testcontainers/testcontainers-go/issues/142

If anybody is up for it

PR #131, opened 2020-01-12
Add ability to configure tmpfs for a container

rieske:
When testing against database containers, most of the time, those containers are thrown away after the tests are finished and we do not care about the data. Docker [provides](https://docs.docker.com/storage/tmpfs/) `tmpfs` mounts that work well for such use cases - the data is only written in memory and not to disk, which in turn allows for tests to run faster.

testcontainers-java [has](https://github.com/testcontainers/testcontainers-java/blob/3cb33792328e49b5ab8d78f747a256d65593dd64/core/src/main/java/org/testcontainers/containers/GenericContainer.java#L1449) an option to configure tmpfs and pass it to docker client too.

This change adds such option for the go library.

I do not know how to best demonstrate in a test that it makes an impact when used.
My local results on a linux computer for the added mysql test (just start a mysql container and wait for msql to be ready for use):
- `PASS: TestContainerWithTmpFs (9.69s)` - with tmpfs
- `PASS: TestContainerWithTmpFs (15.40s)` - when tmpfs is commented out

### DIFF
--- a/docker_test.go
+++ b/docker_test.go
@@ -946,17 +946,12 @@ func TestContainerCreationWithBindAndVolume(t *testing.T) {
 func TestContainerWithTmpFs(t *testing.T) {
 	ctx := context.Background()
 	req := ContainerRequest{
-		Image:        "mysql:8.0.18",
-		ExposedPorts: []string{"3306"},
-		Env: map[string]string{
-			"MYSQL_ROOT_PASSWORD": "password",
-			"MYSQL_DATABASE":      "database",
-		},
-		Tmpfs:      map[string]string{"/var/lib/mysql": "rw"},
-		WaitingFor: wait.ForLog("port: 3306  MySQL Community Server - GPL"),
+		Image: "busybox",
+		Cmd:   []string{"sleep", "10"},
+		Tmpfs: map[string]string{"/testtmpfs": "rw"},
 	}
 
-	mysql, err := GenericContainer(ctx, GenericContainerRequest{
+	container, err := GenericContainer(ctx, GenericContainerRequest{
 		ContainerRequest: req,
 		Started:          true,
 	})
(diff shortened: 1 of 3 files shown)


docker_test.go
```diff
@@ -941,3 +942,32 @@ func TestContainerCreationWithBindAndVolume(t *testing.T) {
 		}
 	}()
 }
+
+func TestContainerWithTmpFs(t *testing.T) {
```
ClaytonNorthey92: what exactly does this test? it looks like it just starts the mysql container then stops it...would it be a good idea to test:
* save something that would be stored in `/var/lib/mysql` in the mysql container
* stop the container
* start another mysql container, ensure that whatever was saved was not persisted (if I understand `tmpfs` correctly)

rieske: It starts a mysql container with its data volume mapped to a tmpfs volume - that includes a database creation that is part of the bootstrap. I mimicked my particular use case for this and the only observable outcome is that the test is executed faster when using tmpfs.

Probably a better test would be to use a minimal Linux container, no mysql, mount tmpfs volume and assert that data can be written to it. I think Java implementation tests it this way.
I'll try to improve this next week.

rieske: Updated.
Just like [here](https://github.com/testcontainers/testcontainers-java/blob/bdb82b76a86b57a9789731430cddf3858f41388a/core/src/test/java/org/testcontainers/junit/GenericContainerRuleTest.java#L142-L158), I try to create a file on tmpfs mount.

ClaytonNorthey92: hey @rieske thank you for getting back to me and working on this.

the more I think about it, we are basically just testing the [Docker client Tmpfs functionality](https://godoc.org/github.com/docker/docker/api/types/container#HostConfig) here, so maybe we shouldn't even have a test?  Sorry for going back and forth.

I think this is a good feature.

gianarb:
Sweet thanks! @ClaytonNorthey92 I like the idea to be defensive in terms of testing, even if almost everything is demanded to docker-sdk the tests that validate its behavior will help us in future version bump or similar procedures :+1: 

Thanks a lot, @rieske for writing it and @ClaytonNorthey92 for your always amazing review!

Issue #133, opened 2020-01-13
Allow Log Following/Consuming


ClaytonNorthey92:
**Describe the bug**
Not a bug, a feature request.
When running containers via testcontainers-go, in order to view the `stdout` from a container you have to call `.Logs(...)` on that container.  I think it would be useful allow an option to forward the `stdout` output from the containers to some sort of log consumer.  For example, you could send this to one that would log `stdout` of the container to `stdout` for the host machine.  This would help with debugging.

The Java implementation uses [follow logs](https://www.testcontainers.org/features/container_logs/#streaming-logs)

**To Reproduce**
Not bug bug, no reproduction steps.

**Expected behavior**
Allow an option to be passed into the `ContainerRequest` to forward `stdout` to a follower, maybe...

```go
type LogConsumer interface {
    Accept(string logMessage) error
}
```

```go
type ContainerRequest struct {
	...
        LogFollowers []LogConsumer
}
```

```go
// an example consumer
type StdoutLogConsumer struct {}

func (s *StdoutLogConsumer) Accept(string logMessage) error {
    fmt.Println(logMessage)
}
```
**Additional context**
This was inspired by [this comment](https://github.com/testcontainers/testcontainers-go/issues/118#issuecomment-567487947).

gianarb:
There is a followoutput method in the testcontainers library:

https://www.testcontainers.org/features/container_logs/

I think it will be good to have something like that. I am not sure if there are interfaces in the std library we can leverage, I am thinking about the `io.Reader` one for example https://golang.org/pkg/io/#Reader to stay a bit more generic and closer to what the community expects

ClaytonNorthey92:
The [Java implementation](https://www.testcontainers.org/features/container_logs/) allows you to consume `stdout`, `stdin`, or both.

I think the best way to achieve this would be to use the Docker API, I don't think we need another library to handle this.

The Docker API itself allows [a "follow" option](https://docs.docker.com/engine/api/v1.30/#operation/ContainerLogs)

This maps to a ["follow" parameter in the Go client when calling ContainerLogs()](https://godoc.org/github.com/docker/docker/api/types#ContainerLogsOptions)

[...]

PR #135, opened 2020-01-24
Feature #133 - Add a FollowOutput method to attach log consumers to

ClaytonNorthey92:
Allow a user to associate `LogConsumer`s to a running container with the `FollowOutput` method.

These each will accept a log type and a content string (just the contents for now, will need to add the type).  The log type is stdout, stderr etc.

Added a test to show how this works.  It runs a server in a docker container that will echo to stdout words that are passed in with the `echo` parameter in the `/ping` endpoint.


logconsumer.go
```diff
@@ -0,0 +1,36 @@
...
+		Content: content,
+		LogType: logType,
+	}
+}
+
+// LogConsumer represents any object that can
+// handle a Log, it is up to the LogConsumer instance
+// what to do with the log
+type LogConsumer interface {
+	Accept(string)
```
ClaytonNorthey92: this should accept a `Log`, rather than just a string.  Right now it just accepts the contents of the log.  But the `Log` object would allow us to view the type as well

docker.go
```diff
@@ -265,6 +285,64 @@ func (c *DockerContainer) Exec(ctx context.Context, cmd []string) (int, error) {
 	return exitCode, nil
 }
 
+func (d *DockerContainer) startLogProducer() {
+	d.stopProducer = make(chan bool)
+	go func() {
```
ClaytonNorthey92: I am ignoring errors that occur in this goroutine, I believe I would just want to log them out, and continue running the goroutine.  is there a better way to handle errors in this goroutine?

ClaytonNorthey92:
@mdelapenya I would like your feedback on this as well if you have time, it is aiming to solve #133

logconsumer_test.go
```diff
@@ -0,0 +1,95 @@
...
+	Msgs  []string
+	Await chan bool
+}
+
+func (g *GetLogConsumer) Accept(m string) {
+	g.Msgs = append(g.Msgs, m)
+	g.Await <- true
+}
+
+func Test_LogConsumerGetsCalled(t *testing.T) {
```
ClaytonNorthey92: Add an example to the README on how these `LogConsumer`s work

docker.go
```diff
@@ -144,11 +148,17 @@ func (c *DockerContainer) Start(ctx context.Context) error {
 		}
 	}
 
+	if c.skipLogProducer == false {
```
mdelapenya: Nit: short form?
```golang
if !c.skipLogProducer {
```

logconsumer_test.go
```diff
@@ -0,0 +1,95 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	ep, err := c.Endpoint(ctx, "http")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	g := GetLogConsumer{
```
mdelapenya: nit: rename struct to `TestLogConsumer`? It's used just here for the tests, right?
wdyt?

logconsumer_test.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	<-g.Await
+
+	if len(g.Msgs) > 0 && g.Msgs[0] == "ready\n" {
+		// hack to get rid of the server "ready message"
+		g.Msgs = g.Msgs[1:]
+	}
+
+	if diff := deep.Equal([]string{"hello\n"}, g.Msgs); diff != nil {
```
mdelapenya: I'd like to use more `assertions` than classic `if/else`, as the mental model for them represents the What we are testing in the test.

reaper.go
```diff
@@ -52,7 +52,8 @@ func NewReaper(ctx context.Context, sessionID string, provider ReaperProvider, r
 		BindMounts: map[string]string{
 			"/var/run/docker.sock": "/var/run/docker.sock",
 		},
-		AutoRemove: true,
+		AutoRemove:      true,
+		SkipLogProducer: true,
```
mdelapenya: What do you think about making `SkipLogProducer` true by default? Then existing client code won't have to change anything on their side.

logconsumer_test.go
ClaytonNorthey92: You are correct, `TestLogConsumer` would be a better name. I will update, thanks!

logconsumer_test.go
ClaytonNorthey92: cool, makes sense, I can update

reaper.go
ClaytonNorthey92: I like that idea.  If the user doesn't need it, it would save resources too (it wouldn't spin up goroutines that have to be managed if they aren't requested).  I will update 👍

docker.go
ClaytonNorthey92: oh yes, I will update 👍

ClaytonNorthey92:
I would like to add the following test before I merge:
- [ ] test for the correct `LogType`

ClaytonNorthey92:
@mdelapenya I updated the way I start and stop producers...I am now requiring the user to call `.StartLogProducer` and `.StopLogProducer` rather than use a flag.  This seems more explicit and clear as to what is going on for the user (developer). What do you think?

logconsumer.go
```diff
@@ -0,0 +1,21 @@
+package testcontainers
+
+// StdoutLog is the log type for STDOUT
+const StdoutLog = "STDOUT"
+
+const StderrLog = "STDERR"
+
+// Log represents a message that was created by a process
+// LogType is either "STDIN", "STDOUT", or "BOTH",
```
ClaytonNorthey92: update these docs, it is either `STDERR` or `STDOUT`

gianarb:
I like that! Well implemented and the interface is clean. I do not have a lot of addition from my side :+1:

docker.go
```diff
@@ -265,6 +278,79 @@ func (c *DockerContainer) Exec(ctx context.Context, cmd []string) (int, error) {
...
+					panic(fmt.Sprintf("received inavlid log type: %d", logType))
+				}
+
+				// a map of the log type --> int representation in the header, notice the first is blank, this is stdin, but the go docker client doesn't allow following that in logs
+				logTypes := []string{"", StdoutLog, StderrLog}
+
+				b := make([]byte, count)
+				_, err = r.Read(b)
+				if err != nil {
+					fmt.Fprintf(os.Stderr, "error occurred reading log with known length %s", err.Error())
```
gianarb: This looks like something that we can solve as soon as we will introduce a logger.
Should we place a `TODO` here about it?

ClaytonNorthey92: yes good idea @gianarb , I can add that here, is there any sort of format we should follow?  so we can search for all places where we should add a logger?

ex. `TODO: add-logger: <description>`

ClaytonNorthey92: that way, when we do add a logger, we can search for "TODO: add-logger:" and find all places to add it

gianarb: Yes!

ClaytonNorthey92:
hey @gianarb I addressed your feedback, this is good to merge whenever you're ready 👍

gianarb:
I would like to get an example as a compiled function, so it will show up in the godoc website, ca you add it please?

ClaytonNorthey92:
yes I will add that 👍

ClaytonNorthey92:
hey @gianarb I added the documentation to appear in the godoc like you suggested, could you take a look?

gianarb:
Thanks a lot!!

Issue #136, opened 2020-01-27
Log progress on containers being started/waited/etc


nikolayk812:
Log progress on containers being started/waited/etc

nikolayk812:
Well, looks like pulling of an image happens before logging that container is being started. Need to be fixed.

mdelapenya:
Hi @nikolayk812 I'm triaging this repository and found this outdated issue, sorry for the late response 😞.

Is it still of your interest adding logs to the pull-image step? If so, would you be willing to send a contribution?

Thanks for your time here

mdelapenya:
BTW, I noticed this was already added in #137. I'm closing this issue as resolved.

Thanks!!

PR #137, opened 2020-01-27
#136 log start-up progress

nikolayk812:
#136 log progress

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"log"
 	"net/url"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


gianarb:
Should we spend some time speaking a bit more about if we should support a proper logger, with maybe our interface? It can be very simple, but I kind of like the idea to leave to the user the ability to set a logger.

But it is also true that we are a library focused on testing. So my expectation is that the interface may be:

```go
type Logger interface {
    Log(args ...interface{})
}
```
to match https://golang.org/pkg/testing/#T.Log

mdelapenya:
Agreed for the Log, although I see it as a follow-up conversation 😊

gianarb:
Let's replace it later as @mdelapenya suggested! Thanks @nikolayk812 !!!

Issue #138, opened 2020-01-27
Network host option works not with expected behavior


Bablzz:
Hello everyone! 

Thanks for excellent library!

I've got stack an one problem. Maybe you can help me. Maybe I don't understand something

Thanks in advance!

PS. Library v.0.0.10

**Describe the bug**
Network host option works not with expected behavior

**To Reproduce**
```
func TestContainerWithHostNetworkOptions(t *testing.T) {
	ctx := context.Background()
	gcr := testcontainers.GenericContainerRequest{ContainerRequest: testcontainers.ContainerRequest{
		Image:    "nginx",
		Networks: []string{"host"},
	},
		Started: true,
	}

	nginxC, err := testcontainers.GenericContainer(ctx, gcr)
	if err != nil {
		fmt.Errorf("Error %s", err)
	}

	host, err := nginxC.Host(ctx)

	resp, err := http.Get("http://" + host + ":80")
	if err != nil {
		t.Error(err)
	}
	t.Log(resp)
}
```
=> **golang.org/x/sys/unix.ECONNREFUSED (111)**

**Expected behavior**
When I use `Networks: []string{"host"}` I expect the same behavior as `docker run --net host`

** docker info **
output of the command:

```
docker info
Client:
 Debug Mode: false

Server:
 Containers: 11
  Running: 2
  Paused: 0
  Stopped: 9
 Images: 12
 Server Version: 19.03.3
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: b34a5c8af56e510852c35414db4c1f4fa6172339
 runc version: 3e425f80a8c931f88e6d94a8c831b9d5aa481657
 init version: fec3683
 Security Options:
  apparmor
  seccomp
   Profile: default
 Kernel Version: 5.0.0-37-generic
 Operating System: Ubuntu 19.04
 OSType: linux
 Architecture: x86_64
 CPUs: 8
 Total Memory: 15.43GiB
 Name: 
 ID: 
 Docker Root Dir: /var/lib/docker
 Debug Mode: false
[...]

PR #140, opened 2020-01-31
#138 Add NetworkMode host 

Bablzz:
Hello everyone!

I debugged this issue and founded out that:
When I start container with network: []string{"host"} docker inspect <id> returns in HostConfig.Network = 'default'

Add property in ContainerRequest struct

### DIFF
--- a/container.go
+++ b/container.go
@@ -2,6 +2,7 @@ package testcontainers
 
 import (
 	"context"
+	"github.com/docker/docker/api/types/container"
 	"io"
 
 	"github.com/docker/docker/pkg/archive"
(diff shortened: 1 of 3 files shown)


gianarb:
Sounds good to me. Yes `-net host` works in a different way and I am glad you figure out how to make it work

Bablzz:
I got an error only in one case:
When use:
```
NetworkMode: "host",
Networks:    []string{"new-network"},
```

gianarb:
Thanks!! :tada:

Bablzz:
My pleasure!

PR #147, opened 2020-02-06
fix retry logic

### LINKED ISSUE #146: HostPortStrategy returns immediately without retrying
PR #144 introduced a bug that effectively breaks the retry mechanism of the HostPostStrategy. Any error is returned immediately, and the rest of the code is skipped entirely.

dmarkhas:
Fixes #146 as well as potential stack overflow issue caused by calling defer inside a loop


gianarb:
Good catch! Thanks a lot

dmarkhas:
@gianarb thanks for merging this! Keep up the great work!

Issue #150, opened 2020-02-11
WaitingFor returns "port not found" if NetworkMode == host


Bablzz:
**Describe the bug**
WaitingFor returns "port not found" if NetworkMode == host

**To Reproduce**
```
func TestContainerWithHostNetworkOptionsAndWaitStrategy(t *testing.T) {
	ctx := context.Background()
	gcr := GenericContainerRequest{ContainerRequest: ContainerRequest{
		Image:       "nginx",
		SkipReaper:  true,
		NetworkMode: "host",
		WaitingFor: wait.ForListeningPort("80/tcp"),
	},
		Started: true,
	}

	nginxC, err := GenericContainer(ctx, gcr)
	if err != nil {
		t.Fatal(err)
	}

	defer nginxC.Terminate(ctx)

	host, err := nginxC.Host(ctx)
	if err != nil {
		t.Errorf("Expected host %s. Got '%d'.", host, err)
	}

	_, err = http.Get("http://" + host + ":80")
	if err != nil {
		t.Errorf("Expected OK response. Got '%d'.", err)
	}
}
```

**Expected behavior**
Port found

**docker info**
```
Client:
 Debug Mode: false

Server:
 Containers: 2
  Running: 2
  Paused: 0
  Stopped: 18
 Images: 35
 Server Version: 19.03.3
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Plugins:
  Volume: local
  Network: bridge host overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: b34a5c8af56e510852c35414db4c1f4fa6172339
 runc version: 3e425f80a8c931f88e6d94a8c831b9d5aa481657
 init version: fec3683
 Security Options:
  apparmor
  seccomp
   Profile: default
 Kernel Version: 5.3.0-29-generic
 Operating System: Ubuntu 19.10
 OSType: linux
 Architecture: x86_64
 CPUs: 8
 Total Memory: 15.34GiB
 Name: 
 ID: 
 Docker Root Dir: /var/lib/docker
 Debug Mode: false
 Registry: https://index.docker.io/v1/
 Labels:
 Experimental: false
 Insecure Registries:
[...]

PR #151, opened 2020-02-11
#150 If networkMode host ForListeningPort does not work

Bablzz:
Hi everybody!

Maybe It is not an ideal solution.


ClaytonNorthey92:
could we add tests for [Endpoint](https://github.com/testcontainers/testcontainers-go/blob/master/docker.go#L57) and [PortEndpoint](https://github.com/testcontainers/testcontainers-go/blob/master/docker.go#L75) that behave with `networkMode="host"`?

Bablzz:
I think I can

ClaytonNorthey92:
@Bablzz what you added looks good 👍 thank you!

This looks good to me, I don't use `networkMode="host"` too often, so I am not sure of all of the side effects besides the port mappings.  But I think the test coverage is good, and the code is written well so that this change is very minimally invasive.

docker_test.go
```diff
@@ -136,6 +136,97 @@ func TestContainerWithNetworkModeAndNetworkTogether(t *testing.T) {
...
+
+func TestContainerWithHostNetworkAndPortEndpoint(t *testing.T) {
+	nginxPort := "80/tcp"
+	ctx := context.Background()
+	gcr := GenericContainerRequest{ContainerRequest: ContainerRequest{
+		Image:       "nginx",
+		SkipReaper:  true,
+		NetworkMode: "host",
+		WaitingFor:  wait.ForListeningPort(nat.Port(nginxPort)),
+	},
```
mdelapenya: I guess this format is not correct, could you fix that? 🙏

Bablzz: Only format?! It's not a problem to fix it. :ok_hand: 
This a strange that **go fmt** does not reformat a code.

Bablzz:
@ClaytonNorthey92 thanks a lot! I hope tests help us :smile:

gianarb:
Thanks a lot @Bablzz !

Bablzz:
you are welcome @gianarb !

Issue #153, opened 2020-02-14
Non-existent container image put test into an infinite loop


nikolayk812:
**Describe the bug**
Test does not fail but stuck in an infinite loop in case of starting a non-existent container

**To Reproduce**
Try to start a test with Postgres version 10.12-alpine started by TC lib. The test never fails but stuck.

**Expected behavior**
The test should wait after a specific timeout

nikolayk812:
How to reproduce: use non-existing image version, put breakpoint to github.com/cenkalti/backoff/retry.go L37 and another one to github.com/testcontainers/testcontainers-go/docker.go L407. It looks like the for-loop never return and it just keeps trying over and over again.

ClaytonNorthey92:
hey @nikolayk812 could you put your test code here? I specifically would like to see the values in `ContainerRequest`

gianarb:
@ClaytonNorthey92 @nikolayk812 
This may be related https://github.com/testcontainers/testcontainers-go/pull/155#pullrequestreview-359577805

erdemtoraman:
The issue was that the retry logic for pulling the image didn't take the context cancellation into account as it didn't use the backoff with context and the backoff's max waiting duration defaults to 15 minutes which gave an impression of getting stuck forever. Additionally there is a way to terminate the retry mechanism by wrapping the error as permanent which should be done if the server return 404 for the requested image

PR #155, opened 2020-02-16
Bugfix: Image not found error causes a very long and uncontrolled stall

erdemtoraman:
#153

1. Pass the context to backoff strategy so that if the ctx is cancelled it doesn't proceed
2.  If the error is a not found error, treat it as a permanent one so that it's not retried and immediately returned


go.mod
```diff
@@ -1,36 +1,17 @@
-module github.com/testcontainers/testcontainers-go
+module github.com/erdemtoraman/testcontainers-go
```
gianarb: Hello!
The change I see in go.mod does not look right :) can you double-check?

gianarb:
@erdemtoraman the difference in go.mod go.sum does not look right, you didn't add any dependency in your code, so my expectation is for those to be untouched. I do not know why they are differnet

erdemtoraman:
> @erdemtoraman the difference in go.mod go.sum does not look right, you didn't add any dependency in your code, so my expectation is for those to be untouched. I do not know why they are differnet

Yeah you're right, I have no idea what might have been there, i'll check out

gianarb:
I am sorry if I look to picky @erdemtoraman but if you look at the commits you pushed on this branch it is clear that something is not right

https://github.com/testcontainers/testcontainers-go/pull/155/commits

You probably started your branch from the wrong point.
Do you mind to figure out why it looks like there is a merge commit https://github.com/testcontainers/testcontainers-go/pull/155/commits/ea1f33afc2be544b267dfaf5c0aff44ef264abc7 and things like that?
Take your time and if you need some help with cherry-picking commits I can help

Bablzz:
It seems to me, in go.sum remove new line (maybe IDE delete new line as default)

ClaytonNorthey92:
just one suggestion, could we squash this into two commits?
* travis version update
* the backoff fix

gianarb:
I cherry-picked a bunch of your commits into this PR https://github.com/testcontainers/testcontainers-go/pull/162

Thanks!

erdemtoraman:
> I cherry-picked a bunch of your commits into this PR #162
> 
> Thanks!

Yeah sorry for being so slow with this, thanks!

PR #164, opened 2020-03-06
fix: delete docker network when skipped reaper

zjhmale:
Originally, `Network#Remove` method will do nothing if we skipped a reaper for it already.


gianarb:
Hello!
My question is: if you choose to skip the reaper (that is the actor in the power of cleaning up), why should the network disappear?

:confused:

mdelapenya:
I think https://github.com/testcontainers/testcontainers-go/pull/320 overlaps with this PR. I'm closing it but please reopen if you consider anything is not addressed yet

Thanks!

PR #166, opened 2020-03-06
Add host for sql waiter URL function.

a-urth:
Issue:
In case container is started in docker which is not in localhost (ci environment with tests already running in container), mapped port is not enough to establish connection to a database.

Solution:
Add host parameter for url function, which is part of wait.ForSQL strategy, so it will be possible to construct connection url with docker host and mapped port.


gianarb:
This is a BC break but I actually like it. @mdelapenya what do yo think?

a-urth:
Hey @gianarb is there any progress for this PR? Is it going to be merged?

gianarb:
this is way easier to merge if we can figure out a way to avoid a bc-break. I think we can using the a with function. Like a `WithHost(string)`. wdyt @a-urth ?

.gitignore
```diff
@@ -1,2 +1,3 @@
+.idea
```
gianarb, line 1: Can you revert this? it should be placed in your global gitignore

mdelapenya:
> this is way easier to merge if we can figure out a way to avoid a bc-break. I think we can using the a with function. Like a `WithHost(string)`. wdyt @a-urth ?

Yes, I like the idea of not forcing a breaking change too

👍

mdelapenya:
I'd waitForMerge("#214") `</joke>` before merging this.

wdyt?

michielboekhoff:
Hey guys, sorry to bring this back from the dead, but I was running into a similar situation. I'd love to get this merged so I can avoid having to revert to what testcontainers-java is doing (waiting for log messages). 

If we can change this PR to preserve backwards compatibility, can we merge? I'm willing to invest some time into this unless @a-urth wants to apply the suggested changes?

michielboekhoff:
I just took a little stab at a backward-compatible way in #310. The reason I didn't just a `With` method is because the URL function itself is also exposed to the calling code. Changing the way it constructs the URL (i.e. changing the URL function to be an interface) might break BC.

mdelapenya:
I'm closing this PR, as it's already been implemented in #524 

Thanks!

PR #171, opened 2020-03-30
docker:getGatewayIp: rewrite with netlink?

GregoryVds:
Hi,

Thanks for the great library, I am using this in many projects, working great :-) 

I am running this lib from within a Container and I keep bumping into the same issue over and over again: I start with a fresh container, without iproutes2, get the `Failed to parse default gateway IP` error, dig into the code, and then I remember that it depends on it :-)

Have you considered replacing this by a netlink call (linux only)? Or parsing /proc/net/route?
I think it would be much cleaner not to rely on external tools such as iproute & awk.

WDYT? I would be happy to make the patch.

Signed-off-by: Gregory Vander Schueren <gregory.vanderschueren@tessares.net>

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -731,6 +731,7 @@ func inAContainer() bool {
 
 func getGatewayIp() (string, error) {
 	// see https://github.com/testcontainers/testcontainers-java/blob/3ad8d80e2484864e554744a4800a81f6b7982168/core/src/main/java/org/testcontainers/dockerclient/DockerClientConfigUtils.java#L27
+	// Replace this with netlink (linux only)? Or something like https://github.com/jackpal/gateway?
 	cmd := exec.Command("sh", "-c", "ip route|awk '/default/ { print $3 }'")
 	stdout, err := cmd.Output()
 	if err != nil {


gianarb:
Hello @GregoryVds ! 

This is a nice issue and thank you for reporting your problem! Currently we use what the Docker SDK provides:

```
func (p *DockerProvider) daemonHost() (string, error) {
	if p.hostCache != "" {
		return p.hostCache, nil
	}

	host, exists := os.LookupEnv("TC_HOST")
	if exists {
		p.hostCache = host
		return p.hostCache, nil
	}

	// infer from Docker host
	url, err := url.Parse(p.client.DaemonHost())
	if err != nil {
		return "", err
	}

	switch url.Scheme {
	case "http", "https", "tcp":
		p.hostCache = url.Hostname()
	case "unix", "npipe":
		if inAContainer() {
			ip, err := getGatewayIp()
			if err != nil {
				return "", err
			}
			p.hostCache = ip
		} else {
			p.hostCache = "localhost"
		}
	default:
		return "", errors.New("Could not determine host through env or docker host")
	}

	return p.hostCache, nil
}
```

I think it is in line with the library to se what provided by docker, but as you can see there is an escape root, you can set an evar `TC_HOST`.

Can you try that solutin first?
Thanks

GregoryVds:
Hi @gianarb 

Thanks for the quick reply, and apologies for my very late answer :-) 
I didn't know about this `TC_HOST` , that's good to know!

Adding `iproutes2` in my container is not an issue for me, so I am not really blocked by this.
I added that package, and now I see that I could alternatively set `TC_HOST` like you suggest.

However I would expect the default to work out of the box, without having to discover that I must either add `iproutes2`, either set the `TC_HOST` var. WDYT? At the moment the library does not work out-of-the box in such a setup.

gianarb:
yep, I understand your concern and I would like to figure out a good way to make this to work for as more use case as possible, at the moment I do not have a lot of time to figure out an easy to maintain solution so I would like to leave this issue open until I can wrap my head around it

namco1992:
Hello @gianarb @GregoryVds, I just saw this PR after I made mine here https://github.com/testcontainers/testcontainers-go/pull/237. I made changes based on the `testcontainers-java` that relies on the docker network inspect to try to get the gateway IP first. Would you mind checking it out and see if it solves the issue?

PR #175, opened 2020-04-05
Force testcontainers to always pull images when creating container

ikolomiyets:
It is sometimes useful when running integration test scenarios to force Testcontainers Go to always pull the image. For instance, when test cases are relying on the image with the latest tag and latest tag is always moved when the new image is pushed to the registry.

This PR adds a property to the ContainerRequest structure that instructs CreateContainer to always pull the image even if it already exists when the property in question is set to true. Otherwise, behaviour does not change.


gianarb:
Thank you for your contribution!

PR #179, opened 2020-04-14
Added ContainerIP to retrieve ip of container from primary network

gianarb:


gianarb:
Thanks @gavinbunney !! Sorry it took me so long!

Issue #152, opened 2020-02-13
golang.org/x/sys/windows".SECURITY_DESCRIPTOR


Ya3Sh:
**Describe the bug**
Project fails to build on windows when I depend on testcontainers-go v0.2.0 with below error:
..\..\..\..\pkg\mod\github.com\docker\docker@v0.7.3-0.20190506211059-b20a14b54661\pkg\system\filesys_windows.go:111:24: cannot use uintptr(unsafe.Pointer(&sd[0])) (type uintptr) as type *"golang.org/x/sys/windows".SECURITY_DESCRIPTOR in assignment

**To Reproduce**
Can you write a unit test to reproduce your issue?

**Expected behavior**

**Additional context**
Tried replacing the docker/docker dependency with different versions. No luck.

Ya3Sh:
Possibly dupe of https://github.com/testcontainers/testcontainers-go/issues/143

Ya3Sh:
#143 resolution didn't work for me.

ImVexed:
Fixed for me by pulling the latest version of docker `go get github.com/docker/docker@aa6a9891b09cce3d9004121294301a30d45d998d`
This seems to stem from the testcontainers using an old version of docker.

tomqwpl:
Also seeing this. Will try and work out how to use the PR

mdelapenya:
Seeing this behaviour here: https://github.com/testcontainers/testcontainers-go/runs/1384925491?check_suite_focus=true

mdelapenya:
I've sent a PR here #235 running the tests in multiple OS: windows and linux (unfortunately, GH does not support Docker on mac)

At this moment, the error described in this issue is fixed

mdelapenya:
Since we do run `go build` in Linux, Windows and Mac, I'd say this issue is no longer reproducible at least since [`v0.10.0`](https://github.com/testcontainers/testcontainers-go/releases/tag/v0.10.0).

Could you please double check with the latest version of tc-go? If still seeing the issue, please reopen so we can dig into the root cause. Thanks in advance!!

Issue #143, opened 2020-02-04
Fails to build on windows after 0.1.0


mrg0lden:
**Describe the bug**
After last release, I started getting this error
```
..\..\go\pkg\mod\github.com\docker\docker@v0.7.3-0.20190506211059-b20a14b54661\pkg\system\filesys_windows.go:111:24: cannot use uintptr(unsafe.Pointer(&sd[0])) (type uintptr) as type *"golang.org/x/sys/windows".SECURITY_DESCRIPTOR in assignment
```

**To Reproduce**
Just use last version `0.1.0` on windows

```
$ docker info
Client:
 Debug Mode: false

Server:
 Containers: 8
  Running: 1
  Paused: 0
  Stopped: 7
 Images: 9
 Server Version: 19.03.5
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: b34a5c8af56e510852c35414db4c1f4fa6172339
 runc version: 3e425f80a8c931f88e6d94a8c831b9d5aa481657
 init version: fec3683
 Security Options:
  seccomp
   Profile: default
 Kernel Version: 4.19.76-linuxkit
 Operating System: Docker Desktop
 OSType: linux
 Architecture: x86_64
 CPUs: 2
 Total Memory: 1.943GiB
 Name: docker-desktop
 ID: TFV5:ZNWP:MAWZ:TZSM:VZNB:DTLZ:PYHL:ADND:4ZIS:YDWB:76H5:MCRM
 Docker Root Dir: /var/lib/docker
 Debug Mode: true
  File Descriptors: 40
  Goroutines: 55
  System Time: 2020-02-04T14:47:00.406826238Z
  EventsListeners: 3
 Registry: https://index.docker.io/v1/
 Labels:
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false
 Product License: Community Engine
```

**Additional context**
Windows 10 1909

mrg0lden:
Managed to solve it by using master version 
`go get github.com\docker\docker@master`

[...]

PR #182, opened 2020-04-16
Update docker version to v19.03.8

ImVexed:
Resolves https://github.com/testcontainers/testcontainers-go/issues/152 & https://github.com/testcontainers/testcontainers-go/issues/143

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/cenkalti/backoff v2.2.1+incompatible
 	github.com/containerd/continuity v0.0.0-20190426062206-aaeac12a7ffc // indirect
 	github.com/docker/distribution v2.7.1-0.20190205005809-0d3efadf0154+incompatible // indirect
-	github.com/docker/docker v17.12.0-ce-rc1.0.20200309214505-aa6a9891b09c+incompatible
+	github.com/docker/docker v17.12.0-ce-rc1.0.20200416190941-2200d938a2d5+incompatible
 	github.com/docker/go-connections v0.4.0
 	github.com/docker/go-units v0.3.3 // indirect
 	github.com/gin-gonic/gin v1.6.2
--- a/go.mod
+++ b/go.mod
@@ -2,6 +2,8 @@ module github.com/testcontainers/testcontainers-go
 
 go 1.13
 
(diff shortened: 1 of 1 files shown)


gianarb:
I think I will never get how gomod works.

https://github.com/testcontainers/testcontainers-go/blob/master/go.mod#L37

This line is supposed to replace the use of docker/docker with docker/engine. So should we remove that as well?

ImVexed:
I also do not understand how gomod works.

Also I thought 19.03.8 worked for me but upon a `go clean --modcache` and deletion of my `go.sum` it broke again until I used the most recent commit off of master. I have no idea why it works.

gianarb:
I think go mod does not manage transient dependencies, so the `replace` will fix the problem you have but it is applied to your application, not to testcontainers. Testcontainer itself as you see from our CI works with the version we pin.

You `replace` docker in your project, and that makes everything to work for your app as you wrote, but from  testcontainer standpoint, unfortunately, we can't fix that

tomqwpl:
Since I'm also having this problem, can someone explain the fix? Can I resolve this locally?
I cloned ImVexed's repository and have used it locally with success (I had to clone so that I could change the package name in go.mod, otherwise go then complained about it not matching the repository name)

gianarb:
This explains what I am trying to say I think: https://github.com/golang/go/wiki/Modules#when-should-i-use-the-replace-directive

> replace directives provide additional control in the top-level go.mod for what is actually used to satisfy a dependency found in the Go source or go.mod files, while replace directives in modules other than the main module are ignored when building the main module.

`replace` directive placed in a library like this one (used by your application) is not used from your main application.

So you can apply the `replace` you need directly in your app. Because adding a `replace` here won't fix the problem you have in your app.

At least this is what I got until now :disappointed:

tomqwpl:
Ah, OK. That's worked for me:

    replace github.com/docker/docker => github.com/docker/docker v17.12.0-ce-rc1.0.20200416190941-2200d938a2d5+incompatible

I don't get why I have to do this though. Are you saying it's not a bug?

gianarb:
No this is how Go designed the `gomod` dependency manager. Have a look at the link I posted it contains a lot of hidden gems about `gomod` itself.

gianarb:
As I think it is clear at this point there is not much we can do about this PR.
I am closing for now. Feel free to open it again or ping me here

Issue #183, opened 2020-04-20
Why port not found?


softshipper:
Hi all

I am trying to use https://github.com/testcontainers/testcontainers-go as follows: 

    func TestGitAutoTag(t *testing.T) {
    
    	path, err := os.Getwd()
    	assert.NoError(t, err, err)
    
    	fmt.Println(path)
    	data := fmt.Sprintf("%s%s", path, "/gogs/data")
    
    	fmt.Println(data)
    
    	//Start gogs container
    	ctx := context.Background()
    	req := testcontainer.ContainerRequest{
    		Image:        "gogs/gogs:0.11.91",
    		ExposedPorts: []string{"3000/tcp"},
    		WaitingFor:   wait.ForHTTP("/"),
    		BindMounts:   map[string]string{data: "/data"},
    		Name:         "gogs",
    	}
    
    	gogs, err := testcontainer.GenericContainer(ctx, testcontainer.GenericContainerRequest{
    		ContainerRequest: req,
    		Started:          true,
    	})
    
    	assert.NoError(t, err, err)
    
    
    	defer gogs.Terminate(ctx)
    
    }

and it shows: 

    TestGitAutoTag: auto_tagger_test.go:38: 
        	Error Trace:	auto_tagger_test.go:38
        	Error:      	Received unexpected error:
        	            	port not found
        	            	github.com/testcontainers/testcontainers-go.(*DockerContainer).MappedPort  

It should be alias to:

    docker run --rm -it -p 3000:3000 --name=gogs --mount type=bind,source=`pwd`/gogs/data,target=/data gogs/gogs

What am I doing wrong? 

Thanks

gianarb:
Hello @softshipper 
thank you for reporting this bug. I am able to reproduce your issue and I think it is related to the `wait.ForHTTP` strategy.

In the meantime can you replace that with `WaitingFor:   wait.ForLog("Listen: http://0.0.0.0:3000")`, it should give you the same effect.

I have a few things to ask you for, your operating system and version. And information about the version of the library you are using

softshipper:
It takes a long time(61.56s):

```
2020/04/20 14:41:45 Starting container id: 7579b1bb43b0 image: quay.io/testcontainers/ryuk:0.2.3
2020/04/20 14:41:46 Waiting for container id 7579b1bb43b0 image: quay.io/testcontainers/ryuk:0.2.3
[...]

PR #184, opened 2020-04-20
docs: add example for wait http

gianarb:
Based on the issue #183 I decided to add this example to explain that by
default wait.HTTP looks for port `80/tcp` but you can override it.


gianarb:
Thanks @mdelapenya for your review!

PR #191, opened 2020-05-08
Bootstrap documentation with mkdocs

gianarb:
I do not have any preference about static side generators and things
like that. But I would like to stay as close as possible with what Java
does.

Ideally my plan is to get this deployed as `go.testcontainers.org` at
some point. I think I will deploy it as github action or something like
that.

- [x] Setup ci/cd
- [ ] Write the actual doc :D


rnorth:
Let me know when you're ready and I'll hook you up with a subdomain 😄

gianarb:
Gonna merge this for now.
@rnorth if you can do the CNAME it will be great

PR #213, opened 2020-06-16
use nginx when possible when writing test

gianarb:
We have to do our best to keep the number of images overall the test
suite as small as possible, to speed up the testsuite itself.


gianarb:
The Parallel test is not passing. So I am mering it right now

PR #214, opened 2020-06-16
test for sql wait strategy

gianarb:
The wait.ForSQL wait strategy does not have a testcase and I forgot
about how to use it as well.

I decided to write it in preparation for the work/discussion ongoing
here: #166


wait/sql.go
```diff
@@ -27,18 +30,24 @@ type waitForSql struct {
...
+}
+
 //WaitUntilReady repeatedly tries to run "SELECT 1" query on the given port using sql and driver.
 // If the it doesn't succeed until the timeout value which defaults to 60 seconds, it will return an error
 func (w *waitForSql) WaitUntilReady(ctx context.Context, target StrategyTarget) (err error) {
 	ctx, cancel := context.WithTimeout(ctx, w.startupTimeout)
 	defer cancel()
 
-	ticker := time.NewTicker(time.Millisecond * 100)
+	ticker := time.NewTicker(time.Second * 1)
```
mdelapenya, line 50: The `* 1` is not needed

wait/sql_test.go
```diff
@@ -0,0 +1,34 @@
...
+	"context"
+	"testing"
+
+	"github.com/docker/go-connections/nat"
+	_ "github.com/go-sql-driver/mysql"
+	"github.com/testcontainers/testcontainers-go"
+	"github.com/testcontainers/testcontainers-go/wait"
+)
+
+func Test_ForSql(t *testing.T) {
```
mdelapenya, line 13: NIT:
```suggestion
func Test_ForSQL(t *testing.T) {
```

mdelapenya (dismissed):
LGTM!

wait/sql.go
```diff
@@ -88,6 +89,11 @@ func (w *waitForSql) WaitUntilReady(ctx context.Context, target StrategyTarget)
 		return fmt.Errorf("sql.Open: %v", err)
 	}
 	defer db.Close()
+
+  db.SetConnMaxLifetime(0)
```
mdelapenya, line 93:
```suggestion
 	db.SetConnMaxLifetime(0)
```

wait/sql.go
```diff
@@ -17,6 +17,7 @@ func ForSQL(port nat.Port, driver string, url func(string, nat.Port) string) *wa
 		Port:           port,
 		URL:            url,
 		Driver:         driver,
+    // Not using the default duration here because it is too low. It will never work
```
mdelapenya, line 20:
```suggestion
		// Not using the default duration here because it is too low. It will never work
```

wait/sql.go
```diff
@@ -17,6 +17,7 @@ func ForSQL(port nat.Port, driver string, url func(string, nat.Port) string) *wa
 		Port:           port,
 		URL:            url,
 		Driver:         driver,
+		// Not using the default duration here because it is too low. It will never work
 		startupTimeout: defaultStartupTimeout(),
```
mdelapenya, line 21:
```suggestion
		startupTimeout: 20 * time.Second,
```

wait/sql.go
```diff
@@ -17,7 +17,8 @@ func ForSQL(port nat.Port, driver string, url func(string, nat.Port) string) *wa
 		Port:           port,
 		URL:            url,
 		Driver:         driver,
-		startupTimeout: defaultStartupTimeout(),
+		// Not using the default duration here because it is too low. It will never work
+		startupTimeout: 20 * time.Second,
 		PollInterval:   defaultPollInterval(),
```
mdelapenya, line 22:
```suggestion
		PollInterval:   time.Second * 1,
```

wait/sql.go
```diff
@@ -88,6 +89,11 @@ func (w *waitForSql) WaitUntilReady(ctx context.Context, target StrategyTarget)
 		return fmt.Errorf("sql.Open: %v", err)
 	}
 	defer db.Close()
+
+ 	db.SetConnMaxLifetime(0)
+	db.SetMaxIdleConns(3)
+	db.SetMaxOpenConns(3)
```
mdelapenya, line 95: Although the code LGTM, I'm a bit reluctant to hardcode this value here without giving the user the ability to set them in the wait strategy, according to their own requirements for the database under test. If we add them as part of the SQL strategy, then I think it's fine.

mdelapenya: Thinking about this twice, these settings should be added in a separate PR. Will send a commit on top removing them, and opening a discussion to make this configurable

PR #215, opened 2020-06-16
wait strategy where not properly set

gianarb:
For a couple of tests the wait strategy where not properly set.
I am trying to figure out if it makes the test suite to pass


gianarb:
🥳

PR #217, opened 2020-06-16
chore: simplify complex docker-compose file

mdelapenya:
## What does this PR do?
It uses Nginx instead of portal, reducing the size from < 2Gb to tens of Mb

## Why is it important?
Both CI builds and local execution of tests will be faster

## Related issues
- Relates #213


gianarb:
Awesome thanks!

PR #221, opened 2020-06-24
refector wait.HTTPStrategy

islishude:
- add method and body for request
- add tls.Config for https
- use time.After instead of time.Sleep


mdelapenya:
Hey @islishude, thank you for this nice contribution! Do you think it would be possible to add a test for the new behavior, specially the TLSConfig part?

Thanks!

islishude:
OK,I will add test cases for it

islishude:
Hi,the tests pass,please review @mdelapenya

wait/http_test.go
```diff
@@ -25,7 +36,88 @@ func ExampleHTTPStrategy() {
...
+			ExpectContinueTimeout: 1 * time.Second,
+		},
+	}
+	resp, err := client.Get(fmt.Sprintf("https://%s:%s", host, port.Port()))
+	if err != nil {
+		t.Error(err)
+		return
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
```
mdelapenya, line 119: What do you think about adding an assertion for the `pong` response? We'd like to check that the response matcher works as intended.

islishude: this variable indicates a state code, and I consider the value of `pong` to be a success state.

in practice, the server can return a response body with a state code representing the current server state,so we can know if the server is ready based on what is specifically returned.

mdelapenya: :) I meant about adding another assertion for the expected response (pong), which is part of the waitFor condition predefined when setting up the test in the response matcher. Of course this is not a requirement by the way, so I'm not opposing to it.

In the same hand, the PR as it's now could be merged, so I want to thank you for your contribution

mdelapenya (approved):

PR #222, opened 2020-06-24
github.com/docker/distribution should be an indirect dependency

islishude:
github.com/docker/distribution should be an indirect dependency.

run by `go mod tidy`


gianarb:
Thanks a lot

PR #223, opened 2020-06-28
Start the ryuk container in privileged mode

ThomasObenaus:
With this tweak it is possible to use the testcontainers-go library in a CI pipeline that is encapsulated in a docker container.
For example I'm using drone ci and a jenkins where in both cases the actual pipeline run is done inside a docker container.

These are running in privileged mode already in order to have the needed permission to mount the docker socket inside the container.
But the same has to be done then when the ryuk container is started. 
If the ryuk container is not started privileged then the start fails with e dial tcp/io timeout while waiting for the ryuk to be up (cause it never started due to missing permissions).

@gianarb: please have a look if this makes sense for you.


gianarb:
This is for sure something we have to do.
testcontainers-java has its own way to inject those configuration
https://www.testcontainers.org/features/configuration/

We do not yet. I don't think we should run `privileged` by default, we should use an envvar has Java does (the same one). The condition for `strings.Containers` is not right because technically you can override the `ReaperImage`

mdelapenya:
Closing, as #449 was already merged, superseding this one.

Thanks!

PR #229, opened 2020-07-14
update docker/docker package version

snowhork:
This PR is for updating docker/docker package in go.mod.

`v0.7.3` is too old, so some secury check fails. (In our team, we use yamory (https://yamory.io/) (japanese))

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/cenkalti/backoff v2.2.1+incompatible
 	github.com/containerd/continuity v0.0.0-20190426062206-aaeac12a7ffc // indirect
 	github.com/docker/distribution v2.7.1-0.20190205005809-0d3efadf0154+incompatible // indirect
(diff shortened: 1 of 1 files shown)


snowhork:
please review @gianarb @alde @mraerino

snowhork:
any updates? @gianarb

kernle32dll:
Actually, this was a massive downgrade!

Take a closer look at the version number as it was: `v0.7.3-0.20190506211059-b20a14b54661`. The referenced version was actually commit hash `b20a14b54661` (https://github.com/moby/moby/commit/b20a14b54661), not `0.7.3`.

The reason for this cryptic version number is the way go modules handle major versions. In short - docker should be importable via e.g. `github.com/moby/moby/13` - **but its not**

So - this PR actually downgraded to `v1.13.1` (https://github.com/moby/moby/releases/tag/v1.13.1), which is a **considerable** downgrade.

TL;DR; - this should be reverted **ASAP**!

Side note: This indirectly causes https://github.com/testcontainers/testcontainers-go/issues/238

PR #237, opened 2020-09-07
Use NetworkInspect to get the gateway's IP and fallback to getDefaultGatewayIP when fails

namco1992:
The `getGatewayIp` originally refers to testcontainers-java [here](https://github.com/testcontainers/testcontainers-java/blob/master/core/src/main/java/org/testcontainers/dockerclient/DockerClientConfigUtils.java#L46) which is deprecated. The new flow is described [here](https://github.com/testcontainers/testcontainers-java/blob/master/core/src/main/java/org/testcontainers/dockerclient/DockerClientProviderStrategy.java#L257): Network inspect for gateway IP -> get default gateway IP -> "localhost".

This PR implemented the new approach and marked the `getDefaultGatewayIP` as deprecated to be consistent with testcontainers-java.


mdelapenya (approved):
Thank you for this contribution! LGTM!

Issue #243, opened 2020-09-17
Unable to execute tests when they are running in Jenkins with Kubernetes plugin when Pods are running in EKS


ikolomiyets:
**Describe the bug**
The pipeline is running in Jenkins instance with Kubernetes plugin. Both Jenkins and Kube plugin pods are running in the EKS.
Tests are failing when trying to create Reaper container, i.e. it times out when waiting for the Reaper to come alive.

**To Reproduce**
Set up Jenkins server with a Kubernetes cloud defined which connectes to a EKS instance.
A test scenario uses Testcontainers to create a container.
A test step just run `go test`.
The following error appears in the output:
```
$ go test
go: downloading github.com/testcontainers/testcontainers-go v0.8.0
2020/09/17 15:50:09 Starting container id: d3eb2438b8d6 image: quay.io/testcontainers/ryuk:0.2.3
2020/09/17 15:50:10 Waiting for container id d3eb2438b8d6 image: quay.io/testcontainers/ryuk:0.2.3
test: 2020/09/17 15:50:10 failed to create container: creating reaper failed: could not start container: port not found
exit status 1
```

**Expected behavior**
Test succeeds

** docker info **
```
$ docker info
Client:
Debug Mode: false

Server:
Containers: 52
  Running: 41
  Paused: 0
  Stopped: 11
Images: 59
Server Version: 19.03.6-ce
Storage Driver: overlay2
  Backing Filesystem: xfs
  Supports d_type: true
  Native Overlay Diff: true
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
Swarm: inactive
Runtimes: runc
Default Runtime: runc
Init Binary: docker-init
containerd version: ff48f57fc83a8c44cf4ad5d672424a98ba37ded6
runc version: dc9208a3303feef5b3839f4323d9beb36df0a9dd
init version: fec3683
Security Options:
  seccomp
   Profile: default
Kernel Version: 4.14.193-149.317.amzn2.x86_64
Operating System: Amazon Linux 2
OSType: linux
Architecture: x86_64
CPUs: 2
Total Memory: 7.69GiB
Name: ip-10-0-2-81.eu-west-1.compute.internal
[...]

PR #244, opened 2020-09-17
Issue #243: Introducing the default network in case "bridge" is disabled

ikolomiyets:
This commit resolves the issue of inability to run the tests in the Docker environment with disabled "bridge" network.
The CreateContainer or CreateNetwork function will check if bridge network exists and if does not it creates a reaper_default network and ensure that container is attached to the reaper_default network.

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -35,6 +35,11 @@ import (
 // Implement interfaces
 var _ Container = (*DockerContainer)(nil)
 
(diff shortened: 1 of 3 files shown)


gianarb:
Very good! Thank you for that

gianarb (approved):

Issue #248, opened 2020-09-23
Cannot create a container with more than one network requested


ikolomiyets:
**Describe the bug**
Docker Provider fails to create a container when ContainerRequest has more than one network specified.

**To Reproduce**
Consider container request like the following:
`	amqpRequest := testcontainers.ContainerRequest{
		Image:        "nginx",
		Name:         "nginx-test",
		ExposedPorts: []string{"80/tcp"},
		Env:          env,
		AutoRemove:   true,
		WaitingFor:   hp,
		Networks:     []string{"network1", "network2"},
	}
`
When executed the response from the docker is "Error response from daemon: Container cannot be connected to network endpoints: network1, network2" (assuming that both networks exist).

**Expected behavior**
Container created.

** docker info **
output of the command:

```
$ docker info
Client:
 Debug Mode: false

Server:
 Containers: 8
  Running: 8
  Paused: 0
  Stopped: 0
 Images: 61
 Server Version: 19.03.12
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 7ad184331fa3e55e52b890ea95e65ba581ae3429
 runc version: dc9208a3303feef5b3839f4323d9beb36df0a9dd
 init version: fec3683
 Security Options:
  seccomp
   Profile: default
 Kernel Version: 4.19.76-linuxkit
 Operating System: Docker Desktop
 OSType: linux
 Architecture: x86_64
 CPUs: 2
 Total Memory: 1.945GiB
 Name: docker-desktop
 ID: YEZ5:Y3SD:YQ2P:CQST:GSQY:FUVV:ZOB4:OWUC:A65B:4LDT:CCPL:I2C3
 Docker Root Dir: /var/lib/docker
 Debug Mode: true
  File Descriptors: 104
  Goroutines: 99
  System Time: 2020-09-23T18:30:17.3452466Z
  EventsListeners: 3
 HTTP Proxy: gateway.docker.internal:3128
 HTTPS Proxy: gateway.docker.internal:3129
 Registry: https://index.docker.io/v1/
[...]

PR #249, opened 2020-09-23
Resolving issue #248

ikolomiyets:
The function CreateContainer has been refactored so when more than one network specified in the container request, only first one is used to call Docker's ContainerCreate function and upon successful creation of the container it is connected the remaining networks using Docker's NetworkConnect function.


mdelapenya (approved):
Hi! Thanks for your contribution, it LGTM

PR #251, opened 2020-10-01
Fix shellcheck issues

Bablzz:
Hello everyone!
I hope you are doing well

In this MR fixed small shellcheck.

Have a nice day

PR #253, opened 2020-10-02
fix: Small fix gofmt and misspell from go report

Bablzz:
Hi there!

I've looked at goreportcard and in my mind it's not to be bad fix a few warn from that website.

If you think that it is not good I am going to close this MR.

Have a great day!

### DIFF
--- a/container.go
+++ b/container.go
@@ -114,7 +114,7 @@ func (t ProviderType) GetProvider() (GenericProvider, error) {
 	return nil, errors.New("unknown provider")
 }
 
-// Validate ensures that the ContainerRequest does not have invalid paramters configured to it
+// Validate ensures that the ContainerRequest does not have invalid parameters configured to it
 // ex. make sure you are not specifying both an image as well as a context
 func (c *ContainerRequest) Validate() error {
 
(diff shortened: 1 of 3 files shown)


gianarb:
Hello @Bablzz thank you for your contribution. The typo is appreciated, the other one does not really add value to me. But it is good to embrace standards because it helps maintainability. But as I said for #251 those standards have to be checked continuously in CI. I never used `goreportcard` but in another project, I use `golangci-lint`. I am not a fan of those or better, I don't have any preference.

@mdelapenya do you have a preference in terms of the static analyzer for Go?

mdelapenya:
I've used golangci-lint, and it works fairly simple. E can enforce it in a pre-commit stage/hook (see

Bablzz:
Maybe use golanglint-ci in travis ci, but now this tool show a few issue:)

gianarb:
Yeah, personally I don't want to merge this PR because it does not add any value. I am happy to merge one that fixes the typos.  I am not that excited about introducing a static check. And I don't want to change what I can't continuously assert. It will just be another thing to look during code review.

Thanks

Bablzz:
if you don't mind, I will fix the typos in another MR

Issue #242, opened 2020-09-16
Migrate from Travis.org to GH actions


mdelapenya:
How does this affect us? https://docs.travis-ci.com/user/migrate/open-source-repository-migration

gianarb:
I am slowly moving my project to GitHub action, so I am happy to see this happening if somebody has the bandwidth to do it

Bablzz:
I did small yaml with actions
```
name: Go

on:
  push:
    branches: [ master ]
  pull_request:
    branches: [ master ]

jobs:
  shellcheck:
    runs-on: [ubuntu-latest]
    steps:

    - uses: actions/checkout@v2
    - name: Run ShellCheck
      uses: ludeeus/action-shellcheck@master
      with:
        scandir: ./scripts
        ignore: testresources
    
    - name: Run gofmt
      run: |
        ./scripts/checks.sh
  test_ubuntu:
    strategy:
      matrix:
        go-version: [1.14.x, 1.15.x, 1.x]
        platform: [ubuntu-latest]
    runs-on: ${{ matrix.platform }}
    steps:

    - name: Set up Go 1.x
      uses: actions/setup-go@v2
      with:
        go-version: ${{ matrix.go-version }}
      id: go

    - name: Check out code into the Go module directory
      uses: actions/checkout@v2

    - name: Get dependencies
      run: |
        go get -v -t -d ./...
        go get gotest.tools/gotestsum
    
    - name: modVerify
      run: go mod verify
      
    - name: modTidy
      run: go mod tidy

    - name: vet
      run: go vet ./...

    - name: gotestsum
      run: gotestsum --format short-verbose ./...

    - name: Test
      run: go test -v ./...
```

mdelapenya:
Done!

PR #257, opened 2020-10-08
feature: move to github action #242

Bablzz:
Hello everyone!

In this MR I add github action. There are the same steps as in .travis.yml

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,53 @@
+name: Go
+
+on: [push, pull_request]
+
+jobs:
+  shellcheck:
+    runs-on: [ubuntu-latest]
+    steps:
+
+      - uses: actions/checkout@v2
+      - name: Run ShellCheck
+        run: |
+          shellcheck scripts/*.sh
+
+      - name: Run gofmt
+        run: |
+          ./scripts/checks.sh
+  test:
+    strategy:
+      matrix:
+        go-version: [1.14.x, 1.15.x, 1.x]
+        platform: [ubuntu-latest]
+    runs-on: ${{ matrix.platform }}
+    steps:
+
+      - name: Set up Go 1.x
+        uses: actions/setup-go@v2
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,53 @@
+name: Go
+
+on: [push, pull_request]
+
+jobs:
+  shellcheck:
```
mdelapenya, line 6: wdyt of renaming the entire job to this, which is more inclusive with other tasks than shellcheck?
```suggestion
  static-analysis:
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,53 @@
+name: Go
```
mdelapenya, line 1:
```suggestion
name: Main pipeline
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,53 @@
...
+        run: go mod tidy
+
+      - name: vet
+        run: go vet ./...
+
+      - name: gotestsum
+        run: gotestsum --format short-verbose ./...
+
+      - name: Test
+        run: go test -v ./...
```
mdelapenya, line 53: I think this stage is already executed by the previous one
```suggestion

```

mdelapenya (commented):
Hi @Bablzz, thank you for this great contribution. I added a few nits about names, but on the other hand it will LGTM after hearing your opinion.

Thanks! 👋

.github/workflows/ci.yml
Bablzz: Agree

mdelapenya (approved):
LGTM 👍 

image

mdelapenya:
@gianarb I think we can keep travis for a while, comparing executions, and once we are happy with GH, then move on removing Travis descriptor.

gianarb:
THanks @Bablzz !!!

Bablzz:
@gianarb @mdelapenya My pleasure! :handshake:

PR #258, opened 2020-10-11
Don't re-use reaper if it failed to initialize

kernle32dll:
If the reaper fails to initialize, it leaves testcontainers in a broken state. To be precise, the `Endpoint` variable is never properly initialized, which causes subsequent usages of testcontainers to fail with `failed to create container: connecting to reaper failed: Connecting to Ryuk on  failed: dial tcp: missing address`.


reaper.go
```diff
@@ -77,11 +77,13 @@ func NewReaper(ctx context.Context, sessionID string, provider ReaperProvider, r
 
 	c, err := provider.RunContainer(ctx, req)
 	if err != nil {
+		reaper = nil
```
mdelapenya, line 80: Hey @kernle32dll, thanks for taking the effort of reporting this!

I see this PR would work although I'm concerned about its maintainability if new code conditions are added in the future.

How would you see delaying the creation of the Reaper struct instance (L51) until the container and the endpoint are created (right before the old L87)? I'd see the code more maintainable in this case. wdyt?

mdelapenya:
@kernle32dll are you still interested in this PR? If not, I'd like to label it as `hacktoberfest` in case anybody is willing to help here.

mdonkers:
With merging https://github.com/testcontainers/testcontainers-go/pull/782, I believe this PR became obsolete and can be closed?

mdelapenya:
I believe so. I'm closing it but feel free to reopen if it persists, thanks!

Issue #260, opened 2020-10-26
Opened PR: Contributing Docs need an update based on error logs


shashank-g172:
**Describe the bug**
While attempting to build the container locally while adding a new feature, I noticed that that the contributing.md doc needs an update to a later version of go, but I'm not sure how that'll break Travis builds.

**To Reproduce**
Run the steps on the docs once exec'd into the container

**Expected behavior**
No build errors

** docker info **
Command> go vet ./...
output of the command:
/root/go/pkg/mod/github.com/containerd/containerd@v1.4.1/errdefs/errors.go:92:9: undefined: "github.com/pkg/errors".Is
# github.com/testcontainers/testcontainers-go/wait
wait/http.go:154:3: unknown field 'ForceAttemptHTTP2' in struct literal of type http.Transport
wait/http.go:185:16: undefined: http.NewRequestWithContext
note: module requires Go 1.13# github.com/docker/docker/pkg/archive
/root/go/pkg/mod/github.com/docker/docker@v17.12.0-ce-rc1.0.20200916142827-bd33bbf0497b+incompatible/pkg/archive/archive_linux.go:155:18: undefined: strings.ReplaceAll
/root/go/pkg/mod/github.com/docker/docker@v17.12.0-ce-rc1.0.20200916142827-bd33bbf0497b+incompatible/pkg/archive/archive_linux.go:242:18: undefined: strings.ReplaceAll
# github.com/testcontainers/testcontainers-go/wait [github.com/testcontainers/testcontainers-go/wait.test]
wait/http.go:154:3: unknown field 'ForceAttemptHTTP2' in struct literal of type http.Transport
wait/http.go:185:16: undefined: http.NewRequestWithContext

```
$ docker info
docker --version
Docker version 18.06.1-ce, build e68fc7a

```

**Additional context**

mdelapenya:
@shashank-g172 I'm still not able to reproduce this is issue, so I'm closing it now.

Please feel free to reopen it if you are still seeing what's described here

Thanks!

PR #261, opened 2020-10-27
Updating the go version for local Travis build and set wait strategy for compose services

shashank-g172:
With the older version of go within the local docker container, I found this issue: https://github.com/testcontainers/testcontainers-go/issues/260 

I'm updating the docs to fix it.


mdelapenya:
> In the interest of speeding up the failures, I ended up removing the failedStrategyMap and failed fast : [2a0da88](https://github.com/shashank-g172/testcontainers-go/commit/2a0da883a042815562a82f0afda504e216aa22f2)

As a consumer of the docker compose version, I'd consider any service in the compose file to be satisfied as a test time dependency. If any is not ready, I'd like it to fail.

compose.go
```diff
@@ -89,6 +105,46 @@ func (dc *LocalDockerCompose) getDockerComposeEnvironment() map[string]string {
...
+
+	cli, err := client.NewClientWithOpts(client.FromEnv)
+	if err != nil {
+		return err
+	}
+
+	for k := range dc.WaitStrategyMap {
+		// Docker compose appends "_1" to every started service by default. Trimming to match running docker container
+		f := filters.NewArgs(filters.Arg("name", strings.TrimSuffix(k.service, "_1")),
+			filters.Arg("publish", strconv.Itoa(k.publishedPort)))
```
mdelapenya, line 118: Nit: indent
```suggestion
		filters.Arg("publish", strconv.Itoa(k.publishedPort)))
```

compose.go
```diff
@@ -106,6 +162,14 @@ func (dc *LocalDockerCompose) WithEnv(env map[string]string) DockerCompose {
 	return dc
 }
 
+// WithExposedService sets the strategy for the service that is to be waited on. If multiple strategies
+// are given for a single service running on different ports, both strategies will be applied on the same container
+func (dc *LocalDockerCompose) WithExposedService(service string, port int, strategy wait.Strategy) DockerCompose {
```
mdelapenya, line 167: Wdyt about renaming this function to `WaitingFor(service, port, strategy)`?

I'd also document why we need the port at this API method, as implemented in the `applyStrategyToRunningContainer`

compose.go
```diff
@@ -89,6 +105,46 @@ func (dc *LocalDockerCompose) getDockerComposeEnvironment() map[string]string {
...
+func (dc *LocalDockerCompose) applyStrategyToRunningContainer() error {
+
+	cli, err := client.NewClientWithOpts(client.FromEnv)
+	if err != nil {
+		return err
+	}
+
+	for k := range dc.WaitStrategyMap {
+		// Docker compose appends "_1" to every started service by default. Trimming to match running docker container
+		f := filters.NewArgs(filters.Arg("name", strings.TrimSuffix(k.service, "_1")),
```
mdelapenya, line 117: What do you think about hiding the `_1` from the service name instead, so that consumers don't have to remember about that, just passing service name?

If so, we should update the tests in consequence.

compose.go
mdelapenya: Thinking about it out loud... Should a consumer waiting for a log entry pass the port? I'd like to simply pass the strategy and then the code should infer what to do there. Do you see any other way to filter/get the service container without passing the port?

mdelapenya (commented):
Hey @shashank-g172, great job here! I'm looking forward to seeing this into the library!

Please let's discuss about my comments in the PR.

Once thing I'd do too is to revert the commits not related to the WaitFor strategy for compose, sending in a separate PR, so that we can review and merge separately, unless you consider them related.

Thanks, and well done again!

shashank-g172:
@mdelapenya, each commit is all part of the same change. I needed to update the local go version in the contributing docs, I had to update the travis yaml file because of a dependency on the docker version, and all the other changes are generated except in compose_test.go and compose.go. I'd need to commit all the files for this to work.

compose.go
shashank-g172: This caused static analysis to fail because the indent wasn't compatible with it's rules: https://github.com/shashank-g172/testcontainers-go/actions/runs/355915148 

Reverted the change in a3e4ee63e6295639f072fd8ae49573ded4e99f4e

mdelapenya: Argg! My visual inspection thought that line was part of another line, not parameters for L117 method. Sorry about that

shashank-g172:
@mdelapenya, before I resolve the conflict, I wanted to verify if everything else looks fine. Since the build works fine locally through the Travis docker image, and there are no issues with the build, I'm going to simply remove the go.mod and go.sum versions from the master branch and override them with what I have in my branch. Does this sound reasonable?

Humbertzhang:
Hey everyone, may I ask if this PR will be merged?

mdelapenya:
> Sorry for the late response, I'm still on vacations and did not check comms that much.

Yeah, I'd like to merge it sooner then later, as it brings a lot of value to the compose feature. I'd still suggest renaming the `withExposedService` method to `waitForService` , although I'd love to hear @gianarb's voice here

gianarb:
Thank you for your contribution @shashank-g172 and I am sorry if it didn't get merged on time and it now requires a rebase. @mdelapenya `withExposedService` is the way TestContainers-Java does it and I think we should keep it consistent 👍 even if I agree `wait` is clearer.

Overall this PR can be merged if rebased

shashank-g172:
@gianarb @mdelapenya @mbroshi - The tests and the functionality works if I'm able to list the running containers so that I can apply the filter to apply the Wait Strategy for the one being specified. If any of you have a Go-Docker SDK agnostic way of listing the containers that I can't seem to find (which is why I had to pin down the version to 1.38.0), I can modify the file. 

https://travis-ci.org/github/testcontainers/testcontainers-go/builds/751999670 (Look at the failed builds without the docker version set)

Considering that there isn't another way to list the containers easily - https://godoc.org/github.com/docker/docker/client#Client.ContainerList

gianarb:
I am not sure if I understand the context of your question @shashank-g172  but I don't think we need to be Docker agnostic when it comes to the scope of Docker Compose, because Docker is a pretty strong compose requirement.
Happy new year btw!

shashank-g172:
Happy New Year @gianarb! 

I think I could phrase my question better:

1. There's one way of listing all the running containers using the Docker GO SDK:  https://godoc.org/github.com/docker/docker/client#Client.ContainerList
2. I'm listing the running  containers the same way while trying to apply the wait strategy to them (in compose.go)
3. However, I think that unless the DOCKER version is set explicitly to 1.38.0 on the CI agent, the functionality doesn't work. I can't seem to find an alternate way of listing running containers with later versions of the DOCKER API using the Go SDK. 
4. Example: When these changes ran- only the agents with the Docker version set 1.38.0 pass, the others fail with this error message:

`one or more wait strategies could not be applied to the running containers: error Error response from daemon: client version 1.40 is too new. Maximum supported API version is 1.38 occured while filtering the service nginx_1: 80 by name and published port`

Is this a result of TestContainers making as assumption about the version of docker?

shashank-g172:
Thoughts?

mdelapenya:
> Thoughts?

Thanks for your time here @shashank-g172. I already expressed my concerns about using the port for any wait strategy type, but I'm not against it: in deed I think this is a really great contribution. We can commit and iterate for improvements.

LGTM 👍

mdelapenya:
Hey @shashank-g172, we removed support for travis. Would you mind rebasing/merging with upstream?

I think this PR brings a lot of value to the docker-compose consumers, so we'd like to review and merge the soonest. Thanks for your patience.

gianarb:
@mdelapenya do you have time to cherry-pick, rebase and help to make this work ready to get merged?

Thanks a lot!

mdelapenya:
> @mdelapenya do you have time to cherry-pick, rebase and help to make this work ready to get merged?
> 
> Thanks a lot!

Yes, I believe I'll have time this Friday morning. Let me do that then. 👍

mdelapenya:
I added a branch in my remote with exactly this PR plus the merge commit and a commit running `go mod tidy`: https://github.com/mdelapenya/testcontainers-go/tree/shashank-g172-master-resolve-conflicts

gianarb:
Ok great, feel free to close this one and open a new one 👍

PR #271, opened 2020-12-18
container information stauff

01101101M:


mdelapenya:
Hi @01101101M, could you share here or in a new issue the motivation for this change? We'd like to better understand the use cases and the real needs to add it.

Thanks in advance!

01101101M:
Hey @mdelapenya, first of all, thank you for your time.

##### Why it's important:
* providing all the features available in the java version (such as [Healthcheck Wait strategy](https://www.testcontainers.org/features/startup_and_waits/#healthcheck-wait-strategy-examples))
* it would be very nice to be able to inspect container properties such as RunningStatus, ExitCode, Health, etc.  
* also in my case, I have a worker container and want to use its output so i need to wait for the container to exit

wait/healt.go
```diff
@@ -0,0 +1,77 @@
+package wait
```
Bablzz, line 1: maybe health.go not healt.go

docker.go
```diff
@@ -186,6 +186,16 @@ func (c *DockerContainer) Terminate(ctx context.Context) error {
 	return err
 }
 
+// update container raw info
+func (c *DockerContainer) inspectContainerFresh(ctx context.Context) (*types.ContainerJSON, error) {
```
mdelapenya, line 190: Wdyt about renaming to `inspectRawContainer`?

mdelapenya (commented):
@01101101M sorry for the delay with the review. I added a tiny rename in a method.

Could you verify that all wait implementation provide the new `State` method?

01101101M:
> @01101101M sorry for the delay with the review. I added a tiny rename in a method.
> 
> Could you verify that all wait implementation provide the new `State` method?

Great! That would be nice.
I'll check it out ASAP.

01101101M:
@mdelapenya
The `State` method is implemented on the `DockerContainer` and is required by `Container` and `StrategyTarget`. Considering that all wait Strategies are using the `StrategyTarget` interface, The `State` method is accessible from all wait implementation and outside the package.

mdelapenya (approved):
LGTM!

PR #272, opened 2020-12-23
feat: add logging on image pull error

tobiaszheller:
Right now when something is wrong with pulling image, for example:
```
Error response from daemon: unauthorized: You don't have the needed permissions to perform this operation, and you may have invalid credentials. To authenticate your request, follow the steps in: https://cloud.google.com/container-registry/docs/advanced-authentication
```
there is no indication that any action is happening. This log will additional debug information if something is wrong.

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -17,15 +17,13 @@ import (
 	"time"
 
 	"github.com/cenkalti/backoff"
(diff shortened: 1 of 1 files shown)


gianarb:
Thanks

gianarb (approved):

PR #283, opened 2021-02-10
Update Docker dependency to Docker 20.10.3.

### LINKED ISSUE #273: "not enough arguments in call to p.client.ContainerCreate"
Installing Testcontainers v0.9.0 currently fails to work, because Go is automatically pulling in Docker v20.10.1 instead of v17.12.0 and the API has changed.

As such, when I run anything I get:
```
/Users/sazzer/.gvm/pkgsets/go1.15.6/global/pkg/mod/github.com/testcontainers/testcontainers-go@v0.9.0/docker.go:645:39: not enough arguments in call to p.client.ContainerCreate
	have (context.Context, *container.Config, *container.HostConfig, *network.NetworkingConfig, string)
	want (context.Context, *container.Config, *container.HostConfig, *network.NetworkingConfig, *v1.Platform, string)
```

I can fix this by running:
```
$ go mod edit -require github.com/docker/docker@v17.12.0-ce-rc1.0.20200916142827-bd33bbf0497b+incompatible
```

But the next time I run `go get -u ./...` to update dependencies then it all breaks again.

michielboekhoff:
This should fix #273, as this changes the version of the Docker dependency to 20.10.3.

The backward-incompatible parameter in the call to `ContainerCreate` is the platform, which is for images with different platforms (e.g. `linux/amd64` vs `linux/arm64`). If not specified, this ends up defaulting to the host platform from what I understand, and is only relevant for machines that are multi-platform capable. At any rate, this is an option we can expose as part of this PR, but I think the use-case is so niche, it might not be worth the effort, as testing that might prove difficult.

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -645,7 +645,7 @@ func (p *DockerProvider) CreateContainer(ctx context.Context, req ContainerReque
 		EndpointsConfig: endpointConfigs,
 	}
 
-	resp, err := p.client.ContainerCreate(ctx, dockerInput, hostConfig, &networkingConfig, req.Name)
+	resp, err := p.client.ContainerCreate(ctx, dockerInput, hostConfig, &networkingConfig, nil, req.Name)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


gianarb:
Thanks a lot for that @michielboekhoff ! I agree with you we do not need to expose that parameter for now!

sirianni:
Thanks @michielboekhoff !  @gianarb is it possible to get a new release with this fix?

Issue #294, opened 2021-03-16
Runtime panic 


danni-popova:
**Describe the bug**
Trying to run a test results in a panic when trying to get mapped ports.

**To Reproduce**
https://github.com/todanni/account/blob/main/internal/repository/repository_test.go

**Expected behavior**
Not panic.

** docker info **
output of the command:

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Build with BuildKit (Docker Inc., v0.5.1-docker)
  scan: Docker Scan (Docker Inc., v0.5.0)

Server:
 Containers: 6
  Running: 0
  Paused: 0
  Stopped: 6
 Images: 7
 Server Version: 20.10.5
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 269548fa27e0089a8b8278fc4fc781d7f65a939b
 runc version: ff819c7e9184c13b7c2607fe6c30ae19403a7aff
 init version: de40ad0
 Security Options:
  seccomp
   Profile: default
 Kernel Version: 5.4.72-microsoft-standard-WSL2
 Operating System: Docker Desktop
 OSType: linux
 Architecture: x86_64
 CPUs: 16
 Total Memory: 25.01GiB
 Name: docker-desktop
 ID: K2UZ:4XFZ:UKRM:UXYT:MSZF:OGA3:QDVU:GVOR:EYU5:54DD:Q3J4:KSCK
 Docker Root Dir: /var/lib/docker
 Debug Mode: false
 Registry: https://index.docker.io/v1/
 Labels:
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false

WARNING: No blkio throttle.read_bps_device support
WARNING: No blkio throttle.write_bps_device support
WARNING: No blkio throttle.read_iops_device support
WARNING: No blkio throttle.write_iops_device support
```

[...]

PR #295, opened 2021-03-18
Wait strategies must poll ports

programminh:
Docker engine 20.10.3 has changed the way ports are returned from InspectContainer. The array is not always populated which triggers a panic.

Related to #294

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -44,7 +44,7 @@ type DockerContainer struct {
 	ID         string
 	WaitingFor wait.Strategy
 	Image      string
-	
+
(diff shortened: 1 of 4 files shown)


gianarb (approved):
Thanks!! good catch, I think caching without invalidation was for sure a poor choice I have made

butuzov:
hi, @gianarb any plans on releasing this? 
I needed to go full circle (from stepping on this issue in the new project to forking repo in order to fix it, only to find it's done by @programminh ). it's really desired fix.

Issue #297, opened 2021-03-20
Unexpected behavior for WaitLog.WithOccurence


FlorianLehmann:
**Describe the bug**
When I define 

NewLogStrategy("<string>").
		WithStartupTimeout(100 * time.Microsecond).
		WithOccurrence(X) 

for a container, it does not wait for X occurences. One occurence is enough to continue.

**To Reproduce**
`func TestWaitShouldFailWithExactNumberOfOccurrences(t *testing.T) {
	target := noopStrategyTarget{
		ioReaderCloser: ioutil.NopCloser(bytes.NewReader([]byte("hello\r\ndude"))),
	}
	wg := NewLogStrategy("dude").
		WithStartupTimeout(100 * time.Microsecond).
		WithOccurrence(2)
	err := wg.WaitUntilReady(context.Background(), target)
	if err == nil {
		t.Fatal("expected error")
	}
}`

**Expected behavior**
it must wait for X occurences in the logs.

**Additional context**

https://github.com/testcontainers/testcontainers-go/blob/6467d9b1599565843074662d1e49dd68112ce1ef/wait/log.go#L75-L95

There is a variable currentOccurence defined before the loop. This variable is incremented in the loop each time the expected log appears. However, logs are read from the beginning. Therefore, the condition based on the number of occurences is always true if the expected log is present only once.

mdelapenya:
@FlorianLehmann should we close this issue? I see the PR was merged

Thanks!!

FlorianLehmann:
Yes, this has been resolved. It can be closed. Thanks

PR #298, opened 2021-03-20
Fix/wait log exact number occurences

FlorianLehmann:
#297

**Describe the bug**
When I define 

NewLogStrategy("<string>").
		WithStartupTimeout(100 * time.Microsecond).
		WithOccurrence(X) 

for a container, it does not wait for X occurences. One occurence is enough to continue.

**To Reproduce**

```go
func TestWaitShouldFailWithExactNumberOfOccurrences(t *testing.T) {
	target := noopStrategyTarget{
		ioReaderCloser: ioutil.NopCloser(bytes.NewReader([]byte("hello\r\ndude"))),
	}
	wg := NewLogStrategy("dude").
		WithStartupTimeout(100 * time.Microsecond).
		WithOccurrence(2)
	err := wg.WaitUntilReady(context.Background(), target)
	if err == nil {
		t.Fatal("expected error")
	}
}
```

**Expected behavior**
it must wait for X occurences in the logs.

**Additional context**

https://github.com/testcontainers/testcontainers-go/blob/6467d9b1599565843074662d1e49dd68112ce1ef/wait/log.go#L75-L95

There is a variable currentOccurence defined before the loop. This variable is incremented in the loop each time the expected log appears. However, logs are read from the beginning. Therefore, the condition based on the number of occurences is always true if the expected log is present only once.


gianarb:
Thanks a lot for your work and sorry if it took me some time

PR #301, opened 2021-03-25
feat: add basic auth support in the wait.ForHTTP strategy

nevins-b:


mdelapenya:
Hi @nevins-b thanks for this contribution. At first sight seems ok, but we'd like to add tests covering this new behavior. Also if you could provide context about the bed for this contribution that could help us in the review process.

Thank you!!

nevins-b:
For context this was initially to support the opendistro ES container (https://opendistro.github.io/for-elasticsearch-docs/docs/install/docker/) which launches with basic auth on all endpoints

wait/testdata/main.go
```diff
@@ -19,13 +21,28 @@ func main() {
 	})
 
 	mux.HandleFunc("/ping", func(w http.ResponseWriter, req *http.Request) {
-		data, _ := ioutil.ReadAll(req.Body)
-		if bytes.Equal(data, []byte("ping")) {
-			w.WriteHeader(http.StatusOK)
-			_, _ = w.Write([]byte("pong"))
-		} else {
-			w.WriteHeader(http.StatusBadRequest)
+		auth := req.Header.Get("Authorization")
```
mdelapenya, line 24: How would you see defining different endpoints for both types:

- `/ping`: old behaviour
- `/authenticated-ping`: with Basic Auth

That would decouple the logic (and the future maintenance) of that code. That would mean adding a separate test hitting the new endpoint.

Does it make sense to you?

mdelapenya: Other than that, I'm OK with this PR 👍

mdelapenya: Hey @nevins-b sorry for the late response. Did you check my comment about having a separate handler for the authenticated ping? Once there, I think this is ready to go.

Thanks!!

mdelapenya: Hey @nevins-b I've pushed a commit on top of yours, separating the two handlers. Hope you can 👀

mdelapenya (dismissed):
I added a few commits on top to separate the HTTP handlers in the test application used to verify the waitFor HTTP feature.

mdelapenya:
@nevins-b I've polished the PR a bit, updating the existing docs for the HTTP wait strategy. So I'm going to merge this PR.

Thanks for your initial work here! 🙇

mdelapenya (approved):

PR #310, opened 2021-04-13
Add waitForSqlWithHost implementation.

michielboekhoff:
This implementation mostly copies the prior waitForSql implementation, the difference being that the `waitForSqlWithHost` allows the user to construct a connection string using the host of the DB container. This is especially useful in a scenario like CI (where tests are being executed in a container) or even remote Docker hosts.

To preserve backwards compatibility, and because of the way that `waitForSql` works, `waitForSqlWithHost` replicates the existing behaviour mostly. In a v1 of testcontainers-go, you would probably want to unify these two differing implementations.

One thing I would like to (personally) see in a v1 of testcontainers-go is that the underlying strategy struct is not exposed to the calling code (instead opting to return the `Strategy` interface) so that we can change its implementation, whereas right now we cannot, as all the fields are exposed to calling code.


michielboekhoff:
Also, we might want to wait for #214 to be merged in before we look at this.

michielboekhoff:
@gianarb - might be great to get your feedback on this if you get a chance!

wait/sql.go
```diff
@@ -79,3 +93,75 @@ func (w *waitForSql) WaitUntilReady(ctx context.Context, target StrategyTarget)
...
+type waitForSqlWithHost struct {
+	URL            func(port nat.Port, host string) string
+	Driver         string
+	Port           nat.Port
+	startupTimeout time.Duration
+	PollInterval   time.Duration
+}
+
+//Timeout sets the maximum waiting time for the strategy after which it'll give up and return an error
+func (w *waitForSqlWithHost) Timeout(duration time.Duration) *waitForSqlWithHost {
```
mniak, line 106: In PR #322 we are considering to change the name of this method to `WithTimeout` in the other wait strategies.
It could  be renamed here as well

michielboekhoff: I think that's a great change, I'll do that when I get a chance to.

mdelapenya:
@michielboekhoff I think this PR has been superseded by #524 

Please let me know if I can close it, as its original intention has been already implemented and merged into main (not released yet)

michielboekhoff:
> @michielboekhoff I think this PR has been superseded by #524 
> 
> Please let me know if I can close it, as its original intention has been already implemented and merged into main (not released yet)

Yep, seems like it. I've closed this PR.

PR #311, opened 2021-04-13
Add instrumentations for containers created with Dockerfile

xicoalmeida:
Add support for [ARGS](https://docs.docker.com/engine/reference/builder/#arg) when using Dockerfile to create images. 
Also allow users to enable/disable build log.


xicoalmeida:
Hey guys! Can you check this PR, please!?

container.go
```diff
@@ -90,7 +91,8 @@ type ContainerRequest struct {
 	ReaperImage     string              // alternative reaper image
 	AutoRemove      bool                // if set to true, the container will be removed from the host when stopped
 	NetworkMode     container.NetworkMode
-	AlwaysPullImage bool // Always pull image
+	AlwaysPullImage bool               // Always pull image
+	BuildArgs       map[string]*string // for specifying network aliases
```
gianarb: Hey! Thank you for your work! I think this fits better as part of the `FromDockerfile` what do you think?

xicoalmeida: For sure! I've changed to there.

xicoalmeida:
Hey @gianarb how's it going? I've added a way to print the build log of Dockerfile. Can you validade if, please?

gianarb:
I have to admit that the idea to print to stdout makes me a bit nervous, I saw we do that in other places because the library does not have a logger, but still it is something we should avoid. Can you tell me a bit more about why you need a log? To figure out why a build failed?! I am wondering if we can, at least for now troubleshoot based on just the error without a full log.

Or we can write a `TODO:` mentioning that at some point we expect a logger (ideally in tests it should be `t.Log()`)

gianarb:
@xicoalmeida can you rebease? I am happy to merge and release a new version of testcontainers

mdelapenya (approved):
As @gianarb mentioned, this PR will be ready to go after a rebase 👍 

Thanks for your contribution!!

xicoalmeida:
Hey @mdelapenya, I've just rebased.

gianarb:
Thanks a lot!

Issue #319, opened 2021-05-10
Reaper leaks


Eun:
**Describe the bug**
The package [goleak](https://github.com/uber-go/goleak) reports leakages in `Reaper`.

**To Reproduce**
Use following test:

```go
package main

import (
	"context"
	"testing"
	"time"

	"github.com/testcontainers/testcontainers-go"
	"github.com/testcontainers/testcontainers-go/wait"
	"go.uber.org/goleak"
)

func TestMain(m *testing.M) {
	goleak.VerifyTestMain(m)
}

func TestFoo(t *testing.T) {
	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
	defer cancel()
	req := testcontainers.ContainerRequest{
		Image:        "postgres:12",
		ExposedPorts: []string{"5432/tcp"},
		AutoRemove:   true,
		Env: map[string]string{
			"POSTGRES_USER":     "postgres",
			"POSTGRES_PASSWORD": "postgres",
			"POSTGRES_DB":       "postgres",
		},
		WaitingFor: wait.ForListeningPort("5432/tcp"),
	}
	container, err := testcontainers.GenericContainer(ctx, testcontainers.GenericContainerRequest{
		ContainerRequest: req,
		Started:          true,
	})
	if err != nil {
		panic(err)
	}
	if err := container.Terminate(ctx); err != nil {
		panic(err)
	}
}

```

**Expected behavior**
A successful test

**docker info**
```output of the command:```

```
$ docker info
Client:
 Context:    default
 Debug Mode: false

Server:
 Containers: 18
  Running: 1
  Paused: 0
  Stopped: 17
 Images: 72
 Server Version: 20.10.2
 Storage Driver: zfs
  Zpool: rpool
  Zpool Health: ONLINE
  Parent Dataset: rpool/var/lib/docker
  Space Used By Parent: 10480205824
  Space Available: 145392234496
  Parent Quota: no
  Compression: lz4
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 
 runc version: 
 init version: 
[...]

PR #320, opened 2021-05-10
close reaper go routines on `DockerContainer.Terminate` & `DockerNetwork.Remove`

Eun:
This fixes the issue #319.

When `Reaper` is used, the `DockerContainer.Terminate` and `DockerNetwork.Remove` functions should close the spawned go routine by `Reaper.Connect`.

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -426,6 +426,11 @@ type DockerNetwork struct {
 
 // Remove is used to remove the network. It is usually triggered by as defer function.
 func (n *DockerNetwork) Remove(ctx context.Context) error {
+	select {
+	// close reaper if it was created
+	case n.terminationSignal <- true:
+	default:
+	}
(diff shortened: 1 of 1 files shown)


mdelapenya:
Hi @Eun thanks for this PR, we'll review it ASAP. Would you mind adding the test you proposed in the original issue? Taht way we will be covered against regressions.

Thanks in advance

Eun:
Hi @mdelapenya, adding a test would not be hegglpful since the other leak at #321 is still open, so adding a test as soon as both issues are fixed would make the most sense.
I guess just adding 
```go
func TestMain(m *testing.M) {
	goleak.VerifyTestMain(m)
}
``` 
would be enough.

mdelapenya (approved):

gianarb:
Thanks a lot for taking the time and open this PR! I will have a look at the open issue around docker connections. I think we really need to add the test you proposed.

PR #323, opened 2021-05-19
Retry follow logs request on closed connection

rmfitzpatrick:
I've noticed an issue in logs producers where the daemon will close the connection due to container logging inactivity.  These changes add an updated retry instead of silently ignoring the error and update the test that demonstrates the problem.


gianarb:
Thanks a lot!! Good catch!

gianarb (approved):

PR #330, opened 2021-06-11
fix(compose): filter containers by service name only

mdelapenya:
## What does this PR do?
It removes the filter by exposed port from the docker client call to retrieve the running containers, leaving the compose service name as the only filter.

I'm adding a unit test representing the bug: an nginx service with no exposed ports, and a wait strategy waiting for a log entry.

> I verified that the new test fails before these changes, and passes after them.

## Why is it important?
It avoid the edge case where the service to wait for is not exposing a port in the compose file. Because of the current implementation, it requires that client code has to expose ports, even if it's not needed.

## Follow-ups
The signature of the `WithExposedService` method is:
- service name
- service port
- wait strategy

We could possibly remove the port, as if needed it would be used as part of the wait strategy


mdelapenya:
@shashank-g172, as the original contributor of the wait-for-service feature, could you please take a look too? 🙏

shashank-g172:
Sorry! Been off the grid for a little while. Thank you for taking care of the merge this feature in @mdelapenya! 

Yes, I will take a look at this!

mdelapenya:
I'm seeing that `Test_LogConsumerGetsCalled` is failing with certain flakiness: in previous runs it failed for 1.14.x, now it's failing for 1.15.x. I'd open an issue for investigation about that test

mdelapenya:
@gianarb I think we should release v0.11.1 including this patch, as it'd fix the need of exposed ports on compose services. Wdyt?

gianarb (approved):

gianarb:
I agree! Thank you for this fix!

PR #334, opened 2021-07-21
fix: store raw information of a container

mdelapenya:
## What does this PR do?
It stores raw information of a container (JSON format) in the Docker Container struct

## Why is it important?
Caused by #271, where we are getting the state of a container from the Docker client, but we did not add the field


gianarb:
Thanks a lot

PR #342, opened 2021-08-07
Support for specifying image platform

amitsaha:
This pull request adds support for specifying a platform of the image to fetch. Equivalent to `docker run --platform ..`.

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -600,7 +600,7 @@ func (p *DockerProvider) CreateContainer(ctx context.Context, req ContainerReque
 	}
 
 	var tag string
-	customPlatform := req.ImagePlatform
+
 	var customOS, customArch string
 	if req.ShouldBuildImage() {
 		tag, err = p.BuildImage(ctx, &req)
@@ -610,9 +610,9 @@ func (p *DockerProvider) CreateContainer(ctx context.Context, req ContainerReque
 	} else {
 		tag = req.Image
 
-		if len(customPlatform) != 0 {
-			customOS = strings.Split(customPlatform, "/")[0]
-			customArch = strings.Split(customPlatform, "/")[1]
+		if len(req.ImagePlatform) != 0 {
+			customOS = strings.Split(req.ImagePlatform, "/")[0]
+			customArch = strings.Split(req.ImagePlatform, "/")[1]
 		}
 		var shouldPullImage bool
 
@@ -634,8 +634,8 @@ func (p *DockerProvider) CreateContainer(ctx context.Context, req ContainerReque
 		if shouldPullImage {
 			pullOpt := types.ImagePullOptions{}
 
-			if len(customPlatform) != 0 {
-				pullOpt.Platform = customPlatform
+			if len(req.ImagePlatform) != 0 {
+				pullOpt.Platform = req.ImagePlatform
 			}
 			if req.RegistryCred != "" {
 				pullOpt.RegistryAuth = req.RegistryCred
@@ -706,13 +706,18 @@ func (p *DockerProvider) CreateContainer(ctx context.Context, req ContainerReque
 		EndpointsConfig: endpointConfigs,
 	}
 
-	platSpec := specs.Platform{}
-	if len(customPlatform) != 0 {
-		platSpec.Architecture = customArch
-		platSpec.OS = customOS
+	var platSpec *specs.Platform
+	if len(req.ImagePlatform) != 0 {
+		platSpec = &specs.Platform{
+			Architecture: customArch,
+			OS:           customOS,
+		}
+	} else {
+		platSpec = nil
 	}
 
-	resp, err := p.client.ContainerCreate(ctx, dockerInput, hostConfig, &networkingConfig, &platSpec, req.Name)
+	resp, err := p.client.ContainerCreate(ctx, dockerInput, hostConfig, &networkingConfig, platSpec, req.Name)
+
 	if err != nil {
 		return nil, err
 	}
--- a/docker.go
+++ b/docker.go
@@ -622,6 +622,10 @@ func (p *DockerProvider) CreateContainer(ctx context.Context, req ContainerReque
 
 		if shouldPullImage {
 			pullOpt := types.ImagePullOptions{}
+
+			if len(req.ImagePlatform) != 0 {
+				pullOpt.Platform = req.ImagePlatform
+			}
 			if req.RegistryCred != "" {
 				pullOpt.RegistryAuth = req.RegistryCred
 			}
--- a/docker.go
+++ b/docker.go
@@ -30,6 +30,8 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/testcontainers/testcontainers-go/wait"
+
+	specs "github.com/opencontainers/image-spec/specs-go/v1"
 )
 
 // Implement interfaces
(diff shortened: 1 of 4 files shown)


mdelapenya:
Hi @amitsaha, thanks for this contribution. It LGTM, although I'd like to see a test if possible.

I'm currently checking if github actions provide support for ARM so that we are able to run tests on ARM.

Could you also check it please?

amitsaha:
> Hi @amitsaha, thanks for this contribution. It LGTM, although I'd like to see a test if possible.
> 
> I'm currently checking if github actions provide support for ARM so that we are able to run tests on ARM.
> 
> Could you also check it please?

Sure, so it looks like github actions can run tests on arm64 as explained [here](https://medium.com/@nabil.servais/using-github-actions-to-test-your-go-code-on-arm64-ce581e646cb). 

Regarding the test, sure i am happy to. I am thinking around the testing strategy though.

My current use case is i specified a x86-64 image for my Macbook M1 - and it works because of the emulation that M1 has.

If we test x86-64 as the image platform, it can only succeed on a x86-64 build host. Similarly for the arm64 image.

Here we really want to ensure that the `ImagePlatform` is propagated to the code that calls out to the docker API. So perhaps we can test that instead of creating a real container which i am not sure is possible given the above explanation.

amitsaha:
I also realised, this only works for images which are to be pulled. For existing images, the platform option is not accounted for. I will push an update once i have got it working. In fact, that may mean we can test this in a nice way as well.

docker.go
```diff
@@ -598,34 +600,46 @@ func (p *DockerProvider) CreateContainer(ctx context.Context, req ContainerReque
 	}
 
 	var tag string
+	customPlatform := req.ImagePlatform
```
mdelapenya, line 603: Super NIT: wdyt about saving this variable and directly use `req.ImagePlatform` elsewhere? Therefore the code itself will be self-contained: I had to double check that the customPlatform variable was in fact initialised with the requested one.

mdelapenya:
@amitsaha could you take a look at the test failures? 🙏 I think we are close to merge :)

amitsaha:
Hi @mdelapenya thanks. Pushed a fix.

amitsaha:
looking at the test failures.

mdelapenya:
No, I haven't that error before in this test suite. Although the error is familiar to me when dealing with multi-platform support: can we ensure that the container uses the same platform (AMD Vs ARM) as the underlying CI machine? 🤷‍♀️

amitsaha:
> No, I haven't that error before in this test suite. Although the error is familiar to me when dealing with multi-platform support: can we ensure that the container uses the same platform (AMD Vs ARM) as the underlying CI machine? 🤷‍♀️

I thought that was the default behavior? I will have a look.

amitsaha:
Fetching upstream master and then rebased fixed the test run. 🤞 let's see how the CI run goes.

docker_test.go
```diff
@@ -1382,6 +1383,26 @@ func TestContainerNonExistentImage(t *testing.T) {
 	})
 }
 
+func TestContainerCustomPlatformImage(t *testing.T) {
+	t.Run("Use a custom image platform", func(t *testing.T) {
```
mdelapenya, line 1387: Wdyt about renaming the test case? Because `windows/arm12` is not a valid platform, I guess that even if running the tests on a Windows ARM machine, this will fail, right? 
```suggestion
	t.Run("Use an invalid custom image platform", func(t *testing.T) {
```

docker_test.go
```diff
@@ -1382,6 +1383,26 @@ func TestContainerNonExistentImage(t *testing.T) {
...
+				SkipReaper:    true,
+				ImagePlatform: "windows/arm12",
+			},
+			Started: false,
+		})
+		if c != nil {
+			defer c.Terminate(context.Background())
+		}
+		if err == nil {
+			t.Fatalf("Expected non-nil error")
```
mdelapenya, line 1400: Because the test expects to raise an error, I would log it with more information:
```suggestion
			t.Fatalf("Expected non-nil error when using an invalid platform: 'windows/arm12'")
```
Maybe extracting the platform to a variable to be reused in this message? wdyt?

mdelapenya (commented):
I made a few comments, but other than that, this LGTM. Waiting for last minute feedback to approve it.

Thank you for your work here!!

docker_test.go
amitsaha: Hi @mdelapenya this was my not so optimal attempt to test the functionality i am adding. I used `windows/arm12` since that's a non-existent image platform so I expect the test to errror out.

mdelapenya: wdyt about `Use a non-existent custom image platform` then?

My intention here is to make the test super explicit to casual readers, specially on failures

docker_test.go
```diff
@@ -1384,20 +1384,25 @@ func TestContainerNonExistentImage(t *testing.T) {
...
+			Started: true,
 		})
 		if c != nil {
 			defer c.Terminate(context.Background())
 		}
 		if err == nil {
-			t.Fatalf("Expected non-nil error")
+			t.Fatalf("Expected an error containing failed to create container: Error response from daemon")
+		}
+		expectedErrorMsgSubstring := "image with reference redis:latest was found but does not match the specified platform"
```
mdelapenya, line 1403: Mmm, not sure about this new verification. Will that message change between Docker versions? If you see it convenient, I'm OK with it too, but I'd like to make sure we are not adding flakiness because of specific message formats. 

```go
if err == nil {
	t.Fatalf("Expected non-nil error with a non-existent platform: %s", nonExistentPlatform)
}
```

amitsaha: Sounds like a good idea. done.

docker_test.go
```diff
@@ -1382,6 +1383,29 @@ func TestContainerNonExistentImage(t *testing.T) {
...
+				ImagePlatform: nonExistentPlatform,
+			},
+			Started: true,
+		})
+		if c != nil {
+			defer c.Terminate(context.Background())
+		}
+		if err == nil {
+			t.Fatalf("Expected an error containing failed to create container: Error response from daemon")
+		}
```
mdelapenya, line 1402: I think this is a left-over, right?
```suggestion
```

amitsaha: Yes, my bad. Fixed now.

mdelapenya (commented):
LGTM! Thanks for your patience while reviewing it!

mdelapenya:
@gianarb what are your thoughts on this one? Testing in ARM is indeed a thing nowadays. Could you take a look? 🙏

docker.go
```diff
@@ -598,34 +600,46 @@ func (p *DockerProvider) CreateContainer(ctx context.Context, req ContainerReque
...
 		tag, err = p.BuildImage(ctx, &req)
 		if err != nil {
 			return nil, err
 		}
 	} else {
 		tag = req.Image
+
+		if len(req.ImagePlatform) != 0 {
+			customOS = strings.Split(req.ImagePlatform, "/")[0]
+			customArch = strings.Split(req.ImagePlatform, "/")[1]
```
mdelapenya, line 615: This is correct, as all platforms are formed by a string in this format: "os/arch"

👍

hairyhenderson:
Any possibility to get this merged soon? This is necessary in some common scenarios (i.e. running the `mysql:5.7` image, which isn't available for arm64 yet)

docker.go
hairyhenderson: A platform may also have 3 parts (os/arch/variant) in some cases. It's probably safest to use containerd's [`platforms.Parse`](https://pkg.go.dev/github.com/containerd/containerd/platforms#Parse) func for this.

hairyhenderson:
I've issued #395 to try and help things move along here - I'm easy with either PR being merged (I needed the changes regardless, so I figured I might as well issue a PR if the commit's in my fork anyway!)

mdelapenya:
I think we can close this one, given #395 was merged

Thanks @amitsaha for bringing the seed for this feature!

PR #344, opened 2021-08-12
chore: add WaitForService method to compose

mdelapenya:
## What does this PR do?
We are providing a simple API to wait for services by service name, only

## Why is it important?
It will allow consumer to wait on a service without having to pass a port.


gianarb (approved):

gianarb:
Thanks a lot!

PR #347, opened 2021-08-17
Added CopyFileFromContainer to DockerContainer

codepitbull:
# Why ?
While working on a k3s-container I needed to extract the kubeconfig from the running container and discovered that there is no CopyFileFromContainer available.

# What ?
Added CopyFileFromContainer to DockerContainer.
Added Unit-test for the new Function.

### DIFF
--- a/container.go
+++ b/container.go
@@ -52,6 +52,7 @@ type Container interface {
 	Exec(ctx context.Context, cmd []string) (int, error)
 	ContainerIP(context.Context) (string, error) // get container ip
 	CopyFileToContainer(ctx context.Context, hostFilePath string, containerFilePath string, fileMode int64) error
+	CopyFileFromContainer(ctx context.Context, filePath string) ([]byte, error)
 }
 
 // ImageBuildInfo defines what is needed to build an image
--- a/container.go
+++ b/container.go
@@ -52,7 +52,7 @@ type Container interface {
 	Exec(ctx context.Context, cmd []string) (int, error)
 	ContainerIP(context.Context) (string, error) // get container ip
 	CopyFileToContainer(ctx context.Context, hostFilePath string, containerFilePath string, fileMode int64) error
-	CopyFileFromContainer(ctx context.Context, filePath string) ([]byte, error)
+	CopyFileFromContainer(ctx context.Context, filePath string) (io.ReadCloser, error)
 }
 
 // ImageBuildInfo defines what is needed to build an image
--- a/docker.go
+++ b/docker.go
@@ -336,7 +336,7 @@ func (c *DockerContainer) Exec(ctx context.Context, cmd []string) (int, error) {
 
 type FileFromContainer struct {
 	underlying *io.ReadCloser
-	tarreader *tar.Reader
+	tarreader  *tar.Reader
 }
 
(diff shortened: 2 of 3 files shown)


mdelapenya (approved):
Thanks for this contribution @codepitbull, this LGTM!

mdelapenya:
As a follow-up, would you see it doable to contribute an implementation to the compose provider? 🤔

codepitbull:
I will checxk to add it to the ComposeProvider

codepitbull:
About the Compose Part: There are no containers exposed.
The compose implementation as it is really just runs docker compose and can check for availability of a container.
It doesn't even allow copying files into a running container.
from what I see this qould take quite a big effort to achieve.

mdelapenya:
> About the Compose Part: There are no containers exposed.
> The compose implementation as it is really just runs docker compose and can check for availability of a container.
> It doesn't even allow copying files into a running container.
> from what I see this qould take quite a big effort to achieve.

Sure thing. I can create an issue to add the support to the compose provider. Meanwhile, this LGTM

mdelapenya (approved):

gianarb:
Hello! Thank you for your contribution! I had a chat with @mdelapenya via Slack and I would like to know what you think about this:

```
	CopyFileFromContainer(ctx context.Context, filePath string) ([]byte, error)
```

I see how your function will may come from what we have done with `CopyFileToContainer` but now that I see it if I think we made a mistake there (I think we should deprecate that in favor of something else).

I think the right signature should be:

```
``` 

This works a bit better over the network and it five more control to the end user who can decide what to do with an `io.Reader`.

I understand the extra code you wrote sounds like a useful utility function but at the end it can be replaced with a few line as explained here https://pkg.go.dev/io/ioutil#ReadAll and I don't think it should be something to include as part of the testcontainer library.

What do you think?

codepitbull:
Hey @gianarb,
thanks for your feedback :)
I have updated my Pr to use the signature you suggested.
With your suggestions in mind I also thought it would be better to eliminate all eager reading.
I wrapped the whole tar reader to enable closing of the underlying ReaderCloser and to return a valid object.
You think this would work better?

Cheers,
Jochen

codepitbull:
Hey @mdelapenya or @gianarb ,
are you interested in this? Anything else you would like to see changed/added in here?
Cheers,
Jochen

mdelapenya (approved):
This LGTM! Thanks again for this contribution and for your patience. I usually wait for @gianarb performing the merges, so let's wait for that review.

mdelapenya:
BTW, the GH action complains about source formatting in the `docker.go` file.

Could you run `./scripts/checks.sh` and verify the issue? 🙏

codepitbull:
@mdelapenya fixed it, forgot to run "go fmt" ...

codepitbull:
And just so you know what this is for:
Thanks to a few hints from @bsideup I was able to implement k3s-based integrationtests for k8s.
The only missing piece was this function.

Here is the example of what I am using it for:
https://github.com/codepitbull/k8sit/blob/main/gok3s/k8sit/testcontainer_test.go

gianarb:
Thank you for this!!

codepitbull:
Thanks to you and the other maintainers!
We are really happy with it for our integration tests :)

PR #349, opened 2021-08-19
Fix default network is not initialize when get gateway ip

mrproliu:
When I try to use docker in docker through the share UNIX socket on GitHub Action, the IP address in the Host Port wait strategy is incorrect.

I found when we trying to get the target host, the default network may not initialize: https://github.com/testcontainers/testcontainers-go/blob/59631817dd3515c7c97e3c84e0201eaa746d6773/docker.go#L895

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -892,6 +892,13 @@ func (p *DockerProvider) GetNetwork(ctx context.Context, req NetworkRequest) (ty
 
 func (p *DockerProvider) GetGatewayIP(ctx context.Context) (string, error) {
 	// Use a default network as defined in the DockerProvider
+	var err error
+	if p.defaultNetwork == "" {
+		p.defaultNetwork, err = getDefaultNetwork(ctx, p.client)
+		if err != nil {
+			return "", err
+		}
+	}
 	nw, err := p.GetNetwork(ctx, NetworkRequest{Name: p.defaultNetwork})
 	if err != nil {
 		return "", err
--- a/docker.go
+++ b/docker.go
@@ -892,8 +892,8 @@ func (p *DockerProvider) GetNetwork(ctx context.Context, req NetworkRequest) (ty
 
 func (p *DockerProvider) GetGatewayIP(ctx context.Context) (string, error) {
 	// Use a default network as defined in the DockerProvider
-	var err error
 	if p.defaultNetwork == "" {
+		var err error
 		p.defaultNetwork, err = getDefaultNetwork(ctx, p.client)
 		if err != nil {
 			return "", err
--- a/docker_test.go
+++ b/docker_test.go
@@ -1513,8 +1513,12 @@ func TestGetGatewayIP(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if _, err := provider.GetGatewayIP(context.Background()); err != nil {
+	ip, err := provider.GetGatewayIP(context.Background())
+	if err != nil {
 		t.Fatal(err)
+	}
+	if ip == "" {
+		t.Fatal("could not get gateway ip")
 	}
 }
 
--- a/docker_test.go
+++ b/docker_test.go
@@ -1506,6 +1506,18 @@ func TestContainerWithReaperNetwork(t *testing.T) {
 	assert.NotNil(t, cnt.NetworkSettings.Networks[networks[1]])
 }
 
+func TestGetGatewayIP(t *testing.T) {
+	// When using docker-compose with DinD mode, and using host port or http wait strategy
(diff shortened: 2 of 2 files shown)


docker.go
```diff
@@ -892,6 +892,13 @@ func (p *DockerProvider) GetNetwork(ctx context.Context, req NetworkRequest) (ty
 
 func (p *DockerProvider) GetGatewayIP(ctx context.Context) (string, error) {
 	// Use a default network as defined in the DockerProvider
+	var err error
+	if p.defaultNetwork == "" {
```
mdelapenya, line 896: I think you can move the error variable inside the block it will be used
```suggestion
	if p.defaultNetwork == "" {
		var err error
```

mdelapenya:
@mrproliu thanks for this contribution! I'd be great if you can share with us logs, screenshots, useful links where we can check the real problem, or even better to contribute a test reproducing the issue (although I know that replicating the DinD + GH action scenario could be difficult).

wdyt?

mrproliu:
> @mrproliu thanks for this contribution! I'd be great if you can share with us logs, screenshots, useful links where we can check the real problem, or even better to contribute a test reproducing the issue (although I know that replicating the DinD + GH action scenario could be difficult).
> 
> wdyt?

We are already trying to run in GitHub Action through DinD.
Here is the CI run link：https://github.com/apache/skywalking-satellite/runs/3358752416?check_suite_focus=true

Here is how we use GitHub Action to implement DinD by sharing socks.
https://github.com/apache/skywalking-infra-e2e/blob/main/action.yaml#L32-L35

When we use docker-compose, we need to bind the ports to the waiting policy to ensure that these ports can be accessed.
https://github.com/apache/skywalking-infra-e2e/blob/de5be41596f27867a15ab75dbad97ea538d863b8/internal/components/setup/compose.go#L144-L167

I hope this content is enough, if you are missing something, please feel free to let me know :)

wu-sheng:
Just say hi to @bsideup :P We are deeply integrating testcontainer for our e2e test infra project, https://github.com/apache/skywalking-infra-e2e

We will incubate the project in the SkyWalking and ASF. Who ever uses the skywalking-infra-e2e, they have 50% chance(or more) are going to use test container too.

docker_test.go
```diff
@@ -1506,6 +1506,18 @@ func TestContainerWithReaperNetwork(t *testing.T) {
 	assert.NotNil(t, cnt.NetworkSettings.Networks[networks[1]])
 }
 
+func TestGetGatewayIP(t *testing.T) {
```
mdelapenya, line 1509: Thanks for contributing a unit test! This is super!

I'd like to ask about the assertions part: according to the PR description:

>the IP address in the Host Port wait strategy is incorrect

Is it enough to verify that no errors are raised, or do we need to verify any other thing? 

BTW It's not a blocker on the review, I just want to make sure we are covering the new behaviour with the tests.

mrproliu: > Is it enough to verify that no errors are raised, or do we need to verify any other thing?

We are not sure about the specific value of the IP address in the GHA test or host test, so we can only verify that the IP address is not empty and there is no error. Do you think we also need to create a container with an export port to verify that this address is available?

mdelapenya: Yeah, if reproducing GHA behaviour is complex I'd keep it like it is now adding the verification of the IP address is not empty. If you can give it a try with a real container, best. But as I said, let's not block this unless needed.

mrproliu: Have added the not empty check :)

mdelapenya: offtopic: it seems the `@dalekliuhan` commiter user does not have associated an email on github

mdelapenya: If the CI passes, this LGTM :)

mdelapenya (approved):
LGTM!

Thanks for helping us out in the review process :)

docker_test.go
mrproliu: > offtopic: it seems the `@dalekliuhan` commiter user does not have associated an email on github

email address on GitHub? I have set it. Do I miss anything? :)

mdelapenya: Look at the commit in the list: it does not have an image associated to it. And when you click on the commit itself, the user is not clickable

mrproliu: Oh, sorry about that, I submit by a different user, not the GitHub user so it cannot be clickable.

wu-sheng: @mdelapenya Does this repo support `squash merge`? If so, GitHub will take care of this issue automatically.

gianarb:
Thanks a lot!

PR #354, opened 2021-09-14
Swap incorrect key and value of BindMounts and VolumeMounts

Lucaber:
It is currently not possible to mount a local path to multiple paths in
the container.

BindMounts are currently stored in a `map[string]string` with the
hostPath being the key and containerPath as the value.
This prevents me from mounting a hostPath to multiple paths in the
container. Using a map actually makes sense because it is not possible
to mount multiple host paths to the same path in the container. Key and
value are just swapped. Same with VolumeMounts.

Yes, i know, this a huge breaking change and probably not the best fix.
We could also add an additional map like `BindMountsFixed` and deprecate
the old one. But as this library is still in version 0, a breaking
change might not be a problem?

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -679,14 +679,14 @@ func (p *DockerProvider) CreateContainer(ctx context.Context, req ContainerReque
 
 	// prepare mounts
 	mounts := []mount.Mount{}
-	for hostPath, innerPath := range req.BindMounts {
+	for innerPath, hostPath := range req.BindMounts {
 		mounts = append(mounts, mount.Mount{
 			Type:   mount.TypeBind,
 			Source: hostPath,
 			Target: innerPath,
 		})
 	}
-	for volumeName, innerPath := range req.VolumeMounts {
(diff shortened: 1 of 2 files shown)


mdelapenya (approved):
Hi @Lucaber this LGTM, thanks for your contribution!

I'm ok with the implications on these breaking changes, as I understand your user case, but I'd like to listen to the community first before merging and releasing a new version.

Please give us a few more time for that 🙏

@gianarb what are your thoughts on this?

Lucaber:
Can we get this merged? @mdelapenya @gianarb

thueske:
@mdelapenya @gianarb any updates on this? :)

gianarb:
Thanks! @bsideup do you mind to tweet about this bc break? Thanks!

bsideup:
Hi @gianarb! I would be happy to retweet, but I am not sure I need to be the one tweeting, or at least I don't see why 😅

PR #358, opened 2021-10-01
Close docker provider client on container terminate

anjmao:


mdelapenya (approved):
This LGTM and thanks for your contribution!

anjmao:
@mdelapenya I'm not sure about coverage report failure. It says that my changed was not covered by tests, but there are many tests which calls Terminate method.

mdelapenya:
> @mdelapenya I'm not sure about coverage report failure. It says that my changed was not covered by tests, but there are many tests which calls Terminate method.

... but probably not entering the `if err != nil` block. So I guess the CI build did not cause the Docker client to fail while closing :)

PR #362, opened 2021-10-04
Expose configuration of logger

rcrowe:
Allow the log instance to be configured independently from other uses of the `log` package.

```go
testcontainers.Logger.SetPrefix(io.Discard)
testcontainers.Logger.SetOutput(io.Discard)
```

This is still global & it would be better to be able to pass in an instance of `*log.Logger` but that can be a follow-up PR. I wanted to land something small & hopefully non-contencious.

### DIFF
--- a/container.go
+++ b/container.go
@@ -3,6 +3,8 @@ package testcontainers
 import (
 	"context"
 	"io"
+	"log"
+	"os"
 
 	"github.com/docker/docker/api/types"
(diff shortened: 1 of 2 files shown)


mdelapenya (approved):

mdelapenya:
Thanks for the contribution! I agree that having a setter/confusion option for the log would be better, so feel free to send it in a follow up PR or even saying more commits on top, whatever you prefer

PR #369, opened 2021-10-23
Bugfix: log-consumer go-routine should recover from closed-connection

Chrisss93:
As best I can tell, a direct commit to the master branch [bc6fdab](https://github.com/testcontainers/testcontainers-go/commit/bc6fdab92dd3c623ad6654bf301b1a659efeb50e) made an attempt at recovering the log-position of a container in the scenario of a closed connection. The logic is all sound, but I think the string used to detect a closed-connection is not (has never been) correct.

The source of the error is `internal/poll`'s `errNetClosing`. It's error string has been set as: `"use of closed network connection"` since the beginning (5 years ago). Of course it would be much cleaner to detect a connection-closed error without string-matching but as stated in the mentioned commit, the error type for this was only exported by the `net` package in go 1.16

### DIFF
--- a/logconsumer_test.go
+++ b/logconsumer_test.go
@@ -7,6 +7,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/docker/docker/client"
 	"github.com/testcontainers/testcontainers-go/wait"
 	"gotest.tools/assert"
 )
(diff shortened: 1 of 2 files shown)


logconsumer_test.go
```diff
@@ -181,3 +182,103 @@ func Test_ShouldRecognizeLogTypes(t *testing.T) {
...
+
+	var consumer TestLogConsumer
+	if err = nginx.StartLogProducer(ctx); err != nil {
+		t.Fatal(err)
+	}
+	defer nginx.StopLogProducer()
+	nginx.FollowOutput(&consumer)
+
+	// Gather the initial container logs
+	time.Sleep(time.Second * 1)
```
mdelapenya, line 243: Do you foresee medium-term, potential issues with the sleep time not being enough?

mdelapenya (approved):
Thanks for this contribution @Chrisss93 ! I do not see anything weird with it, so LGTM

About https://github.com/testcontainers/testcontainers-go/commit/bc6fdab92dd3c623ad6654bf301b1a659efeb50e, it was not a direct commit to master but a PR: https://github.com/testcontainers/testcontainers-go/pull/323 😉

gianarb (approved):
Thanks a lot!

PR #371, opened 2021-10-26
fix: Support compose v2 naming scheme.

jaysonsantos:
It uses `-` as the separator instead of `_`


mdelapenya:
Hey @jaysonsantos I've noticed this PR exists: https://github.com/testcontainers/testcontainers-go/pull/377

Do you see they are overlapping?

PR #373, opened 2021-10-31
feat: support reading DOCKER_HOST from testcontainers props file

mdelapenya:
## What does this PR do?
It adds support for retrieving the Docker host (remote or local) reading the DOCKER_HOST value from Testcontainers' default properties file (which is used at testcontainers-java for quite some time).

That file is located at `~/.testcontainers.properties)`.

The PR includes unit tests for the new methods reading the props file.

## Why is it important?
There are two main reasons:

1. consistency across Testcontainers flavours
2. support running the containers ina remote Docker Host

## Follow-up concerns 
I've not tested this feature with Docker Compose yet (hopefully I can get some feedback from you all first)


docker.go
```diff
@@ -516,17 +518,63 @@ var _ ContainerProvider = (*DockerProvider)(nil)
...
 }
 
+// readTCPropsFile reads from testcontainers properties file, if it exists
+func readTCPropsFile() string {
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return ""
+	}
+	tcProp := path.Join(home, ".testcontainers.properties")
+	content, err := os.ReadFile(tcProp)
```
mdelapenya, line 561: We should use a backwards compatible way of reading files OR properly define a support matrix

mdelapenya, line 561:
```suggestion
	content, err := ioutil.ReadFile(tcProp)
```

docker.go
```diff
@@ -514,19 +516,71 @@ type DockerProvider struct {
...
+	Host      string `properties:"docker.host"`
+	TLSVerify int    `properties:"docker.tls.verify,default=0"`
+	CertPath  string `properties:"docker.cert.path,default="`
+}
+
 // NewDockerProvider creates a Docker provider with the EnvClient
 func NewDockerProvider() (*DockerProvider, error) {
-	client, err := client.NewEnvClient()
+	tcConfig := readTCPropsFile()
+	host := tcConfig.Host
```
bsideup, line 529: `TLSVerify` and `CertPath` seem to be ignored and not set on the `client`

mdelapenya: Yes, I'm not using them yet, although planning to do it in follow-up commits. Will mark this PR as draft meanwhile

mdelapenya: Used in https://github.com/testcontainers/testcontainers-go/pull/373/commits/af1986a417a441e72963e27b370c61d3053105f0

docker_test.go
```diff
@@ -1187,6 +1189,115 @@ func TestEntrypoint(t *testing.T) {
...
+			defer func() {
+				os.Setenv("HOME", oldHome)
+				filet.CleanUp(t)
+			}()
+
+			_ = filet.File(t, path.Join(tmpDir, ".testcontainers.properties"), tt.content)
+
+			config := readTCPropsFile()
+
+			assert.Equal(t, tt.expectedHost, config.Host, "Hosts do not match")
```
mdelapenya, line 1294: @kiview please take a look at the expected values for `Host` in the test struct, as reading the props file with `magiconair` will always set the value as the last entry in the props file.

kiview: Right, that is actually the correct way how `properties` files should be read, so the hand-crafted implementation was flawed in this regard.

kiview (approved):
Looks great from the perspective of aligning more closely with `testcontainers-java`, thanks 👍

bsideup (approved):
👍

docker.go
```diff
@@ -514,19 +516,76 @@ type DockerProvider struct {
 
 var _ ContainerProvider = (*DockerProvider)(nil)
 
+// or through Decode
+type TestContainersConfig struct {
+	Host      string `properties:"docker.host"`
```
mdelapenya, line 521: Here we are setting host as required. Is it correct?

bsideup: ah, I missed this one. The file may exist but `docker.host` can be missing in it, so I would recommend making it optional as well 👍

mdelapenya:
To fix the code coverage warnings in this PR, I think we should add a remote docker daemon, configure it with TLS, and create a DockerProvider struct with a TC properties pointing to that remote. Do you think we can live without that test?

bsideup:
@mdelapenya given the complexity of such test that would delay this PR by a significant amount of time, I suggest we stick to the manual testing of it. That's what we also do in testcontainers-java :)

docker.go
```diff
@@ -514,19 +516,76 @@ type DockerProvider struct {
...
+}
+
 // NewDockerProvider creates a Docker provider with the EnvClient
 func NewDockerProvider() (*DockerProvider, error) {
-	client, err := client.NewEnvClient()
+	tcConfig := readTCPropsFile()
+	host := tcConfig.Host
+
+	opts := []client.Opt{client.FromEnv}
+	if host != "" {
```
bsideup, line 532: if dockerhost is not set, we should ignore other settings (e.g. TLS), as they come together with dockerhost and make no sense without it, so it can be one big `if` :)

mdelapenya:
@gianarb if you agree, I'd like to merge this one by the end of today. Otherwise feel free to add your thoughts

Thanks!

docker.go
mdelapenya: Addressed in https://github.com/testcontainers/testcontainers-go/pull/373/commits/59a295c5c337a81a41d8b4da211eb22974555f24

PR #380, opened 2021-11-15
docs: add snippet to prevent dangling volumes

peter-evans:
Ran into an issue where a container in a compose stack was creating an anonymous volume. The `Down()` method doesn't remove it and I had to pass the `--volumes` option to prevent dangling volumes accumulating on the test runner.

Great library, by the way. Thank you! 🏅

### DIFF
--- a/docs/features/docker_compose.md
+++ b/docs/features/docker_compose.md
@@ -51,3 +51,10 @@ if err != nil {
 return nil
 ```
 
+If your compose stack creates volumes you may need to remove them with the `--volumes` option to prevent dangling volumes accumulating on test runners.
+
+```go
+execError := compose.
+	WithCommand([]string{"down", "--volumes"}).
+	Invoke()
+```


docs/features/docker_compose.md
```diff
@@ -51,3 +51,10 @@ if err != nil {
 return nil
 ```
 
+If your compose stack creates volumes you may need to remove them with the `--volumes` option to prevent dangling volumes accumulating on test runners.
```
mdelapenya, line 54: Do you thing is a flag we should always pass to the `down` method?

mdelapenya (approved):
Thanks for this PR! It's great that we are taking care of the docs. I added a comment about always destroying volumes, wdyt?

Other than that, LGTM

docs/features/docker_compose.md
peter-evans: Yes, it would probably be a good default for testcontainers because it's very unlikely that any users are relying on these volumes persisting between compose runs. For test usage we just want everything to be cleared up.

mdelapenya: Would you mind adding it please? 🙏

peter-evans: Sure. I will send a PR soon.

peter-evans: Please see https://github.com/testcontainers/testcontainers-go/pull/381

peter-evans:
@mdelapenya Should I close this in favour of https://github.com/testcontainers/testcontainers-go/pull/381, or would you like to keep it as an option for now?

mdelapenya:
> @mdelapenya Should I close this in favour of #381, or would you like to keep it as an option for now?

Indeed. Let's close this one and keep #381 which I already approved.

Thanks!

PR #381, opened 2021-11-16
feat: additionally remove volumes by default on compose down

peter-evans:
Following the discussion [here](https://github.com/testcontainers/testcontainers-go/pull/380#discussion_r749166320) this PR adds removal of volumes by default to the compose `Down()` method.

### DIFF
--- a/compose_test.go
+++ b/compose_test.go
@@ -368,6 +368,34 @@ func TestLocalDockerComposeWithMultipleComposeFiles(t *testing.T) {
 	assertContainerEnvironmentVariables(t, containerNameNginx, present, absent)
 }
 
+func TestLocalDockerComposeWithVolume(t *testing.T) {
+	path := "./testresources/docker-compose-volume.yml"
+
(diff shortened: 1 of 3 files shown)


compose_test.go
```diff
@@ -403,7 +431,7 @@ func executeAndGetOutput(command string, args []string) (string, ExecError) {
 	cmd := exec.Command(command, args...)
 	out, err := cmd.CombinedOutput()
 	if err != nil {
-		return "", ExecError{Error: err}
+		return string(out), ExecError{Error: err}
```
peter-evans, line 434: Using `cmd.CombinedOutput()` means that stderr is combined in `out`. I added returning it here so that I could detect the `No such volume` error.

mdelapenya: Neat!

mdelapenya (approved):
Great job @peter-evans here! Thanks for your time contributing to the library :)

mdelapenya:
Merged, thanks!

PR #382, opened 2021-11-17
Read only bind and volume mounts

silh:
Added the ability to specify read-only bind and volume mounts.
That ability is present in docker client itself as well as in java version of testcontainers - see https://www.testcontainers.org/features/files/

### DIFF
--- a/container.go
+++ b/container.go
@@ -88,8 +88,8 @@ type ContainerRequest struct {
 	ExposedPorts    []string // allow specifying protocol info
 	Cmd             []string
 	Labels          map[string]string
-	BindMounts      map[string]string
-	VolumeMounts    map[string]string
+	BindMounts      []Mount
+	VolumeMounts    []Mount
 	Tmpfs           map[string]string
 	RegistryCred    string
 	WaitingFor      wait.Strategy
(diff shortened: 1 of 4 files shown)


mdelapenya:
Thanks for this contribution, it makes total sense.

I wonder whether directly merging it as a breaking change in next minor (we are still in 0.x.y) or include it in an eventual 1.0.0 release 🤔

@gianarb thoughts?

mdelapenya (approved):

gianarb:
I would like to go with the `:ro` `:rw` parsing if possible! That syntax is stable since Docker showed up and I don't think it will change in the future :+1:

silh:
> I would like to go with the `:ro` `:rw` parsing if possible! That syntax is stable since Docker showed up and I don't think it will change in the future 👍

Did it with `:ro` and `:rw: but while rebasing I saw that there was already breaking change which might create confusion with  order of host path and container path in those maps and I think the structure would have been more clear.

silh:
Is anything else needed?
I am not sure about the coverage drop in `wait/http.go` - it seems flacky.

mdelapenya:
Can you double check if there are incompatible changes with https://github.com/testcontainers/testcontainers-go/pull/386?

silh:
I think #386 make this PR obsolete and it can be closed.
There is already a way there to specify if you want a readonly container.

mdelapenya:
Closed as per @silh's comment. Thanks you very much for your time with testcontainers-go!

image

Issue #336, opened 2021-07-29
support podman as alternative to docker


candlerb:
I wondered if you had considered supporting [podman](https://podman.io/) for running containers, as an alternative backend to docker?

Using podman (and buildah for creating container images) lets you get rid of the docker daemon entirely.

There is a [go API](https://podman.io/blogs/2020/08/10/podman-go-bindings.html).  This does involve running a podman service: but unlike docker, the service runs as your own userid, with no elevated privileges, and can be started and stopped by the user themselves.

candlerb:
UPDATE: it looks like very little effort should be required to work with podman 3+.  It is compatible with the docker API, even to the point where you can run native "docker compose".  That works as follows:

* Start a global podman service: `systemctl enable --now podman.socket`
* For RHEL/CentOS/Arch there's a package "podman-docker" which installs symlinks such that `/var/run/docker.sock` points to the podman socket.
* For Debian/Ubuntu you need to set `DOCKER_HOST=unix:///run/podman/podman.sock`: see https://thesynack.com/posts/docker-compose-podman/

If you're not using `docker compose`, you don't need the global podman service anyway.  Start a user service instead:

```
$ systemctl start --user podman.socket
```

This listens on `/run/user/1000/podman/podman.sock` (for example).

However, the `DOCKER_HOST` environment variable doesn't appear to be completely honoured by testcontainers-go.  If I run a test without it, I get a failure as expected:

```
$ go test .
--- FAIL: TestNginxLatestReturn (0.00s)
    main_test.go:25: failed to create container: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x8cfb77]

goroutine 19 [running]:
testing.tRunner.func1.2(0x9445c0, 0xd44680)
	/usr/local/go/src/testing/testing.go:1143 +0x332
testing.tRunner.func1(0xc000082a80)
	/usr/local/go/src/testing/testing.go:1146 +0x4b6
panic(0x9445c0, 0xd44680)
	/usr/local/go/src/runtime/panic.go:965 +0x1b9
example.TestNginxLatestReturn(0xc000082a80)
	/home/nsrc/tmp/testcontainers/main_test.go:27 +0x257
[...]

PR #386, opened 2021-11-24
Replace ContainerRequest.BindMounts and ContainerRequest.VolumeMounts with ContainerRequest.Mounts as dedicated type

baez90:
# What does this PR do?

- introduce dedicated type for mounts
- introduce generic and Docker specific types for different mount sources (bind, volume, tmpfs)
- introduce convenience methods to support common mount scenarios
- add validation to avoid duplicate mount targets

# Why is it important?

The current map based data structure is somehow confusing. This change avoids confusion by introducing dedicated types for all components to get help by IDEs and the compiler.

# How to migrate

## Simple bind mount

```go
req := ContainerRequest {
    		//...
    		BindMounts: map[string]string{
		    "/var/run/docker.sock": "/var/run/docker.sock",
    		}
    		//...
}
```

becomes:

```go
req := ContainerRequest {
    		//...
    		Mounts: Mounts(BindMount("/var/run/docker.sock", "/var/run/docker.sock")),
    		//...
}
```

## Multiple mounts including a volume mount

```go
req := ContainerRequest {
    		//...
    		BindMounts: map[string]string {
		    // container path : host path
		    "/hello.sh", absPath
    		},
    		VolumeMounts: map[string]string {
		    // container path : volume name
		    "/data": volumeName
    		//...
}
```

becomes

```go
req := ContainerRequest {
    		//...
    		Mounts: Mounts(BindMount(absPath, "/hello.sh"), VolumeMount(volumeName, "/data")),
    		//...
}
```

### DIFF
--- a/mounts.go
+++ b/mounts.go
@@ -1,0 +1,136 @@
+package testcontainers
+
+import (
+	"fmt"
+
+	"github.com/docker/docker/api/types/mount"
+)
+
+type BindMounter interface {
+	GetBindOptions() *mount.BindOptions
+}
+
+type VolumeMounter interface {
+	GetVolumeOptions() *mount.VolumeOptions
+}
+
+type TmpfsMounter interface {
+	GetTmpfsOptions() *mount.TmpfsOptions
+}
+
+type ContainerMountSource interface {
+	fmt.Stringer
+	Type() mount.Type
+}
+
+type BindMountSource struct {
+	*mount.BindOptions
+	HostPath string
+}
+
+func (s BindMountSource) String() string {
+	return s.HostPath
+}
+
+func (BindMountSource) Type() mount.Type {
+	return mount.TypeBind
+}
+func (s BindMountSource) GetBindOptions() *mount.BindOptions {
+	return s.BindOptions
+}
+
+type VolumeMountSource struct {
+	*mount.VolumeOptions
+	Name string
+}
+
+func (s VolumeMountSource) String() string {
+	return s.Name
+}
+
+func (VolumeMountSource) Type() mount.Type {
+	return mount.TypeVolume
+}
+
+func (s VolumeMountSource) GetVolumeOptions() *mount.VolumeOptions {
+	return s.VolumeOptions
+}
+
+type TmpfsMountSource struct {
+	*mount.TmpfsOptions
+}
+
+func (s TmpfsMountSource) String() string {
+	return ""
+}
+
+func (TmpfsMountSource) Type() mount.Type {
+	return mount.TypeTmpfs
+}
+
+func (s TmpfsMountSource) GetTmpfsOptions() *mount.TmpfsOptions {
+	return s.TmpfsOptions
+}
+
+type ContainerMountTarget string
+
+func (t ContainerMountTarget) String() string {
+	return string(t)
+}
+
+func BindMount(hostPath string, mountTarget ContainerMountTarget) ContainerMount {
+	return ContainerMount{
+		Source: BindMountSource{HostPath: hostPath},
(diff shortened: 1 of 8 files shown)


mounts.go
```diff
@@ -0,0 +1,169 @@
+package testcontainers
+
+import (
+	"github.com/docker/docker/api/types/mount"
```
silh, line 4: As I understood the already existing code the things related to particular container backend (in this case - docker) are kept in the `docker.go` so that the testcontainers implementation is decoupled from it (although there is right now only one container backend provider).

baez90: I see. To be honest after [this comment](https://github.com/testcontainers/testcontainers-go/issues/336#issuecomment-891667819) I assumed it would be okay to rely on Docker API types but I refactored my draft to be Docker API independent, moved all the Docker related code to `mounts_docker.go` and split the mount sources into generic, container provider agnostic variants and added Docker specific ones in case one would like to be able to provide e.g. volume options.

I hope this matches better the current code structure?

silh (approved):

baez90:
@mdelapenya probably any feedback on this? :blush:

mounts_docker.go
```diff
@@ -0,0 +1,116 @@
+package testcontainers
```
mdelapenya, line 1: NIT: I'd rename this file to `docker_mounts.go` as it's used as `..DockerMounts...` everywhere.

mounts_docker.go
```diff
@@ -0,0 +1,116 @@
...
+	*mount.TmpfsOptions
+}
+
+func (s DockerTmpfsMountSource) GetTmpfsOptions() *mount.TmpfsOptions {
+	return s.TmpfsOptions
+}
+
+// MapToDockerMounts maps the given []ContainerMount to the corresponding
+// []mount.Mount for further processing
+func MapToDockerMounts(containerMounts ContainerMounts) []mount.Mount {
```
mdelapenya, line 83: Do we want to expose this method externally? Not against it, but I'm simply double-checking it's on purpose

mounts_docker.go
```diff
@@ -0,0 +1,116 @@
+package testcontainers
+
+import "github.com/docker/docker/api/types/mount"
+
+var (
+	mountTypeMapping = map[MountType]mount.Type{
+		MountTypeBind:   mount.TypeBind,
+		MountTypeVolume: mount.TypeVolume,
+		MountTypeTmpfs:  mount.TypeTmpfs,
+		MountTypePipe:   mount.TypeNamedPipe,
```
mdelapenya, line 10: I cannot see an implementation for `MountTypePipe` but there are for Bind, Volume and Tmpfs. Is this on purpose?

mdelapenya (approved):
Hey @baez90, thanks for this great PR, I added a few comments, although I'm OK with the changes. Let'd discuss them before moving forward

Thanks in advance and sorry for the late response, I was on PTO past week

baez90:
Thanks @mdelapenya  for your kind words and the feedback!
Don't you worry, I just wasn't sure if there's something missing or if i forgot something or the like :sweat_smile: 

I tried to 'resolve' everything you mentioned.
Anything else that pops into mind?

gianarb:
Can we get a note about a migration for this breaking change? I would like to place it at the top of this PR and in the changelog! Let's get this bc break ready to get merged!

baez90:
Sure thing! I tried to improve the title and added some examples how to migrate in the PR description.
Something like that or did you already have some specific ideas? :blush:

gianarb:
Thanks a lot for your help with this!

PR #387, opened 2021-12-01
Upgrade dependencies to latest versions

baez90:
Some of the dependencies are already one or multiple major versions behind upstream.
For instance `gopkg.in/yaml.v2` in the previous version is known to be vulnerable.

Also `github.com/Flaque/filet` can easily be replaced with functions from the standard library and `github.com/pkg/errors` is an archived project and replaced with Go 1.13 error extensions.

To get rid of at least as many warnings GoLand reports as possible I made all error ignores explicit where they weren't in a `defer` call. The remaining ones could be replaced - at least in test cases - with a neat helper method?


docker_test.go
```diff
@@ -1191,34 +1190,24 @@ func TestEntrypoint(t *testing.T) {
 
 func TestReadTCPropsFile(t *testing.T) {
 	t.Run("HOME is not set", func(t *testing.T) {
-		oldHome := os.Getenv("HOME")
-		os.Unsetenv("HOME")
-		defer func() {
-			os.Setenv("HOME", oldHome)
-		}()
+		tmpDir := t.TempDir()
```
mdelapenya, line 1193: Neat!

docker_test.go
```diff
@@ -1191,34 +1190,24 @@ func TestEntrypoint(t *testing.T) {
 
 func TestReadTCPropsFile(t *testing.T) {
 	t.Run("HOME is not set", func(t *testing.T) {
-		oldHome := os.Getenv("HOME")
-		os.Unsetenv("HOME")
-		defer func() {
-			os.Setenv("HOME", oldHome)
```
mdelapenya, line 1197: Is it required to avoid resetting the env to previous value of the HOME variable?

baez90: Nope, according to [the docs](https://pkg.go.dev/testing#T.Setenv) that's automatically done after the test is completed

baez90: I just saw this is brand new in Go 1.17 :sweat_smile: shall I revert it or is it okay to update the test environment?
I don't expect this to break compatibility for the library itself but I'm not sure about it?

baez90: I replaced the 'fancy' standard lib calls with `gotest.tools` pendants to get compatibility with Go 1.14 again. The dependency was already in place and this way I still can remove filet and the env is automatically managed :blush:

mdelapenya (approved):

baez90:
@mdelapenya  sorry, hat to rebase it once more to switch back to `ioutil.WriteFile` instead of `os.WriteFile` to be compatible with Go 1.13 :see_no_evil:

gianarb (approved):
Thanks a lot for this! Let's get it checked in!

PR #392, opened 2021-12-15
#391: remember service map from all docker compose files

### LINKED ISSUE #391: Creating containers with multiple docker compose file does not returns correct map of services
**Describe the bug**
If you use multiple docker compose files to start containers, the code only keeps the service list mentioned in the last docker-compose file passed. 

**To Reproduce**
I will add a PR to fix this issue with the unit test.

ankan-pfc:
Fixes #391 

I added a new docker-compose file with postgres image. Hope thats okay.
Without this fix, services field keeps map for only the last docker-compose file passed. But there could be service that are not overridden in the last file, and it should keep those values in the map.

### DIFF
--- a/compose.go
+++ b/compose.go
@@ -198,7 +198,13 @@ func (dc *LocalDockerCompose) validate() error {
 			return err
 		}
 
-		dc.Services = c.Services
+		if dc.Services == nil {
+			dc.Services = c.Services
+		} else {
+			for k, v := range c.Services {
+				dc.Services[k] = v
+			}
+		}
 	}
 
 	return nil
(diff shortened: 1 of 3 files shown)


compose.go
```diff
@@ -198,7 +198,13 @@ func (dc *LocalDockerCompose) validate() error {
 			return err
 		}
 
-		dc.Services = c.Services
+		if dc.Services == nil {
+			dc.Services = c.Services
+		} else {
+			for k, v := range c.Services {
+				dc.Services[k] = v
+			}
```
mdelapenya, line 206: Good catch!

mdelapenya (approved):
Hey @ankan-pfc thank you for the fix! It was there looking at us without being noticed :)

LGTM

PR #393, opened 2021-12-24
Logger per container instance

baez90:
# What does this PR do?

Follow up PR for #385 to configure logging instance per container.
To achieve this, it adds variadic options to multiple constructors to propagate the logger finally to the `Container` instance.

## Affected functions

- `NewLocalDockerCompose(...)`
- `GetProvider(...)`
- `NewDockerProvider(...)`

## Options pattern

The idea is that there's a `GenericProviderOption` interface to act exactly as the name intends: all providers should support these options. Options only supported by Docker can be added as `DockerProviderOption`.
There's a converter function `Generic2DockerOptions` to wrap `GenericProviderOption` as `DockerProviderOption` for `NewDockerProvider`.

Everywhere a logger option can be applied the default is the global logger which can be overridden.

The `WithLogger` function takes a `Logger` and returns a option instance that can be used for all option types:

- `GenericProviderOption`
- `DockerProviderOption`
- `LocalDockerComposeOption`

## Test logger

The `TestLogger` function takes a `testing.TB` and wraps it to a `Logger` instance allowing to capture all logs from a container instance as output of a specific test case.

# Why is it important?

This way logs from the container instance can be aggregated with the rest of the test output they were created for.

### DIFF
--- a/compose.go
+++ b/compose.go
@@ -44,6 +44,7 @@ type waitService struct {
 // LocalDockerCompose represents a Docker Compose execution using local binary
 // docker-compose or docker-compose.exe, depending on the underlying platform
 type LocalDockerCompose struct {
+	*LocalDockerComposeOptions
 	Executable           string
 	ComposeFilePaths     []string
 	absComposeFilePaths  []string
@@ -55,6 +56,26 @@ type LocalDockerCompose struct {
 	WaitStrategyMap      map[waitService]wait.Strategy
 }
 
+type (
+	// LocalDockerComposeOptions defines options applicable to LocalDockerCompose
+	LocalDockerComposeOptions struct {
+		Logger Logging
+	}
+
+	// LocalDockerComposeOption defines a common interface to modify LocalDockerComposeOptions
+	// These options can be passed to NewLocalDockerCompose in a variadic way to customize the returned LocalDockerCompose instance
+	LocalDockerComposeOption interface {
+		ApplyToLocalCompose(opts *LocalDockerComposeOptions)
+	}
+
+	// LocalDockerComposeOptionsFunc is a shorthand to implement the LocalDockerComposeOption interface
+	LocalDockerComposeOptionsFunc func(opts *LocalDockerComposeOptions)
+)
+
+func (f LocalDockerComposeOptionsFunc) ApplyToLocalCompose(opts *LocalDockerComposeOptions) {
+	f(opts)
+}
+
 // NewLocalDockerCompose returns an instance of the local Docker Compose, using an
 // array of Docker Compose file paths and an identifier for the Compose execution.
 //
@@ -62,8 +83,16 @@ type LocalDockerCompose struct {
 // Docker Compose execution. The identifier represents the name of the execution,
 // which will define the name of the underlying Docker network and the name of the
 // running Compose services.
-func NewLocalDockerCompose(filePaths []string, identifier string) *LocalDockerCompose {
-	dc := &LocalDockerCompose{}
+func NewLocalDockerCompose(filePaths []string, identifier string, opts ...LocalDockerComposeOption) *LocalDockerCompose {
+	dc := &LocalDockerCompose{
+		LocalDockerComposeOptions: &LocalDockerComposeOptions{
+			Logger: Logger,
+		},
+	}
+
+	for idx := range opts {
+		opts[idx].ApplyToLocalCompose(dc.LocalDockerComposeOptions)
+	}
 
 	dc.Executable = "docker-compose"
 	if runtime.GOOS == "windows" {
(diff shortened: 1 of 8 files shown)


baez90:
Just a rough first draft to clarify the idea and make discussion probably a little bit easier

compose.go
```diff
@@ -55,15 +56,43 @@ type LocalDockerCompose struct {
...
 // NewLocalDockerCompose returns an instance of the local Docker Compose, using an
 // array of Docker Compose file paths and an identifier for the Compose execution.
 //
 // It will iterate through the array adding '-f compose-file-path' flags to the local
 // Docker Compose execution. The identifier represents the name of the execution,
 // which will define the name of the underlying Docker network and the name of the
 // running Compose services.
-func NewLocalDockerCompose(filePaths []string, identifier string) *LocalDockerCompose {
-	dc := &LocalDockerCompose{}
+func NewLocalDockerCompose(filePaths []string, identifier string, opts ...LocalDockerComposeOption) *LocalDockerCompose {
```
mdelapenya, line 86: This change would break the API of those using this method. Have you though maybe creating a fluent builder pattern, something like:

```golang
func (dc *LocalDockerCompose) WithComposeOptions(opts ...LocalDockerComposeOption) *LocalDockerCompose {
     ...
    return dc
}

This way, consumers won't be affected and could be able to customise the compose instance if needed 🤔

Comment (baez90) [reply compose.go:86]: :smile: Honestly I'm not such a huge fan of fluent builders in Go due to the constraints where I **must** put the `.` :smile:  but if you prefer it that way I can try to refactor it :blush: 

As you already noticed I only updated the tests to already use the logger option both for better output and demonstration :blush: 

I think this should not break the API, does it? :thinking:

Comment (mdelapenya) [reply compose.go:86]: Nope, I updated my comment in UPDATE 2 :) So it's fine for me in terms of that API. Will take a deeper look with the description you provided, thanks for updating it

Comment (baez90) [reply compose.go:80]: ah sorry, didn't check when you updated the first comment :sweat_smile:

**Comment by baez90** [comment]
I just added one more change that unifies the logger option to a single one for all option interfaces. Furthermore the convert function now checks if the type already implements the corresponding interface and by renaming the interface methods it's now possible to implement all interfaces for one type :sweat_smile: 

Let me know if you'd like more description/any changes :blush: 
Thanks for your feedback!

File: docker.go [hunk 2]
```diff
@@ -525,11 +527,58 @@ type TestContainersConfig struct {
...
+				o.ApplyGenericTo(opts.GenericProviderOptions)
+			}))
+		}
+
+	}
+
+	return converted
+}
+
+func DockerWithLogger(logger Logging) DockerProviderOption {
```
baez90, line 561: Remove as not needed anymore

logger.go
```diff
@@ -11,4 +12,38 @@ var Logger Logging = log.New(os.Stderr, "", log.LstdFlags)
 // Logging defines the Logger interface
 type Logging interface {
 	Printf(format string, v ...interface{})
-}
\ No newline at end of file
+}
+
+func TestLogger(tb testing.TB) Logging {
+	return testLogger{TB: tb}
```
baez90, line 18: Call `tb.Helper()`

logger.go
```diff
@@ -11,4 +12,38 @@ var Logger Logging = log.New(os.Stderr, "", log.LstdFlags)
...
+func (o LoggerOption) ApplyToLocalCompose(opts *LocalDockerComposeOptions) {
+	opts.Logger = o.logger
+}
+
+type testLogger struct {
+	testing.TB
+}
+
+func (t testLogger) Printf(format string, v ...interface{}) {
+	t.Logf(format, v...)
```
baez90, line 48: Call `tb.Helper()`

baez90:
@mdelapenya anything missing here from my side?
Just confirming I didn't forget to do/add something :blush: no need to hurry but at least I can check this from my notifications :smile:

mdelapenya:
> @mdelapenya anything missing here from my side? Just confirming I didn't forget to do/add something 😊 no need to hurry but at least I can check this from my notifications 😄

Sorry for the delay... a lot of stuff moving on my day-to-day. Added a simple comment and pinged @gianarb for final review, as this change involves non-breaking architectural changes

Thanks for your dedication here @baez90

baez90:
Don't you worry, I didn't mean to put you under stress! This isn't urgent :-)

logger.go
```diff
@@ -11,4 +12,45 @@ var Logger Logging = log.New(os.Stderr, "", log.LstdFlags)
 // Logging defines the Logger interface
 type Logging interface {
 	Printf(format string, v ...interface{})
-}
\ No newline at end of file
+}
+
+// TestLogger returns a Logging implementation for testing.TB
+// This way logs from testcontainers are part of the test output of a test suite or test case
+func TestLogger(tb testing.TB) Logging {
```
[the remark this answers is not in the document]
mdelapenya: Oh I see your point now: providing a test logger for using in client code's... 🤔 I think this makes sense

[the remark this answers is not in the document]
baez90: Probably the current 'Test logger' section in the PR description lacks some details/explanations?
If you could give me a hint what's confusing/missing I could elaborate a bit more on this part?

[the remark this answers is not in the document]
mdelapenya: After reading that section more carefully, I think it's well described. Sorry for the confusion and the time for the review. This PR LGTM. Waiting for @gianarb' final review

mdelapenya (approved):

gianarb (approved):
This is a lot of work! Thank you for that! I am wondering if we should add a chapter in the documentation about logging

PR #394, opened 2022-01-04
feat: add IPAM configs to `NetworkRequest`

### LINKED ISSUE #299: Allow to configure IPAM for networks
Right now I can't configure IPAM parameters for the network that I create, but I see these parameters in this struct (the Go docker client):
```
// NetworkCreate is the expected body of the "create network" http request message
type NetworkCreate struct {
	// Check for networks with duplicate names.
	// Network is primarily keyed based on a random ID and not on the name.
	// Network name is strictly a user-friendly alias to the network
	// which is uniquely identified using ID.
	// And there is no guaranteed way to check for duplicates.
	// Option CheckDuplicate is there to provide a best effort checking of any networks
	// which has the same name but it is not guaranteed to catch all name collisions.
	CheckDuplicate bool
	Driver         string
	Scope          string
	EnableIPv6     bool
	IPAM           *network.IPAM
	Internal       bool
	Attachable     bool
	Ingress        bool
	ConfigOnly     bool
	ConfigFrom     *network.ConfigReference
	Options        map[string]string
	Labels         map[string]string
}
```
It would be nice if the testcontainers allows configuring this parameter.

aivinog1:
I've added a new field for `NetworkRequest`

closes #299

### DIFF
--- a/network_test.go
+++ b/network_test.go
@@ -3,6 +3,8 @@ package testcontainers
 import (
 	"context"
 	"fmt"
+	"github.com/docker/docker/api/types/network"
+	"github.com/stretchr/testify/assert"
 	"github.com/testcontainers/testcontainers-go/wait"
 	"testing"
 	"time"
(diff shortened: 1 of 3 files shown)


network_test.go
```diff
@@ -34,6 +36,51 @@ func ExampleNetworkProvider_CreateNetwork() {
...
+	net, _ := GenericNetwork(ctx, GenericNetworkRequest{
+		NetworkRequest: NetworkRequest{
+			Name:           networkName,
+			CheckDuplicate: true,
+			IPAM:           &ipamConfig,
+		},
+	})
+	defer net.Remove(ctx)
+
+	nginxC, _ := GenericContainer(ctx, GenericContainerRequest{
```
mdelapenya, line 60: Would you mind adding a separate test for getting this container's network IPAM from the request? IIUC this test case, the nginx container is not used not asserted.

aivinog1: Hey @mdelapenya! Sorry for the delay, but what do you mean by:
> a separate test for getting this container's network IPAM from the request

As I understand now there is already an assertion for it: https://github.com/testcontainers/testcontainers-go/pull/394/files#diff-f84ddb94b392ea0e4f707f6928b01545cbd7faddabe60622c9f348c090fc3e45R82

mdelapenya: Ahh you are right, I may have missed it 🤦 Thanks for pointing out!

mdelapenya:
@aivinog1 could you please 🙏 rebase and push to trigger the tests? I do not have permissions yet to force a merge/rebase, so I'd appreciate if you can do it. Sorry for bugging you 😢

aivinog1:
> @aivinog1 could you please 🙏 rebase and push to trigger the tests? I do not have permissions yet to force a merge/rebase, so I'd appreciate if you can do it. Sorry for bugging you 😢 

@mdelapenya it is done, no problem 🙂

mdelapenya:
I see this error in the GH action:

```
=== FAIL: . Test_NetworkWithIPAM (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xac7e40]
```

Could you please take a look?

aivinog1:
@mdelapenya Hey, I fixed it, let's try again :)

mdelapenya (approved):
Great job  here @aivinog1!

PR #396, opened 2022-01-09
fix: improve get container ip

nodece:
### Motivation

Get container ip sometimes empty,  I check code and found that we return the `inspect.NetworkSettings.IPAddress`. If this IP is empty, we should check the `inspect.NetworkSettings.Networks`, it shouldn't empty.


docker.go
```diff
@@ -284,7 +284,17 @@ func (c *DockerContainer) ContainerIP(ctx context.Context) (string, error) {
 		return "", err
 	}
 
-	return inspect.NetworkSettings.IPAddress, nil
+	ip := inspect.NetworkSettings.IPAddress
+	if ip == "" {
+		for _, settings := range inspect.NetworkSettings.Networks {
+			if settings.IPAddress != "" {
+				ip = settings.IPAddress
```
mdelapenya, line 291: What do you think about using a more descriptive variable name?
```suggestion
		for _, network := range inspect.NetworkSettings.Networks {
			if network.IPAddress != "" {
				ip = network.IPAddress
```

nodece: Good point!

mdelapenya (commented):
This looks good to me, although I'm missing a few tests covering the new behaviour.

Would you mind adding them? 🙏 

Thanks!

nodece:
Looks like I should make a mock test, have you any ideas?

baez90:
Have a look at the `TestContainerAttachedToNewNetwork` test in `docker_test.go` it already 'covers' what you are trying to avoid.
I think it has something to do with whether a container is in the default `bridge` network or in another non-default network. In the latter case your way of retrieving the IP is the only one working.

mdelapenya:
I'd say this PR has been already solved in #491. Please let us know if we can close it, thanks!

PR #401, opened 2022-02-02
chore(ci): run tests for more recent Go versions

mdelapenya:
## What does this PR do?
It adds Go 1.16 and 1.17 to the CI matrix to run the tests

## Why is it important?
Go versions coverage

## Other concerns
Should we retire 1.14 (or any other) from the matrix? Which versions should we be compatible with?


gianarb (approved):

gianarb:
Thanks a lot

PR #407, opened 2022-02-21
use var TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE

mniak:


mdelapenya (approved):
Thanks for this contribution @mniak! It LGTM

I'm postponing the merge until I get a deeper understanding of #414

vincentfree:
Can't wait for this feature, I'm using lima on a M1 macbook, the default docker sock won't work out of the box so this would safe a lot of work and time helping others with the setup

vincentfree:
> Thanks for this contribution @mniak! It LGTM
> 
> I'm postponing the merge until I get a deeper understanding of #414

can't this ship without the rest of #414?

kgalli:
@mdelapenya it seems #414 might take some time to have all the edge cases covered. As this is a relatively small change with manageable side effects do you still think this change should wait?

From my point of view there is a lot of value in this change already. Is there any particular info you need to reconsider?

vincentfree:
> @mdelapenya it seems #414 might take some time to have all the edge cases covered. As this is a relatively small change with manageable side effects do you still think this change should wait?
> 
> 
> 
> From my point of view there is a lot of value in this change already. Is there any particular info you need to reconsider?

I second that. This small feature will help with multiple non native docker backends. With lima serving my docker instances I can't use testcontainers until this is implemented.

baez90:
@mdelapenya I assume you mean merging this now and #414 afterwards? Makes sense I think, rebasing #414 shouldn't be that hard and it also narrows the scope of the other PR.

baez90:
Sure, makes sense I think and rebasing this should be straightforward

mdelapenya (approved):

vincentfree:
Can I already use this feature?

baez90:
It's not released yet but you can use it if you update your dependency to latest `main` branch if you really want

Issue #399, opened 2022-01-31
Allow customization of docker.sock path


mniak:
## Feature Request

I would like that the library understands an envbironment variable `TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE` and uses this when starting the reaper container (`ryuk`)

I'll try to contribute some code when I have some spare time
Something like this (using pseudo code on the change section):
on https://github.com/testcontainers/testcontainers-go/blob/9c3076fddb0ca14fceef5bb063690ef87aeeb346/reaper.go#L58-L69

```go
req := ContainerRequest{
    Image:        reaperImage(reaperImageName),
    ExposedPorts: []string{string(listeningPort)},
    Labels: map[string]string{
        TestcontainerLabel:         "true",
        TestcontainerLabelIsReaper: "true",
    },
    SkipReaper: true,
    // THIS ->
    Mounts:     Mounts(BindMount(
        strings.Coalesce(os.GetEnv("TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE"), "/var/run/docker.sock"),
        "/var/run/docker.sock",
    )),
    // <- THIS
    AutoRemove: true,
    WaitingFor: wait.ForListeningPort(listeningPort),
}
```

### Context
Following the announcement that [Docker Desktop is not free anymore](https://www.docker.com/blog/updating-product-subscriptions/) some users started using different tools to manage and their containers.

In some workaround scenarios the socket `docker.sock` is created in non-standard locations.
It's easy to setup the `DOCKER_HOST` environment variable and most applications that use the Docker API will run just fine.

But in [`testcontainers-go`](https://github.com/testcontainers/testcontainers-go/blob/9c3076fddb0ca14fceef5bb063690ef87aeeb346/reaper.go#L66) the reaper container, `ryuk` is [set-up to use the standard location](https://github.com/testcontainers/testcontainers-go/blob/9c3076fddb0ca14fceef5bb063690ef87aeeb346/reaper.go#L66).

[...]

PR #414, opened 2022-03-08
Improve Podman compatibility

baez90:
# What does this PR do?

Although Podman claims to be API compatible with Docker and for instance in #336 the consent seems to be _let's wait until all problems are gone_ :sweat_smile: I thought I'd see what can be done as kind of a compromise.

The problems I could identify (and solve) are:

- The current implementation assumes the default network name is `bridge` which isn't the case for Podman. Even worse, `bridge` is not even a valid network name with Podman because it's also a network mode and Podman does not allow the creation of networks conflicting with network modes.
- The current implementation of the reaper has a hard coded path where it expects the Docker socket. Neither `DOCKER_HOST` env variable nor any other customization the [testcontainers docs](https://www.testcontainers.org/features/configuration/#customizing-ryuk-resource-reaper) are describing are honored.  This is also an issue with rootless Docker.
- Podman requires at least one stream to be attached when a command is executed. It doesn't hurt to attach `stdout` and/or `stderr` but otherwise it breaks the exec wait strategy.

## `DefaultNetwork` option

The `DefaultNetwork` option allows Podman users to configure the name of the default network. If it's not set the previous auto-detection is applied so no breaking change.

## Honoring `DOCKER_HOST` and `TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE`

The `DOCKER_HOST` env variable was already considered when creating the Docker API client. Now its value is propagated (via the `context.Context`) to the reaper as a fallback if `TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE` isnt' set. If both env variables are empty the default socket path `/var/run/docker.sock` will be assumed.

## Default `NetworkMode` if not set

In cases when the `NetworkMode` isn't set it will be set to `bridge` to avoid issues with Podman. Docker apparently assumes `bridge` if nothing's set therefore this shouldn't break anything either.

## Parallel tests

To speed up the test suite I marked all Docker tests to be executed in parallel.

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -323,8 +323,10 @@ func (c *DockerContainer) NetworkAliases(ctx context.Context) (map[string][]stri
 func (c *DockerContainer) Exec(ctx context.Context, cmd []string) (int, error) {
 	cli := c.provider.client
 	response, err := cli.ContainerExecCreate(ctx, c.ID, types.ExecConfig{
-		Cmd:    cmd,
-		Detach: false,
+		Cmd:          cmd,
+		Detach:       false,
+		AttachStderr: true,
+		AttachStdout: true,
 	})
 	if err != nil {
 		return 0, err
@@ -527,9 +529,9 @@ func (n *DockerNetwork) Remove(ctx context.Context) error {
 // DockerProvider implements the ContainerProvider interface
 type DockerProvider struct {
 	*DockerProviderOptions
-	client         *client.Client
-	hostCache      string
-	defaultNetwork string // default container network
+	client    *client.Client
+	host      string
+	hostCache string
 }
 
 var _ ContainerProvider = (*DockerProvider)(nil)
@@ -604,6 +606,10 @@ func NewDockerProvider(provOpts ...DockerProviderOption) (*DockerProvider, error
 
 			opts = append(opts, client.WithTLSClientConfig(cacertPath, certPath, keyPath))
 		}
+	} else if dockerHostEnv := os.Getenv("DOCKER_HOST"); dockerHostEnv != "" {
+		host = dockerHostEnv
+	} else {
+		host = "unix:///var/run/docker.sock"
 	}
 
 	c, err := client.NewClientWithOpts(opts...)
@@ -623,6 +629,7 @@ func NewDockerProvider(provOpts ...DockerProviderOption) (*DockerProvider, error
 	c.NegotiateAPIVersion(context.Background())
 	p := &DockerProvider{
 		DockerProviderOptions: o,
+		host:                  host,
 		client:                c,
 	}
 
@@ -705,24 +712,26 @@ func (p *DockerProvider) CreateContainer(ctx context.Context, req ContainerReque
 
 	// Make sure that bridge network exists
 	// In case it is disabled we will create reaper_default network
-	p.defaultNetwork, err = getDefaultNetwork(ctx, p.client)
-	if err != nil {
-		return nil, err
+	if p.DefaultNetwork == "" {
+		p.DefaultNetwork, err = getDefaultNetwork(ctx, p.client)
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	// If default network is not bridge make sure it is attached to the request
 	// as container won't be attached to it automatically
-	if p.defaultNetwork != Bridge {
+	if p.DefaultNetwork != Bridge {
 		isAttached := false
 		for _, net := range req.Networks {
-			if net == p.defaultNetwork {
+			if net == p.DefaultNetwork {
 				isAttached = true
 				break
 			}
 		}
 
 		if !isAttached {
-			req.Networks = append(req.Networks, p.defaultNetwork)
+			req.Networks = append(req.Networks, p.DefaultNetwork)
 		}
 	}
 
@@ -744,7 +753,7 @@ func (p *DockerProvider) CreateContainer(ctx context.Context, req ContainerReque
 
 	var termSignal chan bool
 	if !req.SkipReaper {
-		r, err := NewReaper(ctx, sessionID.String(), p, req.ReaperImage)
+		r, err := NewReaper(context.WithValue(ctx, dockerHostContextKey, p.host), sessionID.String(), p, req.ReaperImage)
 		if err != nil {
 			return nil, fmt.Errorf("%w: creating reaper failed", err)
 		}
@@ -828,6 +837,9 @@ func (p *DockerProvider) CreateContainer(ctx context.Context, req ContainerReque
 
 	// prepare mounts
 	mounts := mapToDockerMounts(req.Mounts)
+	if req.NetworkMode == "" {
+		req.NetworkMode = Bridge
+	}
 
 	hostConfig := &container.HostConfig{
 		PortBindings: exposedPortMap,
@@ -1001,7 +1013,11 @@ func (p *DockerProvider) CreateNetwork(ctx context.Context, req NetworkRequest) 
 
 	// Make sure that bridge network exists
 	// In case it is disabled we will create reaper_default network
-	p.defaultNetwork, err = getDefaultNetwork(ctx, p.client)
+	if p.DefaultNetwork == "" {
+		if p.DefaultNetwork, err = getDefaultNetwork(ctx, p.client); err != nil {
+			return nil, err
+		}
+	}
 
 	if req.Labels == nil {
 		req.Labels = make(map[string]string)
@@ -1020,7 +1036,7 @@ func (p *DockerProvider) CreateNetwork(ctx context.Context, req NetworkRequest) 
 
 	var termSignal chan bool
 	if !req.SkipReaper {
-		r, err := NewReaper(ctx, sessionID.String(), p, req.ReaperImage)
+		r, err := NewReaper(context.WithValue(ctx, dockerHostContextKey, p.host), sessionID.String(), p, req.ReaperImage)
 		if err != nil {
 			return nil, fmt.Errorf("%w: creating network reaper failed", err)
 		}
@@ -1065,14 +1081,14 @@ func (p *DockerProvider) GetNetwork(ctx context.Context, req NetworkRequest) (ty
 
 func (p *DockerProvider) GetGatewayIP(ctx context.Context) (string, error) {
 	// Use a default network as defined in the DockerProvider
-	if p.defaultNetwork == "" {
+	if p.DefaultNetwork == "" {
 		var err error
-		p.defaultNetwork, err = getDefaultNetwork(ctx, p.client)
+		p.DefaultNetwork, err = getDefaultNetwork(ctx, p.client)
 		if err != nil {
 			return "", err
 		}
 	}
-	nw, err := p.GetNetwork(ctx, NetworkRequest{Name: p.defaultNetwork})
+	nw, err := p.GetNetwork(ctx, NetworkRequest{Name: p.DefaultNetwork})
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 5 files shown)


baez90:
I just noticed there's also #407 which also adds `TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE` support - if that one's merged I will of course adopt these changes :see_no_evil:

mdelapenya:
> I just noticed there's also #407 which also adds `TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE` support - if that one's merged I will of course adopt these changes 🙈

Hey @baez90 thanks again for the PR and the explanation to add context and motivation for it. It makes a difference in the review process.

I think #407 is super simple and straight forward, so we could merge that PR first and then rebase your code with that. Does it makes sense to you?

baez90:
Sure thing, I think I've already an idea how to incorporate the changes from #407 with my own :blush: 

Glad to hear the descriptions are getting a bit better, I'm still learning but I've to admit that I really like the idea!

mdelapenya:
One thing that I'd like you to explore before merging this PR is the idea of running the entire test suite using podman instead of Docker. Do you want to contribute a Github Action workflow specific to it?

baez90:
Interesting idea. I'll have a look into it how much I'd have to adjust in the current test suite to make this possible.
Frankly: some of docker-compose tests aren't working on my machine either way, hence might get a bit hard but I'll see what I can do!

mdelapenya:
The java folks have a PoC for Podman here: https://github.com/testcontainers/testcontainers-java/pull/4583

baez90:
awesome! Thanks for letting me know!
Regarding my previous question: do you mind me adopting the test suite to be able to distinguish between Podman and Docker? At least to be able to configure the `DockerProvider` with the right default network and stuff?

fpozzobon:
That would be great to have this PR merged :)

baez90:
I'm still both waiting for #407 to be merged and working on some improvements (as soon as I can spare some time) to - hopefully - run tests against Podman in the CI - although I'm still not sure if I'll be able to make it 100% compatible.

Sorry for the delay I think I should be able to work on this tomorrow :blush:

baez90:
@mdelapenya now that the pipeline does not fail anymore I'd consider this done. Let me know if I shall split this PR or anything else. I also updated the description to cover the latest changes.

I know that some tests in the Podman pipeline are still a bit flaky but I'd like to wait until Podman 4.x is ready for Ubuntu and revisit the pipeline then once more?

docker.go
```diff
@@ -603,20 +607,15 @@ func Generic2DockerOptions(opts ...GenericProviderOption) []DockerProviderOption
 	return converted
 }
 
-// NewDockerProvider creates a Docker provider with the EnvClient
-func NewDockerProvider(provOpts ...DockerProviderOption) (*DockerProvider, error) {
```
mdelapenya, line 607: This could be a breaking change. Let's label de PR for the release notes

mdelapenya: Ops.. moved below. Nevermind

baez90: :smile: I tried to not break anything this time :smile: but I was wondering if it makes sense to include `func NewDockerClient() (cli *client.Client, host string, err error)` into the public API or if it would be better to make this private.

The reason I made it public in the first place was to be able to use it in the tests but I guess this shouldn't be an issue anyway because there's no test package?

baez90:
I just got the update to Podman 4.1 and it seems like the issue with the missing default network is now resolved. I probably could get rid of some code but the question is do we want to rely for Podman users to have the latest version or shall we test also against a 3.x version? @mdelapenya any thoughts?

network.go
```diff
@@ -17,6 +17,16 @@ type Network interface {
 	Remove(context.Context) error // removes the network
 }
 
+type DefaultNetwork string
+
+func (n DefaultNetwork) ApplyGenericTo(opts *GenericProviderOptions) {
```
mdelapenya, line 22: Is there any usage of `ApplyGenericTo`? I cannot see it anywhere

reaper.go
```diff
@@ -87,12 +96,39 @@ func NewReaper(ctx context.Context, sessionID string, provider ReaperProvider, r
...
 
+func extractDockerHost(ctx context.Context) (dockerHostPath string) {
+	if dockerHostPath = os.Getenv("TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE"); dockerHostPath != "" {
+		return dockerHostPath
+	}
+
+	dockerHostPath = "/var/run/docker.sock"
+
+	var hostRawURL string
+	if h, ok := ctx.Value(dockerHostContextKey).(string); !ok || h == "" {
```
mdelapenya, line 107: Does `dockerHostContextKey` need to be a new type? Could it be just a string? Not a blocker, trying to understand its usage

mdelapenya (commented):
Hey @baez90 this is great and the code LGTM! Thank you very much for your time and dedication to support Podman.

I'll be honest: I need time to download the PR and play with Podman as I never did it in the past.

But the tests seems OK, which is a great step. Let me play with this version of the PR, and hopefully soon we will see this baby merged into the code.

Good job!

reaper.go
baez90: See the [docs](https://pkg.go.dev/context#WithValue) regarding why this is a type :blush: there's also a blog entry on go.dev I think but I can't find it right now. Does this cover your question ?

network.go
baez90: it's the implementation of the `GenericProviderOption` interface

container_test.go
```diff
@@ -252,7 +253,9 @@ func Test_BuildImageWithContexts(t *testing.T) {
 	}
 
 	for _, testCase := range testCases {
+		testCase := testCase
```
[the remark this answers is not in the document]
mdelapenya: (Luke) Using the index is perfect

docker_test.go
```diff
@@ -69,13 +82,14 @@ func TestContainerAttachedToNewNetwork(t *testing.T) {
...
+	})
 
 	nginx, err := GenericContainer(ctx, gcr)
-	if err != nil {
-		t.Fatal(err)
-	}
-	defer nginx.Terminate(ctx)
+
+	require.NoError(t, err)
+	terminateContainerOnEnd(t, ctx, nginx)
```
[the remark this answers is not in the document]
mdelapenya: This is what I needed, thanks for the explanation. LGTM as it is

reaper.go
mdelapenya: >Users of WithValue should define their own types for keys.

LGTM

docker_test.go
```diff
@@ -1620,13 +1565,19 @@ func TestContainerNonExistentImage(t *testing.T) {
 }
 
 func TestContainerCustomPlatformImage(t *testing.T) {
+	if providerType == ProviderPodman {
+		t.Skip("Incompatible Docker API version for Podman")
```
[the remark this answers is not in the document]
baez90: That one's apparently a little bit trickier...so Podman does support Docker API version 1.41 (at least parts of it, cannot tell if everything) but when the client does the negotiation it returns 1.41 which results in the Docker client library claiming that passing the architecture is not supported...I'll create an issue for the Podman guys how to proceed with something like this...of course we could always just set the API version to 1.41 manually for Podman but it feels really ugly...

baez90:
Thank you very much for taking your time to review the PR and giving me feedback and everything!

From my side most things are done now. I've been thinking of removing the Podman provider again but without it the default network detection is still problematic.I don't really know if the special behavior to join a new container always to the `bridge` network first if it is among the networks the container should be joined to is still necessary or not and therefore don't really want to touch it especially because the PR is already so huge...

The current implementation works quite good with Podman 4.1. There are some tests still failing but those are not directly related to the Podman API compatibility but to the fact that Podman does not really handle a lot of parallel container operations really good (no daemon, file system locks for the BoltDB...). I raised an issue for the Docker API version problem, let's see what's coming from them but it shouldn't be a blocker for now I guess?

I'd like to add that not all changes I made are only necessary for Podman but also come in handy if someone's using Docker rootless (e.g. mounting the right socket to the reaper not always `/var/run/docker.sock` ).

Let me know if you encounter any problems or if I should change some things :blush:

docker_test.go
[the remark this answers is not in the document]
baez90: Created an issue for it and referenced it in the `Skip` comment. Hopefully this is resolved soon but might take some time until it's released and available for the pipeline

mdelapenya:
I'm testing this on my Mac, and many tests are failing:

```
DONE 121 tests, 1 skipped, 53 failures in 304.252s
exit status 1
```

I'm not worried at all because the CI is green, but I'd like to know if there is any other conf we should do while working on the tests locally.

cfstras:
Not sure whether this belongs here or with https://github.com/testcontainers/testcontainers-go/issues/399, but I'm using rootless docker ([on lima](https://github.com/lima-vm/lima/blob/master/examples/docker.yaml)) and testcontainers seems to fail starting the reaper.

baez90:
@mdelapenya unfortunately I cannot reproduce your issue with Podman on Windows.
Some of the tests are working but most are also failing but not because of Podman directly but because it's 'rather complicated' to mount the npipe on Windows as a socket to ryuk. I'll see if I can install Go into the Podman WSL but that seems rather hacky.

I tuned the CI pipeline to install Podman 4.1 from Debian experimental (still no Podman 4 for Ubuntu 20.04 or 22.04) and it seems as this would work now as expected including running tests in parallel and without any 'continue on error' setting in place.

Apart from some more experiments with Podman on Windows I'd say this actually looks quite good from my side, what do you think?

mdelapenya:
> @mdelapenya is there something missing here? Of course I'll rebase this PR once more if the other PR is merged

We merged a PR regarding LogConsumer yesterday and caused the conflicts. Could you resolve them 🙏 

> But at some point I'd really appreciate it to get this either merged or I'll close it as there seems to be no interest in this. I don't see any sense investing more time in keeping this up to date without any further progress/feedbackq on how to go on with this PR.

@baez90 please forgive me. It IS of interest and I do appreciate your hard effort in contributing it. I shared my concerns about not being able to make it work on my machine, but it does not mean it's of no interest. I've been focused on smaller PRs with lower impact and less time to review.

Your PR represents the big star for the new release and I'd like to create a great experience using TC-go with Podman. So please please do not hesitate about the importance of your contributions. Sometimes maintainers need to juggle with company work, family time and open source 😞

baez90:
@mdelapenya I didn't change anything I only rebased the branch to resolve conflicts :wink: 
I absolutely know what you mean and normally I try to work in...well defined increments while working on a PR and only amend changes e.g. when working on the pipeline to avoid "for ***** sake - let the pipeline pass' commits :smile: 
From time to time I forget this and appreciate the reminder :blush: 

Regarding your concern testing the PR on MacOS: like I mentioned earlier I had similar problems running the test suite on Windows with Podman Machine due to limitations especially with the reaper because I couldn't find any possibility to mount native Windows pipes as Linux socket into the ryuk conainer. But other tests were running. I don't have access to a Mac so I can't say if there are similar problems to be expected with the way Podman is working on MacOS.

I already added the retry logic you mentioned in another PR (I think) to the Podman pipeline and I'm thinking of doing the same for the Docker pipeline as the last checks didn't even fail in code I touched :sweat_smile: (from what I can tell) and also only for Docker not for Podman...

Apart from that the main change is actually only how the default network is handled to ensure Docker gets its `bridge` network while Podman requires `podman` and I added some code to consider some environment variables although most of this is now already merged by other PRs :sweat_smile: 

I'm also sorry for the probably a bit harsh words earlier. It just happened to be a little bit frustrating.

mdelapenya (approved):
It took a while, and sorry about that. I see you when you expressed your frustration, and as a maintainer it's my mission to keep you engaged, even more after this huge and fantastic contribution. Really well done

mdelapenya:
@gianarb I'd appreciate a final review before merging this one 🙏

gianarb (approved):

mdelapenya:
Big kudos to @baez90 for this humongous job!

Awesome!

image

PR #416, opened 2022-03-19
add ability to stop container

oktalz:
Hi, option to stop containers is not exposed through interface although it exists in docker provider.

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -182,6 +182,28 @@ func (c *DockerContainer) Start(ctx context.Context) error {
 	return nil
 }
 
+// Stop will stop an already started container
+//
+// In case the container fails to stop
+// gracefully within a time frame specified by the timeout argument,
+// it is forcefully terminated (killed).
+//
+// If the timeout is nil, the container's StopTimeout value is used, if set,
+// otherwise the engine default. A negative timeout value can be specified,
+// meaning no timeout, i.e. no forceful termination is performed.
+func (c *DockerContainer) Stop(ctx context.Context, timeout *time.Duration) error {
+	shortID := c.ID[:12]
+	c.logger.Printf("Stopping container id: %s image: %s", shortID, c.Image)
+
+	if err := c.provider.client.ContainerStop(ctx, c.ID, timeout); err != nil {
+		return err
+	}
+
+	c.logger.Printf("Container is stopped id: %s image: %s", shortID, c.Image)
+
+	return nil
+}
+
 // Terminate is used to kill the container. It is usually triggered by as defer function.
 func (c *DockerContainer) Terminate(ctx context.Context) error {
 	select {
--- a/container.go
+++ b/container.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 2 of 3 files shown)


docker.go
```diff
@@ -182,6 +182,28 @@ func (c *DockerContainer) Start(ctx context.Context) error {
...
+// Stop will stop an already started container
+//
+// In case the container fails to stop
+// gracefully within a time frame specified by the timeout argument,
+// it is forcefully terminated (killed).
+//
+// If the timeout is nil, the container's StopTimeout value is used, if set,
+// otherwise the engine default. A negative timeout value can be specified,
+// meaning no timeout, i.e. no forceful termination is performed.
+func (c *DockerContainer) Stop(ctx context.Context, timeout *time.Duration) error {
```
mdelapenya, line 194: Do you think the timeout feature could be done as part of the context, with a deadline, timeout, etc?

oktalz: hm, purpose is not the same.

for example if you use something like this:
```Go
stopCtx, _ := context.WithTimeout(ctx, time.Nanosecond) // deliberately to little :)
err = nginxA.Stop(stopCtx, &stopTimeout)
```
that would result in error:
```
docker_test.go:422: Post "http://%2Fvar%2Frun%2Fdocker.sock/v1.41/containers/283c6cdff4014dd0ef45426573b3944d4d1498d7aba32abd5b7d346c71389b30/stop?t=10": context deadline exceeded
```

then you are unsure did container stopped or not. (did api request was called or not)

for example, if context expires, but API request is already sent, then you can't know if container is up or not because api can also took some time and then docker decides to not stop it for a long period

My initial idea was to replicate underline behavior of docker library (be as transparent as possible)
if you look at the code from [github.com/docker/.../client/container_stop.go#L18](https://github.com/moby/moby/blob/301eba7fa4871862fbc00753c4accfc535e0eedf/client/container_stop.go#L18) that is called, that timeout is not used in context manner, but as query param for docker itself [docs.docker.com....ContainerStop](https://docs.docker.com/engine/api/v1.41/#operation/ContainerStop)

for example 

```Go
stopTimeout := 10 * time.Second
err = nginxA.Stop(stopCtx, &stopTimeout)
```

generates
```
"/v1.41/containers/804f8d1809be90d927d3477eafde83780cdbb6db8f027cbcc0ac16825199fe55/stop?t=10"
```

but if you just have
```Go
err = nginxA.Stop(stopCtx, nil)
```
request will be done without `?t=XYZ`

if you want to stop or terminate container you need this timeout, otherwise you need to additionally terminate it.

In many cases it normal to give some time to container to gracefully stop, and if that takes to long just terminate it.
but having terminate as an only option is not ideal if you decide to start it in test later again.

Additionally, my initial reason for this PR is to test if container can actually stop gracefully or not. (checking logs afterwards that was properly shut down)

I can remove it if that is preferred

mdelapenya: Thanks for such a detailed information. Now it's crystal clear to me 😊

mdelapenya:
Merged, thanks!

image

PR #417, opened 2022-03-19
expose CopyToContainer so file can be created only in memory

oktalz:
this is especially useful in tests, since we do not
need to create a file for testing,
we can just generate it in memory and then
copy it to a container directly

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -395,7 +395,11 @@ func (c *DockerContainer) CopyFileToContainer(ctx context.Context, hostFilePath 
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


docker.go
```diff
@@ -395,7 +395,11 @@ func (c *DockerContainer) CopyFileToContainer(ctx context.Context, hostFilePath
 	if err != nil {
 		return err
 	}
+	return c.CopyToContainer(ctx, fileContent, containerFilePath, fileMode)
```
mdelapenya, line 398: Neat!

mdelapenya (approved):

PR #421, opened 2022-04-04
fix: use 1.18 format to install dependencies

mdelapenya:
## What does this PR do?
It puts gotestsum as a dependency tool, so that the installation is managed by Go dependency manager.

It also pins jinja2 after recent breaking changes (see https://github.com/mkdocs/mkdocs/issues/2799 and what the Java folks have done in https://github.com/testcontainers/testcontainers-java/pull/5186)

## Why is it important?
See https://go.dev/doc/go-get-install-deprecation

> Starting in Go 1.17, installing executables with go get is deprecated. go install may be used instead.
>
> In Go 1.18, go get will no longer build packages; it will only be used to add, update, or remove dependencies in go.mod. Specifically, go get will always act as if the -d flag were enabled.

Reads about `tools.go`:
- https://marcofranssen.nl/manage-go-tools-via-go-modules
- https://play-with-go.dev/tools-as-dependencies_go115_en/
- https://github.com/golang/go/issues/48429

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,13 +35,11 @@ jobs:
       - name: Get dependencies
         if: ${{ matrix.go-version != '1.x' }}
         run: |
-          go get -v -t -d ./...
           go get gotest.tools/gotestsum
 
       - name: Get dependencies for 1.18
         if: ${{ matrix.go-version == '1.x' }}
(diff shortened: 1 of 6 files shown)


gianarb (approved):
Thanks a lot for figuring all of this out

PR #426, opened 2022-04-07
Fix compose v2 wait strategy

oriser:
In compose V2, the container name is `<compose name>-<service name>` instead of `<compose name>_<service name>`

The code do referrer it by having `composeV2ContainerName` along `containerName`, but the  `composeV2ContainerName` is constructed by replacing any `_` from `containerName` with `-`. This is a problem because the service may have `_` in its name and it will cause an incorrect container name that won't be found.

### DIFF
--- a/compose.go
+++ b/compose.go
@@ -135,6 +135,10 @@ func (dc *LocalDockerCompose) getDockerComposeEnvironment() map[string]string {
 	return environment
 }
 
+func (dc *LocalDockerCompose) containerNameFromServiceName(service, separator string) string {
+	return dc.Identifier + separator + service
+}
+
 func (dc *LocalDockerCompose) applyStrategyToRunningContainer() error {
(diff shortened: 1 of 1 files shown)


mdelapenya:
I was exactly testing this feature in a project just now, and realising it was wrong. Thanks for reading my mind 😄 

I thought it was fixed in 0.12, but no, it was not released so I've just released v0.13.0. Will need to merge and release yours as v0.13.1

mdelapenya:
This LGTM, although it could be even greater if you could add a test with a service name including underscores.

oriser:
> This LGTM, although it could be even greater if you could add a test with a service name including underscores.

@mdelapenya There is a test using `WaitForService` with service name with underscores (`TestDockerComposeWithWaitForService`), but it probably not running with newer version of docker-compose (otherwise it would fail).
Worth adding a test running with a newer version of docker-compose

gianarb (approved):

PR #435, opened 2022-04-28
Return a Reader from the Exec function

johanoskarsson:
This should allow the user to read the logs of the command execution. I've tested this quickly locally and it seems to work as expected.
Pushing this so I can get some CI feedback.
I believe this is similar to what was asked for in #126

### DIFF
--- a/logconsumer_test.go
+++ b/logconsumer_test.go
@@ -246,7 +246,7 @@ func TestContainerLogWithErrClosed(t *testing.T) {
 	existingLogs := len(consumer.Msgs)
 
 	hitNginx := func() {
-		i, err := dind.Exec(ctx, []string{"wget", "--spider", "localhost:" + port.Port()})
+		i, _, err := dind.Exec(ctx, []string{"wget", "--spider", "localhost:" + port.Port()})
 		if err != nil || i > 0 {
 			t.Fatalf("Can't make request to nginx container from dind container")
 		}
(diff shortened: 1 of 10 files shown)


logconsumer_test.go
```diff
@@ -246,7 +246,7 @@ func TestContainerLogWithErrClosed(t *testing.T) {
 	existingLogs := len(consumer.Msgs)
 
 	hitNginx := func() {
-		i, err := dind.Exec(ctx, []string{"wget", "--spider", "localhost:" + port.Port()})
+		i, err, _ := dind.Exec(ctx, []string{"wget", "--spider", "localhost:" + port.Port()})
```
mdelapenya, line 249: Is this correct?
```suggestion
		i, _, err := dind.Exec(ctx, []string{"wget", "--spider", "localhost:" + port.Port()})
```

johanoskarsson:
Thanks I had them the wrong way around first!

mdelapenya:
Hey @johanoskarsson, I took a look at your PR and it seems it’s outdated (created before the commit that fixed the netlify deployment in main):

> This branch is 2 commits ahead, 13 commits behind testcontainers/testcontainers-go:main

Could you rebase and push? Thank you in advance for your work here

mdelapenya (approved):
LGTM! Let's wait for the CI

Thanks

mdelapenya:
NIT: before merging this, could you add a test (or an assertion in the existing tests) where we are consuming the Reader? I'm checking that we are always omitting it in the current tests, and would like to see it covered.

mdelapenya:
BTW, I'm tagging this PR as breaking-change because we are changing the signature of an exposed method. Because we are not in 1.0.0, it's fine to break it :)

gianarb (approved):

PR #449, opened 2022-06-10
feat: ability to configure running privileged Ryuk

zregvart:
Adds support for running the Ryuk (reaper) container in privileged mode
either by setting `ryuk.container.privileged=true` in
`$HOME/.testcontainers.properties` or by setting
`TESTCONTAINERS_RYUK_CONTAINER_PRIVILEGED=true` environment variable.

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -569,9 +569,10 @@ var _ ContainerProvider = (*DockerProvider)(nil)
 
 // or through Decode
 type TestContainersConfig struct {
-	Host      string `properties:"docker.host,default="`
-	TLSVerify int    `properties:"docker.tls.verify,default=0"`
-	CertPath  string `properties:"docker.cert.path,default="`
+	Host           string `properties:"docker.host,default="`
+	TLSVerify      int    `properties:"docker.tls.verify,default=0"`
+	CertPath       string `properties:"docker.cert.path,default="`
+	RyukPrivileged bool   `properties:"ryuk.container.privileged,default=false"`
 }
 
 type (
--- a/reaper.go
+++ b/reaper.go
@@ -74,6 +74,11 @@ func NewReaper(ctx context.Context, sessionID string, provider ReaperProvider, r
 		WaitingFor: wait.ForListeningPort(listeningPort),
 	}
 
+	tcConfig := readTCPropsFile()
+	if tcConfig.RyukPrivileged || os.Getenv("TESTCONTAINERS_RYUK_CONTAINER_PRIVILEGED") == "true" {
+		req.Privileged = true
+	}
+
 	// Attach reaper container to a requested network if it is specified
 	if p, ok := provider.(*DockerProvider); ok {
 		req.Networks = append(req.Networks, p.defaultNetwork)


zregvart:
I created this as the work on #223 seems to have stalled.

zregvart:
@mdelapenya I'd appreciate a review of this when you get a chance

docker.go
```diff
@@ -569,9 +569,10 @@ var _ ContainerProvider = (*DockerProvider)(nil)
 
 // or through Decode
 type TestContainersConfig struct {
-	Host      string `properties:"docker.host,default="`
-	TLSVerify int    `properties:"docker.tls.verify,default=0"`
-	CertPath  string `properties:"docker.cert.path,default="`
+	Host           string `properties:"docker.host,default="`
+	TLSVerify      int    `properties:"docker.tls.verify,default=0"`
+	CertPath       string `properties:"docker.cert.path,default="`
+	RyukPrivileged bool   `properties:"ryuk.container.privileged,default=false"`
```
mdelapenya, line 575: Neat! Thanks for honouring the names from the Java folks https://github.com/testcontainers/testcontainers-java/blob/de1324ed2800eff4da326d0c23d281399d006bc0/core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java#L141

reaper.go
```diff
@@ -74,6 +74,11 @@ func NewReaper(ctx context.Context, sessionID string, provider ReaperProvider, r
 		WaitingFor: wait.ForListeningPort(listeningPort),
 	}
 
+	tcConfig := readTCPropsFile()
```
mdelapenya, line 77: I think this will work, although I'd like to read the properties just once, when we create the Docker Provider.

IIRC the Provider is obtained when the container is created. Could you double check it please?

This is not a blocker, so we can merge this as is if we do not find a better path.

zregvart: Yes, that is true. I've added the `Config() TestContainersConfig` to the `DockerProvider` to fetch the configuration that is now read when `NewDockerProvider` is invoked.

zregvart:
My first instinct was to make the change as simple as possible at the cost of computing the `TestContainersConfig` twice. I've now refined this and the `TestContainersConfig` is held in the `config` field of the `DockerProvider` and can be accessed via `Config() TestContainersConfig` now present on `ReaperProvider` as well.

docker.go
```diff
@@ -987,6 +990,12 @@ func (p *DockerProvider) RunContainer(ctx context.Context, req ContainerRequest)
 	return c, nil
 }
 
+// Config provides the TestContainersConfig read from $HOME/.testcontainers.properties or
+// the environment variables
+func (p *DockerProvider) Config() TestContainersConfig {
```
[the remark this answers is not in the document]
zregvart: Thanks, I think that'll be useful. I've added this for a67d762b6c508e17690cde36e3801db72a68961d

container.go
```diff
@@ -29,6 +29,7 @@ type ContainerProvider interface {
 	CreateContainer(context.Context, ContainerRequest) (Container, error) // create a container without starting it
 	RunContainer(context.Context, ContainerRequest) (Container, error)    // create a container and start it
 	Health(context.Context) error
+	Config() TestContainersConfig
```
mdelapenya, line 32: @baez90 please take a look at this for the Podman support 👀

mdelapenya: I'm sorry for that @baez90. Long-running branches have this kind of complexities. I truly appreciate your effort in the Podman support, and I think it's close to be there 🙏 

image

reaper_test.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	tests := []cases{
+		{
+			name:   "non-provileged",
+			req:    createContainerRequest(nil),
+			config: TestContainersConfig{},
+		},
+		{
+			name: "provileged",
```
mdelapenya, line 69: Typo
```suggestion
			name: "privileged",
```

docker.go
```diff
@@ -611,8 +613,8 @@ func WithDefaultBridgeNetwork(bridgeNetworkName string) DockerProviderOption {
 	})
 }
 
-func NewDockerClient() (cli *client.Client, host string, err error) {
-	tcConfig := readTCPropsFile()
+func NewDockerClient() (cli *client.Client, host string, tcConfig TestContainersConfig, err error) {
```
mdelapenya, line 616: We should mark this PR as a breaking change, as we are changing method signature, adding a new returned method. Not a problem though as we are not in 1.0, but at least let's add the label so that it appears as a breaking change in the release notes.

mdelapenya (approved):
LGTM! Thanks for your time in this PR @zregvart!!

Let's fix that tiny typo and we are ready to go 🚀

docker.go
zregvart: Seems that I don't have permissions to set the `type/bc-break` label.

mdelapenya: Ah no worries, only maintainers add the label (we could probably change it if needed). Will add it if you agree that this is a breaking change.

mdelapenya (approved):
Great job with this PR, thank you very much @zregvart!!

Issue #411, opened 2022-02-28
Flaky test: TestContainerLogWithErrClosed


mdelapenya:
I noticed this test is randomly failing. We'd need to further investigate why it fails and in which conditions. It may be written in a way that is not consistent, or maybe not up-to-date. Opening this issue for awareness.

```
=== RUN   TestContainerLogWithErrClosed
2022/02/28 16:27:00 Starting container id: 07de0af25f64 image: docker:dind
2022/02/28 16:27:00 Waiting for container id 07de0af25f64 image: docker:dind
2022/02/28 16:27:17 Container is ready id: 07de0af25f64 image: docker:dind
    logconsumer_test.go:210: port not found
--- FAIL: TestContainerLogWithErrClosed (20.54s)
```

baez90:
I stumbled upon this issue while working on #414 I think it should be resolved in my PR by correctly using the [`PortEndpoint(ctx context.Context port nat.Port, proto string) (string, error)`](https://pkg.go.dev/github.com/testcontainers/testcontainers-go#DockerContainer.PortEndpoint) function

mdelapenya:
Gotestsum support rerunning failed tests: https://github.com/gotestyourself/gotestsum#re-running-failed-tests

We could give it a try for adding stability to the pipeline meanwhile we investigate this flakiness

PR #451, opened 2022-06-13
feature: custom ForSQL query

### LINKED ISSUE #350: Feature request: Enable specifying query for `wait.ForSQL`.
It would be a nice feature if we could specify a query instead of being forced to use the default query of `SELECT 1` [here](https://github.com/testcontainers/testcontainers-go/blob/54854fb15ccb5162fe673bac4b2522e5f84de9a0/wait/sql.go#L75). It is stated that this repo helps with `integration/smoke tests` in the `README`, and I think this would help enable that better. With allowing users to specify what query to run, we could then wait for specific test data to propagate properly in the database before running test cases using only one wait strategy. This could potentially save people from having to specify a bunch of wait strategies for each specific service too (pending set up, number of services, project, use case, etc)!

funvit:
- a new setter added `WithQuery`

Closes #350


mdelapenya:
This LGTM, thanks for taking the time to contribute it! I'd miss a test demonstrating the change, but other than that, great job!

wait/wait.go
```diff
@@ -28,3 +28,7 @@ func defaultStartupTimeout() time.Duration {
 func defaultPollInterval() time.Duration {
 	return 100 * time.Millisecond
 }
+
+func defaultQuery() string {
```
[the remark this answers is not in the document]
funvit: I simply followed the style of other default* methods.

funvit:
Tests added.

wait/sql.go
```diff
@@ -9,6 +9,8 @@ import (
 	"github.com/docker/go-connections/nat"
 )
 
+const defaultForSqlQuery = "SELECT 1"
```
mdelapenya, line 12: Neat! I like it here more, as it's coupled with the SQL code

funvit:
So what to do with failed checks?

mdelapenya:
> So what to do with failed checks?

We have certain flakiness https://github.com/testcontainers/testcontainers-go/issues/411, but had no bandwidth to work on it yet. In any case, I re-ran your builds 🤞

funvit:
> > So what to do with failed checks?
> 
> We have certain flakiness #411, but had no bandwidth to work on it yet. In any case, I re-ran your builds crossed_fingers

Maybe merge my #455 at first?

container_test.go
```diff
@@ -12,9 +12,12 @@ import (
 	"testing"
 	"time"
 
+	"github.com/docker/go-connections/nat"
 	"github.com/stretchr/testify/assert"
 
 	"github.com/testcontainers/testcontainers-go/wait"
+
+	_ "github.com/lib/pq"
```
mdelapenya, line 20: Do we need the import (and the dependency)?

I'll double check it fetching the PR locally

mdelapenya (commented):
My only concern is about adding a dependency on Postgres, which is needed to tun the tests, but will be added to the project dependency on each client code running their tests with tc-go.

I understand the purpose of adding it, although I'd like to avoid adding that dependency. Do you foresee any solution for that?

funvit:
> My only concern is about adding a dependency on Postgres, which is needed to tun the tests, but will be added to the project dependency on each client code running their tests with tc-go.
> 
> I understand the purpose of adding it, although I'd like to avoid adding that dependency. Do you foresee any solution for that?

I can move `TestContainerWithWaitForSQL` to `e2e` directory and create there custom `go.mod`.  This will remove dependency from main package.

After that gitlab CI file must be updated...

e2e/Makefile
```diff
@@ -0,0 +1,8 @@
+
```
mdelapenya, line 1: 💯 for using Make!!

mdelapenya: I think we should include this test execution into the github action. Also creating a Makefile at the root dir, including both test executions, would be great

funvit: Done.

Makefile
```diff
@@ -0,0 +1,9 @@
+
+.PHONY: test
+test: tools
```
mdelapenya, line 3: I'd have a separate Make goal for the unit tests, and another one for the e2e, so that it's possible to call them in separated stages of the build pipeline

e2e/go.mod
```diff
@@ -0,0 +1,11 @@
+module e2e
```
mdelapenya, line 1: @gianarb wdyt about what we proposed here about having a separated module to avoid pushing the postgres dependency to downstream consumers?

Makefile
funvit: Done

e2e/go.mod
funvit: Ok. Readme added.

mdelapenya:
Hey @gianarb I'd like you to take a deep look here because this PR comes with a submodule to avoid including postgres dependency as part of the project. And this is pretty related to the "canned" modules discussion we have in the past: https://github.com/testcontainers/testcontainers-go/issues/112

I'd appreciate your comments here

gianarb:
I think in this case it is fine, there is a separate module and this is not a new feature but an enhancement for something we already support.

Makefile
```diff
@@ -0,0 +1,17 @@
+
+.PHONY: test-all
+test-all: tools test-unit test-e2e
+
+.PHONY: test-unit
+test-unit:
+	@echo "Running unit tests..."
+	go test ./...
```
mdelapenya, line 8: Can you use here the code from the github action for the unit tests, and call the make code there?

mdelapenya: Once this is in place, I think we are ready to merge 😄

e2e/Makefile
```diff
@@ -0,0 +1,8 @@
...
+.PHONY: test
+test: tools
+	go test ./...
```
mdelapenya, line 4: Can we use here `gotestsum`, as we already do in https://github.com/testcontainers/testcontainers-go/pull/451/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR55? 🙏 

This way we will be using the same tools and generating the test reports in the same formats

e2e/README.md
```diff
@@ -0,0 +1,4 @@
+End-to-end tests module.
```
mdelapenya, line 1:
```suggestion
# End-to-end tests module.
```

e2e/README.md
```diff
@@ -0,0 +1,4 @@
...
+
+Main purpose of this module is to have a separated module to avoid pushing
+unexpected external dependencies like `github.com/lib/pq` into main module.
```
mdelapenya, line 4:
```suggestion
unexpected external dependencies like `github.com/lib/pq` into the main module.
```

Makefile
funvit: done

e2e/Makefile
funvit: done

PR #452, opened 2022-06-13
feat: support binds in addition to mounts

zregvart:
This is needed for providing additional options, such as SELinux label.

### DIFF
--- a/container.go
+++ b/container.go
@@ -104,6 +104,7 @@ type ContainerRequest struct {
 	AutoRemove      bool   // if set to true, the container will be removed from the host when stopped
 	AlwaysPullImage bool   // Always pull image
 	ImagePlatform   string // ImagePlatform describes the platform which the image runs on.
+	Binds           []string
 }
 
 type (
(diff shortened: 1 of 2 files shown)


mdelapenya:
Thanks for your contribution! It LGTM, although I'd thank more context in the PR description regarding the original need and the motivation on the change (sorry, our contribution guidelines are not that clear, so please forgive us).

I can see it adds the Binds string array, so that it's possible to add a list of volume bindings for this container (https://github.com/moby/moby/blob/v20.10.17/api/types/container/host_config.go#L393), but I wouldn't like to assume things in advance. Could you elaborate the description a little bit please?

Other than that, the changes looks good to me!

zregvart:
@mdelapenya thanks for having a look. This is what I can't do with Mounts that are already supported:

```go
    Binds: []string{"/src:/dest:Z"},
    // ...
}
```

Which is an equivalent of running:

```sh
$ docker -v /src:/dest:Z ...
```

That is there is [no option](https://pkg.go.dev/github.com/docker/docker@v20.10.16+incompatible/api/types/mount#Mount) to provide the `Z` [SELinux label](https://docs.docker.com/storage/bind-mounts/#configure-the-selinux-label) using the existing functionality.

PR #457, opened 2022-06-13
Enable using stdin when starting docker compose

### LINKED ISSUE #285: Feature request: Enable using stdin when starting docker compose
It would be a nice feature if one can use stdin to inject a docker compose file on the fly instead of using a file path. Docker Compose supports this. Wenn giving ``-`` as filename to command line option ``-f`` docker compose reads the configuration from stdin.

Maybe somewhere [here](https://github.com/testcontainers/testcontainers-go/blob/master/compose.go#L160) this can be done (but I am not good enough to make an acceptable pull request).

mal-as:
Closes #285

### DIFF
--- a/compose.go
+++ b/compose.go
@@ -427,7 +427,7 @@ func (dc *LocalDockerCompose) makeTmpFile() error {
 		return err
 	}
 
-	file, err := ioutil.TempFile("/Users/aleksamalyshev/", "__tmp_")
+	file, err := ioutil.TempFile("", "__tmp_")
 	if err != nil {
 		return err
 	}
--- a/compose.go
+++ b/compose.go
@@ -105,7 +105,7 @@ func NewLocalDockerCompose(filePaths []string, identifier string, opts ...LocalD
 	dc.absComposeFilePaths = make([]string, len(filePaths))
 	for i, cfp := range dc.ComposeFilePaths {
 		if cfp == "-" {
-			if err := dc.makeTmpFile(); err != nil {
+			if err := dc.makeTmpFile(identifier); err != nil {
 				continue
 			}
 			abs, _ := filepath.Abs(dc.tmpFileName)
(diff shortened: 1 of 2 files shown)


compose.go
```diff
@@ -402,3 +416,29 @@ func which(binary string) error {
...
+	return ioutil.ReadAll(os.Stdin)
+}
+
+func (dc *LocalDockerCompose) makeTmpFile() error {
+	data, err := readSTDIN()
+	if err != nil {
+		return err
+	}
+
+	file, err := ioutil.TempFile("/Users/aleksamalyshev/", "__tmp_")
```
mdelapenya, line 430: I think this is a leftover 😅

compose.go
```diff
@@ -402,3 +416,29 @@ func which(binary string) error {
 
 	return err
 }
+
+func readSTDIN() ([]byte, error) {
+	return ioutil.ReadAll(os.Stdin)
+}
+
+func (dc *LocalDockerCompose) makeTmpFile() error {
+	data, err := readSTDIN()
```
mdelapenya, line 425: Better in one line?
```suggestion
func (dc *LocalDockerCompose) makeTmpFile() error {
	data, err := ioutil.ReadAll(os.Stdin)
```

compose.go
```diff
@@ -402,3 +416,29 @@ func which(binary string) error {
...
+	if err != nil {
+		return err
+	}
+
+	file, err := ioutil.TempFile("", "__tmp_")
```
mdelapenya, line 430: Do you foresee any collision between multiple runs? Do you think we could use the compose identifier as part of the file name in that case 🤔?

compose.go
```diff
@@ -103,8 +104,16 @@ func NewLocalDockerCompose(filePaths []string, identifier string, opts ...LocalD
 
 	dc.absComposeFilePaths = make([]string, len(filePaths))
 	for i, cfp := range dc.ComposeFilePaths {
-		abs, _ := filepath.Abs(cfp)
-		dc.absComposeFilePaths[i] = abs
+		if cfp == "-" {
```
mdelapenya, line 107: Mmm I do not get this line. Is it expected to receive a `-` as a file path for a compose file?

mdelapenya (commented):
Thanks for your work with this PR. I left a few review comments before merging it. Please let me know what you think

mdelapenya:
@mal-as  #470 was created yesterday, and I think it collides with this one. What are your thoughts on that PR?

compose.go
mal-as: Yes, it is

compose.go
mal-as: fixed it

mdelapenya:
I think this PR is superseded by #466, therefore I'm closing it. @mal-as could you please reopen it if you feel it's different? Thanks!

Issue #158, opened 2020-02-19
Add final check in CI to double check that environment is clean


gianarb:
In some way I would like to know if at the end of the test suite we still have some containers running, this is to be sure that the cleanup process works

ClaytonNorthey92:
Would it be a good idea to use the Go Docker client to do so?  [Here is a good example](https://godoc.org/github.com/docker/docker/client#hdr-Usage) to list running containers.

Would we want to perform any action if there are still running containers? I assume we would want to stop them?

gianarb:
Yes, I think it sounds like a good idea, it can be a bah script that runs as the last check, it should fail if there are containers running yes.

ClaytonNorthey92:
> Yes, I think it sounds like a good idea, it can be a bah script that runs as the last check, it should fail if there are containers running yes.

@gianarb do we need a bash script? I think we can perform this check in Go with the Go Docker Client, we might be able to run the check as part of our tests, the check could run after a suite has finished

mdelapenya:
My two cents: if the only goal is to ensure that there are no running containers (or networks or volumes) after the test suite runs, I'd simply go with a shell script returning 1. Writing it in Go would mean to maintain it 🤔

gianarb:
@ClaytonNorthey92 I think @mdelapenya got what I am thinking about. Just a small hack to be sure that we get 0 containers running

ClaytonNorthey92:
okay makes sense to me 👍

PR #458, opened 2022-06-13
check that environment is clean

VladimirStepanov:
close https://github.com/testcontainers/testcontainers-go/issues/158

### DIFF
--- a/scripts/check_environment.sh
+++ b/scripts/check_environment.sh
@@ -7,6 +7,6 @@ if [ "$containers" -eq "0" ]; then
    exit 0
 fi
 
-echo "$(docker ps)"
+docker ps
 echo "Number of containers are still running:" "$containers"
 exit 255;
--- a/scripts/check_environment.sh
+++ b/scripts/check_environment.sh
@@ -1,12 +1,12 @@
 #!/usr/bin/env bash
 
-
+docker_out="/tmp/dout.txt"
 containers=$(docker ps --format "{{.ID}}" | wc -l)
 
 if [ "$containers" -eq "0" ]; then
    exit 0
 fi
 
+echo "$(docker ps)"
 echo "Number of containers are still running:" "$containers"
-
 exit 255;
--- a/scripts/check_environment.sh
+++ b/scripts/check_environment.sh
@@ -1,14 +1,14 @@
 #!/usr/bin/env bash
 
 i=0
-while [ $i -ne 15 ]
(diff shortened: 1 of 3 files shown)


VladimirStepanov:
@mdelapenya Hi! Could you check this solution?

mdelapenya:
> @mdelapenya Hi! Could you check this solution?

I think this would work 👍 

On the other hand, I'm missing a contributing guide in the project, not in this PR of course. I'd add the usage of this shell script (and the one for linting) once we create that guide.

scripts/check_environment.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+
+containers=$(docker ps --format "{{.ID}}" | wc -l)
+
+if [ "$containers" -eq "0" ]; then
+   exit 0
+fi
+echo "environment is not clean"
```
mdelapenya, line 9: Could we list the number of containers that are still running? or a least of them?

VladimirStepanov: @mdelapenya done

scripts/check_environment.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+i=0
+while [ $i -ne 10 ]
```
mdelapenya, line 4: IIUC we are going to loop 10 times waiting for the containers to be removed, sleeping 2 secs on each loop, right?

mdelapenya (approved):
LGTM, thanks for this small but important contribution!

scripts/check_environment.sh
VladimirStepanov: @mdelapenya Yes, you understood correctly. I’ve noticed that containers were really removed from ‘docker ps’ output after N seconds( in average in ten seconds).

That’s why I decided to wait 20 seconds for the result. The script checks the Docker output 10 times every 2 seconds to speed up the process . And if the output is not empty after the loop, the script returns an error.

mdelapenya:
In https://github.com/testcontainers/testcontainers-go/runs/6905008920?check_suite_focus=true we see there is a container not stopped:
Screenshot 2022-06-16 at 07 33 59

Could you try to identify it and stopped in the test suite?

VladimirStepanov:
@mdelapenya Fixed:)
Now we pass all tests successfully

VladimirStepanov:
@mdelapenya Could we merge it?

PR #461, opened 2022-06-14
Add SHM size to the container request

EndPositive:
The code-way of specifying --shm-size on the Docker command-line.

E.g.
```golang
var shmsize = 1073741824 // = 1GB
```

I've been using this to start Selenium nodes, for example.

### DIFF
--- a/container.go
+++ b/container.go
@@ -105,7 +105,7 @@ type ContainerRequest struct {
 	AlwaysPullImage bool   // Always pull image
 	ImagePlatform   string // ImagePlatform describes the platform which the image runs on.
 	Binds           []string
-	ShmSize         int64 // Amount of memory shared with the host (in MB)
+	ShmSize         int64 // Amount of memory shared with the host (in bytes)
 }
 
 type (


mdelapenya (changes requested):
Thanks for taking the time to work on this contribution, it's much than appreciated!

My comments: although the documentation for `shm` uses this format:
>Size of /dev/shm. The format is <number><unit>. number must be greater than 0. Unit is optional and can be b (bytes), k (kilobytes), m (megabytes), or g (gigabytes). If you omit the unit, the system uses bytes. If you omit the size entirely, the system uses 64m.

which could advocate for a more elaborated format, I think using a format with just the bytes is great for simplicity.

What I'd take a look is at the consistency with both Docker client and the Java project, which uses `bytes` as default unit.

- Docker client: https://github.com/moby/moby/blob/1a0587bd766e3ab1f6dbe808a94a41a2f54d9126/docs/api/v1.41.yaml#L1111
- TC-java: https://github.com/testcontainers/testcontainers-java/blob/de1324ed2800eff4da326d0c23d281399d006bc0/core/src/main/java/org/testcontainers/containers/GenericContainer.java#L180

Once updated to use `bytes`, I think this PR is perfect!

mdelapenya (approved):

Issue #425, opened 2022-04-06
Support deploying with docker compose using API


bamsammich:
I'm attempting to run tests from a container without docker installed. This isn't an issue for ContainerRequests since it's using the API, but compose relies on `docker-compose` being present on the machine. It would be beneficial to be able to deploy a compose file without the requirement for external dependencies.

mdelapenya:
Thanks for creating this issue! Yes, it's on the plans, exactly the same as the Java folks do with the ContainerisedDockerCompose (https://github.com/testcontainers/testcontainers-java/blob/06fc95f5233b0c11d2ba319b4b02587db0b5028b/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java#L610)

We are currently using the Local docker compose choice, but it would be more than helpful if you want to contribute it, and we will be glad to review it

baez90:
I'm actually wondering why not directly 'including' upstream `docker-compose` as version 2 is implemented in Go. I had a 'quick look' how hard it would be and I think it's actually not that hard to integrate it directly without having to pull any `docker-compose` in a container whatsoever. I'd be willing to give it a shot if there's some interest. This would actually make a few things easier if we'd not have to deal with different `docker-compose` versions as the service names are constructed in different ways and stuff...

fiftin:
Hi @mdelapenya 

I added `NativeDockerComposeExecutor` to my PR which uses docker-compose as library. It works.
But I faced problem with environment variables. We can't pass env vars to it separately from testcontainers.

baez90:
@mdelapenya I know you're not going to like this but: how about another breaking change/deprecating the `DockerCompose` interface? :smile: 

The current interface:

```go
type DockerCompose interface {
	Down() ExecError
	Up() ExecError
	WaitForService(string, wait.Strategy) DockerCompose
	WithCommand([]string) DockerCompose
	WithEnv(map[string]string) DockerCompose
	WithExposedService(string, int, wait.Strategy) DockerCompose
}
```

Isn't really idiomatic Go and doesn't really fit when working with the compose Go API.
I'm thinking about something like this:

```go
type ComposeStack interface {
	Up(ctx context.Context) error
	Down(ctx context.Context) error
	WaitForService(string, wait.Strategy) ComposeStack
	WithEnv(map[string]string) ComposeStack
	WithExposedService(string, int, wait.Strategy) ComposeStack
}
```

[...]

PR #462, opened 2022-06-14
Support deploying with docker compose without local docker-compose

fiftin:
Related to this issue: https://github.com/testcontainers/testcontainers-go/issues/425.

I added support conteinerised docker-compose. To do this I refectored `compose.go` and extracted `ComposeExecutor` which can be:
* `ContainerizedDockerComposeExecutor` - uses docker-compose from docker container.
* `LocalDockerComposeExecutor` - uses local docker-compose.


compose.go
```diff
@@ -76,29 +75,25 @@ func (f LocalDockerComposeOptionsFunc) ApplyToLocalCompose(opts *LocalDockerComp
...
 
-// NewLocalDockerCompose returns an instance of the local Docker Compose, using an
-// array of Docker Compose file paths and an identifier for the Compose execution.
-//
-// It will iterate through the array adding '-f compose-file-path' flags to the local
-// Docker Compose execution. The identifier represents the name of the execution,
-// which will define the name of the underlying Docker network and the name of the
-// running Compose services.
-func NewLocalDockerCompose(filePaths []string, identifier string, opts ...LocalDockerComposeOption) *LocalDockerCompose {
+func NewDockerCompose(
```
mdelapenya, line 78: This signature change comes with a breaking change :) Thankfully we are not in 1.x version, so we can change contracts

mdelapenya: Oh oh wait. Sorry, but I thought this line was changing the `NewLocalDockerCompose` method, but it's not: it's adding a new generic method.

compose_executors.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+
+	if len(cmds) == 0 {
+		cmds = append(cmds, "-f", "/app/docker-compose.yml")
+	}
+
+	args := append(cmds, options.Args...)
+
+	req := ContainerRequest{
+		Image: "docker/compose:1.29.2",
```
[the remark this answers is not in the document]
fiftin: Agree, will do

compose.go
```diff
@@ -116,6 +111,21 @@ func NewLocalDockerCompose(filePaths []string, identifier string, opts ...LocalD
...
+//
+// It will iterate through the array adding '-f compose-file-path' flags to the local
+// Docker Compose execution. The identifier represents the name of the execution,
+// which will define the name of the underlying Docker network and the name of the
+// running Compose services.
+func NewLocalDockerCompose(filePaths []string, identifier string, opts ...LocalDockerComposeOption) *LocalDockerCompose {
+	return NewDockerCompose(filePaths, identifier, LocalDockerComposeExecutor, nil, opts...)
+}
+
+func NewContainerizedDockerCompose(filePaths []string, identifier string, opts ...LocalDockerComposeOption) *LocalDockerCompose {
```
mdelapenya, line 125: Do we still to return a `Local`DockerCompose? It seems weird to me to use a containerised compose method that returns a LocalDC.

I'd be open to using a dedicated DockerisedDockerCompose type, although I think could come with changes in return types and generalisation...

Of course, not a blocker for this PR, but if you are interested in following-up with more refactors to make a stable API, I'm too

compose_test.go
```diff
@@ -90,68 +89,91 @@ func ExampleLocalDockerCompose_WithEnv() {
 	})
 }
 
+func enumDockerComposes(filePaths []string, identifier string, executor func(compose *LocalDockerCompose), t *testing.T) {
+	composes := []*LocalDockerCompose{
+		NewLocalDockerCompose(filePaths, identifier, WithLogger(TestLogger(t))),
+		NewContainerizedDockerCompose(filePaths, identifier, WithLogger(TestLogger(t))),
```
mdelapenya, line 95: I think we are using the new Dockerised compose here. Could you update the docs with examples on how to use it? There is a dedicated section for compose in the docs/features dir.

compose.go
fiftin: I started with creating dedicated `DockerisedDockerCompose` type, but got too many duplications. Using current `LocalDockerCompose` with different executors is a best solution for my opinion.

IMHO: interface `DockerCompose` should be removed and type `LocalDockerCompose` should renamed to `DockerCompose`.

P.S.:

I replaced ref to function to interface `DockerComposeExecutor` with 2 implementations.

compose_executors.go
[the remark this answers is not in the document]
fiftin: I replaced ref to function to interface `DockerComposeExecutor` with 2 implementations.

fiftin:
I replaced ref to function to interface `DockerComposeExecutor` with 2 implementations.

compose_test.go
fiftin: Done

Issue #439, opened 2022-05-12
Use existing container?


SpencerC:
Possibly a newbie question: is there a way to leave containers running between tests and reconnect to them rather than waiting for them to restart every time?

mdelapenya:
Hi @SpencerC, you can keep containers alive after a test run simply passing a bool value to the container request: https://github.com/testcontainers/testcontainers-go/blob/dbb7f60fc0ded54a293adb11ff3097054b2fd495/container.go#L102

What I'm not sure is the main purpose of reusing containers between tests run: could you please elaborate a bit more? Ideally, a container should live alongside the test consuming it. You could define "test lifecycle hooks" beforeSuite/beforeTest depending of your testing framework, to keep containers running for a test suite/case. Please take a look at [Ginko](https://github.com/onsi/ginkgo) or [Godog](https://github.com/cucumber/godog), or use the Main function in the test package to control that.

In any case, it would be great if you can share with us your use case so we can find the best approach that works for you

SpencerC:
Thanks @mdelapenya! I'm building some integration tests which rely on a container running a database, mirroring this pattern: https://github.com/stargate/stargate-grpc-go-client/blob/main/stargate/pkg/client/client_test.go#L45-L59

This particular container takes ~18 seconds to start. It would be nice to start it once, run the tests, have the option to explore the state in DataGrip, then reset it between tests using `DROP` or `TRUNCATE` commands. To be clear, I definitely see the benefit of test-specific containers for final CI testing. Just looking for a way to speed up local development.

Setting `SkipReaper` does indeed leave the container running, but then creates a new container for each test run.

Side note: https://www.youtube.com/watch?v=yi5A3cK1LNA. tldr: it's not a great idea to use external testing frameworks in Go. You can achieve the same effects using [subtests](https://pkg.go.dev/testing#T.Run), [`Cleanup`](https://pkg.go.dev/testing#T.Cleanup), and [Main](https://pkg.go.dev/testing#hdr-Main) (https://stackoverflow.com/questions/23729790/how-can-i-do-test-setup-using-the-testing-package-in-go).

CC: @mpenick

mdelapenya:
>Side note: https://www.youtube.com/watch?v=yi5A3cK1LNA. tldr: it's not a great idea to use external testing frameworks in Go.

[...]

PR #464, opened 2022-06-15
Issue 439: use an existing container

VladimirStepanov:
Close https://github.com/testcontainers/testcontainers-go/issues/439


generic.go
```diff
@@ -58,6 +58,33 @@ func GenericContainer(ctx context.Context, req GenericContainerRequest) (Contain
 	return c, nil
 }
 
+// GenericReusableContainer reuses a container if it exists or creates a new one with parameters
+// Reusing will work only if you pass an existing container's name via 'req.Name' field
+// If the name is empty or it is not in a list of existing containers, the function will create a new generic container
+
```
mdelapenya, line 64:
```suggestion
```

docs/features/creating_container.md
```diff
@@ -81,3 +81,59 @@ func TestIntegrationNginxLatestReturn(t *testing.T) {
...
 }
 ```
+
+## GenericReusableContainer
+
+`testcontainers.GenericReusableContainer` reuses a container if it exists or creates a generic container with parameters.
+Reusing will work only if you pass an existing container's name via 'req.Name' field.
+If the name is empty, or it is not in a list of existing containers, the function will create a new generic container
+
+The following test creates an NGINX container, add a file into it and then reuse the container again for checking the file:
```
mdelapenya, line 91:
```suggestion
The following test creates an NGINX container, adds a file into it and then reuses the container again for checking the file:
```

generic.go
```diff
@@ -58,6 +58,33 @@ func GenericContainer(ctx context.Context, req GenericContainerRequest) (Contain
...
+func GenericReusableContainer(ctx context.Context, req GenericContainerRequest) (Container, error) {
```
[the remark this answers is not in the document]
VladimirStepanov: Yeah, I agree with you. Could you suggest me, is it ok if I add this logic to `GenericContainer` ? Because `GenericReusableContainer ` and `GenericContainer` have similar input arguments and return values. 

If it is ok, i will add new comments and examples to 'GenericContainer'.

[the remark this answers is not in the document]
mdelapenya: Let me check how the Java flavour supports it. Will ping you back here

docs/features/creating_container.md
```diff
@@ -81,3 +81,59 @@ func TestIntegrationNginxLatestReturn(t *testing.T) {
 	}
 }
 ```
+
+## GenericReusableContainer
+
+`testcontainers.GenericReusableContainer` reuses a container if it exists or creates a generic container with parameters.
+Reusing will work only if you pass an existing container's name via 'req.Name' field.
+If the name is empty, or it is not in a list of existing containers, the function will create a new generic container
```
mdelapenya, line 89: Reading this description, which is pretty clear, this thought came to my mind: why adding a new type of container that could be used in a wrong manner, like creating a GenericReusableContainer but with a name that is not reusable?

This is thinking out loud: do you think adding the same logic to the reusable container into the generic container would make sense? Please take a look at how the Java folks do it: https://github.com/testcontainers/testcontainers-java/blob/de1324ed2800eff4da326d0c23d281399d006bc0/core/src/main/java/org/testcontainers/containers/GenericContainer.java#L384

The GenericContainer, if defined as reusable, reuses the container ID. I think this API would be simpler in terms of usability: as a consumer I don't have to remember the different types of Containers, simply configure the generic one.

Again, this is an opinion, as valid as any other, so please tell us your thoughts

VladimirStepanov: @mdelapenya I want to suggest next behavior: we add `Reuse` optional flag to `GenericContainerRequest`, because maybe a user doesn't want to reuse an existing container..

```go
c, err := GenericReusableContainer(ctx, GenericContainerRequest{
		Image:        "nginx:1.17.6",
		ExposedPorts: []string{"80/tcp"},
		WaitingFor:   wait.ForListeningPort("80/tcp"),
		Name:         reusableContainerName,
	},
	Started: true,
	Reuse: true,
})
```

And we will get next possible options:
1) `Reuse` is true. We will reuse an existing container or create a new one if it not exists . If we reuse an existing container, we will ignore flag `Started`, because the container is already started. 
2) `Reuse` is false. Try to create a new container or return an error if a container already exists.

Wdyt?

mdelapenya: I like it! Please go for it 🚀

VladimirStepanov:
@mdelapenya updated:)

docker.go
```diff
@@ -1011,6 +1017,62 @@ func (p *DockerProvider) CreateContainer(ctx context.Context, req ContainerReque
 	return c, nil
 }
 
+func (p *DockerProvider) findContainerByName(ctx context.Context, name string) (*types.Container, error) {
+	if name == "" {
+		return nil, nil
+	}
+	containers, err := p.client.ContainerList(ctx, types.ContainerListOptions{})
```
mdelapenya, line 1024: Is it possible to directly filter by name, instead of the below for-loop?

mdelapenya (approved):

docker.go
```diff
@@ -1011,6 +1019,61 @@ func (p *DockerProvider) CreateContainer(ctx context.Context, req ContainerReque
...
+		return nil, err
+	}
+	if c == nil {
+		return p.CreateContainer(ctx, req)
+	}
+
+	sessionID := uuid.New()
+	var termSignal chan bool
+	if !req.SkipReaper {
+		r, err := NewReaper(ctx, sessionID.String(), p, req.ReaperImage)
```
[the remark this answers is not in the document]
mdelapenya: The reaper will only be created if it does not exist, so the code for the reaper creation is there, but will check if a real reaper container already exists.

PR #463, opened 2022-06-15
add containerised docker compose and tests

mazitovt:
Feature for Issue #425

Added `ContainerisedDockerCompose` struct with embedded `LocalDockerCompose` and override all `DockerCompose` methods. Added two functions `executeInContainer` and `executeComposeInContainer`, that work with docker/compose container.

PR #465, opened 2022-06-15
feat: add reaper for compose

### LINKED ISSUE #372: Containers created by Docker Compose are not cleaned by Ryuk
Containers created and started by testcontainers-go through Docker Compose are not guaranteed to be cleaned up (and this is something I personally observed when running the testcontainers-go test suite on WSL), since Ryuk is not aware of them.

We solved this issue in testcontainers-java by registering the project label added by Docker Compose with Ryuk:
https://github.com/testcontainers/testcontainers-java/blob/76f11b344ce42ad920828b0628a7ec929a9d0eca/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java#L312

oGi4i:
Closes #372

### DIFF
--- a/compose.go
+++ b/compose.go
@@ -16,6 +16,7 @@ import (
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/filters"
 	"github.com/docker/docker/client"
+	"github.com/google/uuid"
 	"gopkg.in/yaml.v3"
 
 	"github.com/testcontainers/testcontainers-go/wait"
@@ -54,6 +55,7 @@ type LocalDockerCompose struct {
 	Services             map[string]interface{}
 	waitStrategySupplied bool
 	WaitStrategyMap      map[waitService]wait.Strategy
+	reaper               *Reaper
 }
 
 type (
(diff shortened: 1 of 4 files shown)


oGi4i:
@mdelapenya
Not the most elegant solution, would love to hear your thoughts on this.

mdelapenya:
Hi @oGi4i, thanks for your first contribution to TC-go.

We are currently reviewing https://github.com/testcontainers/testcontainers-go/pull/463, which could cause this PR to break.

If you agree, I'd like to have your PR _after_ the other one.

oGi4i:
> If you agree, I'd like to have your PR _after_ the other one.

Sure, no problem

mdelapenya:
Hi @oGi4i  we have merged #476, which adds native support for docker compose. Would you mind revisiting this PR and update to that code? I can assist if needed, or we can even ping @baez90 for that.

Thanks in advance

mdelapenya:
@oGi4i I'm closing this issue as stale.

OTOH it would be ideal if the compose APIs allow passing labels to the container CreateOptions (they only provide it for the RunOptions type), which would way simplify this implementation, simplify passing reaper labels as we do in https://github.com/testcontainers/testcontainers-go/blob/main/docker.go#L996-L999

Probably worth it to ping the Docker folks for that support. If they are not responsive, we could try designing this implementation in a similar way that you started here, but adapted to the new code.

In any case, thanks for your time here.

lwaddicor:
Hey, I came across this problem recently. For the time being I fixed it within our code by creating an intermediate docker compose file which with the reaper labels added for all the services.

Again not the most elegent solution, but works nicely and is mostly transparent to the user. Thought i'd post it here in case it is useful in the future

PR #466, opened 2022-06-16
feat: support creating compose Stacks from Readers

goforbroke1006:
https://github.com/testcontainers/testcontainers-go/issues/285
add constructor that takes []io.Reader


mdelapenya:
Hi @goforbroke1006 , thanks for your first contribution to TC-go.

We are currently reviewing https://github.com/testcontainers/testcontainers-go/pull/463, which could cause this PR to be updated.

If you agree, I'd like to have your PR _after_ the other one.

goforbroke1006:
Hi @mdelapenya ,
Ok!

mdelapenya:
Hi @goforbroke1006 we have merged #476, which adds native support for docker compose. Would you mind revisiting this PR and update to that code? I can assist if needed, or we can even ping @baez90 for that.

Thanks in advance

goforbroke1006:
Hi @mdelapenya ,
I take fresh main branch and updated code.

compose.go
```diff
@@ -175,3 +179,41 @@ func NewLocalDockerCompose(filePaths []string, identifier string, opts ...LocalD
 
 	return dc
 }
+
+func NewDockerComposeWithReaders(readers []io.Reader) (*dockerCompose, error) {
```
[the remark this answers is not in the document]
goforbroke1006: Yeah, it's great idea.
I created `WithStackReaders` helper function. And it can be used with `NewDockerComposeWith` contructor.
And of course I remove `NewDockerComposeWithReaders`.
Unit test is updated.

But in `ComposeStackOption.applyToComposeStack` implementation for ComposeStackReaders I did not find another way process errors, so use panic(err). I'm not sure it's good idea, but don't know how to do better.

compose_api.go
```diff
@@ -92,6 +96,45 @@ func (f StackIdentifier) String() string {
...
+
+	// choose directory to keep temporary files
+	// like
+	// 		/tmp/testcontainers-go/my-awesome-service-2f686f6d652f676f666f7262726f6b65
+	projectName := filepath.Base(currentDir)
+	projectHash := fmt.Sprintf("%x", fnv.New32a().Sum([]byte(currentDir)))[:32]
+	tmpDir := filepath.Join(os.TempDir(), "testcontainers-go", fmt.Sprintf("%s-%s", projectName, projectHash))
+	if err := os.RemoveAll(tmpDir); err != nil {
+		panic(err)
+	}
```
mdelapenya, line 115: Is this needed to avoid collisions? I think we can skip this, or at least log the error but continuing

compose_api.go
```diff
@@ -92,6 +96,45 @@ func (f StackIdentifier) String() string {
...
+	currentDir, err := os.Getwd()
+	if err != nil {
+		panic(err)
+	}
+
+	// choose directory to keep temporary files
+	// like
+	// 		/tmp/testcontainers-go/my-awesome-service-2f686f6d652f676f666f7262726f6b65
+	projectName := filepath.Base(currentDir)
+	projectHash := fmt.Sprintf("%x", fnv.New32a().Sum([]byte(currentDir)))[:32]
```
mdelapenya, line 111: We are using `uuid.New().String()` for new UUID. I don't mind using the project hash, but for consistency, I'd advocate for one single manner.

We could send a follow-up PR calculating the project hash everywhere in the compose code.

compose_api.go
```diff
@@ -92,6 +96,45 @@ func (f StackIdentifier) String() string {
...
+	tmpDir := filepath.Join(os.TempDir(), "testcontainers-go", fmt.Sprintf("%s-%s", projectName, projectHash))
```
mdelapenya, line 112: For simplification, and removal of the panics, I'd simply create the tmpDir like this:

```go
	composeID := uuid.New().String()
	tmpDir := filepath.Join(os.TempDir(), composeID)
```

compose_api.go
goforbroke1006: Yes, tried to keep directory clean and don't mixed it with another files in project-name-collision cases.
I removed this part.

compose_api.go
goforbroke1006: Each time `uuid.New().String()` is called, a new hash will be created, new directories will be created like
/tmp/testcontainers-go/my-awesome-project-SOME_NEW_HASH_1/
/tmp/testcontainers-go/my-awesome-project-SOME_NEW_HASH_2/
/tmp/testcontainers-go/my-awesome-project-SOME_NEW_HASH_3/
So we will litter the machine with directories. This reason why I did not use uuid lib.
I think better use hash func that always returns same hash for project name/path.

goforbroke1006: Maybe each temporary docker-compose-???.yml file should contains hash also. For situation when developer runs couple of functional tests in parallel.

mdelapenya: Indeed. Let's keep it deterministic then. Thanks for the clarification

compose_api.go
goforbroke1006: I agree. I pushed this changes already.

compose_api.go
```diff
@@ -122,10 +115,9 @@ func (r ComposeStackReaders) applyToComposeStack(o *composeStackOptions) {
 		if err != nil {
 			panic(err)
 		}
-		name := fmt.Sprintf("docker-compose-%d.yaml", idx)
-		filename := filepath.Join(tmpDir, name)
+		filename := filepath.Join(tmpDir, fmt.Sprintf("docker-compose-%d.yaml", idx))
 		if err := os.WriteFile(filename, content, os.ModePerm); err != nil {
-			continue
+			panic(err)
```
mdelapenya, line 120: Are we sure that we want to `panic` here? Probably there is a motivation that I'm not aware of, but I'd try to find a good reason to do so from the library standpoint not to surprise client code.

goforbroke1006: It can be situation if file exists and access changed for it. Application has no permission write to file and I think no way ignore it. Or situation when no free space on disk.
Panic shows that library user should fix system problem before run tests.

mdelapenya: Thanks for the clarification :)

mdelapenya (approved):
I've tested your PR locally, and this LGTM 👍  Thanks for your patience while reviewing it

I'm going to rename the issue for the changelog, to:

`feat: support creating compose stacks from Readers`

mdelapenya:
@baez90, as contributor of the compose code, would you mind giving a quick look at this PR? 🙏

baez90:
Sorry I gave it a quick look but haven't had time to review it properly.

But what came to my mind is: if tc-go should support `[]io.Reader` (or similar) as source I would rather replace the call to [`ProjectFromOptions(...)`](https://github.com/compose-spec/compose-go/blob/08a21dff3db4be06ac783b3eaea2cd9de9cc0f81/cli/options.go#L316) with a custom implementation (including custom options that allow to either set file paths or 

```go
type NamedReader interface {
    io.Reader
    Name() string
}
```

(by default implemented e.g. by `os.File`)

to be able to create a `types.Project`.

Regarding the option interface we're currently using: it'd be possible to extend the signature to be able to return an error if necessary because the interface should not be implemented outside of tc-go therefore changing the signature shouldn't be a breaking change although I'm not sure if this is necessary considering the alternative to create the `NamedReader` instance(s) (or a convenience wrapper like `func NamedReaderFor(r io.Reader, name string) tc.NamedReader`) and use these to create the `types.Project` instance.
What do you think?

Generally I'm wondering what kind of use case you have for this feature 😅 are you generating stack files or downloading them?

mdelapenya:
Thanks @baez90 for your assistance, I wonder if we could tackle your comments in a follow-up PR of in this one, so that we can see progress on it. What would it be your preference? @goforbroke1006 and yours?

baez90:
If you'd prefer to split it I'd probably _not_ update the signature of the options interface because ideally we don't need an `error` return there and a `panic(err)` until it's refactored would be a fair compromise in the meantime?

IMHO options should not include (much) logic but should only influence logic when they are applied hence the requirement of returning an error indicates a design issue for an option.

I'm still curious what the use case of the whole PR is and if this is actually within the scope of tc-go.
As soon as we accept `io.Reader` instances which we store in temporary files the logical next step would be to also clean them when the `types.Project` is instantiated. Or is that then out of scope? What about naming of the temporary files because I think `docker/compose` is using the names for error reporting when parsing the file (s) and so on :blush:
Having said that I'm not against adding support for `[]io.Reader` but there are some things to consider to keep everything concise and maintainable.

kiview:
I would definitely second @baez90 in asking for the context of the PR. Like this, it is unclear to me why it should be added, besides adding a technical capability.

goforbroke1006:
@baez90 Hi
1. Panic - I decide use this way notify user about system problems (file/dir permission etc). Better handle errors, but interface design does not allow it.
2. Logic in options - I can try to move logic to some helper-function to keep options simple. I have no experience in this kind of question. Hope you give advise.
3. Purpose of PR - if I remember right it was proposal add logic to run tests with Stdin as source for docker-compose. I decide io.Reader better solution to satisfy this request.

baez90:
1. Probably I did not make myself clear: I understand why you're `panic()`-ing here, but the interface design is like this for a reason :wink: options should not need to handle errors
2. it is not about whether you put the logic directly in the options or move it to a helper function but about how options are generally working. In most cases they encapsulate values that replace default values of some 'arguments struct' that is then used e.g. for initializing. Consequently you should not read the `[]io.Reader` instances in the option but extend the `composeStackOptions` to either use `[]io.Reader` or `[]string` as stack file source in the `NewDockerCompose()` function (where you could return an `error` already)
3. I'm not 100% sure if the original request from @normanjaeckel still makes sense with the new compose API. Probably an option to pass an already initialized `types.Project` would make more sense and would be more flexible?

goforbroke1006:
Guys, I see. About light-weight options - I agree.
And maybe this feature with reader is not actual now.
So, you can just reject PR )
Or I can revert all changes and just create example of code with DockerCompose and readers.

Issue #343, opened 2021-08-12
Do I have to specify the port for images with one port exposed?


slsyy:
```dockerfile
# Dockerfile
EXPOSE 11211
CMD ["memcached"]
```

```golang
// Code
port := "11211/tcp"
rq := testcontainers.ContainerRequest{
	Image:        "memcached:1.6.10",
	ExposedPorts: []string{port},
	WaitingFor:   wait.ForListeningPort(nat.Port(port)),
}

memcachedC, err := testcontainers.GenericContainer(context.Background(), testcontainers.GenericContainerRequest{
	ContainerRequest: rq,
	Started:          true,
})

url, _ := memcachedC.Endpoint(context.Background(), "")
```

Is there any way to have the same functionality without specifying the port manually in a code?

ajbouh:
I tried to make this work as well and eventually settled for parsing the container definition itself to discover the port

fiftin:
@slsyy Why not use `NetworkMode: "host` ?

mdelapenya:
I think we can close this one, as https://github.com/testcontainers/testcontainers-go/pull/468 was merged and released in v0.14.0, but please let us know if you need anything else here and feel free to reopen it.

Thanks for your time!

PR #468, opened 2022-06-17
feat: Use exposed ports specified in image if it is not specified in ContainerRequest

fiftin:
Relates to https://github.com/testcontainers/testcontainers-go/issues/343

Example:
```dockerfile
# Dockerfile
EXPOSE 11211
CMD ["memcached"]
```

Old:

```go
port := "11211/tcp"
rq := testcontainers.ContainerRequest{
	Image:        "memcached:1.6.10",
	ExposedPorts: []string{port},
	WaitingFor:   wait.ForListeningPort(nat.Port(port)),
}
```
New:

```go
rq := testcontainers.ContainerRequest{
	Image:        "memcached:1.6.10",
	WaitingFor:   wait.ForListeningPort(""),
}
```

### DIFF
--- a/wait/host_port.go
+++ b/wait/host_port.go
@@ -57,8 +57,23 @@ func (hp *HostPortStrategy) WaitUntilReady(ctx context.Context, target StrategyT
 
 	var waitInterval = 100 * time.Millisecond
 
+	internalPort := hp.Port
+	if internalPort == "" {
+		var ports nat.PortMap
+		ports, err = target.Ports(ctx)
+		if err != nil {
+			return
+		}
+		if len(ports) > 0 {
+			for p := range ports {
+				internalPort = p
+				break
+			}
+		}
+	}
+
 	var port nat.Port
-	port, err = target.MappedPort(ctx, hp.Port)
+	port, err = target.MappedPort(ctx, internalPort)
 	var i = 0
 
 	for port == "" {
@@ -68,7 +83,7 @@ func (hp *HostPortStrategy) WaitUntilReady(ctx context.Context, target StrategyT
 		case <-ctx.Done():
 			return fmt.Errorf("%s:%w", ctx.Err(), err)
 		case <-time.After(waitInterval):
-			port, err = target.MappedPort(ctx, hp.Port)
+			port, err = target.MappedPort(ctx, internalPort)
 			if err != nil {
 				fmt.Printf("(%d) [%s] %s\n", i, port, err)
 			}
@@ -101,7 +116,7 @@ func (hp *HostPortStrategy) WaitUntilReady(ctx context.Context, target StrategyT
 	}
 
 	//internal check
-	command := buildInternalCheckCommand(hp.Port.Int())
+	command := buildInternalCheckCommand(internalPort.Int())
 	for {
 		if ctx.Err() != nil {
 			return ctx.Err()
--- a/wait/host_port.go
+++ b/wait/host_port.go
@@ -16,6 +16,8 @@ import (
 var _ Strategy = (*HostPortStrategy)(nil)
 
 type HostPortStrategy struct {
+	// Port is a string containing port number and protocol in the format "80/tcp"
+	// which
 	Port nat.Port
 	// all WaitStrategies should have a startupTimeout to avoid waiting infinitely
 	startupTimeout time.Duration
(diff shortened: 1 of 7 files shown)


wait/host_port.go
```diff
@@ -57,8 +57,23 @@ func (hp *HostPortStrategy) WaitUntilReady(ctx context.Context, target StrategyT
 
 	var waitInterval = 100 * time.Millisecond
 
+	internalPort := hp.Port
+	if internalPort == "" {
```
mdelapenya, line 61: Can we explicitly mention that if the port is the empty string then the first internal exposed port will be used?

wait/host_port.go
```diff
@@ -57,8 +57,23 @@ func (hp *HostPortStrategy) WaitUntilReady(ctx context.Context, target StrategyT
...
+		var ports nat.PortMap
+		ports, err = target.Ports(ctx)
+		if err != nil {
+			return
+		}
+		if len(ports) > 0 {
```
mdelapenya, line 67: Is it possible that `target.Ports(ctx)` retrieves an empty array, so this condition is never met, therefore internalPort is still empty?

wait/wait.go
```diff
@@ -15,6 +15,7 @@ type Strategy interface {
 
 type StrategyTarget interface {
 	Host(context.Context) (string, error)
+	Ports(ctx context.Context) (nat.PortMap, error)
```
mdelapenya, line 18: I think we are polluting the interface adding this method here:

```
	exitStrategyTarget does not implement StrategyTarget (missing Ports method)
```

Do you foresee another manner to get the ports of the target?

wait/host_port.go
fiftin: Fixed, return error of internalPort is empty.

wait/wait.go
fiftin: Fixed test strategies.

fiftin:
I will be very grateful if you check my PR ASAP. I can not continue to work on the PR after Jul 13. Thank you!

wait/host_port.go
```diff
@@ -57,8 +57,28 @@ func (hp *HostPortStrategy) WaitUntilReady(ctx context.Context, target StrategyT
...
+			for p := range ports {
+				internalPort = p
+				break
+			}
+		}
+	}
+
+	if internalPort == "" {
+		err = fmt.Errorf("no port to waiting")
```
mdelapenya, line 76:
```suggestion
		err = fmt.Errorf("no port to wait for")
```

wait/log_test.go
```diff
@@ -20,6 +20,9 @@ func (st noopStrategyTarget) Host(ctx context.Context) (string, error) {
 	return "", nil
 }
 
+func (st noopStrategyTarget) Ports(ctx context.Context) (nat.PortMap, error) {
+	return nil, nil
+}
```
mdelapenya, line 25: Lint
```suggestion
}

```

mdelapenya (commented):
I added two small nits. The PR looks good to me. My only concern is about the call to the `wait.ForListeningPort("")` method with an empty string. Not sure if adding a new method, something like `wait.ForExposedPort()` would me it more usable though, but I'd like to hear your thoughts on it.

It's not a blocker for this PR, which I will approve once we discuss about it and find ourselves in a closed path.

fiftin:
@mdelapenya ,

We have `HostPortStrategy` struct which has `Port` field. Method `wait.ForListeningPort` just instantiate this struct with provided port.

Yes, we can add method `wait.ForExposedPort` but it would an alias for `wait.ForListeningPort("")`.

mdelapenya:
> Yes, we can add method `wait.ForExposedPort()` but it would an alias for `wait.ForListeningPort("")`.

I want it to be simple to use. I we have to do the mental effort of passing an empty string to wait for a port declared at the Dockerfile, then I think it would make it difficult to remember/easy to forget about it, unless the docs/tests are crystal clear. I think we are missing it in the docs, as you already added the example in the tests 👍 

If you find it having the alias confusing, then it's OK to keep the method with the empty string, but well documented.

fiftin:
Added `wait.ForExposedPort`.

wait/host_port.go
```diff
@@ -16,6 +16,8 @@ import (
 var _ Strategy = (*HostPortStrategy)(nil)
 
 type HostPortStrategy struct {
+	// Port is a string containing port number and protocol in the format "80/tcp"
```
mdelapenya, line 19: 👍

wait/host_port.go
```diff
@@ -39,6 +41,12 @@ func ForListeningPort(port nat.Port) *HostPortStrategy {
 	return NewHostPortStrategy(port)
 }
 
+// ForExposedPort constructs an exposed port strategy. Alias for `NewHostPortStrategy("")`.
+// This strategy waits port exposed in Docker container.
```
mdelapenya, line 45: It's always the first port, right? https://github.com/testcontainers/testcontainers-go/pull/468/files#diff-ad6cdf7fcae968ad7307fe3e8717bcbc7e43c461424e76ca3953cd215ac9395bR77
```suggestion
// This strategy waits for the first port exposed in the Docker container.
```

mdelapenya (approved):
Last minute changes in docs, but other than that LGTM! Thanks @fiftin for your hard work here

Issue #398, opened 2022-01-19
Customise stdout and stderr for compose commands


byrnedo:
Hi, 

I'd like to be able to send all the stdout or stderr for a compose `.Invoke()` to an  `io.Writer` of my choosing. Is there a way to do that currently?

mdelapenya:
Hi @byrnedo, thanks for your question :) I'd say this is currently not possible (see https://github.com/testcontainers/testcontainers-go/blob/master/compose.go#L236-L237), although it could be considered a nice feature request if you explain your use case and possible implementations more in details (maybe here in this issue).

mdelapenya:
Would this implementation satisfy your needs? #420

edebernis:
Hey ! I found myself having the same need than @byrnedo.

Ideally, we should be able to pass options to the `LocalDockerCompose` struct to allow specifying custom pipes for command stdin, stdout and stderr.

Something like this:
```go
func WithStdinPipe(pipe io.Reader) DockerCompose
func WithStdoutPipe(pipe io.Writer) DockerCompose
func WithStderrPipe(pipe io.Writer) DockerCompose
```

Do you want me to implement a PR with this design ?

Thanks !

mdelapenya:
@edebernis we merged https://github.com/testcontainers/testcontainers-go/pull/420. Is that enough for your use case?

edebernis:
with #420, the command outputs are still piped to standards (stdout / stderr).

IMHO, it would be better to allow the user to override this behavior by specifing custom io.Writer / io.Reader to store command outputs.
And use stdout / stderr by default if nothing is specified.

mdelapenya:
> with #420, the command outputs are still piped to standards (stdout / stderr).
> 
> IMHO, it would be better to allow the user to override this behavior by specifing custom io.Writer / io.Reader to store command outputs. And use stdout / stderr by default if nothing is specified.

We will be more than happy to review a PR :)

mdelapenya:
Given this comment by @edebernis https://github.com/testcontainers/testcontainers-go/pull/470#issuecomment-1289608644, do you think we can close this issue?

mdelapenya:
Closing as stale, please feel free to reopen if you need anything else from here.

Thanks!

PR #470, opened 2022-06-20
Add options to redirect compose command stdout and stderr to custom writers

edebernis:
Close https://github.com/testcontainers/testcontainers-go/issues/398

Instead of copying all the command output in a separate buffer and return it when the command stops, this PR allows to stream command output to a specific `io.Writer`

### DIFF
--- a/compose.go
+++ b/compose.go
@@ -1,7 +1,6 @@
 package testcontainers
 
 import (
-	"bytes"
 	"context"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


mal-as (approved):

mdelapenya:
Hi @edebernis we have merged #476, which adds native support for docker compose. Would you mind revisiting this PR and update to that code? I can assist if needed, or we can even ping @baez90 for that.

Thanks in advance

edebernis:
Hey @mdelapenya ! Great news ! Actually, I don't think this MR makes sense anymore right now considering the native support of docker compose.

Closing.

PR #471, opened 2022-06-26
Remove unused import in redis example

doodot:
`_ "github.com/jackc/pgx/v4/stdlib"` seems like CockroachDB related import.


mdelapenya:
@doodot keeping docs up-to-date is super important, so thank you very much for this contribution!!

PR #474, opened 2022-06-30
chore: tell dependabot about the new module

mdelapenya:
## What does this PR do?
It adds the new e2e module to dependabot's configuration file

## Why is it important?
Because otherwise, the PRs from dependabot could fail in Go 1.18

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,7 +7,7 @@ updates:
   open-pull-requests-limit: 3
   rebase-strategy: disabled
 - package-ecosystem: gomod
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -6,3 +6,9 @@ updates:
     interval: daily
   open-pull-requests-limit: 3
   rebase-strategy: disabled
+- package-ecosystem: gomod
+  directory: "/e2e/"
```
mdelapenya, line 10: Remove extra slash
```suggestion
  directory: "/e2e"
```

PR #475, opened 2022-07-01
refactor: read RYUK_CONTAINER_PRIVILEGED once

zregvart:
Now the environment variable `TESTCONTAINERS_RYUK_CONTAINER_PRIVILEGED` is read only once in the `configureTC` function that reads the `.testcontainers.properties` file and accesses the environment variable.

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -663,7 +663,8 @@ func WithDefaultBridgeNetwork(bridgeNetworkName string) DockerProviderOption {
 }
 
 func NewDockerClient() (cli *client.Client, host string, tcConfig TestContainersConfig, err error) {
-	tcConfig = readTCPropsFile()
+	tcConfig = configureTC()
+
 	host = tcConfig.Host
 
 	opts := []client.Opt{client.FromEnv}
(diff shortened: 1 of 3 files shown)


docker.go
```diff
@@ -664,6 +664,10 @@ func WithDefaultBridgeNetwork(bridgeNetworkName string) DockerProviderOption {
 
 func NewDockerClient() (cli *client.Client, host string, tcConfig TestContainersConfig, err error) {
 	tcConfig = readTCPropsFile()
```
mdelapenya, line 666: Wdyt about renaming the `readTCPPropsFile` method to `configureTC` and perform all operations in there: read the Props File and the env vars?

docker.go
```diff
@@ -732,8 +733,8 @@ func NewDockerProvider(provOpts ...DockerProviderOption) (*DockerProvider, error
 	return p, nil
 }
 
-// readTCPropsFile reads from testcontainers properties file, if it exists
-func readTCPropsFile() TestContainersConfig {
+// configureTC reads from testcontainers properties file, if it exists
```
mdelapenya, line 736:
```suggestion
// configureTC reads from testcontainers properties file, if it exists
// it is possible that certain values get overridden when set as environment variables
```

docker.go
```diff
@@ -752,6 +754,11 @@ func readTCPropsFile() TestContainersConfig {
 		return TestContainersConfig{}
```
mdelapenya, line 754: I'm going to merge this one, although I wonder if the struct we are returning here should consider the env variables too 🤔

PR #479, opened 2022-07-05
refactor: more nuanced env vs properties config

zregvart:
The environment variables should trump configuration from `$HOME/.testcontainers.properties` file, but only if they're set and then only if they equal `true` or `false`. This refactors `configureTC` to make sure that environment variables are always considered and indeed trump any configuration in `$HOME/.testcontainers.properties` file, even as parsing the configuration from `$HOME/.testcontainers.properties` completes partially. Also the `HOME is not set` test cases now do not set the `$HOME` so that the `os.UserHomeDir` error path is exercised by the tests.

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -736,30 +736,35 @@ func NewDockerProvider(provOpts ...DockerProviderOption) (*DockerProvider, error
 // configureTC reads from testcontainers properties file, if it exists
 // it is possible that certain values get overridden when set as environment variables
 func configureTC() TestContainersConfig {
+	config := TestContainersConfig{}
+
+	applyEnvironmentConfiguration := func(config TestContainersConfig) TestContainersConfig {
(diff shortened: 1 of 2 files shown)


docker.go
```diff
@@ -736,30 +736,35 @@ func NewDockerProvider(provOpts ...DockerProviderOption) (*DockerProvider, error
 // configureTC reads from testcontainers properties file, if it exists
 // it is possible that certain values get overridden when set as environment variables
 func configureTC() TestContainersConfig {
+	config := TestContainersConfig{}
+
+	applyEnvironmentConfiguration := func(config TestContainersConfig) TestContainersConfig {
+		ryukPrivilegedEnv := os.Getenv("TESTCONTAINERS_RYUK_CONTAINER_PRIVILEGED")
+		if ryukPrivilegedEnv == "true" || ryukPrivilegedEnv == "false" {
```
mdelapenya, line 743: Wdyt if we simply check for true values, otherwise false?

mdelapenya (approved):

PR #483, opened 2022-07-09
fix: return bad error after parallel running

VladimirStepanov:
Hello) I found a small bug in `ParallelContainers` method. Fixed!


parallel.go
```diff
@@ -114,7 +114,7 @@ func ParallelContainers(ctx context.Context, reqs ParallelContainerRequest, opt
 
 	<-waitRes
 
-	if len(errors) == 0 {
+	if len(errors) != 0 {
```
mdelapenya, line 117: Good catch!! Thanks for the quick fix!!

PR #491, opened 2022-07-28
Return container IP when attached to a single named network

leszko:
In many cases the container is attached to a single, named network of the "bridge" mode. In such scenarios, `DockerContainer.ContainerIP()` should resolve this IP and return it.

For example, imaging the following container definition.
```go
    // ...
    Networks:     []string{"some-network"},
}
```

Before this change, the function `container.ContainerIP(ctx)` returns an empty string. After this change, it returns the correct container IP.

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -379,7 +379,18 @@ func (c *DockerContainer) ContainerIP(ctx context.Context) (string, error) {
 		return "", err
 	}
 
-	return inspect.NetworkSettings.IPAddress, nil
+	ip := inspect.NetworkSettings.IPAddress
+	if ip == "" {
+		// use IP from "Networks" if only single network defined
+		networks := inspect.NetworkSettings.Networks
+		if len(networks) == 1 {
+			for _, v := range networks {
+				ip = v.IPAddress
+			}
+		}
+	}
+
+	return ip, nil
 }
 
 // NetworkAliases gets the aliases of the container for the networks it is attached to.
--- a/docker_test.go
+++ b/docker_test.go
@@ -70,6 +70,7 @@ func TestContainerAttachedToNewNetwork(t *testing.T) {
 				networkName: aliases,
 			},
 		},
+		Started: true,
 	}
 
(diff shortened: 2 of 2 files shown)


docker.go
```diff
@@ -379,7 +379,18 @@ func (c *DockerContainer) ContainerIP(ctx context.Context) (string, error) {
 		return "", err
 	}
 
-	return inspect.NetworkSettings.IPAddress, nil
+	ip := inspect.NetworkSettings.IPAddress
+	if ip == "" {
+		// use IP from "Networks" if only single network defined
+		networks := inspect.NetworkSettings.Networks
+		if len(networks) == 1 {
```
mdelapenya, line 386: If len is 1, then I think we can avoid the for loop and directly get the index 0 of the array.

mdelapenya: At the same time, we could create the convention that it will always return the IP from the first network

mdelapenya (commented):
Thanks for your contribution, also updating the tests! At first look, LGTM, will give it a try on CI before a final review.

docker.go
leszko: The problem here is that `networks` is a map, not a slice. Then, I think you can't just take the index 0. Also, taking the always the first network is problematic, because it can easily become non-deterministic. Then, I'd just limit it to returning a value if `len(networks) == 1`.

Btw. I think in the future we could create a function to return all IP addresses, instead of just one.

mdelapenya: Ah ah thanks for the explanation, then it makes sense to do it as you suggest.

leszko:
@mdelapenya If the PR is ok, can we get it merged into `main`?

mdelapenya:
Merged, thanks for your contribution, much much appreciated!

leszko:
> Merged, thanks for your contribution, much much appreciated!

Thanks for merging :)

mdelapenya:
My pleasure!!

image

PR #503, opened 2022-08-24
docs: add docs regarding Podman usage

baez90:


mdelapenya (approved):
THIS!!

Thanks for contributing to the docs, which is super important.

I'm taking a note to myself for starting a design discussion on seeing the provider at the properties level.

Other than that, great job, as always

baez90:
Thanks a lot! I'll try to add some more docs in the upcoming weeks as I just noticed there's also nothing about container mounts and I think this would make sense as well :blush:

Issue #374, opened 2021-11-05
Docker Compose V2: wait strategy cannot be applied: service not found


artamonovkirill:
**Describe the bug**

While running with Docker Compose V2, compose tests with wait strategies like [TestDockerComposeWithWaitForService](https://github.com/testcontainers/testcontainers-go/blob/master/compose_test.go#L156-L179) from this repository:
```
=== RUN   TestDockerComposeWithWaitForService
Network a7b7aa95-7973-4113-b7d9-0b98c56e7557_default  Creating
Network a7b7aa95-7973-4113-b7d9-0b98c56e7557_default  Created
Container a7b7aa95-7973-4113-b7d9-0b98c56e7557-nginx-1  Creating
Container a7b7aa95-7973-4113-b7d9-0b98c56e7557-nginx-1  Created
Container a7b7aa95-7973-4113-b7d9-0b98c56e7557-nginx-1  Starting
Container a7b7aa95-7973-4113-b7d9-0b98c56e7557-nginx-1  Started
    compose_test.go:390: Failed when running []: one or more wait strategies could not be applied to the running containers: service with name nginx_1 not found in list of running containers
```

**To Reproduce**

Either:
* Run tests from this repo on Mac or Windows with [Docker Compose V2 enabled](https://docs.docker.com/compose/cli-command/#install-on-mac-and-windows)
* Run tests from this repo on Linux after [installing Docker Compose V2](https://docs.docker.com/compose/cli-command/#install-on-linux)
* Run tests on [this branch](https://github.com/artamonovkirill/testcontainers-go/tree/test/compose-v2)
    * Changed `docker-compose` to `docker compose`, see [the diff](https://github.com/testcontainers/testcontainers-go/compare/master...artamonovkirill:test/compose-v2?expand=1)
    * [Test results](https://github.com/artamonovkirill/testcontainers-go/runs/4251407429?check_suite_focus=true)

**Expected behavior**

Container startup is awaited.

**docker info**
```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Build with BuildKit (Docker Inc., v0.6.3)
  compose: Docker Compose (Docker Inc., v2.0.0)
  scan: Docker Scan (Docker Inc., v0.8.0)

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 10
 Server Version: 20.10.8
 Storage Driver: overlay2
  Backing Filesystem: extfs
[...]

PR #476, opened 2022-07-01
425 support native docker compose api

baez90:
# What does this PR do?

Deprecate shell-escape based `LocalDockerCompose` implementation and provide new `docker/compose` based one using the Docker API. Introduce a new API that takes `context.Context` into account. Furthermore it exposes more typed options.

#425

PR #514, opened 2022-09-01
fix(compose): wait.ForExit() strategy should work for finished containers

Malinskiy:
Fix for the https://github.com/testcontainers/testcontainers-go/issues/374#issuecomment-1233796935


mdelapenya:
Hi @Malinskiy, thanks for your contribution, I've rebased the branch with upstream and if the CI is still ✅ will merge it. And sorry for the late response 😢

mdelapenya (approved):
LGTM, thanks for this contribution!

I think you should be aware of #476, which refactors compose to be used natively in Go, instead of using a shelled version of compose. So at some point I'd suggest discussing with the original author about possible changes in that PR (I started it [here](https://github.com/testcontainers/testcontainers-go/pull/476/files#r970651927), but you probably find others)

Thanks again for your time here!

PR #517, opened 2022-09-05
feat: copy directory to container

### LINKED ISSUE #516: [Feature]: support copying directories and files in the CopyToContainer method
### Problem

It could be the case the test needs an entire directory to be copied into the container, even before the container is created. With current code, it only checks that the path is a file (we try to simply read it), failing with an error if that path represents a directory

### Solution

We could detect if the file is directory, and tar the entire directory, passing the buffer after it

### Benefit

Users mounting an entire dir will be able to simply copy the dir to the container

### Alternatives

With docker-compose you can simply mount an entire local path to the container, but in scenarios where compose is not the right choice, copying is more relevant

Yes

mdelapenya:
## What does this PR do?
It adds the `CopyDirToContainer` method to the container interface, implementing it for Docker.

This new method uses the standard library to `tar` the source directory and copy it to the container.

We have updated the existing `CopyFileToContainer` method to check if the host path passed is a directory, calling the new method if/when needed.

We have added unit tests for both the tar directory code and for the copy dir, updating the existing tests.

Finally the docs have been updated.

## Why is it important?
First, feature parity with the tc-java project, which supports copying directories to containers.

Second, we detected this feature was missing while contributing to a demo where it needed to upload an entire directory to a place it does not exist in the target container, because the underlying Docker image did not have it on its file system. While trying to find the workaround to it, we identified that this could be a quick win for tc-go, so instead of doing the following workarounds:

- creating a Dockerfile that, using the same Docker image, added the parent directory in a RUN instruction
- rearrange the demo project to upload just one file

it would be much more simple to just upload the directory to its parent.

## Examples
See the changes in the docs ar even the new tests to understand how to use it.

## Related issues
- Closes #516


docker.go
```diff
@@ -481,7 +481,40 @@ func (c *DockerContainer) CopyFileFromContainer(ctx context.Context, filePath st
 	return ret, nil
 }
 
+// CopyDirToContainer copies the contents of a directory to a parent path in the container. This parent path must exist in the container first
+// as we cannot create it
```
eddumelendez, line 485:
```suggestion
// as we can not create it
```

docs/features/copy_file.md
```diff
@@ -39,3 +39,75 @@ nginxC, err := GenericContainer(ctx, GenericContainerRequest{
 	})
 ```
 
+## Copy Directories To Container
+
+It's also possible to copy an entire directory to a container, and that can happen before and/or after the container gets into the "Running" state. As an example, you could need to bulk-copy a set of files, such as a configuration directory that does not exist in the underlying Docker image.
+
+It's important to notice that, when copying the directory to the container, the container path must exists in the Docker image. And this is strong requirement for files to be copied _before_ the container is started, as we cannot create the full path at that time.
+
+Once we understood that, there are two ways to do copy directories to a container. The first one is using the already existing `CopyFileToContainer` method, which will internally check if the host path is a directory, internally calling the new `CopyDirToContainer` method if needed:
```
eddumelendez, line 48:
```suggestion
Once we understood that, there are two ways to copy directories to a container. The first one is using the existing `CopyFileToContainer` method, which will internally check if the host path is a directory, internally calling the new `CopyDirToContainer` method if needed:
```

docs/features/copy_file.md
```diff
@@ -39,3 +39,75 @@ nginxC, err := GenericContainer(ctx, GenericContainerRequest{
 	})
 ```
 
+## Copy Directories To Container
+
+It's also possible to copy an entire directory to a container, and that can happen before and/or after the container gets into the "Running" state. As an example, you could need to bulk-copy a set of files, such as a configuration directory that does not exist in the underlying Docker image.
+
+It's important to notice that, when copying the directory to the container, the container path must exists in the Docker image. And this is strong requirement for files to be copied _before_ the container is started, as we cannot create the full path at that time.
```
eddumelendez, line 46:
```suggestion
It's important to notice that, when copying the directory to the container, the container path must exist in the Docker image. And this is a strong requirement for files to be copied _before_ the container is started, as we cannot create the full path at that time.
```

tar.go
```diff
@@ -0,0 +1,125 @@
...
+	"archive/tar"
+	"bytes"
+	"compress/gzip"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+)
+
+func isDir(path string) (bool, error) {
```
eddumelendez, line 13: should be moved to `file.go` or something like that? IMO, tar.go is not the place for this. WDYT?

docker.go
mdelapenya: https://www.grammarly.com/blog/cannot-or-can-not/#:~:text=Here%20is%20a%20quick%20summary,.%20.%20.%20but%20also.%E2%80%9D 😄

tar.go
mdelapenya: Indeed. Will rename it asap

docker_test.go
```diff
@@ -1855,6 +1856,33 @@ func TestDockerContainerCopyFileToContainer(t *testing.T) {
...
+			WaitingFor:   wait.ForListeningPort(nginxDefaultPort),
+		},
+		Started: true,
+	})
+
+	require.NoError(t, err)
+	terminateContainerOnEnd(t, ctx, nginxC)
+
+	err = nginxC.CopyDirToContainer(ctx, "./testresources/Dockerfile", "/tmp/testresources/Dockerfile", 700)
+	require.Error(t, err) // copying a file using the directory method will reaise an error
```
mdelapenya, line 1876:
```suggestion
	require.Error(t, err) // copying a file using the directory method will raise an error
```

agaudreault-jive:
@eddumelendez @kiview any plan to cut a release soon?

mdelapenya:
> @eddumelendez @kiview any plan to cut a release soon?

Yes, we are very close to it 🙏 , sorry for the inconvenience 😞

We are closing just some bits for the internal ceremonies to release it to the public, so it will eventually happen really soon

PR #536, opened 2022-09-26
docs: add a basic layout for wait strategies in docs

### LINKED ISSUE #314: Wait strategies: use same WithStartupTimeout methods
`sql.go` has a `Timeout` method, but the rest of the wait strategies have `WithStartupTimeout`. To avoid breaking changes, we could add the method and call it from the old one, including a deprecation message.

>Maybe the Strategy interface should force the method 🤔 

Besides that, it would be great if we add docs about the wait for strategies, same as in the Java project: https://www.testcontainers.org/features/startup_and_waits/

mdelapenya:
## What does this PR do?
It creates a basic layout for the wait strategies in the docs

For that:
- it defines a new nav item under features
- this menu item will be the home for all wait strategies, as a parent
- will add wait strategies specific docs under the new parent nav item
- fixes two consistency issues with methods that were not properly implemented:
  - `host_port.go`: it's possible now to set the poll internal, as in any other wait strategy
  - `sql.go`: the WithTimeout function now calls a `WithStartupTimeout` function, which is consistent with the rest of the wait strategies

> I wonder if I should remove the commits for the consistency and send a separate PR, which would be better for the release notes 🤔 

Regarding the website, it's important to notice that the `golang` blocks are automatically pygmented by the existing libraries, so no need to add mkdocs plugins for that.

## Why is it important?
~~It serves as the baseline for adding new docs for the wait strategies (will come in follow-up commits)~~

It documents the wait strategies, which was missing.

## How to test this PR?
Start the local documentation site using the instructions in the docs/contributing_docs.md file

```shell
pip install -r requirements.txt
mkdocs serve
```

Screenshot 2022-09-26 at 20 20 25

## Related issues
- Closes #314 
- Relates #322, which should be updated resolving conflicts

## Follow ups
Automate the local deployment of the docs, using Docker, instead of polluting the local system


mdelapenya:
Still in draft to validate the initial layout. Will add more commits documenting all the wait strategies, and once there, will mark it as ready for review

wait/host_port.go
```diff
@@ -52,6 +54,12 @@ func (hp *HostPortStrategy) WithStartupTimeout(startupTimeout time.Duration) *Ho
 	return hp
 }
 
+// WithPollInterval can be used to override the default polling interval of 100 milliseconds
+func (hp *HostPortStrategy) WithPollInterval(pollInterval time.Duration) *HostPortStrategy {
```
mdelapenya, line 58: Added to keep APIs in sync

wait/sql.go
```diff
@@ -33,8 +33,14 @@ type waitForSql struct {
...
 
 //Timeout sets the maximum waiting time for the strategy after which it'll give up and return an error
+// Deprecated: Use WithStartupTimeout
 func (w *waitForSql) Timeout(duration time.Duration) *waitForSql {
-	w.startupTimeout = duration
+	return w.WithStartupTimeout(duration)
+}
+
+// WithStartupTimeout can be used to change the default startup timeout
+func (w *waitForSql) WithStartupTimeout(startupTimeout time.Duration) *waitForSql {
```
mdelapenya, line 42: Added to keep APIs in sync

docs/features/wait/exec.md
```diff
@@ -0,0 +1,19 @@
+# Exec Wait strategy
+
+The exec wait strategy will check the exit code of a process to be executed in the container, being able to set the following conditions:
+
+- the command and arguments to be executed, as an array of strings.
+- the exit code as a function to resolve a matcher, being the default one `0`.
```
kiview, line 6:
```suggestion
- a function to match a specific exit code, with the default matching `0`.
```

docs/features/wait/exec.md
```diff
@@ -0,0 +1,19 @@
...
+- the startup timeout to be used, default is 60 seconds.
```
kiview, line 7:
```suggestion
- the startup timeout to be used in seconds, default is 60 seconds.
```

docs/features/wait/exec.md
```diff
@@ -0,0 +1,19 @@
...
+- the poll interval to be used, default is 100 milliseconds.
```
kiview, line 8:
```suggestion
- the poll interval to be used in milliseconds, default is 100 milliseconds.
```

docs/features/wait/exec.md
```diff
@@ -0,0 +1,19 @@
+# Exec Wait strategy
+
+The exec wait strategy will check the exit code of a process to be executed in the container, being able to set the following conditions:
```
kiview, line 3: I am a little bit picky here, but I wonder how to consistently talk about _wait strategies_ in the docs. In Java, we are in a comfortable position, since the class name itself is fully descriptive within the domain language (e.g., `HttpWaitStrategy`, `LogMessageWaitStrategy`, etc.). So it follow the pattern of `XWaitStrategy`. 

In Go we generally have `XStrategy` (`ExecStrategy`, `HealthStrategy`) as types, but some are also off from the convention (`waitForSql`, which is also private?).

I know it would be a hard breaking change to align Go with Java here, but maybe we can eventually do it?

docs/features/wait/exit.md
```diff
@@ -0,0 +1,15 @@
+# Exit Wait strategy
+
+The exit wait strategy will check the container is not in the running state, being able to set the following conditions:
```
kiview, line 3:
```suggestion
# Exit Wait strategy

The exit wait strategy will check that the container is not in the running state, and allows to set the following conditions:
```

docs/features/wait/exit.md
```diff
@@ -0,0 +1,15 @@
...
+
+- the exit timeout, default is `0`.
```
kiview, line 5:
```suggestion
- the exit timeout in seconds, default is `0`.
```
?

docs/features/wait/exit.md
```diff
@@ -0,0 +1,15 @@
...
+- the poll interval to be used, default is 100 milliseconds.
```
kiview, line 6:
```suggestion
- the poll interval to be used in milliseconds, default is 100 milliseconds.
```

docs/features/wait/health.md
```diff
@@ -0,0 +1,13 @@
+# Health Wait strategy
+
+The health wait strategy will check the container is in the healthy state, being able to set the following conditions:
```
kiview, line 3:
```suggestion
The health wait strategy will check that the container is in the healthy state and allows to set the following conditions:
```

docs/features/wait/health.md
```diff
@@ -0,0 +1,13 @@
...
+
+- the startupTimeout to be used, default is 60 seconds.
```
kiview, line 5:
```suggestion
- the startup timeout to be used in seconds, default is 60 seconds.
```
Since `exit timeout` hat a space, I thought this is more consistent.

docs/features/wait/health.md
```diff
@@ -0,0 +1,13 @@
...
+- the PollInterval to be used, default is 100 milliseconds.
```
kiview, line 6:
```suggestion
- the poll interval to be used in milliseconds, default is 100 milliseconds.
```

docs/features/wait/host_port.md
```diff
@@ -0,0 +1,31 @@
+# HostPort wait strategy
```
kiview, line 1:
```suggestion
# HostPort Wait strategy
```
Not saying we should write it like this, but it aligns it with the other strategies.

docs/features/wait/host_port.md
```diff
@@ -0,0 +1,31 @@
...
+
+The host-port wait strategy will check if a port is listening in the container, being able to set the following conditions:
+
+- a port to be listening in the container. The port and protocol to be used, which is represented by a string containing port number and protocol in the format "80/tcp".
+- the first exposed port in the container.
```
kiview, line 6: This means instead of specifying a port, just wait for the first exposed port, correct? Needs a bit more clarification.

The example below is actually useful, I just found this bullet point confusing.

docs/features/wait/host_port.md
```diff
@@ -0,0 +1,31 @@
+# HostPort wait strategy
+
+The host-port wait strategy will check if a port is listening in the container, being able to set the following conditions:
+
+- a port to be listening in the container. The port and protocol to be used, which is represented by a string containing port number and protocol in the format "80/tcp".
```
kiview, line 5:
```suggestion
- a port exposed by the container. The port and protocol to be used, which is represented by a string containing the port number and protocol in the format "80/tcp".
```

docs/features/wait/http.md
```diff
@@ -0,0 +1,61 @@
+# HTTP(S) Wait strategy
+
+The HTTP wait strategy will check the result of an HTTP(S) request that is available in the container, being able to set the following conditions:
```
kiview, line 3:
```suggestion
The HTTP wait strategy will check the result of an HTTP(S) request against the container and allows to set the following conditions:
```

docs/features/wait/http.md
```diff
@@ -0,0 +1,61 @@
...
+
+- the port to be used.
+- the path to be used.
+- the HTTP method to be used.
+- the HTTP request body to be sent.
+- the HTTP status code as a function to resolve a matcher.
+- the HTTP response as a function to resolve a matcher.
+- the TLS config to be used for HTTPS.
+- the startup timeout to be used, default is 60 seconds.
+- the poll interval to be used, default is 100 milliseconds.
```
kiview, line 13:
```suggestion
- the startup timeout to be used in seconds, default is 60 seconds.
- the poll interval to be used in milliseconds, default is 100 milliseconds.
```

docs/features/wait/http.md
```diff
@@ -0,0 +1,61 @@
+# HTTP(S) Wait strategy
+
+The HTTP wait strategy will check the result of an HTTP(S) request that is available in the container, being able to set the following conditions:
+
+- the port to be used.
+- the path to be used.
+- the HTTP method to be used.
+- the HTTP request body to be sent.
+- the HTTP status code as a function to resolve a matcher.
+- the HTTP response as a function to resolve a matcher.
```
kiview, line 10:
```suggestion
- the HTTP status code matcher as a function.
- the HTTP response matcher as a function.
```

docs/features/wait/exec.md
```diff
@@ -0,0 +1,19 @@
+# Exec Wait strategy
```
kiview, line 1:
```suggestion
# Exec Wait Strategy
```
After looking through the whole PR, I think we should capitalize `Strategy` in every heading, WDYT?

docs/features/wait/introduction.md
```diff
@@ -0,0 +1,22 @@
+# Wait Strategies
+
+!!! info "Wait strategies vs Startup strategies"
+
+    **Wait strategy:** is the container in a state that is useful for testing. This is generally approximated as 'can we talk to this container over the network'. However, there are quite a few variations and nuances.
```
kiview, line 5:
```suggestion
    **Wait strategy:** Is the container in a state that is useful for testing? This is generally approximated as 'Can we talk to this container over the network?'. However, in practice, there are quite a few variations and nuances.
```

docs/features/wait/introduction.md
```diff
@@ -0,0 +1,22 @@
...
+    
+    **Startup strategy:** did a container reach the desired running state. *Almost always* this just means 'wait until the container is running' - for a daemon process in a container this is the goal. Sometimes we need to wait until the container reaches a running state and then exits - this is the 'one shot startup' strategy, only used for cases where we need to run a one off command in a container but not a daemon.
```
kiview, line 7:
```suggestion
    **Startup strategy:** Did a container reach the desired running state? *Almost always* this just means 'wait until the container is running' - for a daemon process in a container this is the goal. Sometimes we need to wait until the container reaches a running state and then exits - this is the 'one shot startup' strategy, only used for cases where we need to run a one off command in a container but not a daemon.
```

Do we have `one shot startup` strategy in `tc-go`?

docs/features/wait/log.md
```diff
@@ -0,0 +1,20 @@
+# Log Wait strategy
+
+The Log wait strategy will check if a string is present in the container logs with a desired number of ocurrences, being able to set the following conditions:
```
kiview, line 3:
```suggestion
The Log wait strategy will check if a string occurs in the container logs for a desired number of times, and allows to set the following conditions:
```

docs/features/wait/log.md
```diff
@@ -0,0 +1,20 @@
...
+
+- the string to be searched in the container log.
```
kiview, line 5:
```suggestion
- the string to be waited for in the container log.
```

docs/features/wait/log.md
```diff
@@ -0,0 +1,20 @@
...
+- the number of ocurrences of the searched string, default is `1`.
```
kiview, line 6:
```suggestion
- the number of occurrences of the string to wait for, default is `1`.
```

docs/features/wait/log.md
```diff
@@ -0,0 +1,20 @@
...
+- the startup timeout to be used, default is 60 seconds.
+- the poll interval to be used, default is 100 milliseconds.
```
kiview, line 8:
```suggestion
- the startup timeout to be used in seconds, default is 60 seconds.
- the poll interval to be used in milliseconds, default is 100 milliseconds.
```

docs/features/wait/multi.md
```diff
@@ -0,0 +1,20 @@
+# Multi Wait strategy
```
kiview, line 1: Called `WaitAllStrategy` in Java.

docs/features/wait/multi.md
```diff
@@ -0,0 +1,20 @@
...
+
+The Multi wait strategy will hold a list of wait strategies, in order to wait for all of them. It's possible to set the following conditions:
+
+- the startup timeout to be used, default is 60 seconds.
```
kiview, line 5:
```suggestion
- the startup timeout to be used in seconds, default is 60 seconds.
```

docs/features/wait/sql.md
```diff
@@ -0,0 +1,22 @@
+# SQL Wait strategy
+
+The SQL wait strategy will check the result of an HTTP(S) request that is available in the container, being able to set the following conditions:
```
kiview, line 3: Of an HTTP request? I assume of a SQL query?

docs/features/wait/sql.md
```diff
@@ -0,0 +1,22 @@
...
+
+- the SQL query to be used, default is `SELECT 1`.
+- the port to be used.
+- the database driver to be used, as a string.
+- the URL of the database to be used, as a function returning the URL string.
+- the startup timeout to be used, default is 60 seconds.
+- the poll interval to be used, default is 100 milliseconds.
```
kiview, line 10:
```suggestion
- the startup timeout to be used in seconds, default is 60 seconds.
- the poll interval to be used in milliseconds, default is 100 milliseconds.
```

kiview (commented):
Apply suggestions as you see fit. Topics that require a deeper discussion can be discussed outside of this PR.

docs/features/wait/exec.md
mdelapenya: In Go the package provides that context: `wait.ForLog`, `wait.ForSQL` etc, so we are covered there.

I totally agree that the existing code conventions are not consistent (public Vs private, reused fields such as startupTimeout and pollInterval being spread instead in of a separated struct or interface) and I did not change that intentionally. My goal is to face that in separate efforts

docs/features/wait/host_port.md
mdelapenya: Indeed. It would be one or the other, not both. Let me think how to rephrase that.

docs/features/wait/host_port.md
```diff
@@ -0,0 +1,31 @@
...
+req := ContainerRequest{
+    Image:        "docker.io/nginx:alpine",
+    ExposedPorts: []string{"80/tcp"},
+    WaitingFor:   wait.ForListeningPort("80/tcp"),
+}
+```
+
+## First exposed port in the container
+
+The wait strategy will use the first exposed port from the image configuration.
```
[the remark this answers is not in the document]
mdelapenya: The first exposed port at the Dockerfile: the code gets the image config directly from Docker, and retrieves the first port from there.

docs/features/wait/introduction.md
mdelapenya: This has been copied from Java docs :)

docs/features/wait/multi.md
mdelapenya: Indeed. The function call is `wait.ForAll`, which is aligned at the API level. But saying `All` in the docs is less clear imho that multi. But I'm open to rename it

docs/features/wait/sql.md
mdelapenya: Wrong copy&paste? 🤦  Changing it...

mdelapenya: Done

docs/features/wait/multi.md
kiview: I think both are fine:
* We wait for _all_ wait strategies
* We wait for _multiple_ wait strategies

In order to align with Java, I'd be happy if we consider the rename here as well.

docs/features/wait/host_port.md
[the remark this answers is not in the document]
kiview: So if I expose a port in the `ContainerRequest`, this would not be considered?

docs/features/wait/host_port.md
kiview: maybe something like:
`- alternatively, to wait for the first exposed port of the container`

Also, container vs image here 😉

kiview (approved):

docs/features/wait/host_port.md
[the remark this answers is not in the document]
mdelapenya: @kiview I've double checked the code and when the `wait.ForExposedPort` function is used, then it will use the first port from raw representation of the container:

container.NetworkSettings.Ports -> array of nat.Port

If the array is empty, then an error is thrown, which makes me consider a possible unexpected behavior in the case somebody used this method AND there is an exposed port.

Because it's clear in the code (https://github.com/testcontainers/testcontainers-go/blob/6ee7f6a952f3e377de0d70b3b8bc81b3abe5a53d/wait/host_port.go#L68-L86), I think I'm going to explain more in depth this behavior in the docs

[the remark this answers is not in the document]
mdelapenya: Oh, yeah I realized it's "container" definition, not "image" definition. So if the user exposed a port in the container request struct, but used the `ForExposedPort` func, then the container definition will include it, therefore the `Ports` call will retrieve it. Thanks for pointing that out.

Nevertheless, let me explain the behavior in the docs

docs/features/wait/introduction.md
mdelapenya: You are right. Wrong copy&paste from Java docs. I've removed the one-shot strategy to use the `wait.ForExit`, which waits for the container to exit. What this strategy does is simply to verify if the container is not in the `Running` state, which we could improve to check the exit status of the container, like in Java

kiview: A `StartupCheckStrategy` is an additional concept in `testcontainers-java`, that is verified before the configured `WaitStrategy` is checked. So if we don't yet have this additional layer in `tc-go`, I'd recommend removing it from the docs.

docs/features/wait/multi.md
mdelapenya: As discussed, we are going to rename the `multi.go` file to `all.go`:
- consistency with Java
- the exposed API is not changed (wait.ForAll)

PR #547, opened 2022-10-04
docs: add docs regarding Colima usage

hhsnopek:
## Context

This pull request adds documentation around the utilization of Colima with Docker. It includes two ways of integrating with Colima, one of which is suggested by Colima and the other is the workaround defined from a Slack Thread of debugging which also serves as a pre-v0.4.0 Colima solution.

### Preview

screencapture-deploy-preview-547-testcontainers-go-netlify-app-features-using-colima-2022-10-04-17_26_24

### DIFF
--- a/docs/features/using_colima.md
+++ b/docs/features/using_colima.md
@@ -1,0 +1,39 @@
+# Using Colima with Docker
+
+[Colima](https://github.com/abiosoft/colima) is a container runtime which
+integrates with Docker's tooling and can be configured in various ways.
+
+As of Colima v0.4.0 it's recommended to set the active Docker context to use
+Colima. After the context is set testcontainers-go will automatically be
+configured to use Colima.
+
+```bash
+$ docker context ls
+NAME        DESCRIPTION                               DOCKER ENDPOINT                                      KUBERNETES ENDPOINT      ORCHESTRATOR
+colima      colima                                    unix:///Users/foobar/.colima/default/docker.sock
+default *   Current DOCKER_HOST based configuration   unix:///Users/foobar/.colima/docker.sock
+
+$ docker context use colima
+colima
+Current context is now "colima"
+
+$ docker context ls
+NAME       DESCRIPTION                               DOCKER ENDPOINT                                      KUBERNETES ENDPOINT       ORCHESTRATOR
+colima *   colima                                    unix:///Users/foobar/.colima/default/docker.sock
+default    Current DOCKER_HOST based configuration   unix:///var/run/docker.sock
+```
+
+If you're using an older version of Colima or have other applications that are
+unaware of Docker context the following workaround is available:
+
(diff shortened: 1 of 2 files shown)


docs/features/using_colima.md
```diff
@@ -0,0 +1,39 @@
...
+
+$ docker context use colima
+colima
+Current context is now "colima"
+
+$ docker context ls
+NAME       DESCRIPTION                               DOCKER ENDPOINT                                      KUBERNETES ENDPOINT       ORCHESTRATOR
+colima *   colima                                    unix:///Users/foobar/.colima/default/docker.sock
+default    Current DOCKER_HOST based configuration   unix:///var/run/docker.sock
+```
```
eddumelendez, line 24: question, is it required to switch context? setting env var should be enough, right?

hhsnopek: The suggested approach by Colima is to switch context when using Colima v0.4.0+. For Colima v0.3.0 setting both `DOCKER_HOST` and `TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE` are required. This approach of setting the environment variables also works for Colima v0.4.0 but isn't suggested unless you have other applications that do not use Docker context.

I'm open to removing the use of Docker context if you'd like, however, this configures Colima for testcontainers-go without extra configuration. The Docker context solution is left out of the [testcontainers-java docs for Colima](https://www.testcontainers.org/supported_docker_environment/#using-colima). If accepted here I can also open a pull request in testcontainers-java including the Docker context solution.

mdelapenya (commented):
This PR LGTM, thanks @hhsnopek for your contribution!

Once merged, I'd work in placing both Podman and Colima docs under a System Requirements parent page, like in the Java site: https://www.testcontainers.org/supported_docker_environment/#using-colima

docs/features/using_colima.md
kiview: @hhsnopek Could you verify if setting the context will work with `testcontainers-java` directly? To my knowledge, it won't, since this upstream issue in `docker-java` is still open: https://github.com/docker-java/docker-java/issues/1946

eddumelendez: Colima docs are based on `docker cli` context. From testcontainers perspective setting env vars should be enough. At least, when test it in `testcontainers-java` that's what I did.

eddumelendez: makes sense now. thanks for clarifying @hhsnopek !

mdelapenya (approved):
Thanks for your contribution. It's great!

docs/features/using_colima.md
kiview: @hhsnopek No need to check Java if you are not used to it. I am very certain it does not work, as outlined above. 

Having context and ENV variant in the Golang docs makes a lot of sense 👍

Issue #546, opened 2022-10-04
[Enhancement]:  Log docker server info 


eddumelendez:
### Proposal

It should be helpful to log something like lines below when running tests using testcontainers-go

```
16:25:17.129 INFO  org.testcontainers.DockerClientFactory - Connected to docker: 
  Server Version: 70+testcontainerscloud
  API Version: 1.41
  Operating System: Ubuntu 22.04 LTS
  Total Memory: 7396 MB
```

gauravgahlot:
@mdelapenya I would like to work on it.

PR #548, opened 2022-10-05
Log docker server info

gauravgahlot:
Enables - #546 

```sh
github.com/testcontainers/testcontainers-go - Connected to docker: 
  Server Version: 20.10.17
  API Version: 1.41
  Operating System: darwin
  Total Memory: 32768 MB
```


docker.go
```diff
@@ -32,6 +34,7 @@ import (
 	"github.com/magiconair/properties"
 	"github.com/moby/term"
 	specs "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/shirou/gopsutil/v3/mem"
```
gauravgahlot, line 37: I couldn't find a more reliable way of getting the memory. However, I'm unsure if we really need to log the memory. Also, the package brings with it a few more dependencies which I would prefer to avoid.

docker.go
```diff
@@ -770,9 +773,26 @@ func NewDockerProvider(provOpts ...DockerProviderOption) (*DockerProvider, error
...
+	type empty struct{}
+	info := `%v - Connected to docker: 
+  Server Version: %v
+  API Version: %v
+  Operating System: %v
+  Total Memory: %v MB
+`
+
+	sv, _ := p.client.ServerVersion(context.Background())
+	v, _ := mem.VirtualMemory()
```
kiview, line 791: Won't this give you the memory of the process executing the tests, rather than the memory of the Docker host?

kiview (commented):
Can we add a test?

docker.go
kiview: See my inline comment. We need to know the memory of the Docker host, not of the test process host. A good idea would be to check, how we do it in `testcontainers-java`.

docker.go
```diff
@@ -770,9 +773,26 @@ func NewDockerProvider(provOpts ...DockerProviderOption) (*DockerProvider, error
...
+	p.Logger.Printf(info, reflect.TypeOf(empty{}).PkgPath(),
+		sv.Version, sv.APIVersion, runtime.GOOS, v.Total/1024/1024)
```
kiview, line 793: Same as with the memory, `GOOS` won't give you the OS of the Docker host.

mdelapenya:
@gauravgahlot I agree with @kiview: this code will print the information but for the current host, and it could be the case that Docker is connected to a remote daemon.

I'd go with the Docker client libraries, i.e. when we retrieve the Docker client:

```golang
	type empty struct{}
	infoMessage := `%v - Connected to docker: 
  Server Version: %v
  API Version: %v
  Operating System: %v
  Total Memory: %v MB
`
	info, err := p.client.Info(context.Background())
	if err != nil {
		return nil, err
	}

	o.Logger.Printf(infoMessage, reflect.TypeOf(empty{}).PkgPath(),
		info.ServerVersion, c.ClientVersion(), info.OperatingSystem, info.MemTotal/1024/1024)
```

gauravgahlot:
Thank you @kiview @mdelapenya. I almost forgot we can use a remote host, maybe coz we didn't. 😄 
I have updated the log message in description as per the new code.

gauravgahlot:
@mdelapenya Could you add me to the assignees please?

docker.go
```diff
@@ -770,6 +771,26 @@ func NewDockerProvider(provOpts ...DockerProviderOption) (*DockerProvider, error
...
+  Operating System: %v
+  Total Memory: %v MB
+`
+
+		info, err := p.client.Info(context.Background())
+		if err != nil {
+			return nil, err
+		}
+
+		p.Logger.Printf(infoMessage, reflect.TypeOf(empty{}).PkgPath(),
```
mdelapenya, line 789: Super NIT: do you think we need the reflect package her, or just hardcode the "github.com/testcontainers/testcontainers-go" string?

mdelapenya:
> @mdelapenya Could you add me to the assignees please?

Hi @gauravgahlot I'd like to understand the reasons to be in the assignees section. I'm used to using it to define who in the project team is actively working on the issue, not only reviewing but driving the conversation with the original author. That's why I'd like to know your personal preference

docker.go
gauravgahlot: Well, I would prefer a hardcode for this small thing.

gauravgahlot:
> > @mdelapenya Could you add me to the assignees please?
> 
> Hi @gauravgahlot I'd like to understand the reasons to be in the assignees section. I'm used to using it to define who in the project team is actively working on the issue, not only reviewing but driving the conversation with the original author. That's why I'd like to know your personal preference

Oh, okay. I like to maintain a list of contributions that I did over the year and use "assigned" as a filter. That's the only reason. There are other ways to get my contribution list. So, no worries.

docker.go
mdelapenya: Let's use the string here then. I do not thing the package name changes that much 😅

gauravgahlot: Done

mdelapenya (approved):
LGTM thanks for your help!!!

docker.go
```diff
@@ -770,6 +771,25 @@ func NewDockerProvider(provOpts ...DockerProviderOption) (*DockerProvider, error
 		config:                tcConfig,
 	}
 
+	// log docker server info
```
mdelapenya, line 774: With current implementation of the library, this code will be executed with every container.

Could you make sure it's run just once? We are planning a big refactor in how the library is bootstrapped, see #549

gauravgahlot: The first thing that comes to my mind is a simple bool check. But before I make that change, I wanted to ask you if you have any other suggestions/thoughts that inlines with #549.

mdelapenya: I have this at the tip of my head: you could enclose that block in a function, and call it using a sync.Once pattern. wdyt?

gauravgahlot: Sounds better. Will do.

mdelapenya (approved):

PR #552, opened 2022-10-05
Remove obsolete deps/code

baez90:
# What does this PR do?

- remove `gotest.tools/v3` as now everything can be covered with standard library
- replace `io/ioutil` usages with `os` and `io` equivalents


baez90:
😄 it's mostly my OCD that dictates me cleaning this up and I longed for that actually quite some time 😄

Also the first rather small PR in a while 🙈

PR #553, opened 2022-10-06
Support returning all IP addresses of a container

### LINKED ISSUE #513: Support returning all IP addresses of a container instead of just one
_Originally posted by @leszko in https://github.com/testcontainers/testcontainers-go/pull/491#discussion_r932962603_

gauravgahlot:
Fixes #513


docker.go
```diff
@@ -393,6 +393,23 @@ func (c *DockerContainer) ContainerIP(ctx context.Context) (string, error) {
 	return ip, nil
 }
 
+// ContainerIPList gets the IP addresses of all the networks within the container.
+func (c *DockerContainer) ContainerIPList(ctx context.Context) ([]string, error) {
```
mdelapenya, line 397:
```suggestion
// ContainerIPs gets the IP addresses of all the networks within the container.
func (c *DockerContainer) ContainerIPs(ctx context.Context) ([]string, error) {
```

docker.go
```diff
@@ -393,6 +393,23 @@ func (c *DockerContainer) ContainerIP(ctx context.Context) (string, error) {
...
+	ipList := make([]string, 0)
```
mdelapenya, line 398:
```suggestion
	ips := make([]string, 0)
```

docker.go
```diff
@@ -393,6 +393,23 @@ func (c *DockerContainer) ContainerIP(ctx context.Context) (string, error) {
...
+
+	inspect, err := c.inspectContainer(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	networks := inspect.NetworkSettings.Networks
+	for _, endpoints := range networks {
+		ipList = append(ipList, endpoints.IPAddress)
```
mdelapenya, line 407:
```suggestion
		ips = append(ips, endpoints.IPAddress)
```

docker.go
```diff
@@ -393,6 +393,23 @@ func (c *DockerContainer) ContainerIP(ctx context.Context) (string, error) {
...
+	}
+
+	return ipList, nil
```
mdelapenya, line 410:
```suggestion
	return ips, nil
```

mdelapenya (commented):
I've just left a few suggestions on naming. You know, naming thinks...

Thanks for your contribution.

gauravgahlot:
Brings it back to how I originally wrote it. Cool. 👍

mdelapenya:
BTW, could you add tests for the new method? 🙏

gauravgahlot:
> BTW, could you add tests for the new method? 🙏

Should have added one implicitly. Done.

gauravgahlot:
As a side note, I noticed certain lint issues in the code (also verified with golangci-lint). I was thinking may be we should have a `make lint` target for future. If that works for you, please create an issue and assign it to me.

mdelapenya:
> As a side note @mdelapenya, I noticed certain lint issues in the code (also verified with golangci-lint). I was thinking may be we should have a `make lint` target for future. If that works for you, please create an issue and assign it to me.

We'd like to evaluate if [pre-commit](https://pre-commit.com/) works for us, therefore a check would happen at that level.

docker.go
```diff
@@ -393,6 +393,23 @@ func (c *DockerContainer) ContainerIP(ctx context.Context) (string, error) {
...
+	ips := make([]string, 0)
+
+	inspect, err := c.inspectContainer(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	networks := inspect.NetworkSettings.Networks
+	for _, endpoints := range networks {
+		ips = append(ips, endpoints.IPAddress)
```
mdelapenya, line 407: Super NIT.

Because the `network` name is already used by an imported package, let's rename the loop variable:
```suggestion
	for _, nw := range networks {
		ips = append(ips, nw.IPAddress)
```

gauravgahlot: Done

gauravgahlot:
@mdelapenya Is anything required from my end for this PR to be merged?

mdelapenya:
Thanks for your contribution @gauravgahlot !

PR #556, opened 2022-10-06
chore: use hashed versions of test-summary action

mdelapenya:
## What does this PR do?
It uses the commit of the v1 tag in the [`test-summary` action](https://github.com/test-summary/action), pinning it to the `v1` version (see https://github.com/test-summary/action/releases/tag/v1.0)

I also noticed that junit test reports were not generated (added here d0116e469ed1c38a59c38985001a65008cddbe3b but removed during a merge, probably here 897803f9b98e626a5e9953c571f002b5e00859dc). Therefore I'm adding the `--junitfile` flag to `gotestsum`, ignoring the test reports from git, and finally running `go mod tidy` in the e2e module to sync the module dependencies.

## Why is it important?
Security concerns about using dynamic version of GH actions, and recover the summary action


thomaseizinger:
Can we please limit review requests to the language-specific team? I get notified by every request for @testcontainers/oss-team but I only maintain the Rust impl.

mdelapenya:
> Can we please limit review requests to the language-specific team? I get notified by every request for @testcontainers/oss-team but I only maintain the Rust impl.

Makes sense, and sorry for noise. I was not aware that the oss-team included other projects apart from java, go and .NET. Will change patterns ASAP

PR #558, opened 2022-10-06
fix: use regex to find container by name

### LINKED ISSUE #557: [Bug]: Wrong container selected when reuse option enabled
### Testcontainers version

0.14.0

Yes

### Host OS

MAC OS

### Host Arch

ARM

### Go Version

1.18.5

### Docker version

```shell
Client:
 Cloud integration: v1.0.29
 Version:           20.10.17
 API version:       1.41
 Go version:        go1.17.11
 Git commit:        100c701
 Context:           default
 Experimental:      true

 Engine:
  Version:          20.10.17
  Go version:       go1.17.11
  Git commit:       a89b842
  Built:            Mon Jun  6 23:01:01 2022
  OS/Arch:          linux/arm64
  Experimental:     false
 containerd:
  Version:          1.6.8
 runc:
  Version:          1.1.4
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### Docker info

```shell
Client:
 Context:    default
 Debug Mode: false
 Plugins:

Server:
 Containers: 20
  Running: 18
  Paused: 0
  Stopped: 2
 Images: 16
 Server Version: 20.10.17
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 init version: de40ad0
 Security Options:
  seccomp
   Profile: default
  cgroupns
 OSType: linux
 Architecture: aarch64
 CPUs: 4
 Total Memory: 11.68GiB
 Name: docker-desktop
 ID: 74MG:MOIL:KAP7:QXLG:BLFM:QE6G:LHUE:55HE:PIKZ:55KU:KLOH:MMA7
 Debug Mode: false
 Labels:
 Experimental: false
 Insecure Registries:
  hubproxy.docker.internal:5000
  127.0.0.0/8
 Live Restore Enabled: false
```

### What happened?

# Background
I was running a test with the `reuse` option enabled.
Multiple containers are used at the same time during the test.

# Issue
The problem happened when I use multiple containers with similar names. I used the existing container named `test`, and `test1`.
I was expecting to get the information of the `test` container, but the `test1` container info was fetched all the time.

# Problem
The filter that used in `findContainerByName` will not guarantee exact match of the container. 
We should check the name of the containers instead of returning the first element of the containerList.

### Relevant log output

_No response_

### Additional Information

_No response_

hwwwi:
# Issue
fix #557: [Bug]: Wrong container selected when reuse option enabled

# Description
`client.ContainerList` will return all the container which contains the given name in the filter.
~~We should check all the names of the container from the list to find the right container.~~
(Update) The name filter uses the regex pattern to find the container, the order of the container is not guaranteed.


mdelapenya:
@hwwwi I'm surprised that the Docker client is not retrieving the right one, but to be fair not sure if we checked the filter in the case multiple containers are run at the same time with the same prefix name.

Thanks anyway for opening this PR, I think it must be addresses. Could you please add a test demonstrating the bug? 🙏

hwwwi:
> @hwwwi I'm surprised that the Docker client is not retrieving the right one, but to be fair not sure if we checked the filter in the case multiple containers are run at the same time with the same prefix name.
> 
> Thanks anyway for opening this PR, I think it must be addresses. Could you please add a test demonstrating the bug? 🙏

@mdelapenya Thanks for the reply :)
After some research on the name filter of the Docker client, I notice that the filter uses the regular expression.
The result of the `ContainerList` will return all the containers containing the given name. It's same as `docker container ls --filter name=someName`.
Instead of checking all the names of the containers, I add a caret and dollar sign on the name for the exact match of the search.
I also add a test to demonstrate the bug.

kiview:
The fix makes sense from a technical perspective. However, regarding the `reuse` behavior itself, we decided in `testcontainers-java` that relying on the container name is not a stable indicator of equality(with the name itself potentially leading to conflicts) and we, therefore, use the hash of the container create command as the lookup key.

mdelapenya:
> The fix makes sense from a technical perspective. However, regarding the `reuse` behavior itself, we decided in `testcontainers-java` that relying on the container name is not a stable indicator of equality(with the name itself potentially leading to conflicts) and we, therefore, use the hash of the container create command as the lookup key.

@kiview what do you think if we proceed with the merge, and create a discussion/issue to refactor how the reuse container feature works: if I understood correctly, a container should be reused if and only if the _container request_ (Go abstraction) is exactly the same, identified by a unique hash of the struct representing the request.

kiview:
@mdelapenya 

> a container should be reused if and only if the container request (Go abstraction) is exactly the same, identified by a unique hash of the struct representing the request.

Yep, semantically that's it. Indeed, we can align with the `testcontainers-java` implementation here as well. However, you are right that merging the PR does not make the situation worse and, on the contrary, improves on an actual bug. So merging and starting work on a better implementation sounds like a sensible approach to me.

PR #560, opened 2022-10-07
fix container NetworkMode usage

clive-jevons:
docker cannot do container network mode with exposed ports.
in order to make it work, only calculate the default exposed
ports if the NetworkMode is NOT container.


mdelapenya:
This PR is related to #468, as it introduced a change in which all ports will be exposed if no exposed ports were defined at the container request

mdelapenya (approved):
Thanks for identifying this bug and having the time to send a fix

LGTM

PR #570, opened 2022-10-18
feat: add sessionID HTTP Header to the Docker client setup

mdelapenya:
## What does this PR do?
It initialises the sessionID of the containers just once, as in the past it was created per-container, and this is not the expected behavior.

For that, we have created a global sessionID variable that is protected by sync.Once, therefore it's called just once.

This variable is used in the library with a singleton method, which is called from everywhere that a new UUID was required by a container. Therefore, all containers will share the same sessionID during a test execution/session.

Besides that, we detected a minor change in how the reaper was declaring its sessionID, using a variable with a wider scope: we have moved that sessionID variable to its own scope.

Finally, we are adding an HTTP header to the setup of the Docker client, which will allow improving tracing and other operations.

## Why is it important
Better observability at the Docker API level, as we are instrumenting the HTTP requests.

With these changes we are also decoupling bit-by-bit the existing codebase: before them, the calculation of the sessionID happened everytime a container/network was created. Now, it happens just once per test execution. 

## Related issues
- Relates https://github.com/testcontainers/testcontainers-java/pull/5145
- Relates https://github.com/testcontainers/testcontainers-dotnet/commit/0eaea8b0e26e6457e168d160f237c7c33d8f61a7
- Relates https://github.com/testcontainers/testcontainers-go/issues/549

### DIFF
--- a/reaper.go
+++ b/reaper.go
@@ -63,13 +63,17 @@ func NewReaper(ctx context.Context, sessionID string, provider ReaperProvider, r
 		ExposedPorts: []string{string(listeningPort)},
 		NetworkMode:  Bridge,
 		Labels: map[string]string{
-			TestcontainerLabel:         "true",
 			TestcontainerLabelIsReaper: "true",
 		},
 		SkipReaper: true,
 		Mounts:     Mounts(BindMount(dockerHost, "/var/run/docker.sock")),
 		AutoRemove: true,
 		WaitingFor: wait.ForListeningPort(listeningPort),
+	}
+
+	// include reaper-specific labels to the reaper container
+	for k, v := range reaper.Labels() {
+		req.Labels[k] = v
 	}
 
 	tcConfig := provider.Config()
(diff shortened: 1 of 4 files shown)


reaper.go
```diff
@@ -72,6 +71,11 @@ func NewReaper(ctx context.Context, sessionID string, provider ReaperProvider, r
 		WaitingFor: wait.ForListeningPort(listeningPort),
 	}
 
+	// include reaper-specific labels to the reaper container
+	for k, v := range reaper.Labels() {
```
mdelapenya, line 75: We identified that the Labels() function contains more information: about the SessionID

https://github.com/testcontainers/testcontainers-go/pull/570/files#diff-e8ab4ec17e978db949cd15726ba0efa9c89fd714e7e6e0aa537ed7fc1bbde7fdR150-R155

kiview (approved):
Makes semantically sense from my `testcontainers-java` point of view 👍 (can't comment on any Go specifics).

HofmeisterAn (approved):
> Makes semantically sense from my `testcontainers-java` point of view 👍 (can't comment on any Go specifics).

Same for .NET.

PR #569, opened 2022-10-17
feat: add golangci-lint

hhsnopek:
## Context

This pull request introduces golangci-lint into the existing CI/CD Pipeline to increase the project's code health. Applying `go fmt` to all files, removing blank identifier usage in all Wait Strategies for the implicit acceptance of the interface, removes deprecated functions/methods, and fixes _a lot_ of missing error handling.

~Additionally, in 9db062a95da2fadb9d7dcf6dbe3f895f99cfb3f5 I've exposed the `noopStrategyTarget` as `waittest.NopStrategyTarget` for others to consume in their own tests. It was renamed to match the `Nop` naming style used in Go's standard lib. This does not need to be apart of these changes and can be removed from this PR if required.~

PR #580, opened 2022-10-24
feat: implement new MultiStrategy design 

hhsnopek:
## Context

This pull request implements #559, see the design for details.

TODO:
- [x] Add documentation
- ~[ ] Refactor/Clean up Strategies~

While implementing this, I struggled to extend existing strategies due to the use of `Strategy` and `StrategyTarget` abstracts. I have implemented or tested my assumptions but using Functional Options instead of Method Chaining would open the API further and expose less of the inner API details. I may suggest this in a future discussion/iteration. For this implementation, I continued the method chaining and was forced to expose `Strategy#Timeout` to check if the wait strategies had a timeout defined.

This pull request requires #569 to merge as the work is based on those foundational changes. See [c3de1bf](https://github.com/testcontainers/testcontainers-go/pull/580/commits/c3de1bf0f3e43df0936f80884bf07725654e1153) for the only commit used for #559

closes #559


mdelapenya:
Do you think the branch was created from main? I see commits from the golangci-lint P. Could you please remove those commits to simplify the review 🙏 ?

hhsnopek:
@mdelapenya definitely can separate it out once before I transition it from a draft -> ready for review. I mentioned in the PR body that [c3de1bf](https://github.com/testcontainers/testcontainers-go/pull/580/commits/c3de1bf0f3e43df0936f80884bf07725654e1153) is the only commit for #559

hhsnopek:
@mdelapenya I've removed the golanglint-ci commits, feel free to take another peek! I haven't had time to add documentation and clean up the existing strategies. If you're comfortable with merging as is I can follow up with Documentation & a Strategy refactor in separate pull requests

mdelapenya:
@hhsnopek could you update the docs for all the wait strategies here https://golang.testcontainers.org/features/wait/introduction? 🙏 

I think this is the only thing I'm missing at the moment

hhsnopek:
@mdelapenya all set here 🚀

wait/exit.go
```diff
@@ -32,11 +32,11 @@ func NewExitStrategy() *ExitStrategy {
 
 // WithExitTimeout can be used to change the default exit timeout
 func (ws *ExitStrategy) WithExitTimeout(exitTimeout time.Duration) *ExitStrategy {
-	ws.exitTimeout = exitTimeout
+	ws.timeout = &exitTimeout
 	return ws
 }
 
-// WithPollInterval can be used to override the default polling interval of 100 milliseconds
+// WithPollInterval can be used to override the default polling interval of 100 milliseconds func (ws *ExitStrategy) WithPollInterval(pollInterval time.Duration) *ExitStrategy {
```
mdelapenya, line 39: Probably wrong copy&paste 😅 
```suggestion
// WithPollInterval can be used to override the default polling interval of 100 milliseconds
```

wait/all_test.go
```diff
@@ -0,0 +1,97 @@
...
+		ctx    context.Context
+		target StrategyTarget
+	}
+	tests := []struct {
+		name     string
+		strategy Strategy
+		args     args
+		wantErr  bool
+	}{
+		{
```
mdelapenya, line 24: Could you add a few test cases where `wantErr` is true? 🙏

wait/exit.go
```diff
@@ -53,13 +53,16 @@ func ForExit() *ExitStrategy {
...
+}
+
 // WaitUntilReady implements Strategy.WaitUntilReady
 func (ws *ExitStrategy) WaitUntilReady(ctx context.Context, target StrategyTarget) (err error) {
-	// limit context to exitTimeout
-	if ws.exitTimeout > 0 {
-		var cancelContext context.CancelFunc
-		ctx, cancelContext = context.WithTimeout(ctx, ws.exitTimeout)
-		defer cancelContext()
+	if ws.timeout != nil {
```
mdelapenya, line 62: What happens if the user sets timeout to zero? Reading the code of `context.WithTimeout`:

```golang
// WithTimeout returns WithDeadline(parent, time.Now().Add(timeout)).
//
// Canceling this context releases resources associated with it, so code should
// call cancel as soon as the operations running in this Context complete:
//
// 	func slowOperationWithTimeout(ctx context.Context) (Result, error) {
// 		ctx, cancel := context.WithTimeout(ctx, 100*time.Millisecond)
// 		defer cancel()  // releases resources if slowOperation completes before timeout elapses
// 		return slowOperation(ctx)
// 	}
func WithTimeout(parent Context, timeout time.Duration) (Context, CancelFunc) {
	return WithDeadline(parent, time.Now().Add(timeout))
}
```

It seems that the zero would be added to now, cancelling immediately. Is this assumption correct?

wait/nop.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"context"
+	"io"
+	"time"
+
+	"github.com/docker/docker/api/types"
+	"github.com/docker/go-connections/nat"
+)
+
+type NopStrategy struct {
```
mdelapenya, line 12: Do we make this strategy into the public API? I'd suggest not doing it unless we see the value for the downstream users

mdelapenya (commented):
Very well structured, thanks for the change. I'm adding a few minor comments that need to be addressed.

Once resolved, I  think we can move this little one on.

As always, thanks for your hard work here!

wait/nop.go
hhsnopek: This would be available in public API for consumers to use with the existing strategies. I've seen various third-party and internal/std packages which expose a Nop version.

* https://pkg.go.dev/io#NopCloser
* https://pkg.go.dev/github.com/datadog/datadog-go/statsd#NoOpClient
* https://pkg.go.dev/go.uber.org/zap#NewNop

wait/exit.go
hhsnopek: Zero would only be used if explicitly defined by the user, which would be odd. The reason we use a pointer for `timeout` is to differentiate between an undefined (nil) state and user-defined (time.Duration) state.

wait/exit.go
hhsnopek: resolved via 35b1272

wait/all_test.go
hhsnopek: resolved via [adfb44b](https://github.com/testcontainers/testcontainers-go/pull/580/commits/adfb44b9f079b1e7c5691988cc3fdc0927385199)

wait/nop.go
mdelapenya: Indeed. Let's keep the NopStrategy. Cloud you differentiate it from the already existing `noopStrategyTarget`, from the log_test.go file?

I'd keep just one.

hhsnopek: I went with `Nop` instead of `Noop` or `NoOp` since the standard library `io.NopCloser` used it, I tend to prefer what the standard library uses. I have no issues with changing it to `Noop` or `NoOp`

mdelapenya: Ok then, I'd like to stand with standard library too. Let's keep the prefix as `Nop`. Could you apply it everywhere?

wait/wait.go
```diff
@@ -12,6 +12,7 @@ import (
 
 type Strategy interface {
 	WaitUntilReady(context.Context, StrategyTarget) error
+	Timeout() *time.Duration
```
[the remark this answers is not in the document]
hhsnopek: Based on `Strategy` being an interface and not a struct and that Go's interface cannot define attributes, only functions/methods we have two options:
* expose the Timeout as is with a Getter
* create a `Strategy` struct that defines the bare minimum of a `Strategy` (timeout as `*time.Duration`) and expose the Strategy struct through a Getter and then retrieve the timeout

If there's another option I would love to know it, I try my best to prevent breaking changes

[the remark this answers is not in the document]
mdelapenya: I found this reading interesting: https://go.dev/blog/module-compatibility#working-with-interfaces

Do you think the `Timeout` function could be private? If so, we would not break existing implementations when migrating to a new version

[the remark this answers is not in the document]
hhsnopek: After playing around with different implementations I believe I have landed on a backwards compatible change. We could make the `Timeout` function private but Public API users would be unable to use the new MultiStrategy configuration. To support both old and new implementations of a `Strategy` we can expose a second interface `StrategyTarget` which defines the public `Timeout`. All `testcontainers-go/wait` strategies can implement this interface by defining the public `Timeout` function. In `MultiStrategy` we would use a type assertion to see if the Strategy implements StrategyTimeout we can use the new conditional logic for defining a default strategy. If it does not, we ignore it setting a default timeout.

Here's a [proof of concept](https://go.dev/play/p/o54_Yc4OZlZ)

If this is accepted I'll implement it like so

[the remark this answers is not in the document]
mdelapenya: Sounds good to me. It's aligned with what it's explained in https://go.dev/blog/module-compatibility#working-with-interfaces

> Start by creating a new interface with your new method, or identify an existing interface with the new method. Next, identify the relevant functions that need to support it, type check for the second interface, and add code that uses it.
>
> This strategy only works when the old interface without the new method can still be supported, limiting the future extensibility of your module.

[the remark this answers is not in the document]
hhsnopek: resolved via [c594243](https://github.com/testcontainers/testcontainers-go/pull/580/commits/c5942432a584f9209037629e4e5a86dce7d9a1f4)

wait/wait.go
```diff
@@ -7,19 +7,25 @@ import (
...
+	"github.com/testcontainers/testcontainers-go/exec"
 )
 
+// Strategy defines the basic interface for a Wait Strategy
 type Strategy interface {
 	WaitUntilReady(context.Context, StrategyTarget) error
 }
 
+// StrategyTimeout allows MultiStrategy to configure a Strategy's Timeout
+type StrategyTimeout interface {
```
mdelapenya, line 19: As a follow-up PR, let's add docs on how to create new wait strategies, using both interfaces.

mdelapenya (approved):
Thanks for this contribution, including a superb design discussion (#559)

LGTM 👍 

image

PR #597, opened 2022-10-28
add example of using Pulsar with testcontainers-go

clive-jevons:
A quick example of how I used Pulsar in a recent test for another project which others might find useful.

It essentially mimics the way the testcontainers-java Pulsar module sets up Pulsar.

### DIFF
--- a/docs/examples/pulsar.md
+++ b/docs/examples/pulsar.md
@@ -7,12 +7,10 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"testing"
 	"time"
 
 	"github.com/apache/pulsar-client-go/pulsar"
-	. "github.com/onsi/gomega"
 	"github.com/testcontainers/testcontainers-go"
 	"github.com/testcontainers/testcontainers-go/wait"
 )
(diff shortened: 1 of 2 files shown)


mdelapenya:
This is super cool!! Thanks for adding the example.

I'd remove the gomega dependency to avoid associating the example with a test framework. I'd use testing library + require or assert. You can check the ones used in the project as a reference.

In any case, this LGTM!!

mdelapenya:
I've just realized that the example is not shown in the website because the mkdocs descriptor living in the root for of the project, she's not included the new entry. Could you please add it?

docs/examples/pulsar.md
```diff
@@ -0,0 +1,114 @@
...
+	"io"
+	"testing"
+	"time"
+
+	"github.com/apache/pulsar-client-go/pulsar"
+	"github.com/testcontainers/testcontainers-go"
+	"github.com/testcontainers/testcontainers-go/wait"
+)
+
+func TestPulsar(t *testing.T) {
```
mdelapenya, line 18: I think it's fine to have the log consumer in place to demonstrate how to do a test.

What other examples do is to create a method to setup the container (setupPulsar) which only creates the container, and then they create a TestFoo function with the real test.

I think that would be even more helpful, as we decouple the pulsar container creation from its usage in a test. Wdyt?

clive-jevons: Done :)

mdelapenya (approved):
LGTM!!

I've test it locally with success, thanks for this great example on Apache Pulsar usage. Great job

Issue #601, opened 2022-10-31
[Feature]: Allow build image from registries requiring authentication


paulozenida:
### Problem

Currently, it is not possible to build a local Docker image that is built from another one that is fetching from a Docker registry that requires authentication

### Solution

Expose AuthConfig when creating docker requests so that clients may configure the necessary credentials.

### Benefit

It will allow to build Docker images that rely on other Docker images located in registries that require authentication

### Alternatives

Not applicable. I tried using `~/.docker/config.json` but I failed to do make it work.

### Would you like to help contributing this feature?

Yes

mdelapenya:
Closed as already implemented, thanks!

PR #602, opened 2022-10-31
feat: Auth config for build images

paulozenida:
This PR allows for the feature reported in the following issue: https://github.com/testcontainers/testcontainers-go/issues/601


container_test.go
```diff
@@ -127,6 +128,50 @@ func Test_GetDockerfile(t *testing.T) {
 	}
 }
 
+func Test_GetAuthConfigs(t *testing.T) {
```
mdelapenya, line 131: I see the value in this test to verify that the auth credentials are set, and I'd like to see a test for interacting with a real registry. 

Have you explored the idea of starting a Docker registry as a container, configured with security as explained here https://docs.docker.com/registry/deploying/#restricting-access?

paulozenida: Yes. I just tried it and updated this PR. Could you please take a look?

docker_test.go
```diff
@@ -1033,8 +1034,93 @@ func Test_BuildContainerFromDockerfile(t *testing.T) {
...
+		WaitingFor:   wait.ForLog("Ready to accept connections"),
+	}
+
+	redisC, err := prepareRedisImage(ctx, req, t)
+	require.Error(t, err)
+	terminateContainerOnEnd(t, ctx, redisC)
+}
+
+func prepareLocalRegistryWithAuth(t *testing.T) {
+	compose, err := NewDockerCompose("./testresources/docker-compose-auth.yml")
```
mdelapenya, line 1110: I think you can start a generic container here instead of the compose one 🤔

paulozenida: Sure but, do you see an issue with being a docker-compose instead?

mdelapenya: For the pure writing and maintenance of the tests, they would be more clear as a GenericContainer, as the definition of the container would be there in the test method, instead of in a compose file that we have to open elsewhere.

I'd advocate for writing compose tests for the compose code.

paulozenida: Done. I've replaced the docker-compose with a docker container in preparation to the actual test.

docker_test.go
```diff
@@ -1107,17 +1108,52 @@ func Test_BuildContainerFromDockerfileWithAuthConfig_ShouldFailWithoutAuthConfig
 }
 
 func prepareLocalRegistryWithAuth(t *testing.T) {
-	compose, err := NewDockerCompose("./testresources/docker-compose-auth.yml")
-	assert.NoError(t, err, "NewDockerCompose()")
+	ctx := context.Background()
+	wd, err := os.Getwd()
```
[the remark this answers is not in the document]
paulozenida: I don't understand your comment here. I need to refer to the actual folders (data and auth) inside testresources but the mount needs to be an absolute path. How would TempDir work here?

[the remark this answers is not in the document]
mdelapenya: Ah sorry, sorry. I though we were generating the data into a "temp" folder, but you are right: it's actually mounting the testresources dir into the container. 🙏

mdelapenya:
@paulozenida I saw the PR is failing the `TestDockerCreateContainerWithDirs` test because here https://github.com/testcontainers/testcontainers-go/blob/dd2248a01f2d6ded8b51034f37810cf819f24387/docker_test.go#L2566 the existing code is not controlling that a directory could exist under `testresources`. At the moment you added the registry dirs, that test fail, and I think it should be fixed in this PR.

I'd say that simply checking the current srcFile is not a directory, and then read it would be enough. Could you include it in this PR?

Thanks!

paulozenida:
> 

Sure. I've just included the suggested change.

mdelapenya:
> > 
> 
> Sure. I've just included the suggested change.

To avoid causing you more trouble in the pingpong review on failed tests, I added a commit with the same fix in another test: `TestTarDir`. Hopefully this will PASS 🤞

mdelapenya (approved):
Great job @paulozenida in this PR!

I'm updating the branch with current main. Once the CI passes, I'll merge it 🚀 

BTW; the project is missing the same mechanism for pulling images from private registries in other places. As you already contributed this part for the FromDockerfile part, I wonder if you would be interested in contributing the others, such as the Docker Compose code. I can be of any help assisting you identifying where to add it. Please let me know if you'd like to participate.

paulozenida:
> Great job @paulozenida in this PR!
> 
> I'm updating the branch with current main. Once the CI passes, I'll merge it 🚀
> 
> BTW; the project is missing the same mechanism for pulling images from private registries in other places. As you already contributed this part for the FromDockerfile part, I wonder if you would be interested in contributing the others, such as the Docker Compose code. I can be of any help assisting you identifying where to add it. Please let me know if you'd like to participate.

Thank @mdelapenya.

I'm not sure I'll be able to do that right now but if you open an issue explaining the details, I might try to get to it next week or try to find someone in my organisation/team that can try to do so.

PR #603, opened 2022-11-02
feat: provide a tool to generate examples from code

mdelapenya:
## What does this PR do?
It moves the existing examples' code (cockroachDB, nginx, pulsar and redis) to separate example Go modules that will be automatically included into the docs site.

Besides, we have provided a command line tool to generate the scaffolding for the code of a new example. This tool will generate:

- a Go module for the example, including:
    - go.mod and go.sum files
    - a Go package named after the example, in lowercase
    - a Go file for the creation of the container, using a dedicated struct.
    - a Go test file for running a simple test for your container, consuming the above struct.
    - a Makefile to run the tests in a consistent manner
    - a tools.go file including the build tools (i.e. `gotestsum`) used to build/run the example.
- a markdown file in the docs/examples directory including the snippets for both the creation of the container and a simple test.

This tool is not doing the following:
- If the example already exists, it will exit without updating the existing files.
- You have to manually add the markdown entry in the docs to the `mkdocs.yml` file in the root directory of the project. It will generate the navigation menu for the docs website.

In this PR we are adding a Make goal to run all the examples' tests, in order to check if they are still valid and not outdated. The make goal, which calls each example's Makefile, is used in the CI pipeline.

## Why is it important?
Examples embedded into the docs could get outdated and they could not even compile. With this approach, the tests will be part of the test suite.

We expect the code generation tool will simplify how to start an example, creating the scaffolding for it.


eddumelendez (dismissed):
nice!

examples/pulsar/pulsar.go
```diff
@@ -0,0 +1,67 @@
...
+		resp := string(respBytes)
+		return resp == `["standalone"]`
+	}
+	pulsarRequest := testcontainers.ContainerRequest{
+		Image:        "docker.io/apachepulsar/pulsar:2.10.2",
+		ExposedPorts: []string{"6650/tcp", "8080/tcp"},
+		WaitingFor: wait.ForAll(
+			wait.ForHTTP("/admin/v2/clusters").WithPort("8080/tcp").WithResponseMatcher(matchAdminResponse).WithStartupTimeout(defaultTimeout),
+			wait.ForLog("Successfully updated the policies on namespace public/default").WithStartupTimeout(defaultTimeout),
+		).WithStartupTimeout(defaultTimeout),
```
mdelapenya, line 32: @clive-jevons please take a look at this startup timeout

examples/pulsar/pulsar_test.go
```diff
@@ -0,0 +1,72 @@
...
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/apache/pulsar-client-go/pulsar"
+)
+
+func TestPulsar(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTimeout)
```
mdelapenya, line 13: @clive-jevons same here

mdelapenya:
Superseded by #618

PR #624, opened 2022-11-16
fix: do not prepend garbage in the container.Exec response

### LINKED ISSUE #620: [Bug]: Random characters in the output of `io.Reader` returned in `Exec(ctx context.Context, cmd []string)`
### Testcontainers version

0.15.0

Yes

### Host OS

Macos

### Host arch

x86

### Go version

1.19

### Docker version

```shell
Client:
 Cloud integration: v1.0.29
 Version:           20.10.20
 API version:       1.41
 Go version:        go1.18.7
 Git commit:        9fdeb9c
 Built:             Tue Oct 18 18:20:35 2022
 Context:           default
 Experimental:      true

Server: Docker Desktop 4.13.1 (90346)
 Engine:
  Version:          20.10.20
  Go version:       go1.18.7
  Git commit:       03df974
  Built:            Tue Oct 18 18:18:35 2022
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.8
 runc:
  Version:          1.1.4
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### Docker info

```shell
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  compose: Docker Compose (Docker Inc., v2.12.1)
  dev: Docker Dev Environments (Docker Inc., v0.0.3)
  extension: Manages Docker extensions (Docker Inc., v0.2.13)
  scan: Docker Scan (Docker Inc., v0.21.0)

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 12
 Server Version: 20.10.20
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 init version: de40ad0
 Security Options:
  seccomp
   Profile: default
  cgroupns
 Kernel Version: 5.15.49-linuxkit
 OSType: linux
 Architecture: x86_64
 CPUs: 8
 Total Memory: 7.772GiB
 Name: docker-desktop
 ID: XS7M:CTCG:HOAE:PMYH:5IOL:TEXS:CYEV:UCQ6:UISC:4IIN:2U6L:XJ2I
 Debug Mode: false
 Labels:
 Experimental: false
 Insecure Registries:
  hubproxy.docker.internal:5000
  127.0.0.0/8
 Live Restore Enabled: false
```

### What happened?

The `io.Reader` returned by Exec() contains random characters. For my setup I am running a `rabbitmq:management` container and I am running `rabbitmqctl` command to list some connection information, like so,

```go
...
request := tc.GenericContainerRequest{
		ContainerRequest: tc.ContainerRequest{
			Image:        c.image,
			ExposedPorts: []string{"5672/tcp", "15672/tcp"},
			WaitingFor:   wait.ForLog("Server startup complete").WithStartupTimeout(c.timeout),
			AutoRemove: true,
		},
		Started: true,
	}

	container, err := tc.GenericContainer(ctx, request)
	if err != nil {
		return nil, fmt.Errorf("failed to start container: %v", err)
	}
...
_, reader, err := c.Exec(ctx, []string{"rabbitmqctl", "list_connections", "name", "state", "--formatter", "json"})
if err != nil {
return err
}
```

The output in the returned reader seems to contain random characters as such, 
```json
[
�{"name":"172.17.0.1:64168 -> 172.17.0.3:5672","state":"running"}
,{"name":"172.17.0.1:64184 -> 172.17.0.3:5672","state":"running"}
]
```

To rule out a formatter error on `rabbitmqctl`, I tried running the command with a `csv` formatter, and I see the same behavior with random characters in the output of the reader. (Again there are no random characters when the same command is run in the docker container after `exec`ing into it via `docker exec -it <container_id> bash`

```go
_, reader, err := c.Exec(ctx, []string{"rabbitmqctl", "list_connections", "name", "state", "--formatter", "csv"})
if err != nil {
return err
}
```

```csv
o"name","state"
"172.17.0.1:55048 -> 172.17.0.3:5672","running"
"172.17.0.1:55060 -> 172.17.0.3:5672","running"
```

The output looks clean when executed outside of `testcontainers-go` inside the container directly. I see in the source code that the tests for `Exec` don't seem to cover the reader returned from the `Exec` method.

Any ideas why this is happening?

### Relevant log output

_No response_

### Additional information

_No response_

mdelapenya:
## What does this PR do?
This PR provides a why to extract the multiplexed Docker response when executing a command in a container, adding a variadic argument holding the options that a user would need to process the Docker response. In this PR, we are adding just one option for the processor: `Multiplexed`, which uses Docker's built-in `stdcopy.StdCopy` to extract the simple response from the combined response that Docker generates.

For further reference on how this multiplexed response is obtained:
- Docker docs: https://docs.docker.com/engine/api/v1.37/#tag/Container/operation/ContainerAttach
- Tescontainers for Java implementation, via the usage of the `Docker-Java` library: https://github.com/docker-java/docker-java/blob/f6f9d676d9244d6048fdf85c5db6f2c4813d5c23/docker-java-core/src/main/java/com/github/dockerjava/core/FramedInputStreamConsumer.java#L28,
- Testcontainer for .NET, via the usage of the `Docker.DotNet` library: https://github.com/dotnet/Docker.DotNet/blob/30181cc146076ee8dc7ab839535ae89128672531/src/Docker.DotNet/MultiplexedStream.cs#L82

We are adding two unit tests demonstrating the bug, running an Nginx container, and executing `ls /usr/share/nginx`, which should return just one directory: `html`. We are covering both use cases: executing a command in the container with multiplexing the response (the output is sanitized) and without multiplexing it (output contains garbage at the beginning).

As expected, the tests fail before the fix and pass after it.

Given the original function returned a Reader, pushing the responsibility of processing it to the client code, we did not want to update the signature breaking the API, therefore, this solution is adding the ability to request the multiplexed response. This is achieved by replacing the original Reader with the result of Docker's stdcopy. Please notice the stderr Reader is not used at all. In future versions of this library we probably offer a deprecation path to return a struct containing both readers.

## Why is it important?
There is a bug when using the `container.Exec` function, as it prepends garbage to the reponse bytes returned by the Docker client, caused by how the command is created at the library level.

With this fix, it's possible to choose the response format, simplifying how client code is using the Exec function.

## Related issues

- Closes #620

## Follow-ups
Let's create an issue to offer a deprecation path for the current Exec function, which returns a struct containing stdout and stderr, something like what we already do for the Docker Compose module: https://github.com/testcontainers/testcontainers-go/blob/main/compose_local.go#L245


docker.go
```diff
@@ -440,12 +440,15 @@ func (c *DockerContainer) Exec(ctx context.Context, cmd []string) (int, io.Reade
...
 		AttachStderr: true,
+		Tty:          false,
 	})
 	if err != nil {
 		return 0, nil, err
 	}
 
-	hijack, err := cli.ContainerExecAttach(ctx, response.ID, types.ExecStartCheck{})
+	hijack, err := cli.ContainerExecAttach(ctx, response.ID, types.ExecStartCheck{
+		Tty: true,
```
mdelapenya, line 450: This `true` value does not mean that it will use Tty **but** to check if there is Tty 🤷

HofmeisterAn: I am not sure if I understood the concept correct, but for .NET `tty` is always `false`. There is no interaction, we run the command and return the result.

mdelapenya: Indeed: when we create the command, it's always `false` https://github.com/testcontainers/testcontainers-go/pull/624/files#diff-b68159b6168f099a639e8c6f96d75f23333ab0f0505ee22a98268748db17ae4cR443

What I was not aware of, is the ExecAttach part

cyberbeast (approved):
Thank you for taking a look at it. Looks good to me.

mdelapenya:
> Thank you for taking a look at it. Looks good to me.

@cyberbeast we are going to propose another solution for the fix, keeping the test: we verified that Docker defines a protocol to stream responses: https://docs.docker.com/engine/api/v1.37/#tag/Container/operation/ContainerAttach

This protocol is implemented in Docker Java here: https://github.com/docker-java/docker-java/blob/f6f9d676d9244d6048fdf85c5db6f2c4813d5c23/docker-java-core/src/main/java/com/github/dockerjava/core/FramedInputStreamConsumer.java#L28, and in Docker.DotNet here: https://github.com/dotnet/Docker.DotNet/blob/30181cc146076ee8dc7ab839535ae89128672531/src/Docker.DotNet/MultiplexedStream.cs#L82

In Go we are simply returning the Reader, but it could be interesting to hide the streaming protocol implementation detail from the consumers, wrapping the returned reader demultiplexing the response.

kiview (approved):
I think it is the right approach to delegate `github.com/docker/docker/pkg/stdcopy` and the `ProcessOption` interface allows us to be open for future required changes, if they are necessary. 

If we merge the PR like this, the PR description should be updated to reflect the new approach of the PR, else it is quite misleading (it still references the TTY approach). 

Do we plan to introduce an abstraction later, that makes accessing `stdout` and `stderr` distinctively easier?

mdelapenya:
> I think it is the right approach to delegate `github.com/docker/docker/pkg/stdcopy` and the `ProcessOption` interface allows us to be open for future required changes, if they are necessary.
> 
> If we merge the PR like this, the PR description should be updated to reflect the new approach of the PR, else it is quite misleading (it still references the TTY approach).
> 
> Do we plan to introduce an abstraction later, that makes accessing `stdout` and `stderr` distinctively easier?

Yes, need to update the description of the PR.

Indeed, accessing the stderr (stdout is already available in the original reader) would be desired, and I already thought about it but wanted to check this solution first. Maybe logging stderr would be a thing to consider to avoid changing the signature of the Exec function.

kiview:
Then I'd suggest merging this PR as-is to solve the issue at hand and think about approaches for programmatic access to `stderr`. 

Sometimes `stderr` behaves quite unexpected. I remember in the specific case of PostgreSQL, that the database process like to output very normal startup information on `stderr`. So intuitively assuming we can log `stderr` as an error might not be the right approach.

PR #630, opened 2022-11-21
chore: bump version in mkdocs

mdelapenya:
## What does this PR do?
It bumps the version in mkdocs to the latest_version

## Why is it important?
TBH I'm not sure if this extra value is used by mkdocs to render the docs site. According to their docs: https://www.mkdocs.org/user-guide/configuration/#extra

> A set of key-value pairs, where the values can be any valid YAML construct, that will be passed to the template. This allows for great flexibility when creating custom themes.
>
> For example, if you are using a theme that supports displaying the project version, you can pass it to the theme like this:

```yml
extra:
    version: 1.0
```

But I cannot see any occurrence of the `latest_version` key.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -62,4 +62,4 @@ nav:
     - Getting help: getting_help.md
 edit_uri: edit/main/docs/
 extra:
-    latest_version: 0.15.0
+    latest_version: 0.16.0


kiview:
In the case of `testcontainers-java`, we use the `latest_version` variable to template a couple of String in the documentation, what's why it makes sense there.

mdelapenya:
> In the case of `testcontainers-java`, we use the `latest_version` variable to template a couple of strings in the documentation, what's why it makes sense there.

Ok then, I'll keep it here in the case we need it for the future of the examples. Thanks for confirming!

PR #633, opened 2022-11-22
chore: move host-config and endpoint settings to a specific modifiers

mdelapenya:
## What does this PR do?
This PR addresses the need of separating concerns when creating a container, as we want to distinguish between basic and advanced settings.

With basic, we mean settings that are used by the vast majority of our users, exposed as part of the public API of Testcontainers for Go. And with advanced, we mean settings that should still be still available in the public API, but in a hidden manner.

Therefore, we are adding the concept of a `Modifier` to the `ContainerRequest`: a user will be able to define what advanced configurations should be processed before a container is created. 

The advanced settings that we have identified at this moment are those that are related to 3 elements:
1. the `container.Config` Docker type
2. the `container.HostConfig` Docker type
3. the `network.EndpointSettings`

Therefore, we have defined three modifiers:  `ConfigModifier`, `HostConfigModifier` and `EndpointSettingsModifier`.

From Docker docs:

```
// HostConfig the non-portable Config structure of a container.
// Here, "non-portable" means "dependent of the host we are running on".
// Portable information *should* appear in Config.
```

On the contrary, we chose basic settings to be related to the portable information of a container, under the `container.Config` Docker type: 

```
// Config contains the configuration data about a container.
// It should hold only portable information about the container.
// Here, "portable" means "independent from the host we are running on".
// Non-portable information *should* appear in HostConfig.
// All fields added to this struct must be marked `omitempty` to keep getting
// predictable hashes from the old `v1Compatibility` configuration.
```

The technical implementation of this PR comes with:
- we have created three new fields in the ContainerRequest that holds a function to pre-configure advanced settings for the container (hostConfig and endpointSettings):  `ConfigModifier`, `HostConfigModifier` and `EndpointSettingsModifier`. These fields in the request are part of the public API of the container request.
- the `Modifiers` functions are executed before the container is created using the Docker client.
- the `Modifiers` function allow users modify the config, hostConfig and endpointSettings Docker types when they request a container, handy for advances use cases.
- the fields that were present in the ContainerRequest related to the HostConfig, and the EndpointSettings have been `Deprecated` in favor of the different `Modifiers`.
- the fields that relate to the "basic" settings, covered by the `Config` Docker type, are still present in the container request. The modifier will take precedence, as we are going to merge the struct in the modifier with the already created one (from the container request).
- the library will internally check if a request includes each `Modifier`, and provide a default hook including the deprecated fields if it's not present in the request. This way we are keeping backwards compatibility until we remove those fields.
- we have updated the existing tests, using the `Modifiers` when needed, removing the deprecated fields from the tests.
- we have added unit tests for the preCreateModifierHook, to verify that all the fields in play are properly set.

## Why is it important?
We want to allow users configure their containers, but helping them not to commit mistakes caused by a wrong configuration. If the public API directly exposes Docker types in a way that they are easy to consume, a user trying to configure a container with basic settings could be tempted to use advanced setting.

More info about this can be found in the #628 discussion.

Another benefit that comes here is reducing the size of the container request: we do not need to add fields to the public API to map a Docker type: anything can be modified with the Modifiers. We should provide a simple public API, but for advanced use cases, the modifiers assist.

## Related issues

- Relates #628

## How to test this PR
Running the tests would not represent a change in the previous behavior.

Another test could be using a custom project that replaces the testcontainers-go dependency for the one containing this PR in the local workspace. The dependent project must compile and its tests must pass.


container.go
```diff
@@ -101,28 +102,29 @@ type ContainerRequest struct {
 	Cmd             []string
 	Labels          map[string]string
 	Mounts          ContainerMounts
-	Tmpfs           map[string]string
+	Tmpfs           map[string]string // Deprecated: Use PreCreationHook instead
```
kiview, line 105: In Java, we expose this in the 1st level API. Why? Because it is a popular choice for database containers to get some quick performance wins.

container.go
```diff
@@ -101,28 +102,29 @@ type ContainerRequest struct {
...
-	SkipReaper      bool            // indicates whether we skip setting up a reaper for this
-	ReaperImage     string          // alternative reaper image
-	AutoRemove      bool            // if set to true, the container will be removed from the host when stopped
-	AlwaysPullImage bool            // Always pull image
-	ImagePlatform   string          // ImagePlatform describes the platform which the image runs on.
-	Binds           []string
-	ShmSize         int64    // Amount of memory shared with the host (in bytes)
-	CapAdd          []string // Add Linux capabilities
-	CapDrop         []string // Drop Linux capabilities
+	ExtraHosts      []string                                                          // Deprecated: Use PreCreationHook instead
```
kiview, line 110: 1st level in Java. But we don't see any modules using it, not sure how much used by our users and in which use cases.

container.go
```diff
@@ -101,28 +102,29 @@ type ContainerRequest struct {
...
+	Privileged      bool                                                              // Deprecated: Use PreCreationHook instead. For starting privileged container
```
kiview, line 111: 1st level, necessary for certain Docker images (e.g., DB2).

container.go
```diff
@@ -101,28 +102,29 @@ type ContainerRequest struct {
 	Cmd             []string
 	Labels          map[string]string
 	Mounts          ContainerMounts
-	Tmpfs           map[string]string
+	Tmpfs           map[string]string // Deprecated: Use PreCreationHook instead
 	RegistryCred    string
 	WaitingFor      wait.Strategy
 	Name            string // for specifying container name
```
kiview, line 108: I would deprecate this, risk for collisions and generally not needed in the context of TC usage.

container.go
```diff
@@ -101,28 +102,29 @@ type ContainerRequest struct {
...
 	Hostname        string
```
kiview, line 109: We don't have this in 1st level in Java. What are use cases?

container.go
```diff
@@ -101,28 +102,29 @@ type ContainerRequest struct {
...
-	ImagePlatform   string          // ImagePlatform describes the platform which the image runs on.
-	Binds           []string
-	ShmSize         int64    // Amount of memory shared with the host (in bytes)
-	CapAdd          []string // Add Linux capabilities
-	CapDrop         []string // Drop Linux capabilities
+	ExtraHosts      []string                                                          // Deprecated: Use PreCreationHook instead
+	Privileged      bool                                                              // Deprecated: Use PreCreationHook instead. For starting privileged container
+	Networks        []string                                                          // for specifying network names
+	NetworkAliases  map[string][]string                                               // for specifying network aliases
+	NetworkMode     container.NetworkMode                                             // Deprecated: Use PreCreationHook instead
```
kiview, line 114: We have this in 1st level, but I think it is not appropriate there.

container.go
```diff
@@ -101,28 +102,29 @@ type ContainerRequest struct {
...
+	Resources       container.Resources                                               // Deprecated: Use PreCreationHook instead
+	Files           []ContainerFile                                                   // files which will be copied when container starts
+	User            string                                                            // for specifying uid:gid
```
kiview, line 117: Not in 1st level in Java, are there good use cases for this?

container.go
```diff
@@ -101,28 +102,29 @@ type ContainerRequest struct {
...
+	SkipReaper      bool                                                              // indicates whether we skip setting up a reaper for this
+	ReaperImage     string                                                            // alternative reaper image
+	AutoRemove      bool                                                              // Deprecated: Use PreCreationHook instead. If set to true, the container will be removed from the host when stopped
+	AlwaysPullImage bool                                                              // Always pull image
+	ImagePlatform   string                                                            // ImagePlatform describes the platform which the image runs on.
+	Binds           []string                                                          // Deprecated: Use PreCreationHook instead
+	ShmSize         int64                                                             // Deprecated: Use PreCreationHook instead. Amount of memory shared with the host (in bytes)
```
kiview, line 124: We have this in 1st level, since required for Browser images.

kiview (commented):
I tried to match the levels against the current Java implementations. Some are for sure up to discussion, for some we know about their usage, especially in the context of container modules. 

Every container module is a direct use case.

container.go
mdelapenya: This User is not used at all by the library, only in a test that verifies that running `id -u` returns the defined user. I'd use a follow-up PR to discuss whether a field is actually in use or not.

container.go
mdelapenya: Indeed. But I'd do it as part of a new PR, to make it even clearer that passing a name could be seen as an antipattern from testing standpoint

container.go
mdelapenya: Ok, I can moved to the 1st layer. I acted as a robot moving all host config setting to 2nd level

container.go
mdelapenya: let me update the description with all the fields that were pushed from the hostConfig (level 2) to the 1st level

mdelapenya:
As discussed offline with @HofmeisterAn and @eddumelendez having another modifier for the `container.Config`  docker type would be of interest, as then anybody will be able anything they need with advanced scenarios. At the same time we should document it for advanced settings.

lifecycle.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	if req.HostConfigModifier == nil {
+		req.HostConfigModifier = defaultHostConfigModifier(req)
+	}
+	req.HostConfigModifier(hostConfig)
+
+	if req.EnpointSettingsModifier != nil {
+		req.EnpointSettingsModifier(endpointSettings)
+	}
```
HofmeisterAn, line 50: Shouldn't the callbacks be called after the mapping has been completed? For example, lines 71 and 72 modify the instance again, if I have understood it correctly.

mdelapenya: L54:72 are sanitisations that are needed in the case there are no exposed ports, where we read the raw image and extract the ports from there if possible.

HofmeisterAn: Yes, but that means that the modifier methods do not receive the final state of the request.

mdelapenya: That's correct: we could probably move the modifiers after that 🤔

mdelapenya: I've updated it to run the modifiers last, sanitising right before them

mdelapenya: Mmm not sure if I should step back on this change: in the case the hostConfig modifier does not update the exposed ports, neither the original container request does it and the container runs in container network, the original code checked the image ports and exported them. I think we should keep that behaviour and document it

HofmeisterAn: Okay, that is the only thing I noticed. However, please keep in mind that my skills in Go are limited 🤔. Therefore, I was not able to comprehend all the consequences. .NET passes the final [instance](https://github.com/testcontainers/testcontainers-dotnet/blob/fe9f55a84341b57f5e054e631043f590dd2cc74c/src/Testcontainers/Clients/DockerContainerOperations.cs#L173-L208) to the modifier.

docs/features/creating_container.md
```diff
@@ -87,6 +87,17 @@ func TestIntegrationNginxLatestReturn(t *testing.T) {
 }
 ```
 
+### Advanced Settings
+
+The aforementioned `GenericContainer` function and the `ContainerRequest` struct represent a straight forward manner to configure the containers, but you could need to create your containers with more advance settings regarding the config, host config and endpoint settings Docker types. For those more advance settings, _Testcontainers for Go_ offers a way to fully customise the container request and those internal Docker types. These customisations, called _modifiers_, will be applied just before the internal call to the Docker client to create the container.
```
mdelapenya, line 92:
```suggestion
The aforementioned `GenericContainer` function and the `ContainerRequest` struct represent a straightforward manner to configure the containers, but you could need to create your containers with more advance settings regarding the config, host config and endpoint settings Docker types. For those more advance settings, _Testcontainers for Go_ offers a way to fully customise the container request and those internal Docker types. These customisations, called _modifiers_, will be applied just before the internal call to the Docker client to create the container.
```

docs/features/creating_container.md
```diff
@@ -87,6 +87,17 @@ func TestIntegrationNginxLatestReturn(t *testing.T) {
 }
 ```
 
+### Advanced Settings
+
+The aforementioned `GenericContainer` function and the `ContainerRequest` struct represent a straightforward manner to configure the containers, but you could need to create your containers with more advance settings regarding the config, host config and endpoint settings Docker types. For those more advance settings, _Testcontainers for Go_ offers a way to fully customise the container request and those internal Docker types. These customisations, called _modifiers_, will be applied just before the internal call to the Docker client to create the container.
```
kiview, line 92:
```suggestion
The aforementioned `GenericContainer` function and the `ContainerRequest` struct represent a straightforward manner to configure the containers, but you might need to create your containers with more advance settings regarding the config, host config and endpoint settings Docker types. For those more advance settings, _Testcontainers for Go_ offers a way to fully customize the container request and access those internal Docker types. These customizations, called _modifiers_, will be applied just before the internal call to the Docker client to create the container.
```

docs/features/creating_container.md
```diff
@@ -87,6 +87,17 @@ func TestIntegrationNginxLatestReturn(t *testing.T) {
...
+
+<!--codeinclude-->
+[Using modifiers](../../lifecycle_test.go) inside_block:reqWithModifiers
+<!--/codeinclude-->
+
+!!!warning
+	The only special case where the modifiers are not applied last, is when there are no exposed ports in the container request and the container does not use a network mode from a container (e.g. `req.NetworkMode = container.NetworkMode("container:$CONTAINER_ID")`). In that case, _Testcontainers for Go_ will extract the ports from the underliying Docker image and export them.
```
kiview, line 99:
```suggestion
	The only special case where the modifiers are not applied last, occurs when there are no exposed ports in the container request and the container does not use a network mode from a container (e.g. `req.NetworkMode = container.NetworkMode("container:$CONTAINER_ID")`). In that case, _Testcontainers for Go_ will implicitly extract the ports from the underlying Docker image and export them.
```

lifecycle_test.go
```diff
@@ -0,0 +1,299 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestPreCreateModifierHook(t *testing.T) {
+	ctx := context.Background()
+
+	provider, err := NewDockerProvider()
+	require.Nil(t, err)
+
+	t.Run("No exposed ports", func(t *testing.T) {
```
kiview, line 22:
```suggestion
	t.Run("No configured exposed ports should expose default port of image", func(t *testing.T) {
```

lifecycle_test.go
```diff
@@ -0,0 +1,299 @@
...
+		)
+		assert.Equal(
+			t,
+			[]string{"link1", "link2"},
+			inputNetworkingConfig.EndpointsConfig["a"].Links,
+			"Networking config's links should be overwritten by the modifier",
+		)
+	})
+
+	t.Run("No exposed ports and network mode IsContainer", func(t *testing.T) {
```
kiview, line 123:
```suggestion
	t.Run("No exposed ports and network mode IsContainer should not implicitly expose ports", func(t *testing.T) {
```

kiview (approved):

PR #643, opened 2022-11-26
Add toxiproxy example

eddumelendez:
## What does this PR do?
Demostrate how to use `toxiproxy` along with `testcontainers-go`. A proxy is created in order to inject faults to between redis client and redis container.

redis-client -> toxiproxy's proxy -> redis

* Port `8474` is toxiproxy's port
* Port `8666` is a random port for the proxy
* Proxy is created, where `0.0.0.0:8666` will route to `redis:6379`
* Redis' client will use toxiproxy' proxy, it means <toxiproxy-host>:<toxiproxy-proxy-port>
* Read timeout is alter to 2s (default 3s) so the example doesn't take that long when the latency injected is higher. Currently, latency injected is between 900-1100

## Why is it important?
First step for upcoming modules.

## Related issues

- 

Latency can be updated to 3000 ms and then the test will break. In order to fix it the `readTimeout` should be updated too.


mdelapenya (approved):
Fantastic example!

I would not require assertions on how the toxics are added to the requests, as we trust the toxiproxi works as expected.

Super nice contribution!!

PR #652, opened 2022-12-05
fix: do not create garbage in examples nav bar

mdelapenya:
## What does this PR do?
It fixes how the nav bar for the examples is generated, not appending the examples to the "non"-empty array that the generator created to define the examples nav bar.

## Why is it important?
It added empty elements to the nav bar, that the user must remove before submitting the example

## Related issues

-


eddumelendez (approved):
just test it and it works as expected 👍🏽

PR #654, opened 2022-12-05
chore: generate dependabot configs for examples

mdelapenya:
## What does this PR do?
It models dependabot config file, using it to generate dependabot configs for each generated example module.

This generation has this requirements:
- the main module comes first in dependabot.yml
- the existing e2e module comes second
- all the examples comes after, in alphabetical order. This is provided by the `sort.Sort` function, therefore the Updates struct implements the Len, Less and Swap methods of the Sort interface. We can sort updates by its directory.

## Why is it important?
Update example modules with dependencies in an automated manner

## Related issues

- Completes #618


eddumelendez (commented):
besides adding the new section for the new example is removing the quotes to the existing ones

Screenshot 2022-12-05 at 11 59 00

mdelapenya:
Interesting 🤔 let me check how the YAML library is marshaling the strings, as it should include the quotes

eddumelendez:
I think we should remove them first in `main` branch. So, adding a new example is just update the file with the right changes. WDYT?

.github/dependabot.yml
```diff
@@ -12,3 +12,57 @@ updates:
     interval: daily
   open-pull-requests-limit: 3
   rebase-strategy: disabled
+- package-ecosystem: gomod
+  directory: "/examples/cockroachdb"
```
eddumelendez, line 16: remove quotes :)

```suggestion
  directory: /examples/cockroachdb
```

mdelapenya: Removed in [a401b2d](https://github.com/testcontainers/testcontainers-go/pull/654/commits/a401b2d8f471173ec9b6a932ba56f5aeab39f332)

eddumelendez (commented):
when adding a new example the format change. Once this is fixed we are ready to go!

Screenshot 2022-12-05 at 12 30 40

mdelapenya:
@eddumelendez I think go-yaml is still producing valid indents for a YAML file even though the original indents are not honored. Please see https://github.com/go-yaml/yaml/issues/779

eddumelendez (approved):

PR #655, opened 2022-12-05
chore: remove quotes from dependabot updates

mdelapenya:
## What does this PR do?
It removes quotes from dependabot config, preparing for #654

## Why is it important?
YAML marshaling will not quote values unless they change their meaning when unquoted, which is the case

## Related issues

- Blocks #654


eddumelendez (approved):

mdelapenya:
Merging, as it does not affect the CI

PR #702, opened 2022-12-17
chore: reduce concurrent builds

mdelapenya:
## What does this PR do?
It will avoid multiple pushes to have concurrent builds, aborting running builds in the followins scenarios:
- for PRs, a new commit will abort running builds for the workflow
- for branches, because it will use the SHA commit as discriminator, a new commit won't abort the running builds

## Why is it important?
Taking a look at https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration, it seems there is a max number of concurrent builds per organization.

@kiview, we should consider this for any project spanning multiple builds (per lang, per module, per OS...).

## Related issues

-

### DIFF
--- a/.github/workflows/bigtable-example.yml
+++ b/.github/workflows/bigtable-example.yml
@@ -2,6 +2,10 @@ name: Bigtable example pipeline
 
 on: [push, pull_request]
 
+concurrency:
+  group: "${{ github.workflow }}-${{ github.head_ref || github.sha }}"
+  cancel-in-progress: true
+
 jobs:
   test-bigtable:
     strategy:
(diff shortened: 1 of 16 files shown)


HofmeisterAn (approved):

kiview:
Sidenote: In this PR, we see the drawback of duplicating the pipeline config across modules.

kiview (approved):

mdelapenya:
> Sidenote: In this PR, we see the drawback of duplicating the pipeline config across modules.

Indeed, it would represent how we would suffer having a multi-repo, so we have to update just one repository with the new changes.

PR #718, opened 2022-12-30
fix: close response body in http strategy

kruskall:
## What does this PR do?

From to the documentation:
> If the returned error is nil, the Response will contain a non-nil body
which the user is expected to close.

Close the response body according to the documentation.

## Why is it important?

Failed requests (due to wrong response or status code) were leaving open connections.

## Related issues

- None, afaik

### DIFF
--- a/wait/http.go
+++ b/wait/http.go
@@ -217,11 +217,11 @@ func (ws *HTTPStrategy) WaitUntilReady(ctx context.Context, target StrategyTarge
 				continue
 			}
 			if ws.StatusCodeMatcher != nil && !ws.StatusCodeMatcher(resp.StatusCode) {
-				resp.Body.Close()
+				_ = resp.Body.Close()
 				continue
 			}
 			if ws.ResponseMatcher != nil && !ws.ResponseMatcher(resp.Body) {
-				resp.Body.Close()
+				_ = resp.Body.Close()
 				continue
 			}
 			if err := resp.Body.Close(); err != nil {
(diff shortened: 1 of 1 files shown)


wait/http.go
```diff
@@ -217,9 +217,11 @@ func (ws *HTTPStrategy) WaitUntilReady(ctx context.Context, target StrategyTarge
 				continue
 			}
 			if ws.StatusCodeMatcher != nil && !ws.StatusCodeMatcher(resp.StatusCode) {
+				resp.Body.Close()
```
mdelapenya, line 220: I think we should handle the error here too, as in [L227](https://github.com/testcontainers/testcontainers-go/pull/718/files#diff-0f5d7fe7dce1248ede0b1d18ae1c87cb6d3690188f5741191eb6de1077866d8cR227)

mdelapenya (changes requested):
HI @kruskall , thanks for this fix, where I only see that we are not handling the potential error after closing the Body. Could you add that? 🙏

wait/http.go
kruskall: thank you for the feedback! 

the error is used to continue to the next iteration, we're already calling `continue` here. I think I'm missing something

mdelapenya: I know that golangci-lint checks for non-handled errors. So I'd say that we need to handle them in case the Close call fails

kruskall: > I know that golangci-lint checks for non-handled errors. 

I'm not able to reproduce this. I'm running golangci-lint v1.50.1 and I can see `errcheck` is enabled but nothing is being reported. 
I've looked at the codebase and it seems there are other places where the error is being assigned to an empty var. I've pushed some changes to do the same here. Let me know what you think :ping_pong:

mdelapenya: Sorry for the 🏓 (please see https://github.com/testcontainers/testcontainers-go/discussions/653)

if golangci does not complain, I'm good with it.

kruskall: No worries, congratulations! :smile:

mdelapenya (approved):

Issue #636, opened 2022-11-23
[Modules]: add examples/modules on the most uses services/images


mdelapenya:
### Proposal

Given there is a tool to generate the scaffolding for new examples as Go modules (https://golang.testcontainers.org/examples), let's use it to start adding examples.

We have elaborated the list below from Testcontainers for Java. If you're interested in adding any example, please comment here the module you want to work on (we'll add your username to next to it in the list), use the tool to generate the scaffolding for the module you are interested in, and send a PR. We'll link your PR here next to the item in the list.

Our recommendation is to visit the Java module and start porting its behavior and settings to the Go example module.

## Modules
- [ ] azure cosmosdb
- [x] cassandra @anilsenay #1726
- [x] clickhouse @anilsenay https://github.com/testcontainers/testcontainers-go/pull/1372
- [x] cockroachdb @mdelapenya #618
- [x] consul @bo-er https://github.com/testcontainers/testcontainers-go/pull/722 @Chris-AS1 https://github.com/testcontainers/testcontainers-go/pull/2194
- [x] @alihanyalcin #932 https://github.com/testcontainers/testcontainers-go/pull/876
- [ ] db2
- [ ] dragonflydb
- [ ] dynalite
- [x] elasticsearch @mdelapenya https://github.com/testcontainers/testcontainers-go/pull/1574
- [x] gcloud
  - [x] datastore @eddumelendez https://github.com/testcontainers/testcontainers-go/pull/638
  - [x] firestore @eddumelendez https://github.com/testcontainers/testcontainers-go/pull/637
  - [x] pubsub @eddumelendez https://github.com/testcontainers/testcontainers-go/pull/640
  - [x] spanner @eddumelendez https://github.com/testcontainers/testcontainers-go/pull/642
- [ ] hivemq
- [x] inbucket @jespino #2142
- [x] influxdb
- [x] k3s @mritunjaysharma394 #1194 
- [x] k6 @pablochacin https://github.com/testcontainers/testcontainers-go/pull/1721
- [x] kafka @noctarius https://github.com/testcontainers/testcontainers-go/pull/1610
- [x] localstack @mdelapenya #795
- [x] mariadb @mdelapenya #1548
- [x] minio @adoublef  https://github.com/testcontainers/testcontainers-go/pull/2132
- [x] mockserver @jespino https://github.com/testcontainers/testcontainers-go/pull/2085
- [x] mongodb @ravilushqa #723 and #726
[...]

PR #723, opened 2023-01-04
MongoDB example

ravilushqa:
## What does this PR do?

Added example for MongoDB

## Why is it important?

MongoDB doesn't have example yet

## Related issues

- Relates https://github.com/testcontainers/testcontainers-go/issues/636


.github/workflows/mongodb-example.yml
```diff
@@ -1,4 +1,4 @@
-name: Mongodb example pipeline
+name: MongoDB example pipeline
```
mdelapenya, line 1: This is a good point, as the generator does not distinguish this notation including initials. I'll take a note to include a way to provide a variant that overrides the titled names

examples/mongodb/mongodb.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/testcontainers/testcontainers-go/wait"
+)
+
+// mongodbContainer represents the mongodb container type used in the module
+type mongodbContainer struct {
+	testcontainers.Container
+}
+
+// setupMongoDB creates an instance of the mongodb container type
+func setupMongoDB(ctx context.Context) (*mongodbContainer, error) {
```
mdelapenya, line 16: Were you able to take a look at the [Java implementation](https://github.com/testcontainers/testcontainers-java/blob/main/modules/mongodb/src/main/java/org/testcontainers/containers/MongoDBContainer.java), or do you think this simple example is enough?

mdelapenya: In any case, this LGTM. I'll merge it as is and we can iterate and evolve the example in follow ups, if you are interested

mdelapenya (approved):
Thanks for your time adding this example! Much appreciated :)

examples/mongodb/mongodb.go
ravilushqa: I have a look into Java implementation, looks like it's supports replica set that looks like for me as an option. I doesn't need it, an I believe if someone will need it, I believe that it could be added as an option, as container option.
From my perspective it is good enough, as it just works and doesn't require overcomplicating it at this stage.

If we will add these rows below test, it will work, and this is most common usecase for tests, just perform operations
```go
res, err := mongoClient.Database("test").Collection("test").InsertOne(ctx, bson.M{"test": "test"})
if err != nil {
	t.Fatal(fmt.Errorf("error inserting document into mongo: %w", err))
}

if res.InsertedID == nil {
	t.Fatal("expected InsertedID to not be nil")
}
```

wdyt?

ravilushqa:
thank you for maintaining this great project and rapid review ❤️

examples/mongodb/mongodb.go
mdelapenya: Sounds good to me 👏

PR #726, opened 2023-01-04
simplify mongodb test

ravilushqa:
## What does this PR do?

cleaner example of using mongo test

## Related issues

-


ravilushqa:
I'm just started investigating your library, and I've found that code from my previous PR could be a cleaner 😄

mdelapenya (approved):
LGTM

mdelapenya:
> also can you please release new version so I will be able to use it 🙏

Because the example modules are not exported modules, you can simply copy&paste the module code in your own project, so no need to release it

Issue #764, opened 2023-01-17
[Bug]: Ryuk reaper container not properly re-initialized after it got terminated


mdonkers:
### Testcontainers version

0.15.0

### Using the latest Testcontainers version?

Yes

### Host OS

Linux

### Host arch

x86

### Go version

1.19

### Docker version

```shell
$ docker version
Client: Docker Engine - Community
 Version:           23.0.0-rc.1
 API version:       1.42
 Go version:        go1.19.4
 Git commit:        139e924
 Built:             Thu Dec 22 23:37:13 2022
 OS/Arch:           linux/amd64
 Context:           default

Server: Docker Engine - Community
 Engine:
  Version:          23.0.0-rc.1
  API version:      1.42 (minimum version 1.12)
  Go version:       go1.19.4
  Git commit:       cba986b
  Built:            Thu Dec 22 23:37:13 2022
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.14
  GitCommit:        9ba4b250366a5ddde94bb7c9d1def331423aa323
 runc:
  Version:          1.1.4
  GitCommit:        v1.1.4-0-g5fd4c4d
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### Docker info

```shell
$ docker info
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.9.1
    Path:     /home/miel/.docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.14.2
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 2
  Running: 0
  Paused: 0
  Stopped: 2
 Images: 25
 Server Version: 23.0.0-rc.1
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
[...]

PR #782, opened 2023-01-25
Verify Reaper state to create new or return existing instance

mdonkers:
## What does this PR do?

Resolves https://github.com/testcontainers/testcontainers-go/issues/764 and also makes https://github.com/testcontainers/testcontainers-go/pull/258 (PR) obsolete.

As the issue describes, the Reaper container might exit if there's no connection for a longer time. For example when running multiple integration tests, some not using TestContainers. When that happens the next test will fail as Reaper is created as singleton but no connection can be created.

## Why is it important?

Have reliable tests

## Related issues

- Closes https://github.com/testcontainers/testcontainers-go/issues/764
- Supersedes https://github.com/testcontainers/testcontainers-go/pull/258


mdonkers:
@mdelapenya this is the PR for better Reaper consistency.

I opted to use the `Container.State()` for checking if running, instead of using a connection, because it seems nicer using Container functions. The caveat is in some scenarios, there is a very slight chance the container exits between checking its state and running the `Connect()` function.

mdonkers:
@mdelapenya it seems like this PR is awaiting approval to run tests (again?). Maybe because I fixed a linting error and then rebased and pushed again?

reaper_test.go
```diff
@@ -193,3 +185,31 @@ func Test_ReaperForNetwork(t *testing.T) {
 	assert.Equal(t, "reaperImage", provider.req.Image)
 	assert.Equal(t, "reaperImage", provider.req.ReaperImage)
 }
+
+func Test_ReaperReusedIfHealthy(t *testing.T) {
+	if testing.Short() {
+		t.Skip("Skipping integration test")
+	}
```
mdelapenya, line 192: Do we need this separation of concerns, short tests Vs non-short tests?

reaper.go
```diff
@@ -28,7 +28,7 @@ type reaperContextKey string
 
 var (
 	dockerHostContextKey = reaperContextKey("docker_host")
-	reaper               *Reaper // We would like to create reaper only once
+	reaperSingleton      *Reaper // We would like to create reaper only once
```
mdelapenya, line 31: Wdyt if we call this variable as `reaperInstance`. I know that singleton is a well-known word in the industry, but in the comments we always use the instance word: https://github.com/testcontainers/testcontainers-go/pull/782/files#diff-e8ab4ec17e978db949cd15726ba0efa9c89fd714e7e6e0aa537ed7fc1bbde7fdR48

reaper_test.go
mdonkers: Not sure, I saw it being used in other places as well. I think if you run the Makefile `test-all` target, they don't run.
But happy to remove it as well.

reaper.go
mdonkers: sure, will rename

reaper_test.go
mdelapenya: Mmmm I see them being used in the example modules for nginx, redis and cocroachdb 🤔 Probably worth it to remove them all in a follow-up PR

mdonkers: As all checks are succeeding, please let me know whether to remove this now, or you will handle it in a separate PR.
Because then I believe this PR is good to merge?

mdelapenya: Let's remove this here as it's completely new, and the rest in a follow up PR 🙏 

Once there, I think we are ready to go

mdelapenya:
Sorry for the merge conflicts 😞  , we needed to extract the calculation of the docker host to an internal package as part of certain refactors (#796). Do you want me to resolve them?

mdonkers:
> Sorry for the merge conflicts disappointed , we needed to extract the calculation of the docker host to an internal package as part of certain refactors (#796). Do you want me to resolve them?

Conflicts resolved. Wasn't that bad luckily :-)

reaper.go
```diff
@@ -38,22 +38,39 @@ type ReaperProvider interface {
...
+func reuseOrCreateReaper(ctx context.Context, sessionID string, provider ReaperProvider, opts ...ContainerOption) (*Reaper, error) {
 	mutex.Lock()
 	defer mutex.Unlock()
-	// If reaper already exists re-use it
-	if reaper != nil {
-		return reaper, nil
+	// If reaper already exists and healthy, re-use it
+	if reaperInstance != nil {
+		// Verify this instance is still running by checking state.
+		// Can't use Container.IsRunning because the bool is not updated when Reaper is terminated
```
mdelapenya, line 52: Meanwhile the CI runs (tests will pass) can you elaborate a bit on this comment? Do you think that bool should be updated in consequence (probably here? in a follow-up?

mdelapenya (approved):
Shaun approves 😎 

image

[This comment](https://github.com/testcontainers/testcontainers-go/pull/782/files#r1091665195) does not affect the review, so will eventually merge this one today

Thank you very much for your support!

mdelapenya:
Mmm I've checkout out the PR locally to verify one weird behaviour that I'm seeing in another issue, which happens when there are multiple packages in the Go project running the tests. How go test works is that it runs each package separately, creating as many Ryuk containers as packages in the execution. I've verified that even with this PR, there are multiple Ryuk containers. Let me do some verifications on accessing the mutex and will let you know asap

reaper.go
mdonkers: So the `Container.IsRunning` is simply a boolean flag (https://github.com/testcontainers/testcontainers-go/blob/fb646713ffeaede8472fb1dd28dce495c6f3ab6f/docker.go#L63) that gets updated when a container is stopped / terminated (https://github.com/testcontainers/testcontainers-go/blob/fb646713ffeaede8472fb1dd28dce495c6f3ab6f/docker.go#L233). And while the Reaper is a 'regular' container, it is never terminated via the `Stop()` or `Terminate()` methods. Instead, it terminates automatically by itself by closing the connection to it (https://github.com/testcontainers/testcontainers-go/blob/fb646713ffeaede8472fb1dd28dce495c6f3ab6f/docker.go#L241).
Which makes the `IsRunning()` unreliable for the Reaper container as it would never return `false`.

That's why I chose to poll Docker and get the state from there, as it should be reliable.

What I think would be better, is if the `IsRunning()` method (https://github.com/testcontainers/testcontainers-go/blob/fb646713ffeaede8472fb1dd28dce495c6f3ab6f/docker.go#L89) gets updated to what I'm doing below; checking from Docker if the state is running or not. That would make it reliable and not depend on a boolean flag internally updated.
I did not want to touch this code in the PR because I don't know the history behind it. Perhaps `IsRunning` is regularly checked and would be relatively heavy on the Docker calls?

mdelapenya: I see, thanks for such a clear explanation, now I see what you mean.

I think checking the state from Docker would be more accurate than the one we have now, and I understand your concerns about performance if that method is called many times. I've checked their usages and are just a few of them at this moment (particularly worried on the GenericContainer function calling it).

Adding that code, we could also take the situation to a place where multiple goroutines in tests are trying to read the container state at the same time, and one decides to see it as not running, and the next one as running, which could be worse.

I'd keep this simple as in your submission, and evaluate if we need to check for the state more frequently.

mdelapenya:
> I've verified that even with this PR, there are multiple Ryuk containers. Let me do some verifications on accessing the mutex and will let you know asap

This is the expected behaviour of `go test ./...`: creates a test binary for each package, therefore no synchronisation across them will happen and, effectively, as many reaper instances will be running for each package in the test execution.

Said that, if you agree, we can merge this one, as it covers the scenario of running just one reaper instance per test binary.

mdonkers:
All good from my side. Would be happy to have it merged.

emetsger:
> Mmm I've checkout out the PR locally to verify one weird behaviour that I'm seeing in another issue, which happens when there are multiple packages in the Go project running the tests.

@mdelapenya I believe this is because the reaper container can be exposed by the Docker API before the reaper is ready to accept connections.

I've opened the following PRs to address this:
1. Add a HEALTHCHECK to the reaper Docker image (https://github.com/testcontainers/moby-ryuk/pull/128)
2. Update testcontainers-go to consider the health status of the reaper container before returning it (https://github.com/testcontainers/testcontainers-go/pull/2508)

PR #794, opened 2023-01-30
chore: run "go mod tidy" automatically when creating examples

mdelapenya:
- chore: do not generate go.sum in examples
- chore: print success messages in main function
- chore: run go mod tidy automatically

## What does this PR do?
It removes the go.sum template as this PR is automating the execution of the "go mod tidy" command in the directory for the new example.

For that we leverage the exec package of stdlib, setting its directory to the new example dir. As a result, we are extracting the success message from the generate function, and instead moving it to the main one.

The success message has been changed in consequence

## Why is it important?
Do even less extra work when generating an example!


examples/main.go
```diff
@@ -110,6 +112,19 @@ func main() {
 		fmt.Printf(">> error generating the example: %v\n", err)
 		os.Exit(1)
 	}
+
+	log.Printf("Running command and waiting for it to finish...")
```
mdelapenya, line 116:
```suggestion
```

eddumelendez (approved):

PR #799, opened 2023-01-31
chore: restructure Docker helper methods

mdelapenya:
- chore: add InAContainer
- chore: use new internal method for inAContainer
- chore: extract default gateway calculation to the internal package
- chore: move sessionID to the internal packages
- chore: simplify getting the string from the session ID
- chore: expose DaemonHost in the Docker provider

## What does this PR do?
this PR leverages the new internal `testcontainersdocker` package to move there a few utility functions used to check certain state in Docker, such as
- default gateway
- check if it's running in a container
- retrieve the session ID

We were not able to move the calculation of the daemonHost to the internal package because it's coupled with the DockerProvider struct, and moving the function to the internal would have caused a package cycle.
Therefore we are simply exposing the DaemonHost function in the DockerProvider struct, so that it can be consumed from anybody being able to obtain a DockerProvider instance using the NewDockerProvider function.
This new exposed function will call the already existing one, which was already private and easy to rename/move.

## Why is it important?
These changes come with two major improvements:
- refactor of the code, moving internal parts of the library to the internal packages
- add visibility to the upcoming example modules to certain functions that calculate the DaemonHost. As an example, the localstack module, which needs it to prepare the EnvVars to be used to run the container.

## Related issues

- Required by #795


internal/testcontainersdocker/docker_host.go
```diff
@@ -38,3 +57,17 @@ func ExtractDockerHost(ctx context.Context) (dockerHostPath string) {
...
 	}
 }
+
+// InAContainer returns true if the code is running inside a container
+// See https://github.com/docker/docker/blob/a9fa38b1edf30b23cae3eade0be48b3d4b1de14b/daemon/initlayer/setup_unix.go#L25
+func InAContainer() bool {
+	return inAContainer("/.dockerenv")
+}
+
+func inAContainer(path string) bool {
```
mdelapenya, line 67: Making it testable

eddumelendez (approved):

PR #800, opened 2023-02-02
feat: add localstack module

### LINKED ISSUE #795: Create localstack example module

mdelapenya:
- chore: add localstack basic example
- chore: add wait for log
- chore: add Docker socket binding
- chore: support defining legacy mode
- chore: support passing services and configuring them
- chore: simplify the creation of a localstack container request
- chore: support configuring the AWS region
- chore: expose localstack container
- chore: support for retrieving the endpoint of a given service
- chore: store enabled services in the localstack container
- chore: store the region in the localstack container
- chore: apply default region when needed
- chore: support passing version and legacyMode as functional options
- chore: support for overriding the container request with a functional option
- chore: remove useless method for retrieving the internal endpoint of a service
- chore: simplify unit tests
- chore: create aws session using v1
- chore: add S3 tests
- chore: rename start localstack function
- chore: define default functions as vars
- chore: add tests for legacy mode

## What does this PR do?
It adds an example module for localstack, but instead of keeping it closed, it exposes the functions that are needed to work with it.

The module include some tests that have been borrowed from the Java implementation, although not them all. In particular, a few for the S3 service.

## Why is it important?
It will serve as the first example to be promoted to module, exposing an API to interact with localstack.

## Related issues

- Closes #795


examples/localstack/localstack.go
```diff
@@ -0,0 +1,206 @@
...
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/session"
```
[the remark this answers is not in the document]
mdelapenya: Got it, let's use it only in the test code

I thought it could be handy to provide a way to get the AWS session

examples/localstack/go.mod
```diff
@@ -0,0 +1,61 @@
+module github.com/testcontainers/testcontainers-go/examples/localstack
+
+go 1.18
+
+require (
+	github.com/aws/aws-sdk-go v1.44.192
```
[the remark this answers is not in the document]
mdelapenya: I started with tests (examples) using v1 as they were the simplest ones that I found in the localstack's examples repo (https://github.com/localstack/localstack-aws-sdk-examples/tree/main/go).

Once we are happy we could eventually merge this one. We can then add more tests, or even look for contributors for the module to help us in maintaining it

examples/localstack/localstack.go
```diff
@@ -0,0 +1,190 @@
...
+// StartContainer creates an instance of the LocalStack container type
+func StartContainer(ctx context.Context, overrideReq overrideContainerRequestOption, opts ...localStackContainerOption) (*LocalStackContainer, error) {
+	credentials := Credentials{
+		AccessKeyID:     defaultAccessKeyID,
+		SecretAccessKey: defaultSecretAccessKey,
+		Token:           defaultToken,
+	}
+
+	req := testcontainers.ContainerRequest{
+		Image:      "localstack/localstack",
```
eddumelendez, line 82: what if there is a custom image? As part of the example this is good but thinking about modules this will not allow to use custom images, not only those hosted in a registry but what if my image is called `my-localstack`.

Also, I don't see a test which can override this image with, for example, `localstack/localstack:0.13.0`

examples/localstack/types.go
```diff
@@ -0,0 +1,306 @@
...
+		for port := range servicesPorts {
+			exposedPorts = append(exposedPorts, fmt.Sprintf("%d/tcp", port))
+		}
+
+		req.ExposedPorts = exposedPorts
+	}
+}
+
+// WithDefaultVersion uses the default version for the container, which is "0.11"
+var WithDefaultVersion = WithVersion(defaultVersion)
```
eddumelendez, line 276: IMO, given this is part of what is going to become a module, we should avoid providing default versions in the api. This is something deprecated in tc-java.

eddumelendez (commented):
just leaving two comments which I would like to discuss before going further

examples/localstack/localstack.go
mdelapenya: Because it's possible to pass a `OverrideContainerRequest` optional func, where you can customise as much as you need from the original container request, I did not add a explicit optional func to change it. Please take a look at the unit tests for that func type:

```go
func TestWithContainerRequest(t *testing.T) {
		Env:          map[string]string{},
		Image:        "foo",
		ExposedPorts: []string{},
		SkipReaper:   false,
		WaitingFor: wait.ForNop(
			func(ctx context.Context, target wait.StrategyTarget) error {
				return nil
			},
		),
		Networks: []string{"foo", "bar", "baaz"},
		NetworkAliases: map[string][]string{
			"foo": {"foo0", "foo1", "foo2", "foo3"},
		},
	}

	merged := OverrideContainerRequest(testcontainers.ContainerRequest{
		Env: map[string]string{
			"FOO": "BAR",
		},
		Image:    "bar",
		Networks: []string{"foo1", "bar1"},
		NetworkAliases: map[string][]string{
			"foo1": {"bar"},
		},
		WaitingFor: wait.ForLog("foo"),
		SkipReaper: true,
	})(req)

	assert.Equal(t, "BAR", merged.Env["FOO"])
	assert.True(t, merged.SkipReaper)
	assert.Equal(t, "bar", merged.Image)
	assert.Equal(t, []string{"foo1", "bar1"}, merged.Networks)
	assert.Equal(t, []string{"foo0", "foo1", "foo2", "foo3"}, merged.NetworkAliases["foo"])
	assert.Equal(t, []string{"bar"}, merged.NetworkAliases["foo1"])
	assert.Equal(t, wait.ForLog("foo"), merged.WaitingFor)
}
```

Therefore, because passing the `OverrideContainerRequest` functional opt to the `StartContainer` function, in which the functional opt receive a way to change the image, then we are good to go.

A different story would be that we want to explicitly make certain functions public, like `WithImage`.

examples/localstack/types.go
mdelapenya: You mean making it internal, and internally check if a version was not provided to apply the default?

examples/localstack/localstack.go
```diff
@@ -0,0 +1,190 @@
...
+
+	p, err := nat.NewPort("tcp", fmt.Sprintf("%d", internalPort))
+	if err != nil {
+		return "", err
+	}
+
+	return l.MappedPort(ctx, p)
+}
+
+func runInLegacyMode(version string) bool {
```
mdelapenya, line 55: As discussed online, we should remove the support for legacy mode: it's a new module and we would like to define the minimal version that is supported, so no legacy mode should be possible.

It will simplify the code significantly

mdelapenya: Another option would be to keep it, and create an issue for future deprecation of the legacy mode, and notify consumers with semver

examples/localstack/localstack.go
```diff
@@ -0,0 +1,190 @@
...
+	c := &LocalStackContainer{
+		Container:       container,
+		Credentials:     credentials,
+		EnabledServices: enabledServices,
+		Region:          localStackReq.region,
+	}
+	return c, nil
+}
+
+func configure(req *LocalStackContainerRequest) (reason string, err error) {
```
mdelapenya, line 155: configureHostname instead?

docs/examples/localstack.md
```diff
@@ -0,0 +1,142 @@
...
+- Redshift
+- Route53
+- S3
+- SES
+- SNS
+- SQS
+- SSM
+- STS
+- SecretsManager
+- StepFunctions
```
eddumelendez, line 137: we should point to the docs instead of listing all services available by LocalStack. Otherwise, list need to be revisited to keep it up-to-date.

examples/localstack/localstack.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	// finally we apply the options
+	for _, opt := range opts {
+		opt(&localStackReq)
+	}
+
+	/*
+		Do not run in legacy mode when the version is a valid semver version greater than the v0.11 and legacyMode is false
+			| runInLegacyMode | legacyMode | result |
```
eddumelendez, line 80:
```suggestion
			| version < 0.11 | legacyMode | result |
```

examples/localstack/localstack.go
```diff
@@ -0,0 +1,150 @@
...
+	if !strings.HasPrefix(version, "v") {
+		version = fmt.Sprintf("v%s", version)
+	}
+
+	if semver.IsValid(version) {
+		return semver.Compare(version, "v0.11") < 0 // version < v0.11
+	}
+
+	fmt.Printf("Version %s is not a semantic version, LocalStack will run in legacy mode.\n", version)
+	fmt.Printf("Consider using \"StartContainer(ctx, NoopOverrideContainerRequest, WithLegacyMode)\" constructor if you want to enable legacy mode.")
```
eddumelendez, line 41: I think those message are confusing a little bit. In line 37, my understanding is if version is a valid version and less than v0.11 then it will run in legacy mode. So, the print messages should be 

```suggestion
	fmt.Printf("Version %s is not a semantic version, LocalStack will not run in legacy mode.\n", version)
	fmt.Printf("Consider using \"StartContainer(ctx, NoopOverrideContainerRequest, WithLegacyMode)\" constructor if you want to enable legacy mode.")
```

Please, correct me if I am wrong

docs/examples/localstack.md
mdelapenya: Are all the services available in the Go/Java code? If not, we'll have to visit the code to update it when a new service comes. But in any case, it's safer to point to the docs. I'll change that ASAP

examples/localstack/localstack.go
mdelapenya: I took them from Java: https://github.com/testcontainers/testcontainers-java/blob/main/modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java#L112

.github/workflows/localstack-example.yml
```diff
@@ -0,0 +1,46 @@
+name: LocalStack example pipeline
```
eddumelendez, line 1: Also, rename the file

```suggestion
name: LocalStack Module pipeline
```

docs/examples/localstack.md
```diff
@@ -0,0 +1,134 @@
+# LocalStack
```
eddumelendez, line 1: file should be under modules?

modules/localstack/localstack.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	// finally we apply the options
+	for _, opt := range opts {
+		opt(&localStackReq)
+	}
+
+	if isLegacyMode(localStackReq.Image) {
+		return nil, fmt.Errorf("version=%s. Testcontainers for Go does not support running LocalStack in legacy mode. Please use a version >= 0.11.0", localStackReq.Image)
+	}
```
eddumelendez, line 78: If for some reason want to use an older version it's only about to override the ContainerRequest with the right ports which in this case should be manually instead.

modules/localstack/localstack_legacy_mode_test.go
```diff
@@ -0,0 +1,29 @@
+package localstack
```
eddumelendez, line 1: should be removed?

modules/localstack/types.go
```diff
@@ -0,0 +1,265 @@
...
+func (s Service) servicePort() int {
+	return defaultPort
+}
+
+// List of AWS services, ordered alphabetically
+
+// APIGateway is the API Gateway service
+var APIGateway = Service{
+	name: "apigateway",
+	port: 4567,
```
eddumelendez, line 80: if legacy mode is not supported then port can be remove it. If want to use it can override the ContainerRequest adding this port manually.

docs/examples/localstack.md
mdelapenya: I was tempted to do it, but then the changes would be much bigger. I'm working in a PR that separate modules and examples at the generator level

modules/localstack/localstack_legacy_mode_test.go
mdelapenya: Nope, because we want to check that it will raise the error

modules/localstack/localstack.go
```diff
@@ -0,0 +1,141 @@
...
+
+	"github.com/testcontainers/testcontainers-go"
+	"github.com/testcontainers/testcontainers-go/internal/testcontainersdocker"
+	"github.com/testcontainers/testcontainers-go/wait"
+	"golang.org/x/mod/semver"
+)
+
+const defaultPort = 4566
+const defaultRegion = "us-east-1"
+const defaultVersion = "0.11.2"
```
thrau, line 17: Hey! This is a pretty old LocalStack version. It would be awesome to upgrade this to something post v1.0. we are releasing 1.4 next week :)

mdelapenya: @thrau I'm bumping it on https://github.com/testcontainers/testcontainers-go/pull/800/commits/bb4e2a87a165331588948c2026393f0065a976a8 to 1.3.1. Thanks for the suggestion!

modules/localstack/localstack.go
mdelapenya: Indeed, if we want to support legacy mode, then it should not raise an error. My question is: do we want to provide backwards compatibility with legacy mode? If so, I'd not remove the ports from the Service struct, in order to make it easier for developers to switch to legacy.

mdelapenya: A valid question would also be: do we want to support legacy mode in this brand new module?

modules/localstack/localstack.go
mdelapenya: We updated to 1.4.0!

modules/localstack/types.go
mdelapenya: We removed everything regarding Services

modules/localstack/localstack.go
mdelapenya: We agreed in not supporting legacy mode

docs/examples/localstack.md
mdelapenya: I will reorganise the docs in a follow-up PR, as it needs to update the code generation tool

.github/workflows/localstack-example.yml
mdelapenya: Done

Issue #811, opened 2023-02-07
[Bug]: Concurrent map writes in dockerCompose.Up


joshua-hill-form3:
### Testcontainers version

0.17.0

### Using the latest Testcontainers version?

Yes

### Host OS

Darwin

### Host arch

amd64

### Go version

1.18.10

### Docker version

```shell
not relevant
```

### Docker info

```shell
not relevant
```

### What happened?

Given a test set up that uses the Docker Compose module
And there are multiple wait strategies
When Up is called
Then a panic occurred with output `fatal error: concurrent map writes`

### Relevant log output

```shell
fatal error: concurrent map writes
fatal error: concurrent map writes

goroutine 167 [running]:
runtime.throw({0x29dfd2f?, 0x70?})
	/opt/actions-runner/_work/_tool/go/1.18.10/x64/src/runtime/panic.go:992 +0x71 fp=0xc0005d3c78 sp=0xc0005d3c48 pc=0x43aa91
runtime.mapassign_faststr(0x2582120, 0xc000a35e60, {0x29b9e06, 0x5})
	/opt/actions-runner/_work/_tool/go/1.18.10/x64/src/runtime/map_faststr.go:295 +0x38b fp=0xc0005d3ce0 sp=0xc0005d3c78 pc=0x4158ab
github.com/testcontainers/testcontainers-go/modules/compose.(*dockerCompose).lookupContainer(0xc00093b8c0, {0x2f88698, 0xc000c11840}, {0x29b9e06, 0x5})
	/opt/actions-runner/_work/REDACTED/REDACTED/vendor/github.com/testcontainers/testcontainers-go/modules/compose/compose_api.go:300 +0x4ea fp=0xc0005d3f20 sp=0xc0005d3ce0 pc=0x1c7766a
github.com/testcontainers/testcontainers-go/modules/compose.(*dockerCompose).Up.func1()
	/opt/actions-runner/_work/REDACTED/REDACTED/vendor/github.com/testcontainers/testcontainers-go/modules/compose/compose_api.go:233 +0x4d fp=0xc0005d3f78 sp=0xc0005d3f20 pc=0x1c76b4d
golang.org/x/sync/errgroup.(*Group).Go.func1()
	/opt/actions-runner/_work/REDACTED/REDACTED/vendor/golang.org/x/sync/errgroup/errgroup.go:75 +0x64 fp=0xc0005d3fe0 sp=0xc0005d3f78 pc=0xb51644
runtime.goexit()
[...]

PR #812, opened 2023-02-07
fix: Synchronise writes to dockerCompose.containers map

joshua-hill-form3:
## What does this PR do?

Prevent concurrent writes to the `dockerCompose.containers` map using a mutex.

## Why is it important?

To avoid occasional panics when running tests that use Docker Compose with multiple wait strategies.

## Related issues

- Fixes https://github.com/testcontainers/testcontainers-go/issues/811


mdelapenya (approved):
Thanks for the big report and the quick fix! Much appreciated

LGTM

PR #813, opened 2023-02-09
chore: update Docker labels for containers

mdelapenya:
## What does this PR do?
This PR adds a Go file in the internal package with the constants representing the Docker labels that will be applied to all the containers produced by this library.

We are deprecating the old ones, as they were part of the public API, and using the new ones internally (production and test code). We are keeping the usage of the old labels for backwards compatibility in the case any of our users was using the old labels.

For the version of the library, which we do not want to expose at this moment, it lives in the internal package.

## Why is it important?
It will allow us to identify containers created by testcontainers for Go in a more consistent manner.

### DIFF
--- a/docker_test.go
+++ b/docker_test.go
@@ -30,6 +30,7 @@ import (
 	"github.com/docker/docker/api/types/filters"
 	"github.com/docker/docker/client"
 
+	"github.com/testcontainers/testcontainers-go/internal/testcontainersdocker"
 	"github.com/testcontainers/testcontainers-go/wait"
 )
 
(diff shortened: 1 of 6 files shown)


HofmeisterAn (approved):

docker_test.go
```diff
@@ -378,7 +379,7 @@ func TestContainerStartsWithoutTheReaper(t *testing.T) {
 	terminateContainerOnEnd(t, ctx, container)
 
 	resp, err := client.ContainerList(ctx, types.ContainerListOptions{
-		Filters: filters.NewArgs(filters.Arg("label", fmt.Sprintf("%s=%s", TestcontainerLabelSessionID, container.SessionID()))),
+		Filters: filters.NewArgs(filters.Arg("label", fmt.Sprintf("%s=%s", testcontainersdocker.LabelSessionID, container.SessionID()))),
```
mdelapenya, line 382: I'm thinking about this 🤔 for backwards compatibility, we should keep the old label in the filter

mdelapenya: Never mind, this is for tests only

Issue #532, opened 2022-09-16
[Feature]: RegistryCred multiple private registry credentials support


szaher:
### Problem

At the moment testcontainers supports providing registry credentials via `RegistryCred` in `ContainerRequest` but this is problematic when using multiple private docker registries specially with multi-stage docker files. 

```

FROM ACCOUNT1.dkr.ecr.us-east-1.amazonaws.com/myimages/app:v1.0
FROM ACCOUNT2.dkr.ecr.us-west-2.amazonaws.com/mirror/alpine:4.0.0 as curl

RUN apk add --no-cache --update curl

FROM curl as curl_container
ENV GRPC_HEALTH_PROBE_VERSION v0.2.2
RUN curl -fsSL https://github.com/grpc-ecosystem/grpc-health-probe/releases/download/${GRPC_HEALTH_PROBE_VERSION}/grpc_health_probe-linux-amd64 -o /grpc_health_probe \
  && chmod +x /grpc_health_probe


```
### Solution

RegistryCred parameter's type in the container request can be map[string]map[string]string 

```
var registryCreds map[string]map[string]string
registryCreds["ACCOUNT1.dkr.ecr.us-east-1.amazonaws.com"] = map[string]string{"username": "Saad", "password": "123"}
```
or 

```
import types
func GetRegistryCreds() {
	var registryCreds map[string]types.AuthConfig

	auth1 := types.AuthConfig{
		Username: "Saad",
		Password: "2407",
	}
	registryCreds["ACCOUNT1.dkr.ecr.us-east-1.amazonaws.com"] = auth1
	
	auth2 := types.AuthConfig{
		Username:      "User2",
		Password:      "1808",
	}
	registryCreds["ACCOUNT2.dkr.ecr.us-west-2.amazonaws.com"] = auth2
}


```
### Benefit

It enables support for multiple private registries 

### Alternatives

user must have all his/her images on private repo

### Would you like to help contributing this feature?

Yes

mdelapenya:
@szaher do you think that the work in https://github.com/testcontainers/testcontainers-go/pull/869 can close this issue? Now that the docker auth is transparently obtained from the credential helpers, it's totally possible to have multiple credentials. Wdyt?

mdelapenya:
Given the docker auth credentials are now obtained from the helpers, I'd say it's fine to close this one.

Please reopen it if you consider it's not fixed. Thanks in advance 🙇

Issue #324, opened 2021-05-21
How to control user account that launches images. 


naveenchlsn:
We are seeing test containers throttled by dockerhub while downloading images. We have a docker hub pro account. Hence we were not expecting to get throttled by the docker hub. 

Which user does the test containers launch the dockers with? 
Is it possible to change it? 

Thanks,
Naveen

gianarb:
Technically testcontains is just a wrapper around the docker/client. It depends on how you run tests. On its own, it relays on the process who starts the test suite.

I am wondering if we have to do something that we do not do to authenticate the docker/client 😕

mdelapenya:
@naveenchlsn does https://github.com/testcontainers/testcontainers-go/issues/324#issuecomment-849426711 resolve your use case?

mdelapenya:
@naveenchlsn I'm closing this issue as there is an open PR (#869) that provides a way to access the docker registry using the credential helpers.

Please reopen it if you do not see it resolved, thanks in advance for your support!

PR #869, opened 2023-02-21
feat: support for reading auth credentials from docker credential helpers

### LINKED ISSUE #609: Support for authentication when pulling Docker images everywhere
The project is missing the same mechanism as the one added in #602 for pulling images from private registries in other places.

I wonder if anybody could be interested in contributing the others, such as the Docker Compose code. I can be of any help assisting you identifying where to add it.

mdelapenya:
## What does this PR do?
This PR exposes a new method:
- AuthFromDockerConfig(registry string): retrieves the auth config type for a registry, obtained from the docker config

To implement it, we will internally check, in this particular order:

1. the `DOCKER_AUTH_CONFIG` environment variable, unmarshalling the string value from its JSON representation and using it as the Docker config.
2. the `DOCKER_CONFIG` environment variable, as an alternative path to the Docker config file.
3. else it will load the default Docker config file, which lives in the user's home, e.g. `~/.docker/config.json`
4. it will use the right Docker credential helper to retrieve the authentication (user, password and base64 representation) for the given registry.

For that, we are using https://github.com/cpuguy83/dockercfg, which already handles the location of the Docker config file using the `DOCKER_CONFIG` environment variable.

Finally, now that we are able to read the Docker Auth configs, we are **deprecating** all previous manners to add them to the container definition (the container request, the buildFromDockerfile struct, and the reaper). Instead of using those fields, we are going to use the aforementioned extraction of the Docker Auth configs to populate them when needed.

In the case of building from a Dockerfile including a FROM clause consuming an image living in a private registry, the code will parse the Dockerfile alongside the build args, to extract all Docker images in it, trying to locate the Docker auth configs for each image. The build args will be interpolated in the case the image contains them.

Docs page has been updated with how Testcontainers for Go will autodiscover the Docker auth configs.

## Why is it important?
It will allow consumers of the library to discover the auth credentials when working with images living in private registries.

We decided to provide that capability instead of automatically discovering it, but I'd not be opposed to do it if desired.

## Related issues

- Closes #609
- Relates #532
- Relates #324


mdelapenya:
@HofmeisterAn came with the idea of reading `DOCKER_AUTH_CONFIG` as a possible input for the JSON value of the Docker config. I think it could be added to this PR with easy, although not sure if that should be added to the original library: https://github.com/cpuguy83/dockercfg

Thoughts?

docs/features/docker_auth.md
```diff
@@ -0,0 +1,52 @@
...
+	_Testcontainers for Go_ uses [https://github.com/cpuguy83/dockercfg](https://github.com/cpuguy83/dockercfg) to retrieve the authentication from the credential helpers.
+
+To retrieve the credentials for a given registry, _Testcontainers for Go_ exposes the `AuthFromDockerConfig` function, which takes the registry name as a parameter and returns the authentication for that registry:
+
+```go
+auth, err := testcontainers.AuthFromDockerConfig("myregistry.com")
+if err != nil {
+	// do something with err
+}
+```
```
HofmeisterAn, line 23: You cannot resolve the credentials internally (automatic), based on the image name?

HofmeisterAn (approved):
👏

docker_auth_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+		authCfgs := cfg.AuthConfigs
+
+		if _, ok := authCfgs[indexDockerIO]; !ok {
+			t.Errorf("Expected to find %s in auth configs", indexDockerIO)
+		}
+	})
+
+	t.Run("with DOCKER_CONFIG env var pointing to a non-existing file raises error", func(t *testing.T) {
+		t.Setenv("DOCKER_CONFIG", filepath.Join(testDockerConfigDirPath, "non-existing"))
```
kiview, line 32: I assume `Setenv` is only persisted in the context of the test and does not pollute into the whole process for other tests, right?

docker_auth_test.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+const exampleAuth = "https://example-auth.com"
+const indexDockerIO = "https://index.docker.io/v1/"
+
+var testDockerConfigDirPath = filepath.Join("testresources", ".docker")
+
+func TestGetDockerConfig(t *testing.T) {
+	t.Run("without DOCKER_CONFIG env var retrieves default", func(t *testing.T) {
```
kiview, line 17: This test assumes that the default file as written by Docker is present, correct? I wonder if it is worth protecting this environment assumption with an early assert, so we would understand through this assert failing, that the environment failed the assumption (which is something else from our code failing).

docs/features/docker_auth.md
kiview: I was wondering the same. In testcontainers-java, this process is transparent for the user (in the same way it is present for the user when using Docker CLI).

kiview (commented):
I was wondering the same thing as mentioned by @HofmeisterAn, can't testcontainers-go load the correct credentials automatically without the user explicitly configuring it?

docker_auth_test.go
mdelapenya: Exactly that, testing library will handle it

docs/features/docker_auth.md
mdelapenya: I can include it automatically :)
> We decided to provide that capability instead of automatically discovering it, but I'd not be opposed to do it if desired.

I can imagine this workflow:
- if the user provided `req.RegistryCred`, use it
- else, extract the registry name from the image, and load them

Is this correct?

docker_auth_test.go
mdelapenya: Do you mean checking that the file does exist? This is what L19:20 does

docs/features/docker_auth.md
HofmeisterAn: For .NET we always extract the registry name from the image.

HofmeisterAn: .NET had a similar feature. With the help of various credential providers, I removed it. However, there are still occasional discussions about whether we should support it or not, as seen in this GitHub issue: https://github.com/testcontainers/testcontainers-dotnet/issues/755. Since I understand how the helper and other components work, I prefer the approach of automatically detecting the configuration. However, I can understand that it may be difficult for users to grasp. Although in most environments, users probably do not notice it because it just works.

HofmeisterAn (commented):
I have two questions:

1. Does the implementation cover the PAT? I am not sure if .NET is outdated, but we have something like [this](https://github.com/testcontainers/testcontainers-dotnet/blob/6688b92d217e710a773e87126fc98c78b27de253/src/Testcontainers/Configurations/AuthConfigs/DockerRegistryAuthenticationConfiguration.cs#L41-L47), that sets the `IdentityToken` property.
2. Don't you need to execute the [creds store](https://github.com/testcontainers/testcontainers-dotnet/blob/6688b92d217e710a773e87126fc98c78b27de253/src/Testcontainers/Builders/DockerCredentialProcess.cs#L8-L41) somewhere to receive the credentials? Or is that covered by the dependency?

internal/testcontainersdocker/images.go
```diff
@@ -0,0 +1,125 @@
...
+	for scanner.Scan() {
+		lines = append(lines, scanner.Text())
+	}
+	if scanner.Err() != nil {
+		return nil, scanner.Err()
+	}
+
+	// extract images from dockerfile
+	for _, line := range lines {
+		if !strings.HasPrefix(strings.ToUpper(line), "FROM") {
```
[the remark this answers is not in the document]
mdelapenya: Indeed. Let's be conservative and trim the line. Will add it asap

mdelapenya:
> I have two questions:
> 
> 1. Does the implementation cover the PAT? I am not sure if .NET is outdated, but we have something like [this](https://github.com/testcontainers/testcontainers-dotnet/blob/6688b92d217e710a773e87126fc98c78b27de253/src/Testcontainers/Configurations/AuthConfigs/DockerRegistryAuthenticationConfiguration.cs#L41-L47), that sets the `IdentityToken` property.

PAT == Personal Access Token, right? There are two scenarios:
1) building from dockerfile: we receive the docker auth configs from the credentials helper or the config file, and we build a types.AuthConfig struct (a Docker type) with what comes from the config. That object is entirely passed to the ImageBuildOptions docker type, so whatever Docker client's `ImageBuild` method does internally with the PAT is transparent for us. If it's there, it will be passed.
2) pulling an image: we receive the docker auth configs from the credentials helper or the config file, and we JSON encode that struct, finally passing it to the ImagePullOptions Docker type. The options are directly passed to the Docker client's `ImagePull` method.

in any case I'm going to give it a try tomorrow with a local test and a personal token, and check how it goes 🤔 Thanks for raising this concern!

> 2. Don't you need to execute the [creds store](https://github.com/testcontainers/testcontainers-dotnet/blob/6688b92d217e710a773e87126fc98c78b27de253/src/Testcontainers/Builders/DockerCredentialProcess.cs#L8-L41) somewhere to receive the credentials? Or is that covered by the dependency?

Indeed, it's the dependency the responsible of calling the credential helpers. See https://github.com/cpuguy83/dockercfg/blob/main/auth.go#L51

internal/testcontainersdocker/images.go
```diff
@@ -0,0 +1,125 @@
...
+			continue
+		}
+
+		// remove FROM
+		line = strings.TrimPrefix(line, "FROM")
+		parts := strings.Split(strings.TrimSpace(line), " ")
```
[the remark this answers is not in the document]
mdelapenya: This is covered, as TrimSpace will remove all leading and trailing whitespaces

HofmeisterAn (approved):

docker_auth_test.go
mdelapenya: Oh I understand now: you mean checking that the default file does exist in the scope of the test, right? I've added the check here: https://github.com/testcontainers/testcontainers-go/pull/869/commits/402471b8ac6f1095fb2cdcbb7b917dbf8b7e5f1c

mdelapenya:
I've tested this PR using private registries and private images on Docker Hub, and all our tests pass. Merging it

srabraham:
Hey @mdelapenya, we're users of this package, and thanks for your work on it. We'd been relying on setting Docker credentials via RegistryCred, and after some mangling, I now have our code setting the DOCKER_AUTHCONFIG environment variable instead. One piece of feedback though: this PR claims to have "deprecated" RegistryCred, when in fact it made that feature stop doing anything at all. That presents quite confusing behavior for consumers. If you'd deleted those fields entirely, that at least we would've gotten a compilation error. As it it, there's an option (RegistryCred, and maybe a few others too) that exist but do nothing now

mdelapenya:
@srabraham oh I'm sorry if this PR broke you (and all the users). I really thought we covered the depreciation path for the field, but it seems not. We could release a patch release bringing back the value of the deprecated fields if needed. How would you see that?

I'd like to thank you for your honesty in this feedback, and would like to also recognize that deprecating things are not that easy.

At the same time, we are trying to move to a much more stable state of the code and APIs to release a 1.0.0  version, and that could come with some breaking changes (unfortunately not this time 😥). I wish we'd communicate those changes in the best possible way through the release notes.

srabraham:
@mdelapenya Thanks for that! We did figure out how to get it working; it just took a little time. The solution for us was to replace this, which we were using to set RegistryCred
image

image

with this, prior to making a GenericContainerRequest 
image
image

mdelapenya:
@srabraham we could include that code snippet in the release notes for 0.19.0 as a workaround

srabraham:
> @srabraham I went ahead and updated the release notes for 0.19.0, tagging this PR as a breaking change. Again thanks for sharing your feedback in a positive way. We much appreciate it, and again, my apologies for the non-properly communicated breaking change.

Thanks for all your assistance with this! Yep, users of pre-v1 software need to be cautious, and that's ok! If you or anyone else would like my above snippets in code form (rather than screenshots), here they are: https://gist.github.com/srabraham/52e65ea37db5580a091334e3d065fed9

PR #876, opened 2023-02-25
add couchbase module

### LINKED ISSUE #932: Implement couchbase module

alihanyalcin:
## What does this PR do?
This module initialize couchbase cluster and create buckets for testing purposes. Similar to [testcontainers-java couchbase module](https://github.com/testcontainers/testcontainers-java/tree/main/modules/couchbase/src/main/java/org/testcontainers/couchbase)

## Why is it important?
In java projects, it is really easy to write integration tests for couchbase. So, this module has nearly all functionality supported in testcontainers-java couchbase module.

## Related issues
- Closes #932


mdelapenya (commented):
I added a few minor comments fixing what #877 caused, but it's amazing you used the code generation tool for the module 👏 

I'm going to start the review today, and want to double check with the rest of the team for commonalities with other implementations (e.g. [Java](https://github.com/testcontainers/testcontainers-java/blob/main/modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java)). Did you use it by any case as the reference implementation for Go?

Other than that, thanks for the initial PR, I will be pinging you with questions though it.

The first ping is regarding the docs: as the module is a more elaborated than an example one, I'd recommend extending the docs, like in the LocalStack module (https://golang.testcontainers.org/modules/localstack/) where we expand on the API; or in the upcoming Apache Pulsar module (still in PR stage): https://deploy-preview-872--testcontainers-go.netlify.app/modules/pulsar/

.github/dependabot.yml
```diff
@@ -12,18 +12,6 @@ updates:
         interval: weekly
       open-pull-requests-limit: 3
       rebase-strategy: disabled
```
mdelapenya, line 14: I can see this change is breaking unit tests for the module generator. Did you update it manually? I can see the `modules/compose` has been removed. If not, then I think there is a bug in the generator 🤔

modules/couchbase/couchbase.go
```diff
@@ -0,0 +1,593 @@
...
+		body["cbasSSL"] = cbasSSL.Port()
+	}
+
+	if contains(c.config.enabledServices, eventing) {
+		eventingAdminPort, _ := c.MappedPort(ctx, EVENTING_PORT)
+		eventingSSL, _ := c.MappedPort(ctx, EVENTING_SSL_PORT)
+
+		body["eventingAdminPort"] = eventingAdminPort.Port()
+		body["eventingSSL"] = eventingSSL.Port()
+	}
```
mdelapenya, line 276: This block is not covered by the tests. Could you include them in them?

modules/couchbase/couchbase.go
```diff
@@ -0,0 +1,593 @@
...
+
+	if contains(c.config.enabledServices, eventing) {
+		waitStrategy = append(waitStrategy, wait.ForHTTP("/api/v1/config").
+			WithPort(EVENTING_PORT).
+			WithBasicCredentials(c.config.username, c.config.password).
+			WithStatusCodeMatcher(func(status int) bool {
+				return status == http.StatusOK
+			}))
+	}
```
mdelapenya, line 346: Same here: this block is not covered by the tests. Could you include them in them?

wait/http.go
```diff
@@ -109,6 +111,12 @@ func (ws *HTTPStrategy) WithPollInterval(pollInterval time.Duration) *HTTPStrate
 	return ws
 }
 
+func (ws *HTTPStrategy) WithBasicCredentials(username, password string) *HTTPStrategy {
```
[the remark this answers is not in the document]
mdelapenya: BW, I've just merged #301, so you could rebase and get its benefits :)

.github/dependabot.yml
alihanyalcin: No, I didn't update it manually

wait/http.go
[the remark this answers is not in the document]
alihanyalcin: Yes I saw this PR, but since it was quite old, I implemented my own version 😅

.github/dependabot.yml
mdelapenya: Ok, I'll investigate why and if found, will push a fix for the generator, thanks!

modules/couchbase/couchbase.go
```diff
@@ -0,0 +1,593 @@
...
+
+	c.config.isEnterprise = gjson.Get(string(response), "isEnterprise").Bool()
+
+	if !c.config.isEnterprise {
+		if contains(c.config.enabledServices, analytics) {
+			return errors.New("the Analytics Service is only supported with the Enterprise version")
+		}
+		if contains(c.config.enabledServices, eventing) {
+			return errors.New("the Eventing Service is only supported with the Enterprise version")
```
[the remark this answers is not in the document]
alihanyalcin: Yes, community edition does not support analytics and eventing service.

alihanyalcin:
Hi @mdelapenya, I added requested test and Couchbase documentation.

.github/dependabot.yml
```diff
@@ -24,6 +24,12 @@ updates:
         interval: monthly
       open-pull-requests-limit: 3
       rebase-strategy: disabled
+    - package-ecosystem: gomod
```
mdelapenya, line 27: In the last merge, the dependabot config file is sorted by path, therefore this new module would fall at the end, in the modules block 🙏

mdelapenya (commented):
I left a few minor comments regarding methods belonging to the container struct.

Other than that, once discussed, this PR LGTM, as it matches the Java implementation. Possibly @eddumelendez can verify the Java part.

modules/couchbase/couchbase.go
```diff
@@ -0,0 +1,593 @@
...
+		"flushEnabled":  flushEnabled,
+		"replicaNumber": strconv.Itoa(bucket.numReplicas),
+	}
+
+	_, err := c.doHttpRequest(ctx, MGMT_PORT, "/pools/default/buckets", http.MethodPost, body, true)
+
+	return err
+}
+
+func (c *CouchbaseContainer) doHttpRequest(ctx context.Context, port, path, method string, body map[string]string, auth bool) ([]byte, error) {
```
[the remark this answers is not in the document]
alihanyalcin: If I extract it, I need to call `getURL` method every time I call `doHttpRequest` to get couchbase container url, and code below will be duplicated: 

```
url, err := c.getUrl(ctx, port, path)
if err != nil {
	return nil, err
}
```

modules/couchbase/couchbase.go
```diff
@@ -0,0 +1,593 @@
...
+	request, err := http.NewRequestWithContext(ctx, method, url, strings.NewReader(form.Encode()))
+	if err != nil {
+		return nil, err
+	}
+
+	request.Header.Add("Content-Type", "application/x-www-form-urlencoded")
+
+	if auth {
+		request.SetBasicAuth(c.config.username, c.config.password)
+	}
```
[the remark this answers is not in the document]
mdelapenya: Sure, my point is, instead of passing `auth bool`:

```go
if config.username != "" && config.password != "" {
    request.SetBasicAuth(config.username, config.password)
}
```

passing the config instead of the boolean.

But in any case this is a nit, I'd prefer avoiding calls passing `true|false`, but it'd be fine if we all agree in keeping this method signature.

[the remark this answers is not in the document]
mdelapenya: Ohh I see now, there are methods using basic auth and others not using it. Got it, then I think it's fine keeping it as is

Thanks for the feedback!

modules/couchbase/couchbase.go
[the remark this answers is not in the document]
mdelapenya: Ok, following with the previous discussion on passing `auth bool`, it makes sense to keep it here as you explain. Thanks for the clarification, now I see that we need the container to get it's URL

modules/couchbase/go.mod
```diff
@@ -0,0 +1,60 @@
...
+	golang.org/x/term v0.5.0 // indirect
+	golang.org/x/tools v0.1.12 // indirect
+	google.golang.org/genproto v0.0.0-20220617124728-180714bec0ad // indirect
+	google.golang.org/grpc v1.47.0 // indirect
+	google.golang.org/protobuf v1.28.0 // indirect
+	gotest.tools/v3 v3.4.0 // indirect
+)
+
+replace (
+	github.com/docker/docker => github.com/docker/docker v20.10.3-0.20221013203545-33ab36d6b304+incompatible // 22.06 branch
```
mdelapenya, line 58: No need to replace Docker anymore. See #824 
```suggestion
```

modules/couchbase/couchbase.go
```diff
@@ -0,0 +1,597 @@
...
+	_, err := c.doHttpRequest(ctx, MGMT_PORT, "/node/controller/setupAlternateAddresses/external", http.MethodPut, body, true)
+
+	return err
+}
+
+func (c *CouchbaseContainer) configureIndexer(ctx context.Context) error {
+	storageMode := "forestdb"
+	if c.config.isEnterprise {
+		storageMode = "memory_optimized"
+	}
```
eddumelendez, line 291: I think it would be great to consider [this](https://github.com/testcontainers/testcontainers-java/pull/3428) in the new Go implementation

eddumelendez (commented):
LGTM regarding to the actual status in tc-java. I've left one comment which I think could avoid breaking changes in the future for Go implementation. Consider it as a nice to have :)

modules/couchbase/couchbase.go
mdelapenya: @alihanyalcin do you think we can contribute that part? As this module will fall under the next release lifecycle, we could merge this PR as is, and you could do what @eddumelendez suggests in a follow-up. Wdyt?

alihanyalcin: Hi, I have contributed the requested section. BTW, the enterprise edition of the Couchbase Server supports both the Plasma and Memory Optimized storage engines, while the community edition only supports the ForestDB storage engine.

mdelapenya: Thanks for getting so fast! We cannot be experts in all the supported technologies so we are more than thankful for seeing this kind of improvements 👏👏👏

modules/couchbase/storage_mode.go
```diff
@@ -0,0 +1,9 @@
+package couchbase
+
+type indexStorageMode string
+
+const (
+	MemoryOptimized indexStorageMode = "memory_optimized"
```
mdelapenya, line 6: Could we copy the comments from the Java PR here, so each constant is self-explanatory with the comments? 🙏

alihanyalcin: Done 👍

mdelapenya (approved):
It took a while, but I think this PR LGTM

Thanks @alihanyalcin for your support and your time during the review. 🙇

lmitelman:
@mdelapenya A new vulnerability was found in the [crypto](https://pkg.go.dev/vuln/GO-2023-1621) dependency, which is used by testcontainers!

I strongly recommend bumping that dependency, as SonarQube and other tools do not let pipelines through!

```
Vulnerability #1: GO-2023-1621
  The ScalarMult and ScalarBaseMult methods of the P256 Curve may
  return an incorrect result if called with some specific
  unreduced scalars (a scalar larger than the order of the curve).
  This does not impact usages of crypto/ecdsa or crypto/ecdh.

  More info: https://pkg.go.dev/vuln/GO-2023-1621

  Standard library
    Found in: crypto/internal/nistec@go1.20.1
    Fixed in: crypto/internal/nistec@go1.20.2
```

mdelapenya:
@lmitelman could you open a separate report for this? We'll label with security to address it asap

Thanks for reporting it!

PR #906, opened 2023-03-01
chore: monthly updates for dependabot

mdelapenya:
## What does this PR do?
It updates dependabot's config file and the code generator that updates dependabot.

## Why is it important?
With the raise of modules, we receive multiple PRs from dependabot, so better to handle them month by month


mdelapenya:
Merging, as it's harmless to the CI

PR #910, opened 2023-03-02
remove volumes on down

### LINKED ISSUE #644: [Enhancement]: Remove created volumes on "down"
### Proposal

When calling the `Down()` method of the `LocalDockerCompose` created through `NewLocalDockerCompose()`, volumes created when "upping" the instance are deleted through the `--volumes` option:

https://github.com/testcontainers/testcontainers-go/blob/20df64ea3d8b80668604487e8aff3bfb0e6da2ea/compose_local.go#L84

It would be nice to have that behavior on `ComposeStack.Down()` created through `NewDockerCompose()` as well. I believe  this can be easily achieved by adding `Volumes: true` on the following line:

https://github.com/testcontainers/testcontainers-go/blob/20df64ea3d8b80668604487e8aff3bfb0e6da2ea/compose_api.go#L158

BTW, I thank all the contributors of this awesome library. It's been really useful to me.

cgoodsirsmyth-pp:
## What does this PR do?

Optionally cleans up volumes when using `compose.Down()`

## Why is it important?

Cleaning up after your tests is important. ;)

## Related issues

- Closes #644
- Closes #690


modules/compose/compose_api_test.go
```diff
@@ -365,6 +367,29 @@ func TestDockerComposeAPIWithVolume(t *testing.T) {
...
+	t.Cleanup(cancel)
+
+	err = compose.Up(ctx, Wait(true))
+	assert.NoError(t, err, "compose.Up()")
+
+	err = compose.Down(context.Background(), RemoveOrphans(true), RemoveVolumes(true), RemoveImagesLocal)
+	assert.NoError(t, err, "compose.Down()")
+
+	volumeListFilters := filters.NewArgs()
+	volumeListFilters.Add("name", fmt.Sprintf("%s_mydata", identifier))
```
mdelapenya, line 386: NIT: Could you add a comment mentioning that this volume comes from the compose referenced above? 🙏

mdelapenya (approved):
LGTM! Thanks for making Testcontainers for Go awesome!

modules/compose/compose_api_test.go
mdelapenya, line 386:
```suggestion
	// the "mydata" identifier comes from the "testresources/docker-compose-volume.yml" file
	volumeListFilters.Add("name", fmt.Sprintf("%s_mydata", identifier))
```

modules/compose/compose_api_test.go
```diff
@@ -365,6 +367,29 @@ func TestDockerComposeAPIWithVolume(t *testing.T) {
 	assert.NoError(t, err, "compose.Up()")
 }
 
+func TestDockerComposeAPIVolumesDeletedOnDown(t *testing.T) {
+	identifier := uuid.New().String()
+	stackFiles := WithStackFiles("./testresources/docker-compose-volume.yml")
```
mdelapenya, line 372: Could you use filepath to create a OS-agnostic path? 🙏

mdelapenya (changes requested):
I took a second look, and I think we are only missing this last change before merging the PR

Thanks again for your work here

modules/compose/compose_api_test.go
mdelapenya: I know it's not there for any method, if you want to add them all, in a separate PR, that would be awesome

cgoodsirsmyth-pp: Updated all tests to use `filepath` 👍

mdelapenya: This is a tiny thing, but I'd appreciate if you extract the file names to a constant, as Sonar is "complaining" about it: https://sonarcloud.io/project/issues?id=testcontainers_testcontainers-go&pullRequest=910&resolved=false&types=CODE_SMELL 🙏

mdelapenya: @cgoodsirsmyth-pp I this is the only thing left 🙏  As I said, it's a tiny one, but if you think we could live with the sonar warning which, to me, in testing sometimes is desired to have repetition, then it's fine to merge it as is.

cgoodsirsmyth-pp: Moved simple, complex and volume compose to a `const` as well as the `testresources` package.

cgoodsirsmyth-pp: Did not notice the sonarcloud code smell apologies.

mdelapenya: Ah no worries, we are experimenting with sonarcloud atm. Not very strict with them, but checking if it provides value for us. Thanks anyway for updating the PR. I think we are ready to go now :)

mdelapenya (approved):
Thanks for your time here, nice contribution!

PR #928, opened 2023-03-08
fix: use right paths in the release script

mdelapenya:
## What does this PR do?
This PR fixes the release script:

- it now handles the next version in go modules properly, as it was replacing too many things, such as the replace directive in the Go modules
- it uses the right path to each module when replacing the current version

## Why is it important?
Make the script as autonomous as possible


mdelapenya:
Merging as it does not affect the CI

PR #939, opened 2023-03-13
chore: always close Docker client

### LINKED ISSUE #321: Docker connection not closed
**Describe the bug**
The `ProviderType.GetProvider()` function creates a docker client, but never closes it, which leads to a leaking connection.

Similar to https://github.com/testcontainers/testcontainers-go/issues/319

**To Reproduce**
Use following test:

```go
package main

import (
	"context"
	"testing"
	"time"

	"go.uber.org/goleak"
)

func TestMain(m *testing.M) {
	goleak.VerifyTestMain(m)
}

func TestFoo(t *testing.T) {
	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
	defer cancel()
		Image:        "postgres:12",
		ExposedPorts: []string{"5432/tcp"},
		AutoRemove:   true,
		Env: map[string]string{
			"POSTGRES_USER":     "postgres",
			"POSTGRES_PASSWORD": "postgres",
			"POSTGRES_DB":       "postgres",
		},
		WaitingFor: wait.ForListeningPort("5432/tcp"),
	}
	container, err := testcontainers.GenericContainer(ctx, testcontainers.GenericContainerRequest{
		ContainerRequest: req,
		Started:          true,
	})
	if err != nil {
		panic(err)
	}
	if err := container.Terminate(ctx); err != nil {
		panic(err)
	}
}

```

**Expected behavior**
A successful test

**docker info**
```output of the command:```

```
$ docker info
Client:
 Context:    default
 Debug Mode: false

Server:
 Containers: 18
  Running: 1
  Paused: 0
  Stopped: 17
 Images: 72
 Server Version: 20.10.2
 Storage Driver: zfs
  Zpool: rpool
  Zpool Health: ONLINE
  Parent Dataset: rpool/var/lib/docker
  Space Used By Parent: 10480205824
  Space Available: 145392234496
  Parent Quota: no
  Compression: lz4
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 
 runc version: 
 init version: 
 Security Options:
  apparmor
  seccomp
   Profile: default
 Kernel Version: 5.8.0-50-generic
 Operating System: Ubuntu 20.04.2 LTS
 OSType: linux
 Architecture: x86_64
 CPUs: 8
 Total Memory: 15.53GiB
 Name: tobias-talon-T480s
 ID: UMBP:HB2G:5S3P:JMX3:VZ46:ATMB:64RC:4WXO:YJ6L:M7UK:DWGB:JJ25
 Debug Mode: false
 Labels:
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false
```

**Additional context**
It might be necessary to adjust `GenericProvider` to include a `Close()` function that closes the used provider.

mdelapenya:
## What does this PR do?
This PR improves how the Docker client is handled:

- it simplifies how the API version is negotiated when retrieving a Docker client: instead of forcing a call to the `NegotiateAPIVersion(ctx)` method, it's passed as a client Opt when getting the new client.
- it unifies how a Docker client is retrieved, creating an internal function to it. There, we will centralise the creation of the Docker client.
- it also adds a `Close` method to the DockerProvider struct. With it, we have identified the usages of the DockerProvider, and its underlying Docker client, to close the client and avoid leaking Docker connections.

As part of the PR we are moving the provider related code to a separate Go file, and identified that all `compose_test.go` paths were not using `filepath` to create cross-OS paths.

## Why is it important?
It makes sure that we are not leaking Docker client connections, making sure it's closed everywhere, also providing a way to close the Docker Provider wrapper.

## Related issues

- Closes #321

## Follow-ups
@Eun I'm not adding the test you provided in the original issue because I dit not see any difference running it against this branch or main (without the closes). Do you think we must add them to be covered? My only concern would be plugin the leak detector to the TestMain (`goleak.VerifyTestMain(m)`, see https://pkg.go.dev/go.uber.org/goleak#readme-quick-start):

> Instead of checking for leaks at the end of every test, goleak can also be run at the end of every test package by creating a TestMain function for your package.

### DIFF
--- a/container.go
+++ b/container.go
@@ -28,15 +28,6 @@ type DeprecatedContainer interface {
 	Terminate(ctx context.Context) error
 }
 
-// ContainerProvider allows the creation of containers on an arbitrary system
-type ContainerProvider interface {
-	CreateContainer(context.Context, ContainerRequest) (Container, error)        // create a container without starting it
-	ReuseOrCreateContainer(context.Context, ContainerRequest) (Container, error) // reuses a container if it exists or creates a container without starting
-	RunContainer(context.Context, ContainerRequest) (Container, error)           // create a container and start it
-	Health(context.Context) error
-	Config() TestContainersConfig
-}
-
 // Container allows getting info about and controlling a single container instance
 type Container interface {
 	GetContainerID() string                                         // get the container id from the provider
@@ -132,30 +123,6 @@ type ContainerRequest struct {
 	EnpointSettingsModifier func(map[string]*network.EndpointSettings) // Modifier for the network settings before container creation
 }
 
-type (
-	// ProviderType is an enum for the possible providers
-	ProviderType int
-
-	// GenericProviderOptions defines options applicable to all providers
-	GenericProviderOptions struct {
-		Logger         Logging
-		DefaultNetwork string
-	}
-
-	// GenericProviderOption defines a common interface to modify GenericProviderOptions
-	// These options can be passed to GetProvider in a variadic way to customize the returned GenericProvider instance
-	GenericProviderOption interface {
-		ApplyGenericTo(opts *GenericProviderOptions)
-	}
-
-	// GenericProviderOptionFunc is a shorthand to implement the GenericProviderOption interface
-	GenericProviderOptionFunc func(opts *GenericProviderOptions)
-)
-
-func (f GenericProviderOptionFunc) ApplyGenericTo(opts *GenericProviderOptions) {
-	f(opts)
-}
-
 // containerOptions functional options for a container
 type containerOptions struct {
 	ImageName           string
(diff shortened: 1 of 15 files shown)


Eun:
I don't think adding my tests would help, I just added them to have a fast reproducible result.
However since this is not reproducible at all (anymore), I would not bother with tests at the moment.

so LGTM from my side.  

As a side note:
the goleak package was changed so it doesn't produce the output anymore.
Maybe its worth checking the options available or using an older version of goleak to reproduce the issue at least once and see that it is resolved.

PR #942, opened 2023-03-17
feat: add Vault Module

alihanyalcin:
## What does this PR do?
Support for Vault containers

## Why is it important?
[Vault](https://www.vaultproject.io/) is an open-source tool designed for securely storing, accessing, and managing secrets and sensitive data such as passwords, certificates, API keys, and other confidential information.


modules/vault/options.go
```diff
@@ -0,0 +1,82 @@
...
+import "fmt"
+
+// Option is a function type for modifying a Vault configuration
+type Option func(*Config)
+
+// Config is a struct that contains various configuration options for the Vault
+type Config struct {
+	imageName    string              // The name of the Docker image to use for the Vault
+	token        string              // The root token to use for the Vault
+	port         int                 // The port number to use for the Vault
```
eddumelendez, line 12: port is 8200 by default in Vault image. I don't think it is useful to change the port. Do you have any use case in mind? In Java, it is already deprecated

modules/vault/options.go
```diff
@@ -0,0 +1,82 @@
...
+	secrets      map[string][]string // A map of secret paths to their respective secret values
+	initCommands []string            // A list of commands to execute on Vault initialization
+	logLevel     LogLevel            // The level of logging to use for the Vault
```
eddumelendez, line 15: This is a env var which can be passed by override the `ContainerRequest`. I know this is handy but in previous conversation we have discussed this. At least there should be some logic around the env var such as in databases in order to take this into account. In Java, should be deprecated.

modules/vault/options.go
```diff
@@ -0,0 +1,82 @@
...
+// Option is a function type for modifying a Vault configuration
+type Option func(*Config)
+
+// Config is a struct that contains various configuration options for the Vault
+type Config struct {
+	imageName    string              // The name of the Docker image to use for the Vault
+	token        string              // The root token to use for the Vault
+	port         int                 // The port number to use for the Vault
+	secrets      map[string][]string // A map of secret paths to their respective secret values
+	initCommands []string            // A list of commands to execute on Vault initialization
```
eddumelendez, line 14: I think this was based on the Java implementation but TBH it was confusing some times, see [this comment](https://github.com/testcontainers/testcontainers-java/issues/5510#issuecomment-1179451354). IMO, I would rely more on [executing a command](https://golang.testcontainers.org/features/override_container_command/)  or [copying a script](https://golang.testcontainers.org/features/copy_file/) and executing it later. WDYT?

eddumelendez (commented):
Thanks for your contribution @alihanyalcin ! I have left some comments based on the experience in Java modules implementation in general. I understand the implementation has been based on `Testcontainers for Java` but let's build a better version of it in Go 😉

modules/vault/options.go
alihanyalcin: good idea but `Cmd` is running before `WaitingFor` so it try to execute a command before container running. Is there any way to execute command after `WaitingFor`? If there is not, copying a script is more sense.

modules/vault/options.go
alihanyalcin: I tried to add all functionality Java has. I didn't notice it is deprecated, I will remove it 👍

modules/vault/options.go
alihanyalcin: I solved it using `WaitingFor`. Can you review it again?

docs/modules/vault.md
```diff
@@ -0,0 +1,35 @@
...
+
+### Use HTTP API to read data from Vault container:
+<!--codeinclude-->
+[Use HTTP API to read data](../../modules/vault/vault_test.go) inside_block:TestVaultFirstSecretPathWithHTTP
+<!--/codeinclude-->
+
+### Use client library to read data from Vault container:
+<!--codeinclude-->
+[Use library to read data](../../modules/vault/vault_test.go) inside_block:TestVaultFirstSecretPathWithClient
+<!--/codeinclude-->
```
mdelapenya, line 35: Could you add a section here describing the different opts? You can find an example here: https://golang.testcontainers.org/modules/localstack/#module-reference

mdelapenya (approved):
@alihanyalcin another great PR for the vault module, thank you for your love to Testcontainers!

I'll wait for @eddumelendez to check the comments he had before merging the PR, but other than that, great job!

modules/vault/vault.go
```diff
@@ -0,0 +1,104 @@
...
+		req.Env["VAULT_TOKEN"] = token
+	}
+}
+
+// WithLogLevel is an option function that sets the logging level for the Vault
+func WithLogLevel(logLevel LogLevel) ContainerOptions {
+	return func(req *testcontainers.ContainerRequest) {
+		req.Env["VAULT_LOG_LEVEL"] = string(logLevel)
+	}
+}
```
eddumelendez, line 75: I thought the `WithLogLevel` fn will be removed because of this [comment](https://github.com/testcontainers/testcontainers-go/pull/942#discussion_r1140881048)

modules/vault/vault.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/testcontainers/testcontainers-go/wait"
+	"strings"
+
+	"github.com/docker/docker/api/types/container"
+	"github.com/testcontainers/testcontainers-go"
+)
+
+const (
+	defaultPort      = "8200"
+	defaultImageName = "vault:1.13.0"
```
eddumelendez, line 15: Because of the notice below (source: https://hub.docker.com/_/vault) we should point to `hashicorp/vault` instead.

> DEPRECATION NOTICE
Upcoming in Vault 1.14, we will stop publishing official Dockerhub images and publish only our Verified Publisher images. Users of Docker images should pull from [hashicorp/vault](https://hub.docker.com/r/hashicorp/vault) instead of [vault](https://hub.docker.com/_/vault). Verified Publisher images can be found at https://hub.docker.com/r/hashicorp/vault.

```suggestion
	defaultImageName = "hashicorp/vault:1.13.0"
```

alihanyalcin:
Hi @mdelapenya @eddumelendez I've added requested changes.

mdelapenya (approved):

PR #943, opened 2023-03-18
Add MySQL module

eddumelendez:
The module allow to start a MySQL database. By default, the value
for username, password and database is test but those can be
set using the proper functions `WithUsername()`, `WithPassword` and
`WithDatabase()`. It also provides fns to get those values and
build the connection string.


modules/mysql/mysql_test.go
```diff
@@ -0,0 +1,154 @@
...
+		" `col_1` VARCHAR(128) NOT NULL, \n" +
+		" `col_2` VARCHAR(128) NOT NULL, \n" +
+		" PRIMARY KEY (`col_1`, `col_2`) \n" +
+		")")
+	if err != nil {
+		t.Errorf("error creating table: %+v\n", err)
+	}
+}
+
+//func TestMySQLWithConfigFile(t *testing.T) {
```
eddumelendez, line 110: having issues with this test 🤔

.github/dependabot.yml
```diff
@@ -108,6 +108,12 @@ updates:
         interval: monthly
       open-pull-requests-limit: 3
       rebase-strategy: disabled
+    - package-ecosystem: gomod
```
mdelapenya, line 111: We would need to remove the entry for the MySQL example module

modules/mysql/options.go
```diff
@@ -0,0 +1,34 @@
...
+	configFile string
+}
+
+func WithUsername(username string) Option {
+	return func(config *Config) {
+		config.username = username
+	}
+}
+
+func withPassword(password string) Option {
```
mdelapenya, line 18: Export this option for users
```suggestion
func WithPassword(password string) Option {
```

modules/mysql/options.go
```diff
@@ -0,0 +1,34 @@
...
+	return func(config *Config) {
+		config.password = password
+	}
+}
+
+func withDatabase(database string) Option {
```
mdelapenya, line 24: Same here
```suggestion
func WithDatabase(database string) Option {
```

modules/mysql/options.go
```diff
@@ -0,0 +1,34 @@
...
+	return func(config *Config) {
+		config.database = database
+	}
+}
+
+func withConfigFile(configFile string) Option {
```
mdelapenya, line 30: And here
```suggestion
func WithConfigFile(configFile string) Option {
```

modules/mysql/mysql.go
```diff
@@ -0,0 +1,82 @@
...
+const defaultDatabaseName = "test"
+
+// MySQLContainer represents the MySQL container type used in the module
+type MySQLContainer struct {
+	testcontainers.Container
+	config *Config
+}
+
+// StartContainer creates an instance of the MySQL container type
+func StartContainer(ctx context.Context, image string, opts ...Option) (*MySQLContainer, error) {
```
mdelapenya, line 23: I think you could probably create a new Option for `WithImage`, that sets it in the config struct. That would simplify the method signature and we will be closer to the design we have in mind for an eventual refactor

eddumelendez: I was wondering about it because, for me at least and based on tc-java, it would be nice to declare the image to use when declaring `StartContainer` instead of relying on defaults. `Testcontainers for Java` used to have defaults in the past and now it promotes to always declare the image. Do you think this is something to take into account? We can also save to pass another option.

mdelapenya: Interesting feedback. Do you suggest that the StartContainer func should include a parameter for the image?

eddumelendez: in the current approach, yes. It also save to have an additional option 😄

Also, didn't mention it but in Testcontainers for Java an assertion is performed against the image name because the module works in this case with `mysql` but if you pass `my-custom-mysql` then we can not make sure it will work unless there is an option to say it is also compatible with mysql. This is already implemented in Java.

The way I see it, with this proposal we can be ready to introduce it smoothly.

mdelapenya: Indeed, it saves the option confusing the default but forces to add a validation on the image name, unless we let it fall

modules/mysql/mysql.go
```diff
@@ -17,61 +16,53 @@ const defaultDatabaseName = "test"
...
-	}
-	if len(config.password) != 0 && config.password != "" {
-		mysqlEnv["MYSQL_PASSWORD"] = config.password
-		mysqlEnv["MYSQL_ROOT_PASSWORD"] = config.password
-	} else if strings.EqualFold(rootUser, config.username) {
-		mysqlEnv["MYSQL_ALLOW_EMPTY_PASSWORD"] = "yes"
-	} else {
-		return nil, fmt.Errorf("empty password can be used only with the root user")
-	}
+type PostgresContainerOption func(req *testcontainers.ContainerRequest)
```
mdelapenya, line 24:
```suggestion
type MySQLContainerOption func(req *testcontainers.ContainerRequest)
```

modules/mysql/mysql.go
```diff
@@ -17,61 +16,53 @@ const defaultDatabaseName = "test"
...
 
+// StartContainer creates an instance of the MySQL container type
+func StartContainer(ctx context.Context, image string, opts ...PostgresContainerOption) (*MySQLContainer, error) {
```
mdelapenya, line 27:
```suggestion
func StartContainer(ctx context.Context, image string, opts ...MySQLContainerOption) (*MySQLContainer, error) {
```

modules/mysql/mysql.go
```diff
@@ -0,0 +1,90 @@
...
+		},
+		WaitingFor: wait.ForLog("port: 3306  MySQL Community Server"),
+	}
+
+	opts = append(opts, func(req *testcontainers.ContainerRequest) {
+		username := req.Env["MYSQL_USER"]
+		password := req.Env["MYSQL_PASSWORD"]
+		if strings.EqualFold(rootUser, username) {
+			delete(req.Env, "MYSQL_USER")
+		}
```
mdelapenya, line 44: This block is probably not needed, as setting the env var as `root` could be something harmless

modules/mysql/mysql.go
```diff
@@ -0,0 +1,90 @@
...
+		if len(password) != 0 && password != "" {
+			req.Env["MYSQL_ROOT_PASSWORD"] = password
+		} else if strings.EqualFold(rootUser, username) {
+			req.Env["MYSQL_ALLOW_EMPTY_PASSWORD"] = "yes"
+			delete(req.Env, "MYSQL_PASSWORD")
+		}
+	})
```
mdelapenya, line 51: I like this sanitisation 👍

modules/mysql/mysql_test.go
```diff
@@ -0,0 +1,223 @@
...
+	})
+
+	// perform assertions
+
+	host, _ := container.Host(ctx)
+
+	p, _ := container.MappedPort(ctx, "3306/tcp")
+	port := p.Int()
+
+	connectionString := fmt.Sprintf("%s:%s@tcp(%s:%d)/%s?tls=skip-verify",
```
mdelapenya, line 36: In the postgres module we are proposing a method in the container retrieving the connection string, which I consider handy for databases. Do you think we could replicate it for MySQL (and eventual future databases?)

eddumelendez: sure! I can check it later

modules/mysql/mysql.go
eddumelendez: it is indeed 😅 otherwise, the container will not start. That's the reason why the env var should not be present at all

modules/mysql/mysql.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	username, ok := req.Env["MYSQL_USER"]
+	if !ok {
+		username = rootUser
+	}
+	password := req.Env["MYSQL_PASSWORD"]
+
+	if len(password) == 0 && password == "" && !strings.EqualFold(rootUser, username) {
+		return nil, fmt.Errorf("empty password can be used only with the root user")
```
mdelapenya, line 64: We could define a custom and exported error here, something like 
```go
var ErrEmptyPasswForRoot = errors.New("empty password can be used only with the root user")
```
and finally in https://github.com/testcontainers/testcontainers-go/pull/943/files#diff-02c2cf13c522643c66ea754bfc71eeaaec57856bf6abdb78f50a93205f429325R65 check for the error with `errors.Is` https://pkg.go.dev/errors#Is

mdelapenya: But I'm not sure about this, as it would mean adding Error types for each module to the public API, which although interesting at some point, could come with different maintainability concerns, like naming, which error types support, etc

modules/mysql/mysql.go
```diff
@@ -0,0 +1,90 @@
...
+	testcontainers.Container
+	username string
+	password string
+	database string
+}
+
+type MySQLContainerOption func(req *testcontainers.ContainerRequest)
+
+// StartContainer creates an instance of the MySQL container type
+func StartContainer(ctx context.Context, image string, opts ...MySQLContainerOption) (*MySQLContainer, error) {
```
mdelapenya, line 27: Do we still provide the image as part of the start function, or do we advocate for using a default + WithImage option?

eddumelendez: for consistency in tc-go, I'll add WithImage

modules/mysql/mysql.go
```diff
@@ -79,16 +79,19 @@ func StartContainer(ctx context.Context, opts ...MySQLContainerOption) (*MySQLCo
 	return &MySQLContainer{container, username, password, database}, nil
 }
 
-func (c *MySQLContainer) Username() string {
-	return c.username
-}
+func (c *MySQLContainer) ConnectionString(ctx context.Context) (string, error) {
```
mdelapenya, line 82: Question: do we want to allow consumers to customize the the connection string, probably adding parameters to it? Please see https://golang.testcontainers.org/modules/postgres/#connectionstring

eddumelendez: done!

mdelapenya (approved):
LGTM! Great module, as always!

PR #944, opened 2023-03-21
feat: add status check to wait strategies

### LINKED ISSUE #938: [Enhancement]: Fails the wait when the container stops
### Proposal

Some strategies assume the container is running.
However, a container may terminate due to an unexpected error, etc. Even in this case, `testcontainers-go` will keep waiting until the timeout.
It might be a good idea to check the status of the container while polling and return an error if it was stopped.

Example:

In the following code, mysql container immediately stops due to missing environment variables, but [wait.ForSQL](https://pkg.go.dev/github.com/testcontainers/testcontainers-go/wait#ForSQL) continues to wait until the timeout.

```go
package main

import (
	"context"
	"log"
	"net"

	"github.com/go-sql-driver/mysql"
)

func main() {
	ctx := context.Background()

		Image: "mysql:8.0",
		Env: map[string]string{
			"MYSQL_DATABASE": "app",
			"MYSQL_USER":     "user",
			"MYSQL_PASSWORD": "password",
			// "MYSQL_ALLOW_EMPTY_PASSWORD": "yes",
		},
		ExposedPorts: []string{"3306/tcp"},
		WaitingFor: wait.ForSQL("3306", "mysql", func(host string, port nat.Port) string {
			cfg := mysql.NewConfig()
			cfg.Net = "tcp"
			cfg.Addr = net.JoinHostPort(host, port.Port())
			cfg.DBName = "app"
			cfg.User = "user"
			cfg.Passwd = "password"
			return cfg.FormatDSN()
		}),
	}

	container, err := testcontainers.GenericContainer(ctx, testcontainers.GenericContainerRequest{
		ContainerRequest: req,
		Started:          true,
	})
	if err != nil {
		log.Fatal(err)
	}
	defer container.Terminate(ctx)
}
```

frozenbonito:
## What does this PR do?

This PR adds container status check to wait strategies while polling.

## Why is it important?

Some strategies assume the container is running.
However, a container may terminate due to an unexpected error, etc. Even in this case, testcontainers-go will keep waiting until the timeout.

This PR enables fail-fast when a container stops.

## Related issues

- Closes #938


wait/log.go
```diff
@@ -100,7 +102,9 @@ LOOP:
 			}
 
 			logs := string(b)
-			if strings.Count(logs, ws.Log) >= ws.Occurrence {
+			if logs == "" && checkErr != nil {
```
mdelapenya, line 105: Should we check for error here or in L91, to avoid retrieving and reading target's logs?

wait/wait.go
```diff
@@ -29,6 +31,28 @@ type StrategyTarget interface {
...
+}
+
+func checkState(state *types.ContainerState) error {
+	switch {
+	case state.Running:
+		return nil
+	case state.OOMKilled:
+		return errors.New("container crashed with out-of-memory (OOMKilled)")
+	case state.Status == "exited":
+		return fmt.Errorf("container exited with status %d", state.ExitCode)
```
mdelapenya, line 50:
```suggestion
		return fmt.Errorf("container exited with code %d", state.ExitCode)
```

wait/http_test.go
```diff
@@ -247,3 +250,212 @@ func TestHTTPStrategyWaitUntilReadyNoBasicAuth(t *testing.T) {
...
+func TestHttpStrategyFailsWhileGettingPortDueToOOMKilledContainer(t *testing.T) {
+	var mappedPortCount int
+	target := &wait.MockStrategyTarget{
+		HostImpl: func(_ context.Context) (string, error) {
+			return "localhost", nil
+		},
+		MappedPortImpl: func(_ context.Context, _ nat.Port) (nat.Port, error) {
+			defer func() { mappedPortCount++ }()
+			if mappedPortCount == 0 {
+				return "", errors.New("port not found")
```
mdelapenya, line 263: We could probably extract `errors.New("port not found")` to a const/var to avoid duplications. It could be also interesting adding it to http.go/host_port.go/sql.go as an exported var to be reused as a type. wdyt?

mdelapenya (commented):
@frozenbonito it LGTM, thanks for your contribution!

I added a few comments, minors, but wanted to check with you first before approving the PR.

wait/log.go
frozenbonito: The logs just before the container stops may contain the desired entries.
Therefore, I think we should check all its logs even if the container is stopped.
To ensure that all logs read, here I confirm that the logs are empty after checking the status.

wait/wait.go
frozenbonito: Thank you for your comment, I fixed it.

wait/http_test.go
frozenbonito: This error mimics the error the code below (in L171).

https://github.com/testcontainers/testcontainers-go/blob/447099b908d12a0dd64f4110e995c51ea2933751/docker.go#L144-L172

Perhaps the const/var should be defined in the `testcontainers-go` package, but referencing it from `wait` package is a cycle import.

Since this is just a error for testing, I don't think it makes much sense to avoid duplication.

frozenbonito: Thank you for comment.
I extracted the error into var.
How is this?

wait/wait_test.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+var ErrPortNotFound = errors.New("port not found")
+
+type MockStrategyTarget struct {
+	HostImpl       func(context.Context) (string, error)
+	PortsImpl      func(context.Context) (nat.PortMap, error)
+	MappedPortImpl func(context.Context, nat.Port) (nat.Port, error)
+	LogsImpl       func(context.Context) (io.ReadCloser, error)
+	ExecImpl       func(context.Context, []string, ...tcexec.ProcessOption) (int, io.Reader, error)
```
mdelapenya, line 20: I'm checking that `ExecImpl` is not used anywhere else but in https://github.com/testcontainers/testcontainers-go/pull/944/files#diff-01e1c8168e5c89a7f968caae722f60ee322b183eddf9169560ba0ccef4abf0bdR41.

Do you think we need it, or just having a NOOP is enough for the mock strategy executing commands?

mdelapenya: Oh oh I'm sorry 🙏 you're right, the diff was not open. I'm resolving this comment now

mdelapenya (approved):
LGTM, thanks for your patience during the review

We really appreciate your love for OSS and Testcontainers for Go!

PR #971, opened 2023-03-23
docs: extend Couchbase docs

mdelapenya:
## What does this PR do?
It adds a module reference section in the docs, describing each exposed function in the public API, including code snippets extracted from the code.

At the same time, we are adding Go docs for the public API.

Finally, we are moving the test methods to a test package, so that we demonstrate the real usage of the module from inside.

Render URL: https://deploy-preview-971--testcontainers-go.netlify.app/modules/couchbase

## Why is it important?
Make the module more usable through docs

## Related issues

- Completes #876


mdelapenya:
@alihanyalcin could you please take a look? I'm extending the docs for the recently added Couchbase module.

Thanks in advance!

PR #985, opened 2023-03-27
Reorder docs language links

leocross:
## What is this PR doing?
It's reordering the supported langs in the docs site, to have NodeJS before Python.

## Why is it important?
Consistency in the order of the languages


mdelapenya (approved):
LGTM! Thanks @leocross 

I've updated the description of the PR to match the intention of this PR, as we consider a best practice to share context to eventual readers about what this PR does and why it's needed.

PR #994, opened 2023-03-28
Add support for LocalStack v2

eddumelendez:
`HOSTNAME_EXTERNAL` env var is deprecated and will be replaced by
`LOCALSTACK_HOST` in the upcoming v2.


modules/localstack/localstack_test.go
```diff
@@ -165,3 +175,60 @@ func TestStartWithNetwork(t *testing.T) {
...
+			Name: "localstack-network",
+		},
+	})
+	require.Nil(t, err)
+	assert.NotNil(t, nw)
+
+	localstack, err := StartContainer(
+		ctx,
+		OverrideContainerRequest(testcontainers.ContainerRequest{
+			Image:          "localstack/localstack:latest",
```
eddumelendez, line 193: Replace by proper tag. It should be `v2.0`

```suggestion
			Image:          "localstack/localstack:v2.0",
```

modules/localstack/localstack.go
```diff
@@ -40,6 +41,25 @@ func isLegacyMode(image string) bool {
...
+	if version == "latest" {
+		return true
+	}
+
+	if !strings.HasPrefix(version, "v") {
+		version = fmt.Sprintf("v%s", version)
+	}
+
+	if semver.IsValid(version) {
+		return semver.Compare(version, "v2.0") > 0 // version < v0.11
```
mdelapenya, line 57:
```suggestion
		return semver.Compare(version, "v2.0") > 0 // version >= v2.0
```

mdelapenya:
Localstack tests are failing. Could you take a look 🙏 ?

modules/localstack/localstack_test.go
```diff
@@ -165,3 +175,60 @@ func TestStartWithNetwork(t *testing.T) {
...
+	require.Nil(t, err)
+	assert.NotNil(t, cli)
+
+	c, output, err := cli.Exec(ctx, []string{"/usr/local/bin/aws", "sqs", "create-queue", "--queue-name", "baz", "--region", "eu-west-1",
+		"--endpoint-url", "http://localstack:4566", "--no-verify-ssl"})
+	if err != nil {
+		t.Fatal(err)
+	}
+	if c != 0 {
+		t.Fatalf("AWS CLI command was executed, expected return code 1, got %v", c)
```
mdelapenya, line 224: The tests are currently failing here. I'm debugging them locally atm

mdelapenya: Double checked that it fails locally. Will try to understand why and propose a fix

eddumelendez: This is interesting! the same config works in tc-java. I was wondering if the network name was making those tests flaky but look like is not. So, if I understood correctly the suggestion is to add the wait strategies as part of the container request, right? Although, the check is against the cli output. I think we can store the result in a file and check it later but doesn't seem to simplify the pr

modules/localstack/localstack.go
```diff
@@ -16,6 +16,7 @@ const defaultPort = 4566
 const defaultRegion = "us-east-1"
 const defaultVersion = "1.4.0"
```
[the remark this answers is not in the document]
eddumelendez: I would suggest to leave it as is for some minor versions and then bump to the latest.

modules/localstack/localstack_test.go
mdelapenya: I'd still create the CLI container, but instead of directly using the Exec method of the container, move it to a wait.ForExec(...).WithResponseMatcher(). If the container is not created without error, then the tests should fail

Does it sound good to you?

eddumelendez: I think we have the same issue 🤔

mdelapenya (approved):
LGTM, thanks!

modules/localstack/localstack_test.go
mdelapenya: 🤔 Could it be caused by the startup timeout not being enough depending on the machine?

I've seen some issues regarding that in the couchbase module in a non-deterministic manner. Restarting its CI build fixes it. Could it be the case for this?

PR #995, opened 2023-03-28
chore: render logos from the base location

mdelapenya:
## What does this PR do?
It removes the trailing slash from the SRC attribute of the social links in the website.

## Why is it important?
They do not get rendered when in a multi-deployment scenario (main, v0.19.0, v0.18.0, etc)


mdelapenya:
@leocross could you take a look? 🙏

mdelapenya:
Render URL generates the right logos. I'm merging this.

PR #1031, opened 2023-04-01
Added example for RabbitMQ

Tingaev:
## What does this PR do?

Added example for RabbitMQ

## Why is it important?

RabbitMQ doesn't have example yet

## Related issues

- Relates #636


examples/rabbitmq/rabbitmq_test.go
```diff
@@ -0,0 +1,38 @@
...
+	// Clean up the container after the test is complete
+	t.Cleanup(func() {
+		if err := container.Terminate(ctx); err != nil {
+			t.Fatalf("failed to terminate container: %s", err)
+		}
+	})
+
+	// perform assertions
+
+	amqpConnection, err := amqp.Dial(fmt.Sprintf("amqp://guest:guest@%s", container.endpoint))
```
mdelapenya, line 28: It's great having the amqp connection here. Could you extend the example on how to put/get values on/from the queue? 🙏

Tingaev: Ok
I'm just wondering what's the best way to do this?
1. Write and read in one connection
2. Or separate

mdelapenya: I'm not an expert in RabbitMQ, so I'd suggest adding the most educational example on how to do it.

mdelapenya:
FYI, we have submitted a fully-working module for RabbitMQ here: #1618

PR #1035, opened 2023-04-04
feat: support for wait.ForExec with response matcher

mdelapenya:
## What does this PR do?
It adds `WithResponseMatcher` to the wait.ForExec strategy, so that it behaves as in the wait.ForHTTP strategy.

## Why is it important?
Provide support for the use case where we want to run a command in a container, and verify the output of that command.

## Related issues

- Discovered while testing #994

### DIFF
--- a/wait/exec.go
+++ b/wait/exec.go
@@ -2,7 +2,10 @@ package wait
 
 import (
 	"context"
+	"io"
 	"time"
+
+	tcexec "github.com/testcontainers/testcontainers-go/exec"
 )
 
 // Implement interface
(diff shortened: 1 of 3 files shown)


wait/exec.go
```diff
@@ -72,13 +82,16 @@ func (ws *ExecStrategy) WaitUntilReady(ctx context.Context, target StrategyTarge
 		case <-ctx.Done():
 			return ctx.Err()
 		case <-time.After(ws.PollInterval):
-			exitCode, _, err := target.Exec(ctx, ws.cmd)
+			exitCode, resp, err := target.Exec(ctx, ws.cmd, tcexec.Multiplexed())
```
mdelapenya, line 85: We used multiplexed to remove the trim the trailing bytes from the Docker protocol (garbage). See #624

PR #1036, opened 2023-04-05
feat: provide container lifecycle hooks

mdelapenya:
## What does this PR do?
This PR opens the gates to the extension of a container, defining lifecycle hooks that can be leveraged by the consumers of the library to extend how a container is defined.

We have defined 8 lifecycle hooks:

1. Creating, executed right before the container is actually created by the Docker engine
2. Created, executed right after the container is created by the Docker engine
3. Starting, executed right before the container enters the Running state
4. Started, executed right after the container enters the Running state
5. Stopping, executed right before the container is stopped
6. Stopped, executed right after the container is stopper
7. Terminating, executed right before the container is terminated, removed from the Docker engine
8. Terminated, executed right after the container is terminated

For that, we are defining two types of hooks: one getting access to the container request (which defines the container), and the second getting access to the running container created by the aforementioned request. 

Each hook will be backed by an array of functions that returns an error and receive the Go context as first argument plus a second argument:
1) the container request for the `Creating` hook, as the representation of the container type does not exist at this moment.
2) the container for the rest of the hooks, as the representation of the container type has just been created.

If a hook returns an error, then the lifecycle will be stopped at that moment. For that reason, I'm not using `defer` to execute the Post hooks.

To add the hooks, we have defined a `LifecycleHooks` fiels in the ContainerRequest struct, so that it's possible to define how the container will behave when requesting a container.

With this implementation users would be able to define as many functions as they need to customize the container, e.g. passing an array of functions to the PreCreates field of the lifecycle, and/or an array of functions to the PostStarts field of the lifecycle. I've provided with a test demonstrating how it's possible to leverage the lifecycle: the tests adds a "log entry" everytime the life cycle hooks are called, and we assert that the life cycle is honoured (creating > created > starting > started > stopping > stopped > starting > started > terminating > terminated)

Finally, the docs have been updated to include the lifecycle hooks explanation in the "Create Container" section. Render URL here: https://deploy-preview-1036--testcontainers-go.netlify.app/features/creating_container/#lifecycle-hooks

## Why is it important?
As a user of the library creating a container, I want to be able to execute code just before/after the container is created/started/stopped/terminated.

## Follow-ups
I'd like to identify all code that happens out of the lifecycle hooks defined here but acting as a hook, wrapping them into hooks that will be part of the lifecycle

### DIFF
--- a/docs/features/creating_container.md
+++ b/docs/features/creating_container.md
@@ -87,6 +87,25 @@ func TestIntegrationNginxLatestReturn(t *testing.T) {
 }
 ```
 
+### Lifecycle callbacks
+
+_Testcontainers for Go_ allows you to define your own lifecycle callbacks for better control over your containers. You just need to define functions that return an error and implement the `testcontainers.Lifecycle` interface. The `testcontainers.Lifecycle` interface has the following methods:
+
+* `Creating` - called before the container is created
+* `Created` - called after the container is created
+* `Starting` - called before the container is started
+* `Started` - called after the container is started
+* `Stopping` - called before the container is stopped
+* `Stopped` - called after the container is stopped
+* `Terminating` - called before the container is terminated
+* `Terminated` - called after the container is terminated
+
+In the following example, we are going to create a container with a lifecycle callback that will print a message when any of the lifecycle hooks is called:
+
+<!--codeinclude--> 
+[Extending container with life cycle hooks](../../lifecycle_test.go) inside_block:lifecycleHooks
+<!--/codeinclude-->
+
 ### Advanced Settings
 
 The aforementioned `GenericContainer` function and the `ContainerRequest` struct represent a straightforward manner to configure the containers, but you could need to create your containers with more advance settings regarding the config, host config and endpoint settings Docker types. For those more advance settings, _Testcontainers for Go_ offers a way to fully customize the container request and those internal Docker types. These customisations, called _modifiers_, will be applied just before the internal call to the Docker client to create the container.
--- a/docs/features/creating_container.md
+++ b/docs/features/creating_container.md
@@ -87,9 +87,9 @@ func TestIntegrationNginxLatestReturn(t *testing.T) {
 }
 ```
 
-### Lifecycle callbacks
+### Lifecycle hooks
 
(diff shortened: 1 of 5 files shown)


docs/features/creating_container.md
```diff
@@ -87,6 +87,25 @@ func TestIntegrationNginxLatestReturn(t *testing.T) {
 }
 ```
 
+### Lifecycle callbacks
```
mdelapenya, line 90:
```suggestion
### Lifecycle hooks
```

docs/features/creating_container.md
```diff
@@ -87,6 +87,25 @@ func TestIntegrationNginxLatestReturn(t *testing.T) {
...
+
+_Testcontainers for Go_ allows you to define your own lifecycle callbacks for better control over your containers. You just need to define functions that return an error and implement the `testcontainers.Lifecycle` interface. The `testcontainers.Lifecycle` interface has the following methods:
```
mdelapenya, line 92:
```suggestion
_Testcontainers for Go_ allows you to define your own lifecycle hooks for better control over your containers. You just need to define functions that return an error and receive the Go context as first argument, and a `ContainerRequest` for the `Creating` hook, and a `Container` for the rest of them. The `testcontainers.ContainerLifecycleHooks` struct has the following methods:
```

docs/features/creating_container.md
```diff
@@ -87,6 +87,25 @@ func TestIntegrationNginxLatestReturn(t *testing.T) {
...
+* `Creating` - called before the container is created
+* `Created` - called after the container is created
+* `Starting` - called before the container is started
+* `Started` - called after the container is started
+* `Stopping` - called before the container is stopped
+* `Stopped` - called after the container is stopped
+* `Terminating` - called before the container is terminated
+* `Terminated` - called after the container is terminated
+
+In the following example, we are going to create a container with a lifecycle callback that will print a message when any of the lifecycle hooks is called:
```
mdelapenya, line 103:
```suggestion
In the following example, we are going to create a container using all the lifecycle hooks, all of them printing a message when any of the lifecycle hooks is called:
```

PR #1039, opened 2023-04-05
feat: Add ability to disable internal checks for `wait.HostPortStrategy`

### LINKED ISSUE #1038: [Enhancement]: Add ability to disable internal port check in `HostPortStrategy`
### Proposal

When using base images without shells such as Chainguard or Distroless the `wait.HostPortStrategy`internal checks always fail as a shell is not installed.

It should be possible to disable these internal checks and only check externally

RiverPhillips:
## What does this PR do?

Adds the ability to disable the internal check for the `wait.HostPortStrategy`

## Why is it important?

When using images without a shell the `wait.HostPortStrategy` fails due to the internal checks. This makes it possible to disable these internal checks

## Related issues

- Closes #1038


wait/host_port.go
```diff
@@ -171,7 +195,6 @@ func (hp *HostPortStrategy) WaitUntilReady(ctx context.Context, target StrategyT
 			return errors.New("/bin/sh command not executable")
```
mdelapenya, line 195: Wdyt if we create an error type for the 126 exit code, and we check for that error in L140 (https://github.com/testcontainers/testcontainers-go/pull/1039/files#diff-ad6cdf7fcae968ad7307fe3e8717bcbc7e43c461424e76ca3953cd215ac9395bR140), so that we always automatically detect the `/bin/sh` command is not executable? It will cause removing the new `internalCheckDisabled` fields/funcs, and will be transparent for the users. Probably a log warn?

RiverPhillips: I've updated that and I am now logging. I couldn't see any sort of structured logging library in use, but I may have missed it.

mdelapenya: You could use the default Logger instance from https://github.com/testcontainers/testcontainers-go/blob/main/logger.go#L13

wait/host_port.go
```diff
@@ -123,11 +126,25 @@ func (hp *HostPortStrategy) WaitUntilReady(ctx context.Context, target StrategyT
 		}
 	}
 
+	if err := externalCheck(ctx, ipAddress, port, target, waitInterval); err != nil {
+		return err
+	}
+
+	err = internalCheck(ctx, internalPort, target)
+	if err != nil && errors.Is(errShellNotExecutable, err) {
+		log.Printf("Shell not executable in container, only external port check will be performed")
```
mdelapenya, line 135: We probably want to add a `\n`
```suggestion
		log.Println("Shell not executable in container, only external port check will be performed")
```

mdelapenya:
Hey @RiverPhillips I'm going to apply the suggestion and merge this PR, as we are planning a release and I'd like to inclide this feature.

Thanks for taking the time to elaborate the issue alongside this PR 🙇

PR #1041, opened 2023-04-07
docs: document tc_host inside the networking section

sgloutnikov:
## What does this PR do?

Document the `TC_HOST` environment variable in the container networking section.

## Why is it important?

Our CI Kubernetes runners were incorrectly returning `localhost` via the `Host` call. After going through the `docker.go` code I noticed there is a manual override via the `TC_HOST` ENV variable. Document this so it's more visible in the future.

### DIFF
--- a/docs/features/networking.md
+++ b/docs/features/networking.md
@@ -44,6 +44,8 @@ It is normally advisable to use `Host` and `MappedPort` together when constructi
 [Getting the container host and mapped port](../../docker_test.go) inside_block:buildingAddresses
 <!--/codeinclude-->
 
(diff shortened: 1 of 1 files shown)


docs/features/networking.md
```diff
@@ -44,6 +44,8 @@ It is normally advisable to use `Host` and `MappedPort` together when constructi
 [Getting the container host and mapped port](../../docker_test.go) inside_block:buildingAddresses
 <!--/codeinclude-->
 
+Setting the `TC_HOST` environment variable overrides the host of the docker daemon where the container port is exposed. For example, `TC_HOST=172.17.0.1`.
```
mdelapenya, line 47:
```suggestion
!!! info
    Setting the `TC_HOST` environment variable overrides the host of the docker daemon where the container port is exposed. For example, `TC_HOST=172.17.0.1`.
```

PR #1042, opened 2023-04-10
feat: use an interface for container customization in modules

mdelapenya:
## What does this PR do?
This PR leverages a new interface for container customization: `ContainerCustomizer`, which has one single method: `Customize`.

Therefore we are adding a polymorphic behaviour for customization, which opens the gate to module authors to create an in-module struct implementing the interface and pass it as a customizer.

As a result, we are updating all existing modules to follow this design.

## Why is it important?
Accepting interfaces as options allow users to define their own customizers, making it more flexible the creation of modules.

## Related issues

- Continues #1036 #1037

### DIFF
--- a/docs/modules/index.md
+++ b/docs/modules/index.md
@@ -57,11 +57,18 @@ We are going to propose a set of steps to follow when adding types and methods t
 - Make sure a `RunContainer` function exists and is public. This function is the entrypoint to the module and will define the initial values for a `testcontainers.GenericContainerRequest` struct, including the image, the default exposed ports, wait strategies, etc. Therefore, the function must initialise the container request with the default values.
 - Define container options for the module leveraging the `testcontainers.ContainerCustomizer` interface, that has one single method: `Customize(req *GenericContainerRequest)`.
 - We consider that a best practice for the options is define a function using the `With` prefix, that returns a function returning a modified `testcontainers.GenericContainerRequest` type. For that, the library already provides with a `testcontainers.CustomizeRequestOption` type implementing the `ContainerCustomizer` interface, and we encourage you use this type for creating your own customizer functions.
-- At the same time, you could need to create your own container customizers for your module. Make sure they implement the `testcontainers.ContainerCustomizer` interface.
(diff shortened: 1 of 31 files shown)


modules/couchbase/couchbase.go
```diff
@@ -48,31 +54,57 @@ type CouchbaseContainer struct {
...
+	for _, srv := range initialServices {
+		opts = append(opts, withService(srv))
+	}
+
+	for _, opt := range opts {
+		opt.Customize(&genericContainerReq)
+
+		// transfer options to the config
+
+		if bucketCustomizer, ok := opt.(bucketCustomizer); ok {
```
mdelapenya, line 87: @fbiville this is how I'm transferring the functional option state to the container config

modules/couchbase/couchbase.go
```diff
@@ -581,19 +654,35 @@ func (c *CouchbaseContainer) checkAllServicesEnabled(rawConfig []byte) bool {
...
+	}
+}
+
+// withService creates a serviceCustomizer for the given service.
+// It's private to prevent users from creating other services than the Analytics and Eventing services.
+func withService(service Service) serviceCustomizer {
+	return serviceCustomizer{
+		enabledService: service,
 	}
+}
```
mdelapenya, line 673: @fbiville this is how I'm creating a custom implementation of a customizer for services, holding a state for the services to enable, and adding a Customize request method for the customizer which modifies the container request.

The services, because they are not part of the request, need to be transferred to the config above.

docs/modules/index.md
```diff
@@ -55,11 +55,59 @@ We are going to propose a set of steps to follow when adding types and methods t
...
+- Define container options for the module leveraging the `testcontainers.ContainerCustomizer` interface, that has one single method: `Customize(req *GenericContainerRequest)`.
+- We consider that a best practice for the options is define a function using the `With` prefix, that returns a function returning a modified `testcontainers.GenericContainerRequest` type. For that, the library already provides with a `testcontainers.CustomizeRequestOption` type implementing the `ContainerCustomizer` interface, and we encourage you use this type for creating your own customizer functions.
+- At the same time, you could need to create your own container customizers for your module. Make sure they implement the `testcontainers.ContainerCustomizer` interface. Defining your own customizer functions is useful when you need to transfer certain state that is not present at the `ContainerRequest` to the container, possibly using an intermediate Config struct.
```
mdelapenya, line 60: @fbiville explained here in the docs

PR #1058, opened 2023-04-15
feat: Add Redpanda module

### LINKED ISSUE #1033: [Enhancement]: Kafka module
### Proposal

testcontainers-java supports a built-in Kafka module, to allow users to manage Kafka without a lot of the boilerplate. Is this something you'd be willing to accept? If so, I'm more than willing to create a PR for it.

I suppose this might wait to wait for #1016.

weeco:
## What does this PR do?

This PR adds a module for Redpanda.

## Why is it important?

Redpanda is a Kafka API compatible streaming solution. Compared to Apache Kafka this starts very quickly, is easier to operate, doesn't require ZooKeeper, it's very lightweight and also comes with schema registry on board which makes it the ideal testcontainer for users that want to test their Kafka services.

## Related issues

- Closes #1033 

Run the provided go test.


eddumelendez (commented):
Thanks for the contribution, @weeco! In Testcontainers for Java, [Redpanda module](https://github.com/testcontainers/testcontainers-java/blob/6338802332df0874e04371d3bbcfbd6a16d79abf/modules/redpanda/src/main/java/org/testcontainers/redpanda/RedpandaContainer.java) was introduced and the image used to simplify the configuration was `docker.redpanda.com/vectorized/redpanda:v22.2.1`. We already use `docker.redpanda.com/redpandadata/redpanda`. I was wondering if, in order to keep consistency cross-language implementation we can do the same. Recently, there have been some changes in `Testcontainers for Go` in order to introduces container lifecycle, see https://github.com/testcontainers/testcontainers-go/pull/1036 https://github.com/testcontainers/testcontainers-go/pull/1037. I also wonder how the `redpanda.yaml.tpl` plays with the config we already have in `Testcontainers for Java` in order to enable more features.

weeco:
Hey @eddumelendez ,
thanks for taking a look at this PR. This module uses the right docker registry and the Java module should be updated to use this registry and version by default, too. In general I expect this Go module to be the source of truth, because the majority of our engineers use Go. Once this module will be used internally, it will be maintained by more people but it will take some extra time until we get there. I can ping our Java contributors to update the image in the Java module if needed.

Regarding container lifecycles - I totally forgot about that; I'll look into them later today. I think the PostStart hook may work here.

> I also wonder how the redpanda.yaml.tpl plays with the config we already have in Testcontainers for Java in order to enable more features.

In general I'd say that the Go module supports way more features. The Java module seems to rely on our `rpk` cli to take care of the configuration. Because this module is more complex and therefore requires more configuration this approach wouldn't work here anymore.

mdelapenya (dismissed):
@weeco thanks for this great module! Looks pretty neat!

I only have one small nit, that we could address in a follow up PR (or here, whatever you prefer): after merging with upstream, there is a need to run `go mod tidy`, which is trivial.

I'm also adding a suggestion for the docs, as we added a tag icon with the version in which a module was introduced.

Other than that, great job!! Thanks!

docs/modules/redpanda.md
```diff
@@ -0,0 +1,95 @@
+# Redpanda
+
+Redpanda is a streaming data platform for developers. Kafka API compatible. 10x faster. No ZooKeeper. No JVM!
+This testcontainer provides three APIs:
+
+- Kafka API
+- Schema Registry API
```
mdelapenya, line 7:
```suggestion
Not available until the next release of testcontainers-go <a href="https://github.com/testcontainers/testcontainers-go"><span class="tc-version">:material-tag: main</span></a>

## Introduction

Redpanda is a streaming data platform for developers. Kafka API compatible. 10x faster. No ZooKeeper. No JVM!
This Testcontainers module provides three APIs:
```

docs/modules/redpanda.md
```diff
@@ -0,0 +1,95 @@
+# Redpanda
+
+Redpanda is a streaming data platform for developers. Kafka API compatible. 10x faster. No ZooKeeper. No JVM!
+This testcontainer provides three APIs:
```
mdelapenya, line 4:
```suggestion
Not available until the next release of testcontainers-go <a href="https://github.com/testcontainers/testcontainers-go"><span class="tc-version">:material-tag: main</span></a>

## Introduction

Redpanda is a streaming data platform for developers. Kafka API compatible. 10x faster. No ZooKeeper. No JVM!
This Testcontainers module provides three APIs:
```

modules/redpanda/redpanda.go
```diff
@@ -0,0 +1,308 @@
...
+	"context"
+	_ "embed"
+	"fmt"
+	"os"
+	"text/template"
+	"time"
+
+	"github.com/docker/go-connections/nat"
+
+	"github.com/redpanda-data/redpanda/src/go/rpk/pkg/api/admin"
```
eddumelendez, line 14: usually, modules should not be couple to a client library.

weeco: This is just a HTTP client. I could write the HTTP calls on my own, but I don't think that would be better. The HTTP call is needed to create a service account when creating a Redpanda module with authentication enabled.

weeco:
Thanks for reviewing @mdelapenya and @eddumelendez .

I committed your suggested change, also tidied the Go modules and squashed my commits. I'm unsure how to address @eddumelendez's comment as this client is just a HTTP wrapper to create the Redpanda users once the cluster is started. This is only needed if the user uses authentication. If we remove this feature the user would be in charge of creating the users on their own via the Redpanda admin client that is being used.

mdelapenya:
> I'm unsure how to address @eddumelendez's comment as this client is just a HTTP wrapper to create the Redpanda users once the cluster is started

As commented online, I good step to avoid a dependency would be a little of copying :) I wonder if you could copy the code you need from the `rpk/admin` package into the module 🤔

weeco:
@mdelapenya I replaced the rpk dependency by adding a little helper client which currently only supports this specific endpoint for adding users. Appreciate another pass on this.

modules/redpanda/redpanda.go
```diff
@@ -0,0 +1,303 @@
...
+
+var (
+	//go:embed mounts/redpanda.yaml.tpl
+	nodeConfigTpl string
+
+	//go:embed mounts/bootstrap.yaml.tpl
+	bootstrapConfigTpl string
+
+	//go:embed mounts/entrypoint-tc.sh
+	entrypoint []byte
```
mdelapenya, line 26: I liked this to embed the static resources! 👏

mdelapenya (approved):
@weeco thanks for this contribution. I liked a lot reviewing the templating part, which I think will help with an upcoming support for different versions.

LGTM!

PR #1079, opened 2023-04-18
feat: use credential helper in docker config, even if auth is empty in .docker/config.json

### LINKED ISSUE #1078: [Bug]: Failed authentication from private repository with credHelpers
### Testcontainers version

0.19.0

Yes

### Host OS

macOS Intel

### Host arch

x86

### Go version

1.20

### Docker version

```shell
Client:
 Cloud integration: v1.0.31
 Version:           20.10.24
 API version:       1.41
 Go version:        go1.19.7
 Git commit:        297e128
 Built:             Tue Apr  4 18:21:21 2023
 Context:           desktop-linux
 Experimental:      true

Server: Docker Desktop 4.18.0 (104112)
 Engine:
  Version:          20.10.24
  Go version:       go1.19.7
  Git commit:       5d6db84
  Built:            Tue Apr  4 18:18:42 2023
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.18
  GitCommit:        2456e983eb9e37e47538f59ea18f2043c9a73640
 runc:
  Version:          1.1.4
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### Docker info

```shell
Client:
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.10.4)
  compose: Docker Compose (Docker Inc., v2.17.2)
  dev: Docker Dev Environments (Docker Inc., v0.1.0)
  extension: Manages Docker extensions (Docker Inc., v0.2.19)
  init: Creates Docker-related starter files for your project (Docker Inc., v0.1.0-beta.2)
  scan: Docker Scan (Docker Inc., v0.25.0)
  scout: Command line tool for Docker Scout (Docker Inc., v0.9.0)

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 2
 Server Version: 20.10.24
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 2456e983eb9e37e47538f59ea18f2043c9a73640
 init version: de40ad0
 Security Options:
  seccomp
   Profile: default
  cgroupns
 Kernel Version: 5.15.49-linuxkit
 OSType: linux
 Architecture: x86_64
 CPUs: 4
 Total Memory: 7.773GiB
 Name: docker-desktop
 ID: HB76:PKZ2:CB3Q:W7PV:HK2Q:VUCW:TQMH:KNTZ:FJBF:WNIM:DMU2:ZLPV
 Debug Mode: true
  File Descriptors: 42
  Goroutines: 49
  System Time: 2023-04-17T18:32:48.325708796Z
  EventsListeners: 7
 Labels:
 Experimental: false
 Insecure Registries:
  hubproxy.docker.internal:5555
  127.0.0.0/8
 Live Restore Enabled: false
```

### What happened?

I was trying to create a container from an image stored in a private GCP container registry but got an unauthorized error. I'm authenticating with the recommended method from google, using the [credential helper](https://cloud.google.com/container-registry/docs/advanced-authentication#gcloud-helper). Since my ~/.docker/config.json doesn't have any `auths` field, testcontainers-go doesn't get the credentials from the credential helper:
On line https://github.com/rokjoana/testcontainers-go/blob/main/docker_auth.go#L53 the code only goes through the array of `cfg.AuthConfigs` to get the credentials. If the array is empty, it doesn't call the method to get the credentials using the docker credential helper. This method should check also if the field `cfg.CredentialHelpers` exist if `cfg.AuthConfigs` is empty and call `dockercfg.GetRegistryCredentials()`.
 

### Relevant log output

```shell
2023/04/17 17:07:28 Failed to get image auth for eu.gcr.io. Setting empty credentials for the image: eu.gcr.io/kramphub-toolshed-shared/pubsub-emulator:2.0.0. Error is:credentials not found in native keychain
2023/04/17 17:07:29 Failed to pull image: Error response from daemon: unauthorized: You don't have the needed permissions to perform this operation, and you may have invalid credentials. To authenticate your request, follow the steps in: https://cloud.google.com/container-registry/docs/advanced-authentication, will retry
```

### Additional information

_No response_

rokjoana:
## What does this PR do?

If when retrieving the docker configuration, when there's no `auth` field on the user's docker config and if the user has configured credential helpers, the code will use the credential helpers to retrieve the auth information for the image repository instead of setting it to empty. 

## Why is it important?
Credential helpers are often used by third party docker repository providers to perform access control on the img repository.
In my case I'm using [GCP](https://cloud.google.com/container-registry/docs/advanced-authentication#gcloud-helper) but this is the case also for aws and azure. This change will extend the possibility of usage of images that are located in those repositories.

## Related issues
Closes #1078 

## How to test this PR
This is a tricky one to test because of the way the credentials are retrieved using these helpers. 
`cpuguy83/dockercfg` replicates the usage the docker CLI command: it calls an executable located in the PATH, which needs to output the credentials to stdout. This is tricky to emulate on go because one would need to use a script to link an executable that would mock the behaviour of a credential helper to the PATH when running the test. 
The other solution would involve having a private registry setup in a private registry provider such as GCP and use the credential helper of the provider. That would also imply some dependencies on the tests. 
Fortunately I had access to a private registry, so I was able to verify it by just using my fork in the project where I was using an image of the priv repository in my tests.


mdelapenya:
@rokjoana this LGTM at first sight, but wonder if you could cover the changes with some tests if possible (I could understand that mocking GCR would be difficult, but in any case, let's think if a test could be added).

In any case, thanks for the quick fix!

rokjoana:
I've described two test possibilities but both are tricky and dirty. I can try the dirtiest one, where we could define a mock credential helper executable that we build before the tests and add it to the path so that cpuguy83/dockercfg can pick it up when calling in the path using the name in the configurations.

mdelapenya:
> I've described two test possibilities but both are tricky and dirty. I can try the dirtiest one, where we could define a mock credential helper executable that we build before the tests and add it to the path so that cpuguy83/dockercfg can pick it up when calling in the path using the name in the configurations.

Dirty but working :) Thanks for taking the time to explain and develop the fix

mdelapenya:
BTW there is a file that has not been gofmt-ed, as golangci-lint is complaining in the GH action. Could you take a look 🙏 ?

mdelapenya:
@rokjoana I think it would be great to resolve this bug for the next release. Are you able to continue or would you need more time for it? I can take the lead and continue on top of your work if needed.

docker_auth.go
```diff
@@ -81,6 +81,19 @@ func getDockerAuthConfigs() (map[string]types.AuthConfig, error) {
 		cfgs[k] = ac
 	}
 
+	// in the case where the auth field in the .docker/conf.json is empty, and the user has credential helpers registered
+	// the auth comes from there
+	if len(cfg.AuthConfigs) == 0 && len(cfg.CredentialHelpers) > 0 {
+		for k, _ := range cfg.CredentialHelpers {
```
mdelapenya, line 87:
```suggestion
		for k := range cfg.CredentialHelpers {
```

docker_auth.go
```diff
@@ -81,6 +81,19 @@ func getDockerAuthConfigs() (map[string]types.AuthConfig, error) {
 		cfgs[k] = ac
 	}
 
+	// in the case where the auth field in the .docker/conf.json is empty, and the user has credential helpers registered
+	// the auth comes from there
+	if len(cfg.AuthConfigs) == 0 && len(cfg.CredentialHelpers) > 0 {
```
mdelapenya, line 86: I think this check is simpler now, as the body of the inner for/range loop will not be evaluated
```suggestion
	if len(cfg.AuthConfigs) == 0 {
```

moneymikeMD:
I have a similar but different problem and was hoping this PR would fix it. Sadly, it did not.

I tested this PR locally but my issue persisted. I am on a Mac (Ventura 13.4) and was getting an error:
```
Failed to get image auth for docker.io. Setting empty credentials for the image: docker.io/testcontainers/ryuk:0.5.1. Error is:credentials not found in native keychain
```
I dug into it a little more. My current theory is in the [docker_auth.go](../docker_auth.go) file, the [DockerImageAuth](../docker_auth.go#L17) function is looking for the registry for the image. The default registry on Mac is ` https://index.docker.io/v1/` while the ryuk image starts with [docker.io](http://docker.io/) . The means the check for [cfgs[registry]](../docker_auth.go#L26) is never finding a matching AuthConfig and thus returns `dockercfg.ErrCredentialsNotFound`.
So, possible solutions are to make the registry matching looser, or to make the fallback condition of [testcontainersdocker.ExtractRegistry](../internal/testcontainersdocker/images.go#L88) easier to hit.

rokjoana:
I do apologize for the delay. I had some personal stuff to deal with. I'll have a look at your comments and finish the PR soon.

mdelapenya:
> Failed to get image auth for docker.io. Setting empty credentials for the image: docker.io/testcontainers/ryuk:0.5.1. Error is:credentials not found in native keychain

@moneymikeMD this log message is printed out in the case there is an error getting the DockerAuth, BUT the pull will take place:

Please see https://github.com/testcontainers/testcontainers-go/blob/c0e1c32cc709ea176d6ba0ebf7b53421df2f86e7/docker.go#L962-L977

Is Ryuk effectively pulled into your system?

mdelapenya:
> I do apologize for the delay. I had some personal stuff to deal with. I'll have a look at your comments and finish the PR soon.

I totally get it, no worries at all 😊

rokjoana:
@mdelapenya I had a look and the only way for us to specifically test this is to, during the tests, register an executable in the PATH that would output the credentials of a mocked private registry repository (just like the docker credential helpers do).
Not only you can't do this on go, I think its something that's prone to be unsafe in the long run (the mocked binary can be easily tampered). I checked also the testcontainer libraries for Java and I couldn't find a direct test for this case (I'm also not sure if it was corrected in them). I don't know what we should do in this case but I leave up to you the decision of merging the fix.

rokjoana:
@moneymikeMD I can confirm that the log occurs but on my machine the images are pulled. As far as the issue is about, I think it could be something unrelated. Maybe we should continue the discussion on the issue itself.

mdelapenya (approved):
LGTM! Thanks for your patience during the review, it took a while, having summer and parental leave in between, but here we go 🚀

mdelapenya:
@moneymikeMD I'd appreciate if you can double check what @rokjoana asked [here](https://github.com/testcontainers/testcontainers-go/pull/1079#issuecomment-1606084696), in order to try to debug the error you are experiencing: 

> Could you share the format of your .docker/config.json?

It will need opening a different GH issue for tracking, I can do the linking if needed 🙏

Issue #511, opened 2022-08-29
[Enhancement]: Release docs from release tag


anuraaga:
### Proposal

I noticed this page in the docs that mentions the `Files:` pattern of precopying files

https://golang.testcontainers.org/features/copy_file/

But this isn't in the latest release tag v0.13.0 so it's confusing that the API can't be used. I think the docs are tagged together with the code in this repo, so perhaps they could be released from the tag as well.

mdelapenya:
Hi @anuraaga thanks for proposing this change, which I can share with you is a topic of high interest to us internally.

I agree with you it could lead to confusion and eventual disappointment, but there are reasons nowadays that impede us to do it in a different manner.

First, having the docs living in the same repository makes it difficult (although not impossible) to sync releases and docs. Difficult in the sense that we do not want to cause overhead for contributors making them send separated PRs for the docs, or even split existing ones including them, as it could slow down the review&merge cycle.

Second, the process to merge those docs PRs at release time would need more manual work for the OSS team, or even create automation that does not exist at this moment, to make those docs PRs into the release.

I want to share with you that revamping the docs is on our radar as a team, not only for the Go project but for all the available languages, so this process will be eventually updated.

In the meantime, I'd ask for understanding there could be inconsistencies that hopefully do not prevent you from using the product 🙏

anuraaga:
Thanks for the context @mdelapenya - I had just noticed that the `docs` folder is in this repository together with the code, which I believe is the website. Given that, isn't the versioning issue quite easy since a new feature can add docs in the exact same PR, and the code and docs would be tagged with a release together when ready? It looks fairly easy to keep in sync unless I'm missing something

leg100:
This has tripped me up as well (The docs say you can use `TESTCONTAINERS_RYUK_DISABLED=true` to disable the reaper but support for this was only added _after_ the latest tagged release and thus is not available when one installs your lib using `@latest`.

I noticed you're using Material for MkDocs. There is a simple solution to having the website show the latest tagged version by default and not main:

https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/

PR #1081, opened 2023-04-19
docs: include the testcontainers-go version in which a module was added

mdelapenya:
- chor: support for material design icons
- docs: include the version where the module started
- docsd: consistent starting of module docs, removing product images
- chore: generate new modules including the introduction
- docs: pulsar started in 0.19.0
- docs: use main as not-released tag for modules
- chore: generate modules including a disclaimer about not being released yet
- docs: extend info about the generated docs for a module
- chore: automate version-since in the non-released modules

## What does this PR do?
This PR does a few things:

- Updates each module in the docs to include the version in which it was released.
- New modules created with the code generator will display a message indicating they won't be available until the next release
- The release script modifies the modules page of those modules that are new to the release, using the new version as the version-since.
- The message in the docs page for a module will contain a tag icon next to the tc-go version, or `main` if the module has not been released yet.
- Extend the docs for both releasing and generating modules.

## Why is it important?
We currently have an issue with the docs, as they are deployed using the main branch. Therefore, anything in that branch will be deployed to the docs site. And this is causing noise as there will be code and modules that are not released yet but appearing in the docs.

With this PR we are trying to mitigate that, displaying a label/tag with the version of Testcontainers for Go where the module was added.

## Related issues

- Relates #511


docs/modules/couchbase.md
```diff
@@ -1,8 +1,10 @@
 # Couchbase
 
-<img src="https://cdn.worldvectorlogo.com/logos/couchbase.svg" width="300" />
```
mdelapenya, line 3: Removing the image for standarisation

docs/modules/vault.md
```diff
@@ -1,8 +1,10 @@
 # Vault
 
-<img src="https://cdn.worldvectorlogo.com/logos/vault-enterprise.svg" width="200" />
```
mdelapenya, line 3: Removing the image for standardization

mkdocs.yml
```diff
@@ -27,6 +27,10 @@ markdown_extensions:
     - pymdownx.snippets
     - toc:
         permalink: true
+    - attr_list
+    - pymdownx.emoji:
+        emoji_index: !!python/name:materialx.emoji.twemoji
+        emoji_generator: !!python/name:materialx.emoji.to_svg
```
mdelapenya, line 33: Needed because of https://squidfunk.github.io/mkdocs-material/reference/icons-emojis/#configuration

PR #947, opened 2023-03-21
fix: don't panic when logs waits for more than 5 seconds

martin-sucha:
This should fix https://github.com/testcontainers/testcontainers-go/issues/946

## What does this PR do?

This removes panic when logs endpoint takes more than 5 seconds to respond. The panic happened at least with podman when no new logs appear when using follow and since parameters.

We keep retrying until the context is canceled (the retry request would fail anyway) or the producer is stopped, whichever comes first. This makes the retry behavior consistent.

## Why is it important?

Panicking the whole test suite breaks all my tests if the container ends early or there is some temporary error fetching the logs.

## Related issues

- Closes #946
- Closes #606

PR #1085, opened 2023-04-24
fix: align header while log producers are reading

Dirk007:
## What does this PR do?

This PR fixes a bug in the log handling because incomplete reads were ignored. Fixed by replacing `io.ReadCloser.Read()` with `io.ReadFull()` to ensure that always a complete header and complete-log entries are read.

## Why is it important?

As `Read()` does not guarantee to read all the bytes requested, one unsuccessful read brought the header-handling out of sync  so random headers were read consequently after ignoring that first one. As a result "random"-length log-entries were tried to read (with exceptional sizes) as well.

TL;DR we must not ignore incomplete log-header reads.

## Related issues

None


mdelapenya:
Hi @Dirk007 thanks for the fix! We are about to merge https://github.com/testcontainers/testcontainers-go/pull/947, so please consider potential changes in the behaviour of the log consumers (at first sight, it does not, but I'd ask you for double checking within the scope of this PR).

Once #947 is merged, I'll attend this one, thanks! 🙏

Dirk007:
Hey Manuel. This PR does not interfer with #947 in any way.

Cheers, Dirk

docker.go
```diff
@@ -667,17 +667,22 @@ func (c *DockerContainer) StartLogProducer(ctx context.Context) error {
...
-					// this explicitly ignores errors
-					// because we want to keep procesing even if one of our reads fails
-					continue
+					if errors.Is(err, context.DeadlineExceeded) {
+						// Probably safe to continue here
+						continue
+					}
+					_, _ = fmt.Fprintf(os.Stderr, "container log error: %+v", err)
+					// if we would continue here, the next header-read will result into random data...
+					<-stop
```
mdelapenya, line 684: `stop` is not valid anymore in this goroutine scope, as #947 refactored to call it without passing the `c.stopProducer`. Besides, the `c.stopProducer`'s type was changed to `context.CancelFunc`.

docker.go
```diff
@@ -667,17 +667,22 @@ func (c *DockerContainer) StartLogProducer(ctx context.Context) error {
...
 						goto BEGIN
 					}
-					// this explicitly ignores errors
-					// because we want to keep procesing even if one of our reads fails
-					continue
+					if errors.Is(err, context.DeadlineExceeded) {
+						// Probably safe to continue here
+						continue
+					}
+					_, _ = fmt.Fprintf(os.Stderr, "container log error: %+v", err)
```
mdelapenya, line 682: Probably a nit, but do we want to skip the error here?

docker.go
Dirk007: I removed the stop writing right away in a hurry as I am in a workshop today. It's _not_ the ultimate solution as `r` stays open. Idea: `defer r.Close()`?

docker.go
mdelapenya: Mmmm well, if the error happens at the Printf scope, then I think it could be fine proceed as in https://github.com/testcontainers/testcontainers-go/pull/1085/files#diff-b68159b6168f099a639e8c6f96d75f23333ab0f0505ee22a98268748db17ae4cR681: probably safe to continue here

So this LGTM

docker.go
```diff
@@ -697,11 +701,16 @@ func (c *DockerContainer) StartLogProducer(ctx context.Context) error {
...
+				_, err = io.ReadFull(r, b)
 				if err != nil {
 					// TODO: add-logger: use logger to log out this error
 					_, _ = fmt.Fprintf(os.Stderr, "error occurred reading log with known length %s", err.Error())
-					continue
+					if errors.Is(err, context.DeadlineExceeded) {
+						// Probably safe to continue here
+						continue
+					}
+					// we can not continue here as the next read most likely will not be the next header
```
mdelapenya, line 712: wdyt if we use the logger here to explain that log consumer won't take place, and the reason why?

Dirk007: I tried to explain that by a common message now.

mdelapenya:
@Dirk007 I'm postponing the merge of this PR for two things on my side of the table:
1. dependabot updates 🤦 
2. This comment https://github.com/testcontainers/testcontainers-go/pull/947#issuecomment-1531133129

I'd appreciate your patience while debugging 2) 🙏

Dirk007:
ping ^^

mdelapenya:
> ping ^^

🤦  thanks for the ping!!! Will make it to this PR today (currently working from a train, hopefully we have it in before arriving to my destination 😅 ) 🙏

mdelapenya (approved):
LGTM, thanks!

mdelapenya:
@Dirk007 I'm merging this now, as I consider very difficult adding a test demonstrating the bug here. In any case, if you see it doable, let's do it in a follow-up PR 🤔 

Thanks for your patience while contributing to testcontainers-go!

Dirk007:
Unless the `io.Read/io.ReadAll` is not abstracted to a mockable object, I think testing for this is impossible. Maybe I'll find some time in the future to make a suggestion in this direction.

Thanks for merging and have a good trip @mdelapenya !

PR #1133, opened 2023-05-04
chore: run jobs if and only if the upstream workflow completed with success

mdelapenya:
- chore: run workflow if and only if the upstream workflow_run completed with success
- fix: rename job for modulegen pipeline

## What does this PR do?
It adds an IF condition on each job so that the upstream workflow needs to complete with success in order to trigger the current workflow.

## Why is it important?
A workflow run is triggered regardless of the conclusion of the previous workflow

See https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-a-workflow-based-on-the-conclusion-of-another-workflow

## Related issues

- Completes #1130

### DIFF
--- a/.github/workflows/bigtable-example.yml
+++ b/.github/workflows/bigtable-example.yml
@@ -1,6 +1,10 @@
 name: Bigtable example pipeline
 
 on:
+  workflow_run:
+    workflows: [Module generator pipeline]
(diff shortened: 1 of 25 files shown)


mdelapenya:
Superseded by #1414

PR #1165, opened 2023-05-10
First implementation of DockerRegitry module

davidesalerno:
## What does this PR do?

This PR is adding the new Docker Registry module. This module is wrapping the  registry image and it will be useful for those who needs to interact with a container registry to check if images are pushed or not after a build (i.e. local integration test of projects like kaniko or buildah).

### DIFF
--- a/modules/dockerregistry/dockerregistry_test.go
+++ b/modules/dockerregistry/dockerregistry_test.go
@@ -110,8 +110,6 @@ func TestDockerRegistryWithData(t *testing.T) {
 	t.Cleanup(cancel)
 }
 
-/**/
-
 func TestDockerRegistryWithAuth(t *testing.T) {
 	ctx := context.Background()
 	wd, err := os.Getwd()
@@ -136,6 +134,35 @@ func TestDockerRegistryWithAuth(t *testing.T) {
 	req, _ := http.NewRequest("GET", "http://"+ipAddress+":"+port.Port()+"/v2/_catalog", nil)
 	req.SetBasicAuth("testuser", "testpassword")
 	resp, _ := h.Do(req)
+	defer resp.Body.Close()
+	_, cancel := context.WithCancel(context.Background())
+	t.Cleanup(cancel)
+}
+
+func TestDockerRegistryWithAuthWithUnauthorizedRequest(t *testing.T) {
+	ctx := context.Background()
+	wd, err := os.Getwd()
+	assert.NoError(t, err)
+	container, err := RunContainer(ctx, WithAuthentication(wd+"/../../testdata/auth"))
+
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Clean up the container after the test is complete
+	t.Cleanup(func() {
+		if err := container.Terminate(ctx); err != nil {
+			t.Fatalf("failed to terminate container: %s", err)
+		}
+	})
+
+	port, ipAddress := getRegistryPortAndAddress(t, err, container, ctx)
+
+	// Let's simply check that the registry is up and running with a GET to http://localhost:5000/v2/_catalog
+	h := http.Client{}
+	req, _ := http.NewRequest("GET", "http://"+ipAddress+":"+port.Port()+"/v2/_catalog", nil)
+	resp, err := h.Do(req)
+	require.Equal(t, resp.StatusCode, 401)
 	defer resp.Body.Close()
 	_, cancel := context.WithCancel(context.Background())
 	t.Cleanup(cancel)


.github/workflows/module-registry.yml
```diff
@@ -0,0 +1,56 @@
...
+        uses: actions/setup-go@v3
+        with:
+          go-version: ${{ matrix.go-version }}
+        id: go
+
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v3
+
+      - name: modVerify
+        working-directory: ./modules/registry
```
mdelapenya, line 37: I think we should name the directory `dockerregistry`. Thoughts?

mdelapenya (commented):
Great contribution! Not sure if you already took a look at the existing tests including an authenticated docker registry (please see https://github.com/testcontainers/testcontainers-go/blob/main/docker_auth_test.go#L240), which I think it could be a great addition to this module, something like adding a customizer for `WithAuthentication` or similar.

Wdyt?

mdelapenya:
Not to mention that we could eventually merge this PR at this state, without the authentication, and create a follow-up PR with that work. At the end of the day, the module won't be available until the next release

modules/registry/go.mod
```diff
@@ -0,0 +1,53 @@
...
+)
+
+require (
+	github.com/Azure/go-ansiterm v0.0.0-20210617225240-d185dfc1b5a1 // indirect
+	github.com/Microsoft/go-winio v0.5.2 // indirect
+	github.com/cenkalti/backoff/v4 v4.2.0 // indirect
+	github.com/containerd/containerd v1.6.19 // indirect
+	github.com/cpuguy83/dockercfg v0.3.1 // indirect
+	github.com/dnephin/pflag v1.0.7 // indirect
+	github.com/docker/distribution v2.8.1+incompatible // indirect
```
mdelapenya, line 17: Let's add this security fix too, that was merged today in #1189 
```suggestion
	github.com/docker/distribution v2.8.2+incompatible // indirect
```

.github/workflows/module-registry.yml
mdelapenya: Indeed, that would be, using `DockerRegistry` as title maybe?

mdelapenya:
Hi @davidesalerno did you take a look at my comment here https://github.com/testcontainers/testcontainers-go/pull/1165#pullrequestreview-1420650610?

modules/dockerregistry/dockerregistry.go
```diff
@@ -0,0 +1,45 @@
...
+
+	container, err := testcontainers.GenericContainer(ctx, genericContainerReq)
+	if err != nil {
+		return nil, err
+	}
+
+	return &DockerRegistryContainer{Container: container}, nil
+}
+
+func WithAuthentication(dockerAuthConfig string) testcontainers.CustomizeRequestOption {
```
mdelapenya, line 38: Let's define the expectations on how to use this module: that would a user expect if the registry runs with authentication?
- should it be possible to pass an arbitrary `htpassw` file? See https://github.com/testcontainers/testcontainers-go/blob/main/testdata/auth/htpasswd
- should it be possible to load/push images to it?

I'd try to start with a container request like the one in this test: https://github.com/testcontainers/testcontainers-go/blob/main/docker_auth_test.go#L275-L299, allowing customising the auth file.

Then, the test for the module should check that's is posible to log in into that registry.

Does it make sense to you?

davidesalerno: Yes it is making sense to me.

Do we want also a realm as a parameter?

mdelapenya: I do not have a strong opinion... I think we can start simple, and iterate with a follow-up PR for the realm if you want. Or if you see it important, add it here. What is your preference? At the same time, if you consider it important for users of the module, then we can include it since the beginning.

mkdocs.yml
```diff
@@ -68,6 +68,7 @@ nav:
         - modules/pulsar.md
         - modules/redis.md
         - modules/redpanda.md
+        - modules/dockerregistry.md
```
mdelapenya, line 71: This entry will be probably located at L63 https://github.com/testcontainers/testcontainers-go/pull/1165/files#diff-98d0f806abc9af24e6a7c545d3d77e8f9ad57643e27211d7a7b896113e420ed2R63

.github/dependabot.yml
```diff
@@ -138,6 +138,12 @@ updates:
         interval: monthly
       open-pull-requests-limit: 3
       rebase-strategy: disabled
+    - package-ecosystem: gomod
```
mdelapenya, line 141: This entry will be probably located at L93: https://github.com/testcontainers/testcontainers-go/pull/1165/files#diff-dd4fbda47e51f1e35defb9275a9cd9c212ecde0b870cba89ddaaae65c5f3cd28R93

modules/dockerregistry/dockerregistry.go
```diff
@@ -0,0 +1,96 @@
...
+// DockerRegistryContainer represents the DockerRegistry container type used in the module
+type DockerRegistryContainer struct {
+	testcontainers.Container
+}
+
+// RunContainer creates an instance of the DockerRegistry container type
+func RunContainer(ctx context.Context, opts ...testcontainers.ContainerCustomizer) (*DockerRegistryContainer, error) {
+	req := testcontainers.ContainerRequest{
+		Image:        "registry:2",
+		ExposedPorts: []string{"5000:5000/tcp"},
```
mdelapenya, line 19: I think we do not want to always force a fixed port for the container:
```suggestion
		ExposedPorts: []string{"5000/tcp"},
```

modules/dockerregistry/dockerregistry.go
```diff
@@ -0,0 +1,96 @@
...
+				HostPath: htpasswdPath,
+			},
+			Target: "/auth",
+		}
+		req.Mounts = append(req.Mounts, htpasswdMount)
+	}
+
+}
+
+// WithData customizer that will retrieve a data directory and mount it inside the registry container
```
mdelapenya, line 66: Do we need a Docker mount for it? I wonder if copying the files/dir into the container, using the `req.Files` attribute would be something to consider.

@eddumelendez is there a preference when creating modules: using mounts or just copying files before the container start? 🤔

modules/dockerregistry/dockerregistry.go
```diff
@@ -0,0 +1,96 @@
...
+
+	}
+}
+
+// WithImage customizer that will override the registry image used
+func WithImage(image string) testcontainers.CustomizeRequestOption {
+	return func(req *testcontainers.GenericContainerRequest) {
+		req.Image = image
+	}
+}
```
mdelapenya, line 96: The core library already exposes a `testcontainers.WithImage` option, which can be used in the modules.
```suggestion
```

modules/dockerregistry/dockerregistry_test.go
```diff
@@ -0,0 +1,291 @@
...
+	}
+	defer resp.Body.Close()
+	_, cancel := context.WithCancel(context.Background())
+	t.Cleanup(cancel)
+}
+
+func TestDockerRegistryWithCustomImage(t *testing.T) {
+	ctx := context.Background()
+
+	container, err := RunContainer(ctx, WithImage("docker.io/registry:latest"))
```
mdelapenya, line 54:
```suggestion
	container, err := RunContainer(ctx, testcontainers.WithImage("docker.io/registry:latest"))
```

modules/dockerregistry/dockerregistry_test.go
```diff
@@ -0,0 +1,291 @@
...
+	redisContainer, err := testcontainers.GenericContainer(ctx, testcontainers.GenericContainerRequest{
+		ContainerRequest: req,
+		Started:          true,
+	})
+	require.Nil(t, err)
+	terminateContainerOnEnd(t, ctx, redisContainer)
+	_, cancel := context.WithCancel(context.Background())
+	t.Cleanup(cancel)
+}
+func getRegistryPortAndAddress(t *testing.T, err error, container *DockerRegistryContainer, ctx context.Context) (nat.Port, string) {
```
mdelapenya, line 268: I'd create a method at the registry container level, so that any user would be able to get both. Please take a look at how mysql or postgres modules expose the ConnectionString

modules/dockerregistry/dockerregistry_test.go
```diff
@@ -0,0 +1,291 @@
...
+	t.Setenv("DOCKER_AUTH_CONFIG", `{
+		"auths": {
+				"localhost:5000": { "username": "testuser", "password": "testpassword", "auth": "`+base64+`" }
+		},
+		"credsStore": "desktop"
+	}`)
+	ctx := context.Background()
+	wd, err := os.Getwd()
+	assert.NoError(t, err)
+	container, err := RunContainer(ctx, WithAuthentication(wd+"/../../testdata/auth"), WithData(wd+"/../../testdata/data"), WithImage("docker.io/registry:latest"))
```
mdelapenya, line 237: I'd copy/duplicate the resources that exist in the core library to the module, to avoid relative paths and decouple the module from the upstream core library just a little bit more.

We could decide later if we want to test that feature in the core or in the module.

modules/dockerregistry/tools/tools.go
```diff
@@ -0,0 +1,11 @@
+//go:build tools
```
mdelapenya, line 1: Please take a look at #1303, which removed the dependency with gotestsum at the package dependency level: it's now a responsibility of our CI to install it, simplifying our dependency tree.

Therefore, this file should be removed

modules/dockerregistry/dockerregistry.go
eddumelendez: The preference is to copy files.

.github/workflows/module-dockerregistry.yml
```diff
@@ -0,0 +1,56 @@
+name: DockerRegistry module pipeline
```
mdelapenya, line 1: Please take a look at #1414: this module file has been removed in favour of an array of modules in the `ci.yml` workflow.

modules/dockerregistry/dockerregistry.go
mdelapenya: @davidesalerno then I'd advocate for converting all the Mounts into Files 🙏

mdelapenya:
@davidesalerno we have sent https://github.com/testcontainers/testcontainers-go/pull/2341 creating a module for the docker registry. It supersedes this one, so I marked it to close this one.

I'd thank you a lot if you could take a look to that PR and come with your feedback, so it should cover your use case 🙏 

Cheers!

PR #1243, opened 2023-05-24
chore: Removes the refercence about docker 22.06 from the docker-compose docs

stubents:
Also, updates the replace directive for the compose module.

## What does this PR do?

As discussed in https://github.com/testcontainers/testcontainers-go/discussions/1231 the docs are out of sync with the actual replace directive.
Note that it might be necessary merge this PR only after the next release of docker-compose as it reflects the replace on the current v2 branch.

## Why is it important?

The docs should be updated so that new users don't get confused by the warning in the how to guide.


mdelapenya (approved):
Thanks for cleaning up the docs! We really appreciate it! ❤️

PR #1252, opened 2023-06-01
docs: fix passing env vars example

borod108:
Doc fix, the current given example does not compile, suggested a fix.


mdelapenya (approved):
Good catch! LGTM

I need to update the code snippets to directly use code that does exist in the current tests.

mdelapenya:
Thank you for the fix! 🙇 

The docs will automatically deployed to our site using Netlify, so it should be there in a couple of minutes

image

PR #1287, opened 2023-06-14
fix: use correct path for gotestsum output

sebastianbuechler:
## What does this PR do?

Use the correct path for the gotestsum output in the module-postgres pipeline

## Why is it important?

It's broken.

## Related issues

<!-- Recommended

- Relates #11246

<!-- Recommended
## How to test this PR

Check module-postgres pipeline

<!-- Optional
## Follow-ups

If this fix works we can check other pipelines as well if they need a fix too.

### DIFF
--- a/.github/workflows/module-postgres.yml
+++ b/.github/workflows/module-postgres.yml
@@ -52,5 +52,5 @@ jobs:
       - name: Test Summary
         uses: test-summary/action@4ee9ece4bca777a38f05c8fc578ac2007fe266f7
         with:
-          paths: "**/TEST-postgres*.xml"
+          paths: "**/TEST-unit*.xml"
         if: always()


.github/workflows/module-postgres.yml
```diff
@@ -52,5 +52,5 @@ jobs:
       - name: Test Summary
         uses: test-summary/action@4ee9ece4bca777a38f05c8fc578ac2007fe266f7
         with:
-          paths: "**/TEST-postgres*.xml"
+          paths: "**/TEST-unit*.xml"
```
mdelapenya, line 55: Your are right! https://github.com/testcontainers/testcontainers-go/blob/b76e5fb92637edeb419cf0d934a3d7bdb47176f5/commons-test.mk#L13 the make goal generates the file using unit as variable 🤦‍♂️

Could you apply this change to all the modules, including the template in the module generator? 🙏

sebastianbuechler: Done, only the `module-compose.yml` did not need a change as the junitfile was already properly configured. I also adjusted the template in the module generator.

sebastianbuechler: I just saw that my linter made changes to the `on` conditions of the workflows. Do you have an idea how to easily fix this as you probably will use a different linter?

mdelapenya: I'm using vscode + Go plugins. Also golangci-lint, both locally and at CI time. I'd suggest renting back the indents, keeping the jUnit change alone.

mdelapenya (commented):
Pending the final approval to the rollback of the indents, but other than that, LGTM! Thanks for the quick fix! 👏

.github/workflows/module-postgres.yml
sebastianbuechler: Prettier was linting the yml files per default on my setup and since the project does not have its own config it just used a default one.

sebastianbuechler:
> Pending the final approval to the rollback of the indents, but other than that, LGTM! Thanks for the quick fix! 👏

Reverted the linting. Thanks for the package!

mdelapenya:
The unit test that is failing is https://github.com/testcontainers/testcontainers-go/blob/main/modulegen/main_test.go#LL479C46-L479C55

It covers that the templates are properly rendered for each module. Would you mind updating the test too? 🙏

mdelapenya (approved):
LGTM! Thanks for detecting this subtle bug and proposing a quick fix! Much appreciated 👏

Issue #1249, opened 2023-05-28
[Bug]: Failing with "protocol not available, failed to create Docker provider" on Windows


danielorbach:
### Testcontainers version

0.20.1

### Using the latest Testcontainers version?

Yes

### Host OS

Windows

### Host arch

x86

### Go version

1.20

### Docker version

```shell
Client:
 Cloud integration: v1.0.31
 Version:           23.0.5
 API version:       1.42
 Go version:        go1.19.8
 Git commit:        bc4487a
 Built:             Wed Apr 26 16:20:14 2023
 OS/Arch:           windows/amd64
 Context:           default

Server: Docker Desktop 4.19.0 (106363)
 Engine:
  Version:          23.0.5
  API version:      1.42 (minimum version 1.12)
  Go version:       go1.19.8
  Git commit:       94d3ad6
  Built:            Wed Apr 26 16:17:45 2023
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.20
  GitCommit:        2806fc1057397dbaeefbea0e4e17bddfbd388f38
 runc:
  Version:          1.1.5
  GitCommit:        v1.1.5-0-gf19387a
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### Docker info

```shell
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.10.4
    Path:     C:\Program Files\Docker\cli-plugins\docker-buildx.exe
  compose: Docker Compose (Docker Inc.)
    Version:  v2.17.3
    Path:     C:\Program Files\Docker\cli-plugins\docker-compose.exe
  dev: Docker Dev Environments (Docker Inc.)
    Version:  v0.1.0
    Path:     C:\Program Files\Docker\cli-plugins\docker-dev.exe
  extension: Manages Docker extensions (Docker Inc.)
    Version:  v0.2.19
    Path:     C:\Program Files\Docker\cli-plugins\docker-extension.exe
  init: Creates Docker-related starter files for your project (Docker Inc.)
    Version:  v0.1.0-beta.4
    Path:     C:\Program Files\Docker\cli-plugins\docker-init.exe
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc.)
    Version:  0.6.0
    Path:     C:\Program Files\Docker\cli-plugins\docker-sbom.exe
[...]

PR #1289, opened 2023-06-15
fix: support detecting if k3s module is running from inside a Docker container

rfratto:
The k3s module previously didn't work properly inside of a Docker container; the caller would connect to its HTTPS server via the Docker network gateway, which wasn't included in the list of SANS from the TLS certificate exposed by the k3s container. This would cause requests to fail due to TLS validation errors:

```
Get "https://172.17.0.1:32776/api/v1": tls: failed to verify certificate: x509: certificate is valid for 10.43.0.1, 127.0.0.1, 172.17.0.6, ::1, not 172.17.0.1
```

This change determines what the DaemonHost address is prior to creating the container request, and then injects that address as one of the TLS SANs.

When not running in a Docker container, the k3s module continues to function as it did before, as the DaemonHost address in that scenario is "localhost".

I have successfully tested this PR both locally and verifying that it fixes the tests in our environment (see grafana/agent#4161).

### DIFF
--- a/modules/k3s/k3s.go
+++ b/modules/k3s/k3s.go
@@ -28,6 +28,11 @@ type K3sContainer struct {
 
 // RunContainer creates an instance of the K3s container type
 func RunContainer(ctx context.Context, opts ...testcontainers.ContainerCustomizer) (*K3sContainer, error) {
+	host, err := getContainerHost(ctx, opts...)
+	if err != nil {
+		return nil, err
+	}
+
 	req := testcontainers.ContainerRequest{
 		Image: "docker.io/rancher/k3s:v1.27.1-k3s1",
 		ExposedPorts: []string{
@@ -47,7 +52,7 @@ func RunContainer(ctx context.Context, opts ...testcontainers.ContainerCustomize
 		Cmd: []string{
 			"server",
 			"--disable=traefik",
-			"--tls-san=localhost",
+			"--tls-san=" + host, // Host which will be used to access the Kubernetes server from tests.
 		},
 		Env: map[string]string{
 			"K3S_KUBECONFIG_MODE": "644",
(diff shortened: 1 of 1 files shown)


modules/k3s/k3s.go
```diff
@@ -72,6 +77,31 @@ func RunContainer(ctx context.Context, opts ...testcontainers.ContainerCustomize
...
+	var req testcontainers.GenericContainerRequest
+	for _, opt := range opts {
+		opt.Customize(&req)
+	}
+
+	logging := req.Logger
+	if logging == nil {
+		logging = testcontainers.Logger
+	}
+	p, err := req.ProviderType.GetProvider(testcontainers.WithLogger(logging))
```
mdelapenya, line 91: I'm taking a look to check if it's possible to get a Provider instance from the library, instead of from the container request + customisers. In fact in https://github.com/testcontainers/testcontainers-go/pull/1251/files#diff-e9b78e9b34e14f076ba250ebd1042b315e63673569f82372a8064cece5f034bfR25 I'm deprecating the field in order to get them automatically from the environment.

In any case, I think this is good enough for now :)

mdelapenya (approved):
Thanks for this contribution! It makes sense to distinguish the execution context: from inside a container or not.

Apart from the PR, I'm interested in knowing more about your use case for this module: could you share with us a little bit of how you use it and which use cases it does resolve in your projects? 🙏

mdelapenya:
@mritunjaysharma394, as original contributor of the module, could you also take a look? 🙏

rfratto:
> Apart from the PR, I'm interested in knowing more about your use case for this module: could you share with us a little bit of how you use it and which use cases it does resolve in your projects? 🙏

Sure! We have a Kubernetes operator where we run a Kubernetes cluster to do integration tests. We previously embedded k3d as a library for bringing up a Kubernetes, but started to look for another solution when we realized importing k3d was causing some dependency hell issue. (We have a [ton](https://github.com/grafana/agent/blob/b7db5f3cddbecd42e7c8f3e5838f011702d271be/go.mod#L5-L582) of dependencies already).

We were already using testcontainers-go for bringing up a Vault server for integration tests, so [switching out](https://github.com/grafana/agent/pull/4147) our k3d code for the k3s module made a ton of sense for us. 

We'll be making more heavy use of the k3s module over time, as we're starting to add more Kubernetes controllers into our non-operator binary, where integration tests will help with having confidence in the stability of our code.

mdelapenya:
@rfratto we plan a release in the short term, but we need to resolve one issue with Windows first (see https://github.com/testcontainers/testcontainers-go/issues/1249)

At the moment, each module follows the core library release cycle, as we don't want to fall in the trap of maintaining multiple release branches (one per module), so the next full release will include these changes. 🙏

rfratto:
No worries! We'll use the main branch until is a release is available. Thanks for the update!

mdelapenya:
@rfratto just sharing that v0.21.0 is out, including the k3s module 😉

PR #1294, opened 2023-06-15
fix: prefer cross-platform default DOCKER_HOST

### LINKED ISSUE #1249: [Bug]: Failing with "protocol not available, failed to create Docker provider" on Windows

danielorbach:
Instead of blindly assuming a Linux operating system we can rely on the docker client package to tell us the appropriate default docker host.

## What does this PR do?

I've changed the default `DOCKER_HOST` socket from its Linux value by relying on Docker's client library to supply their cross-platform values.

I chose to parse their value as a URL (I say `parseURL` doing the same) to separate between the schema and the path.

I was defensive because I am unfamiliar with the codebase - that's the reason for the suffix/prefix validations. Also, upon errors (url parsing, or unknown schema) I fallback to the predefined default rendering the changes of this PR a no-op.

## Why is it important?

Windows users really want to use test-containers too! 🪟🐳 

Before release `v0.20.1` the host was not explicitly set to a default, hence it fell-back to the docker library's default anyways.

## Related issues

- Closes #1249 

## How to test this PR

I haven't been able to run all the tests because I currently use a Windows machine. However, I ran some container-based tests (that broke with release `v0.20.1`) on my machine and they have passed 🤗

### DIFF
--- a/internal/testcontainersdocker/docker_socket.go
+++ b/internal/testcontainersdocker/docker_socket.go
@@ -1,5 +1,12 @@
 package testcontainersdocker
 
+import (
+	"net/url"
+	"strings"
+
+	"github.com/docker/docker/client"
+)
+
 // DockerSocketSchema is the unix schema.
 var DockerSocketSchema = "unix://"
 
@@ -11,3 +18,29 @@ var DockerSocketPathWithSchema = DockerSocketSchema + DockerSocketPath
 
 // TCPSchema is the tcp schema.
 var TCPSchema = "tcp://"
+
+func init() {
+	const DefaultDockerHost = client.DefaultDockerHost
+
+	u, err := url.Parse(DefaultDockerHost)
+	if err != nil {
+		// unsupported default host specified by the docker client package,
+		// so revert to the default unix docker socket path
+		return
+	}
+
+	switch u.Scheme {
+	case "unix", "npipe":
+		DockerSocketSchema = u.Scheme + "://"
+		DockerSocketPath = u.Path
+		if !strings.HasPrefix(DockerSocketPath, "/") {
+			// seeing as the code in this module depends on DockerSocketPath having
+			// a slash (`/`) prefix, we add it here if it is missing.
+			// for the known environments, we do not foresee how the socket-path
+			// should miss the slash, however this extra if-condition is worth to
+			// save future pain from innocent users.
+			DockerSocketPath = "/" + DockerSocketPath
+		}
+		DockerSocketPathWithSchema = DockerSocketSchema + DockerSocketPath
+	}
+}


internal/testcontainersdocker/docker_socket.go
```diff
@@ -11,3 +18,24 @@ var DockerSocketPathWithSchema = DockerSocketSchema + DockerSocketPath
 
 // TCPSchema is the tcp schema.
 var TCPSchema = "tcp://"
+
+func init() {
```
mdelapenya, line 22: I had a local branch doing something similar, but creating a `docker_socket_windows.go` file, directly assigning the values we need:

```go
// DockerSocketSchema is the npipe schema.
var DockerSocketSchema = "npipe://"

// DockerSocketPath is the path to the docker socket under Windows systems.
var DockerSocketPath = "//./pipe/docker_engine"
```

I'm not sure if we should use this simpler approach, or instead delegate to the docker client as in this PR. What are your thoughts on this?

danielorbach: That would've solved it for sure! 🛬

I'm in favour of delegating to the Docker library because it honours their responsibility, in my mind.

However, you've already chosen to use a hard-coded value, therefore I'd support your approach blindly 😎 you definitely know best.

Let me know how you'd like to proceed. I'm available to make the desired changes.

mdelapenya: >I'm in favour of delegating to the Docker library because it honours their responsibility, in my mind.

I like this too. I think we can stick to it. Thanks for your support!

mdelapenya (dismissed):
Given the Docker client already does what I suggested in https://github.com/testcontainers/testcontainers-go/pull/1294/files#r1231789163, I think it's safer to honour the Docker client, as proposed in this PR.

Shaun approves 😎 

image

internal/testcontainersdocker/docker_socket.go
```diff
@@ -11,3 +18,24 @@ var DockerSocketPathWithSchema = DockerSocketSchema + DockerSocketPath
...
+		// unsupported default host specified by the docker client package,
+		// so revert to the default unix docker socket path
+		return
+	}
+
+	switch u.Scheme {
+	case "unix", "npipe":
+		DockerSocketSchema = u.Scheme + "://"
+		DockerSocketPath = u.Path
+		if !strings.HasPrefix(DockerSocketPath, "/") {
```
mdelapenya, line 36: One question: when does the `DockerSocketPath`, read from `u.Path` will not start with `/`?

Do you think this check is possible, given the docker client returns two possible values here?:
- npipe:////./pipe/docker_engine
- unix:///var/run/docker.sock

danielorbach: I proceeded with caution seeing as the code in this module depends on the presence of said `/`.
For the known environments I do not see how it should miss the slash, however I think this extra if-condition is worth to save future pain from innocent users.

mdelapenya: OK, let's keep it, although I'd appreciate adding there a comment explaining why 🙏 

Once added, this is ready to be shipped! 🛥️

danielorbach: Done 🗯️

danielorbach: You are correct, I've pulled latest changes from `main` however I forgot to commit the comment @mdelapenya

mdelapenya:
@mehdihadeli we are going to release a new version soon, hopefully at the end of this week

mehdihadeli:
Thanks

danielorbach:
> @danielorbach Hi, thanks for the fix, Did you publish a new package release for this fix?

@mdelapenya I haven't a clue how to do this 😅

PR #1301, opened 2023-06-23
Bump Redpanda container version to the latest release

RafalKorepta:
## What does this PR do?

Bump the latest container tag that mach latest Redpadna release https://github.com/redpanda-data/redpanda/releases/tag/v23.1.13.

## Why is it important?

It has bug fix included.

## Related issues

- None

### DIFF
--- a/modules/redpanda/redpanda.go
+++ b/modules/redpanda/redpanda.go
@@ -41,7 +41,7 @@ func RunContainer(ctx context.Context, opts ...testcontainers.ContainerCustomize
 	// Some (e.g. Image) may be overridden by providing an option argument to this function.
 	req := testcontainers.GenericContainerRequest{
 		ContainerRequest: testcontainers.ContainerRequest{
-			Image: "docker.redpanda.com/redpandadata/redpanda:v23.1.7",
+			Image: "docker.redpanda.com/redpandadata/redpanda:v23.1.13",
 			User:  "root:root",
 			// Files: Will be added later after we've rendered our YAML templates.
 			ExposedPorts: []string{


modules/redpanda/redpanda.go
```diff
@@ -11,7 +11,6 @@ import (
 
 	"github.com/docker/go-connections/nat"
 
-	"github.com/testcontainers/testcontainers-go"
```
mdelapenya, line 14: I believe this import must be kept here 😄

RafalKorepta: Oh my goland did something unexpected :) will fix it in the next push

RafalKorepta:
@mdelapenya Can I ask you for review? Should I do something more?

eddumelendez:
Hi @RafalKorepta, [Manuel is currently on parental leave](https://github.com/testcontainers/testcontainers-go/discussions/653). In general, I don't use to bump default images because don't want to change the behavior current users are having. The recommended way is to define always `withImage`. This is from my perspective in Testcontainers for Java. I see Manu did a comment on the PR and probably he is right with it. I'll leave the PR open but just wanted to let you know about Manu's status and my thoughts on this.

mdelapenya:
> The recommended way is to define always withImage

Indeed! We try to define a lower limit for the version with the default version in the code, so if we start bumping the version, it would increase that lower limit representing, creating the sense that previous version is not supported anymore.

mdelapenya:
Closing the PR because of https://github.com/testcontainers/testcontainers-go/pull/1301#issuecomment-1609895707

Nevertheless, thanks for opening the PR. Once we have a clear support matrix on modules and versions, then it would make sense to have this kind of PRs 🙏 

Hope you understand

Cheers!

mdelapenya:
Oh I've just realised, but you could add a test testing that new version if you wish!

PR #1305, opened 2023-06-26
ci(deps): add github-actions and pip package-ecosystems

mmorel-35:
## What does this PR do?

Generates the `dependabot.yml` configuration file

## Why is it important?

As the number of modules or package system growth it keeps the generation easy

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,5 +1,11 @@
 version: 2
 updates:
+    - package-ecosystem: github-actions
+      directory: /
+      schedule:
+        interval: monthly
+      open-pull-requests-limit: 3
+      rebase-strategy: disabled
     - package-ecosystem: gomod
       directory: /
       schedule:
@@ -140,6 +146,12 @@ updates:
       rebase-strategy: disabled
     - package-ecosystem: gomod
       directory: /modules/vault
+      schedule:
+        interval: monthly
+      open-pull-requests-limit: 3
+      rebase-strategy: disabled
+    - package-ecosystem: pip
+      directory: /
       schedule:
         interval: monthly
       open-pull-requests-limit: 3


Makefile
```diff
@@ -1,4 +1,5 @@
 include ./commons-test.mk
+include mk/dependabot.mk
```
mdelapenya, line 2: Could you move the directory to be hidden: `.mk`?

BTW I liked this approach and we should move all Make helper files into that folder.

mk/dependabot.mk
```diff
@@ -0,0 +1,27 @@
+SHELL:=/bin/bash
```
mdelapenya, line 1: Could you also document how to run these commands?

mdelapenya (commented):
Thanks for taking the time to create this PR, I appreaciate your efforts doing it.

The dependabot file is currently controlled by the `modulegen` module, which uses a Go template for the descriptor. Please see https://golang.testcontainers.org/modules/#interested-in-adding-a-new-module

I'd say this helper is useful, but would get into conflicts with the current code generation tools.

Do you see a way to combine both approaches

mk/dependabot.mk
mmorel-35: This is the way the SHELL can be defined, the $${dir:1} requires to be executed with /bin/bash that's why it's defined here. See [this link](https://www.gnu.org/software/make/manual/html_node/Choosing-the-Shell.html) for more details

Makefile
mmorel-35: Done !

mk/dependabot.mk
mdelapenya: Sorry, I meant in the PR description, how to execute the commands, for further reference

mmorel-35:
Hi @mdelapenya ,
The only "conflict" I could find with the modulegen module was the presence of the comment I added at the beginning of the dependabot file. So I removed it.

If I do understand modulegen's behavior, , it will append the config for a new module. So it's a one time task ? What appends if dependabot file is "theorivally" lost and you want to rebuild it ? Can you execute modulegen again and recreate the dependabot file.

What `make generate/dependabot` is recreate the file from the actual repository files. So if you add a module, remove a module or remove all dependencies of a module you can update the dependabot file in one time.

mdelapenya:
I'm starting to think that we need a `cmd` package, in which we add the different commands that we have:
- generate a module, moving the `modulegen` package into `cmd`
- regenerate module files: dependabot, ci.yml, mkdocs.yml

I think all the building blocks do exist in the current `modulegen` code, so I'd like to ask your opinion on refactoring this PR to use a programatic approach with Go code instead of shell/make.

mmorel-35:
Having a cmd package is a nice idea indeed.
I think it shall be addressed in another PR.
This one could just be focused on adding the github-actions and pip package-ecosystems.

What do you think ?

mdelapenya:
> Having a cmd package is a nice idea indeed. I think it shall be addressed in another PR. This one could just be focused on adding the github-actions and pip package-ecosystems.
> 
> What do you think ?

You mean keeping this PR's work for the dependabot update, without the makefile, right? If that's the case, LGTM

> I don't know much about this but are you aware of go.work files and have you considered using it ?

Indeed, I've been exploring locally with the go.work file, still need to get more used to it. But I see it as a nice improvement for working with the entire multi-project repo.

In regards of the `cmd` package, I'm able to peer and work with you in an implementation, if you want to contribute it.

mmorel-35:
> You mean keeping this PR's work for the dependabot update, without the makefile, right? If that's the case, LGTM


Yes I do

> In regards of the cmd package, I'm able to peer and work with you in an implementation, if you want to contribute it. 

I can give it a try during next week yes.

mdelapenya (approved):
thanks for this!

PR #1339, opened 2023-07-04
Add default labels when Ryuk is disabled

eddumelendez:
Currently, when Ryuk is disabled, no default labels are provided in the container.
This commit introduces the following default labels:

```
"org.testcontainers": "true",
"org.testcontainers.lang": "go",
"org.testcontainers.version": "0.22.0"
```

Also, for consistency with the other Testcontainers lang implementations,
label `org.testcontainers=true` is added.


reaper_test.go
```diff
@@ -55,11 +54,9 @@ func createContainerRequest(customize func(ContainerRequest) ContainerRequest) C
 		ReaperImage:  "reaperImage",
 		ExposedPorts: []string{"8080/tcp"},
 		Labels: map[string]string{
-			TestcontainerLabel:                "true",
-			TestcontainerLabelIsReaper:        "true",
-			testcontainersdocker.LabelReaper:  "true",
-			testcontainersdocker.LabelLang:    "go",
-			testcontainersdocker.LabelVersion: internal.Version,
+			TestcontainerLabel:               "true",
+			TestcontainerLabelIsReaper:       "true",
```
mdelapenya, line 58: I think we should deprecate this variable, as there exists an internal one with the same key: `testcontainersdocker.LabelReaper`

mdelapenya (dismissed):
I added a few comments, but other than that, LGTM, thanks!

mdelapenya (approved):
LGTM, thanks @eddumelendez ! 🚀

Issue #1357, opened 2023-07-10
[Bug]: hijacked connection returned by ContainerExecAttach is not closed


AlexanderYastrebov:
### Testcontainers version

v0.21.0

### Using the latest Testcontainers version?

Yes

### Host OS

Linux

### Host arch

x86

### Go version

1.20

### Docker version

```shell
-
```

### Docker info

```shell
-
```

### What happened?

Here https://github.com/testcontainers/testcontainers-go/blob/f5a4a541f71a324d1bf81f40700e5426259dbae6/docker.go#L485-L492
cli.ContainerExecAttach returns hijacked connections which [should be closed by the caller](https://pkg.go.dev/github.com/docker/docker/client#Client.ContainerExecAttach):
> It's up to the called to close the hijacked connection by calling types.HijackedResponse.Close.

Unfortunately Exec returns io.Reader https://github.com/testcontainers/testcontainers-go/blob/f5a4a541f71a324d1bf81f40700e5426259dbae6/container.go#L51

I think it should return io.ReadCloser that closes hijacked connection on Close()

See related #321

PR #1358, opened 2023-07-10
all: fix goroutine leaks

AlexanderYastrebov:
Fixes various goroutine leaks by moving around and adding missing Close calls.

Leaks were detected by github.com/AlexanderYastrebov/noleak by adding

```go
// go get github.com/AlexanderYastrebov/noleak@latest
// cat main_test.go
package testcontainers

import (
        "os"
        "testing"

        "github.com/AlexanderYastrebov/noleak"
)

func TestMain(m *testing.M) {
        os.Exit(noleak.CheckMain(m))
}
```

Not all leaks could be fixed e.g. due to #1357 therefore this change does not add leak detector, only fixes.

Updates #321

## Related issues

-


AlexanderYastrebov:
Fixed golangci-lint `Error return value of `n.Remove` is not checked (errcheck)`

docker.go
```diff
@@ -245,6 +245,14 @@ func (c *DockerContainer) Stop(ctx context.Context, timeout *time.Duration) erro
 
 // Terminate is used to kill the container. It is usually triggered by as defer function.
 func (c *DockerContainer) Terminate(ctx context.Context) error {
+	select {
+	// close reaper if it was created
+	case c.terminationSignal <- true:
+	default:
+	}
+
+	defer c.provider.client.Close()
```
mdelapenya, line 254: What will happen if a potential error happens while closing the client? Will it be swallowed?

docker_test.go
```diff
@@ -200,9 +200,11 @@ func TestContainerWithHostNetworkOptions_UseExposePortsFromImageConfigs(t *testi
 		t.Errorf("Expected server endpoint. Got '%v'.", err)
 	}
 
-	_, err = http.Get(endpoint)
+	resp, err := http.Get(endpoint)
 	if err != nil {
 		t.Errorf("Expected OK response. Got '%d'.", err)
+	} else {
+		resp.Body.Close()
 	}
```
mdelapenya, line 208: As in https://github.com/testcontainers/testcontainers-go/pull/1358/files#diff-037b83426e676aa3786c08b334b0db7bb70235792961cbe4ed3340bc0a60a7a8R728
```suggestion
	}
	defer resp.Body.Close()
```

mdelapenya (commented):
I left a few minor comments, will postpone the final review until we discuss them.

Other than that, LGTM. Thanks for your dedication!

docker_test.go
AlexanderYastrebov: I put it into the else branch because we can only call (deferred) resp.body.Close only if err is nil otherwise it would panic. Here error check uses t.Error that does not stop execution of the test unlike t.Fatal in the example you mention.

AlexanderYastrebov: Changed this to defer body close and test response status code like elsewhere.

docker.go
AlexanderYastrebov: TBH the whole approach of closing provider (to close client idle connections) in a hundred of places feels wrong.
Maybe there should be a terminal state of the container (after Terminate call?) that would close client and all other methods would return error if they are called after.

AlexanderYastrebov: TBH the whole approach of closing provider (to close client idle connections) in a hundred of places feels wrong.
Maybe there should be a terminal state of the container (after Terminate call?) that would close client and all other methods would return error if they are called after.

mdelapenya: I'm going to start a discussion for an upcoming V1 release. This kind of discussions are welcome for that thread

AlexanderYastrebov:
> Not all leaks could be fixed e.g. due to https://github.com/testcontainers/testcontainers-go/issues/1357 therefore this change does not add leak detector, only fixes.

There is another tricky case - reused containers.
In the TestParallelContainersWithReuse reused container [has reaper](https://github.com/testcontainers/testcontainers-go/pull/464#pullrequestreview-1564157142) that would not be closed:
https://github.com/testcontainers/testcontainers-go/blob/5011016830e1a1a2339b67b279b825e7e42bf1c8/parallel_test.go#L163-L169
Closing all containers produces error as there is only one real container.

In the TestGenericReusableContainer reused container also can not be closed but it calls Exec:
https://github.com/testcontainers/testcontainers-go/blob/5011016830e1a1a2339b67b279b825e7e42bf1c8/generic_test.go#L90-L94
which creates an idle connection in the client and the problem is that reused containers have their own provider:
https://github.com/testcontainers/testcontainers-go/blob/5011016830e1a1a2339b67b279b825e7e42bf1c8/generic.go#L125-L138
https://github.com/testcontainers/testcontainers-go/blob/5011016830e1a1a2339b67b279b825e7e42bf1c8/docker.go#L1139-L1149

docker_test.go
```diff
@@ -1473,7 +1493,7 @@ func TestContainerNonExistentImage(t *testing.T) {
 			t.Fatalf("err should be a ctx cancelled error %v", err)
 		}
 
-		terminateContainerOnEnd(t, ctx, c)
+		terminateContainerOnEnd(t, context.Background(), c) // use non-cancelled context
```
mdelapenya, line 1496: Good catch!

mdelapenya (approved):

AlexanderYastrebov:
@mdelapenya Thank you.

> would you participate in an eventual redesign

Just tag me and I'll think whether I could contribute anything meaningful.

PR #1360, opened 2023-07-12
feat(redpanda): Add option to enable topic auto-creation

ladislavmacoun:
## What does this PR do?

Adds an option to enable topic auto-creation

## Why is it important?

The option to enable automatic topic creation is particularly useful for testing environments, allowing users to produce data into topics without the need to create them beforehand.

### DIFF
--- a/modules/redpanda/options.go
+++ b/modules/redpanda/options.go
@@ -23,6 +23,9 @@ type options struct {
 	// You must use SCRAM-SHA-256 as algorithm when authenticating on the
 	// Kafka API.
 	ServiceAccounts map[string]string
+
+	// AutoCreateTopics is a flag to allow topic auto creation.
+	AutoCreateTopics bool
 }
 
 func defaultOptions() options {
@@ -32,6 +35,7 @@ func defaultOptions() options {
 		KafkaAuthenticationMethod:          "none",
 		SchemaRegistryAuthenticationMethod: "none",
 		ServiceAccounts:                    make(map[string]string, 0),
+		AutoCreateTopics:                   false,
 	}
 }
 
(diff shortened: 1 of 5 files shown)


eddumelendez:
Thanks for your contribution! So far, I haven't been able to find any reference about .bootstrap.yml. can you share why it is needed in both places? I only found the redpanda.yml [config](https://docs.redpanda.com/docs/21.11/cluster-administration/configuration/) but when enabled only there nothing happens. I would like to understand more about it.

ladislavmacoun:
> Thanks for your contribution! So far, I haven't been able to find any reference about .bootstrap.yml. can you share why it is needed in both places? I only found the redpanda.yml [config](https://docs.redpanda.com/docs/21.11/cluster-administration/configuration/) but when enabled only there nothing happens. I would like to understand more about it.

The bootstrapping mechanism is mentioned in the [production deployment documentation](https://docs.redpanda.com/docs/deploy/deployment-option/self-hosted/manual/production/production-deployment/#configure-bootstrapping), there is a reference in the [code](https://github.com/testcontainers/testcontainers-go/blob/main/modules/redpanda/redpanda.go#L227C29-L227C29) for bootstrap.yaml [configuration](https://docs.redpanda.com/docs/reference/cluster-properties/).

My understanding of this is that the bootstrap.yaml is responsible for the configuration of the entire cluster, and it is only assessed at the time of the initial start-up. On the other hand, adjustments can be made to the redpanda.yaml config as needed.

modules/redpanda/options.go
```diff
@@ -23,6 +23,9 @@ type options struct {
 	// You must use SCRAM-SHA-256 as algorithm when authenticating on the
 	// Kafka API.
 	ServiceAccounts map[string]string
+
+	// AutoCreateTopicsEnabled is a flag to allow topic auto creation.
+	AutoCreateTopicsEnabled bool
```
mdelapenya, line 28: I think the name is clear enough. Could you propagate this across all references? 🙏 
```suggestion
	AutoCreateTopics bool
```

ladislavmacoun: Changed 👍

mdelapenya (commented):
This PR LGTM! Although I'd like to postpone the final approval to @eddumelendez, who had some other concerns.

Thanks for your time and patience!

mdelapenya:
Great work! 💪

PR #1372, opened 2023-07-29
feat: add clickhouse module

anilsenay:
## What does this PR do?

This PR adds ClickHouse modules for golang testcontainers

## Why is it important?

It is important as part of a task of https://github.com/testcontainers/testcontainers-go/issues/636 to support ClickHouse modules too.

## Related issues

Relates to https://github.com/testcontainers/testcontainers-go/issues/636


modules/clickhouse/clickhouse.go
```diff
@@ -0,0 +1,184 @@
...
+func (c *ClickHouseContainer) ConnectionString(ctx context.Context, args ...string) (string, error) {
+	containerPort, err := c.MappedPort(ctx, nativePort)
+	if err != nil {
+		return "", err
+	}
+
+	host, err := c.Host(ctx)
+	if err != nil {
+		return "", err
+	}
```
mdelapenya, line 57: Do you think we can use the `ConnectionHost` method here?

docs/modules/clickhouse.md
```diff
@@ -0,0 +1,125 @@
...
+If you need to set a different ClickHouse Docker image, you can use `testcontainers.WithImage` with a valid Docker image
+for ClickHouse. E.g. `testcontainers.WithImage("clickhouse/clickhouse-server:23.3.8.21-alpine")`.
+
+#### Wait Strategies
+
+If you need to set a different wait strategy for ClickHouse, you can use `testcontainers.WithWaitStrategy` with a valid wait strategy
+for ClickHouse.
+
+!!!info
+The default deadline for the wait strategy is 60 seconds.
```
mdelapenya, line 51:
```suggestion
    The default deadline for the wait strategy is 60 seconds.
```

docs/modules/clickhouse.md
```diff
@@ -0,0 +1,125 @@
...
+options.
+
+<!--codeinclude-->
+
+[Custom Database initialization](../../modules/clickhouse/clickhouse_test.go) inside_block:customInitialization
+
+<!--/codeinclude-->
+
+!!!info
+The default values for the username is `default`, for password is `clickhouse` and for the default database name is `clickhouse`.
```
mdelapenya, line 77:
```suggestion
    The default values for the username is `default`, for password is `clickhouse` and for the default database name is `clickhouse`.
```

mdelapenya (commented):
Minor comments about docs format

modules/clickhouse/clickhouse_test.go
```diff
@@ -0,0 +1,274 @@
...
+	// Clean up the container after the test is complete
+	t.Cleanup(func() {
+		assert.NoError(t, container.Terminate(ctx))
+	})
+
+	connectionString, err := container.ConnectionHost(ctx)
+	assert.NoError(t, err)
+
+	conn, err := ch.Open(&ch.Options{
+		Addr: []string{connectionString},
```
mdelapenya, line 190: Ditto:
```suggestion
	connectionHost, err := container.ConnectionHost(ctx)
	assert.NoError(t, err)

	conn, err := ch.Open(&ch.Options{
		Addr: []string{connectionHost},
```

mdelapenya (commented):
Could you please take a look at the variable names? 🙏

modules/clickhouse/clickhouse_test.go
```diff
@@ -0,0 +1,274 @@
...
+
+	// connectionString {
+	connectionString, err := container.ConnectionString(ctx, "debug=true")
+	assert.NoError(t, err)
+	// }
+
+	opts, err := ch.ParseDSN(connectionString)
+	assert.NoError(t, err)
+
+	conn, err := ch.Open(opts)
```
mdelapenya, line 137: Wdyt if you define a different test function for the code from here to the end of the current test function? Something to represent that it creates tables, inserts data and retrieves it: `TestClickHouseCRUD`.

At the same time, I see duplication in the query constants and the init-db script. Why not using the `WithInitScripts`. I can imagine you want to the the `conn.Exec` method works, right? I'd call a `DELETE` or `UPDATE`  (I'm not a ClickHouse user at all!! 🙏 ) to check the `Exec` call, and then the `conn.Query` method.

But this is not a blocker, I'd like to know your thoughts on this.

modules/clickhouse/clickhouse_test.go
```diff
@@ -0,0 +1,274 @@
...
+
+		assert.NoError(t, err)
+
+		data = append(data, r)
+	}
+
+	assert.Len(t, data, 1)
+}
+
+func TestClickHouseWithConfigFile(t *testing.T) {
```
mdelapenya, line 219: I'm missing a test with the YAML config file, as it's exposed in the API. The YAML file would be added to the testdata dir, and I'd also recommend (to avoid much duplication) to enclose the exec/query code block into a function that would be called else where. Thoughts?

mdelapenya (commented):
This PR is great and almost perfect to be merged. Please read my comments, as I'd love to get your insights before.

Thanks you so much for your enormous work here 👏

modules/clickhouse/clickhouse_test.go
```diff
@@ -0,0 +1,274 @@
...
+	})
+	assert.NoError(t, err)
+	assert.NotNil(t, conn)
+	defer conn.Close()
+
+	err = conn.Ping(context.Background())
+	assert.NoError(t, err)
+}
+
+func TestClickHouseIpPort(t *testing.T) {
```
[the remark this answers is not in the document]
anilsenay: First of all, thank you for your comments 🙏

You are right, `TestClickHouseConnectionHost` would be better function name. I just wanted to be clear about `host` means `ip:port` but not needed I guess.

I left duplicate crud operations on purpose. I just want to show that all methods works in same way with same operations. But I am not sure if it is necessary actually. Maybe it would better to create another test for CRUD operations just like you mentioned next comment. I am a little confused there 😅

modules/clickhouse/clickhouse_test.go
anilsenay: About `TestClickHouseCRUD`, I wrote about it in my previous comment.

I tested `WithInitScripts` in `TestClickHouseWithInitScripts` function; and in other tests, I prefer to not use it so I wanted to use Exec like using it in a regular application code. I think testing `Exec` is fine now because I wanted to do similar work as in other test functions. I don't think testing `Exec` with `DELETE` or `UPDATE` makes any difference anyway. 

They are my thoughts, but I would like to know if you think I am wrong in some place.

modules/clickhouse/clickhouse_test.go
anilsenay: To be honest, I thought that if it works with `config.xml`, it would probably work with YAML too 😆 

I would add it, not problem.

I am aware about duplications but like I said in my first comment I left them on purpose. But I will work on it if it seems noisy

modules/clickhouse/clickhouse_test.go
mdelapenya: > I prefer to not use it so I wanted to use Exec like using it in a regular application code

This makes total sense, thanks for the explanation :) 

And I agree with you: I prefer duplication in tests for the sake of clarity. Let's keep them as they are.

anilsenay:
@mdelapenya I refactor tests, and I decided to avoid duplications after all then create functions to perform crud :D

Would you take a look at my last commit?

modules/clickhouse/clickhouse.go
```diff
@@ -0,0 +1,179 @@
...
+
+// RunContainer creates an instance of the ClickHouse container type
+func RunContainer(ctx context.Context, opts ...testcontainers.ContainerCustomizer) (*ClickHouseContainer, error) {
+	req := testcontainers.ContainerRequest{
+		Image: defaultImage,
+		Env: map[string]string{
+			"CLICKHOUSE_USER":                      defaultUser,
+			"CLICKHOUSE_PASSWORD":                  defaultUser,
+			"CLICKHOUSE_DB":                        defaultDatabaseName,
+			"CLICKHOUSE_DEFAULT_ACCESS_MANAGEMENT": "1",
```
eddumelendez, line 150: Do you know what is the effect of this? I can not find any in the [docs](https://hub.docker.com/r/clickhouse/clickhouse-server/). I did some tests in java and it is like has no effect.

anilsenay: Which part did you ask for? All of these env variables is written in [docs](https://hub.docker.com/r/clickhouse/clickhouse-server/) you linked.

eddumelendez: I found the [PR](https://github.com/ClickHouse/ClickHouse/pull/20139). I think the env var should be set by the user if needed and not update the default value provided by the image.

anilsenay: Ah, I guess I misunderstand purpose of this variable acording to [docs](https://hub.docker.com/r/clickhouse/clickhouse-server/). You are right this setting does not needed to set as default.

eddumelendez (approved):

PR #1361, opened 2023-07-17
fix: upgrade docker client

Slijkhuis:
## What does this PR do?

Upgrades the `github.com/docker/docker` package.

~~The `AuthConfig` type was moved: https://github.com/moby/moby/pull/43885.~~ 👈 not relevant since I'm not upgrading to v24 anymore, but rather use the backported fix on v23.

Possibly https://github.com/testcontainers/moby-ryuk must be upgraded as well. 

## Why is it important?

You cannot use `testcontainers-go` with Go 1.20.6 currently, this PR should fix that.

## Related issues

- Closes #1359

PR #1394, opened 2023-08-01
chore(deps): bump docker dependencies to v24.0.5+incompatible

### LINKED ISSUE #1359: [Bug]: go 1.20.6: "http: invalid Host header, host port waiting failed"
### Testcontainers version

0.21.0

Yes

### Host OS

OSX and Linux

### Host arch

OSX arm64, Linux amd64

### Go version

1.20.6

### What happened?

With Go 1.20.6, I start getting an error "http: invalid Host header, host port waiting failed" when setting up new containers. From doing some debugging, it seems to be coming from the ryuk setup. This appears to be the relevant change in the standard library: https://github.com/golang/go/issues/61076, since it introduces that check that returns "invalid Host header". From doing the debugging in the screenshot below, the Host that Go is choking on is "/var/run/docker.sock".

image
image
image

### Relevant log output

_No response_

### Additional information

_No response_

mdelapenya:
## What does this PR do?
It bumps the core library and all the modules to Docker v24.0.5+incompatible

## Related issues

- Supersedes and Closes #1361
- Closes #1359

### DIFF
--- a/examples/bigtable/go.mod
+++ b/examples/bigtable/go.mod
@@ -25,7 +25,7 @@ require (
 	github.com/containerd/containerd v1.6.19 // indirect
 	github.com/cpuguy83/dockercfg v0.3.1 // indirect
 	github.com/docker/distribution v2.8.2+incompatible // indirect
-	github.com/docker/docker v23.0.5+incompatible // indirect
+	github.com/docker/docker v24.0.5+incompatible // indirect
 	github.com/docker/go-connections v0.4.0 // indirect
 	github.com/docker/go-units v0.5.0 // indirect
 	github.com/envoyproxy/go-control-plane v0.11.0 // indirect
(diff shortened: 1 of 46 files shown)


docker_auth_test.go
```diff
@@ -42,7 +42,7 @@ func TestGetDockerConfig(t *testing.T) {
 		require.Nil(t, err)
 		require.NotNil(t, cfg)
 
-		assert.Equal(t, 1, len(cfg.AuthConfigs))
+		assert.Greater(t, len(cfg.AuthConfigs), 0)
```
mdelapenya, line 45: I tried this in a codespace, and the value was always 3, for the three entries that are present in that machine:
- docker.pkg.github.com
- ghcr.io
- https://index.docker.io/v1/

AlexanderYastrebov: FWIW I have it empty on my machine and saw this test failing
```
~$ jq .auths .docker/config.json 
{}
```

mdelapenya: Mmm interesting... I probably have it populated locally because of having run `docker login`. I guess the test would pass if you log in. But in any case, the test should be idempotent no matter the state.

Would you be willing to participate in a quick fix for that?

AlexanderYastrebov: Created #1413

davideme:
Any ETA on a release that include the fix?

mdelapenya:
> Any ETA on a release that include the fix?

Hi @davideme I'm planning it for the end of this week, so hopefully we can have it tomorrow 🤞

PR #1413, opened 2023-08-02
test: fix TestGetDockerConfig

AlexanderYastrebov:
## What does this PR do?

Do not assume non-empty auths section in local config file but test that getDockerConfig returns default config when DOCKER_CONFIG env variable is not set.

Also remove t.Cleanup calls that restore env variables because t.Setenv restores them itself.

## Why is it important?

Fixes failing test that depends on local environment.

## Related issues

- Relates #1394


mdelapenya (approved):
LGTM, thanks!!!

PR #1375, opened 2023-08-01
chore(ci): add a Windows pipeline for running the tests

mdelapenya:
- chore(ci): add a GH pipeline to trigger the windows worker
- chore(ci): remove windows worker from the regular pipeline

## What does this PR do?
It adds two GH workflows: one for listening to GH comments as commands:
- It will listen to the `/windows-test` command as a GH comment for maintainers, triggering the pipeline for windows, using the Azure self-hosted workers.
- A second workflow with the real test execution on our self-hosted workers on Azure.

It also removes the windows build from the main pipeline, as it's not needed anymore.

## Why is it important?
Test on Windows is key for many users, so we must be aware of potential bugs on that platform, which is widely used by our community.

## Related issues

- Relates https://github.com/testcontainers/testcontainers-go/issues/1367
- Relates https://github.com/testcontainers/testcontainers-go/issues/948
- Closes https://github.com/testcontainers/testcontainers-go/issues/837

PR #1415, opened 2023-08-04
chore(ci): set LongPathsEnabled on WIndows workers

mdelapenya:
## What does this PR do?
It executes a Powershell command on the Windows worker to enable long paths.

This is a temporary fix until the target machine have it as a built-in

## Why is it important?
The git checkout fails and we think this is the most plausible solution

## Related issues

- Continues #1375

## How to test this PR

In a PR, i.e. this one, add the `/windows-test` GH comment so that the Windows workflow gets triggered.

### DIFF
--- a/.github/workflows/ci-windows.yml
+++ b/.github/workflows/ci-windows.yml
@@ -25,6 +25,11 @@ jobs:
               context: 'Windows tests',
             })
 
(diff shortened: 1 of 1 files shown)


mdelapenya:
Because the code needs to be in the main branch to be accesible, I'm merging this one now

PR #1441, opened 2023-08-06
internal/testcontainersdocker: close unused client

AlexanderYastrebov:
## What does this PR do?

Closes unused docker client.

## Why is it important?

Docker client uses idle connections and if it is not closed it triggers goroutine leak detector.
I've created #1358 previously but while it is in discussion I propose to fix this obvious case which would also be enough for our project, see https://github.com/zalando/skipper/pull/2436

## Related issues

- Relates #1358


mdelapenya (approved):
Good catch!

AlexanderYastrebov:
Thank you.

PR #1442, opened 2023-08-07
chore: remove temporary workaround for windows

mdelapenya:
## What does this PR do?
Removes the temporary workaround to apply long paths to Windows 

## Why is it important?
It failed so we added to the self-hosted machine.


mdelapenya:
Merging because it won't affect the tests.

PR #1446, opened 2023-08-07
chore: run dependabot on sundays

mdelapenya:
## What does this PR do?
This PR does:
- updates the dependabot file to run only monthly on Sundays
- updates the test assertions for dependabot generation
- removes an outdated if clause when generating modules
- updates the code generation to use the `os` package instead of deprecared APIs in `ioutil`
- autogenerates dependabot code properly, moving github-actions and pip to the first two elements

## Why is it important?
Idempotency in the code generation, and more predictability on when the dependabot updates will be called, using the weekend window to avoid affecting the rest of the teams.


cristianrgreco (approved):
Thanks @mdelapenya! 🙏

PR #1454, opened 2023-08-07
[WIP] Create Cmd

mmorel-35:
Hi @mdelapenya ,
Here is a first draft .
I had to update devcontainer image as it is not working as.

I created a cmd but the main.go file needs to be in a main package so it cannot stay in the root . Maybe create another module like module-admin and move the pkg and cmd I created.

This first command has been created with the help of cobra-cli at first but was refarctored. It allows to append an update to the dependabot config file with this:

```sh
go run main.go dependabot addExample --name newmodule
```

Two things I noticed. There isno mecanism to handle duplication with an already existing update (directory and package-ecosystem).

The sorting mecanism needed to be update tohandle package-ecosystem as you can see.

What do you think ?


main.go
```diff
@@ -0,0 +1,7 @@
+package main
```
mdelapenya, line 1: I'd move this file to the cmd package, as we do not want to interfere with the `testcontainers` package of the library

go.mod
```diff
@@ -14,6 +14,7 @@ require (
 	github.com/magiconair/properties v1.8.7
 	github.com/moby/term v0.5.0
 	github.com/opencontainers/image-spec v1.1.0-rc4
+	github.com/spf13/cobra v1.7.0
```
mdelapenya, line 17: As a consequence of my previous comment, cobra would not be added as dependency to the core, but instead to the `cmd` package/module

mmorel-35:
Hi @mdelapenya,

I have separated elements between this three folders
* _templates : provides the templates with a dedicated folder per type of elements
*  cmd: provides the cobra commant
* package: provides the different packages for different elements

The different elements are 
* dependabot
* make
* mkdocs 
* modules: Go modules 
* templates: Utils to generate templates
* workflows: Handles Github Wofkflows configuration

mmorel-35:
Let’s not merge this. I suggest to do things more progressively , one package at a time so the jump is going to be easy

PR #1459, opened 2023-08-08
Support groups in dependabot updates

mmorel-35:
Support groups in dependabot updates

### DIFF
--- a/modulegen/dependabot.go
+++ b/modulegen/dependabot.go
@@ -16,6 +16,12 @@ type DependabotConfig struct {
 	Updates Updates `yaml:"updates"`
 }
 
+type Group struct {
+	Patterns []string `yaml:"patterns"`
+}
+
+type Groups map[string]Group
+
 type Schedule struct {
 	Interval string `yaml:"interval"`
 	Day      string `yaml:"day"`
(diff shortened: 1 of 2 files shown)


mdelapenya:
@mmorel-35 thanks for this! Do you envision adding groups now? I'd like to read more about it and create logical groups (i.e. all the Google Cloud emulators could go together, or identify from previous PRs what deps are present in multiple modules).

Would like to help me in identifying those groups?

mmorel-35:
I don't want to add them necessarily now but in case you would try to add them manually in the dependabot file this would avoid an unexpected deletion when a new example or module is added

mdelapenya:
I think we could start with a group per module, having a wildcard `*` for all the dependencies in the module. This will create a single PR per module, which is much better than the current situation.

An edge case could be a dependency that is present in multiple modules... but we already live in that world

mmorel-35:
What do you think about this @mdelapenya  ?

modulegen/dependabot.go
```diff
@@ -39,6 +46,11 @@ func NewUpdate(example Example) Update {
 			Interval: updateSchedule,
 			Day:      "sunday",
 		},
+		Groups: Groups{
+			"all": Group{
+				Patterns: []string{"*"},
```
mdelapenya, line 51: Could you double check that this will generate:
```yaml
      groups:
        all:
          patterns:
            - "*"
```

and no:
```yaml
      groups:
        all:
          patterns:
            - *
```
🙏 

From modules dir:
```shell
go run . --name questdb --image "questdb/questdb:latest" --title QuestDB --as-module
```

mmorel-35: Ok, that shall be alright

mdelapenya:
Merge, thanks! 👏👏👏

PR #1460, opened 2023-08-09
Simplify dependabot updates sorting

mmorel-35:
Use the sorter capabilities to handle the sorting.
It uses the package-ecosystem and the directory. 
I know that the result is a bit different from the actual.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,14 +8,14 @@ updates:
         day: sunday
       open-pull-requests-limit: 3
       rebase-strategy: disabled
-    - package-ecosystem: pip
+    - package-ecosystem: gomod
       directory: /
       schedule:
         interval: monthly
         day: sunday
       open-pull-requests-limit: 3
       rebase-strategy: disabled
-    - package-ecosystem: gomod
+    - package-ecosystem: pip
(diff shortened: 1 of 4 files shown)


mdelapenya:
IIUC it will use gh-actions > gomod > pip, so the code is sorting by package-ecosystem too. I like it! TBH I considered enforcing the extras (gha, pip) first and then the go code, but thinking about it twice, why? 🙈 

I do like this approach mode, as it more deterministic and based on a personal bias. Thanks for this!

mdelapenya (approved):
LGTM! Thanks

mdelapenya:
My only concern (unrelated) is that any time the dependabot file is modified, it triggers an update 🤦 So be ready for YADW: Yet Another Dependabot Wave

mmorel-35:
You shouldn't be concerned about this. Maybe the amount of allowed PR per module is too low or the check frequency too long.
The config need help you address every update as soon as possible. If there are new updates every time there is a problem with the configuration.

mdelapenya:
I was expecting @dependabot to be smart enough to honour the monthly schedule, instead of triggering an update every time the file is modified, which happened today after this merge

PR #1470, opened 2023-08-09
chore: add a GH action for release drafter

mdelapenya:
## What does this PR do?
It copies tc-node's GH workflow for the release drafter,

## Why is it important?
in order to have a consistent process for drafting releases.


mdelapenya:
Merging as this is not causing any test to fail

PR #1474, opened 2023-08-09
chore: triple max timeout for the workflow run, which takes +10m

mdelapenya:
## What does this PR do?
max timeout *3 for the workflow run, which now takes +10m on Windows

## Why is it important?
Avoid killing the worker too early

## Related issues

-


mdelapenya:
Not impacting the build, merging

PR #1490, opened 2023-08-10
chore: use Info instead of Ping to check if Docker is live

mdelapenya:
## What does this PR do?
It replaces the calls to the Docker client's Ping endpoint to Info. For that we have defined our own wrapper around the Docker client, which implements Docker's `client.SystemAPIClient`, specially of interest the Info method: we will cache all the calls to Info from this wrapper, and reuse the cached data. As a consequence, we have deprecated the old `NewDockerClient`, which exposed the plain Docker client. The new method is called `NewDockerClientWithOpts`.

At the same time, the already public `LogDockerServerInfo` function has been deprecated and it produces NOOP (No-Operation): its logic has been moved to the custom Info code that we are implementing in this PR, therefore, it will be cached and printed just once, the first time the Docker client is obstained.

### Deprecation path for `NewDockerClient`
We have internally replaced all the occurrences of `NewDockerClient` to `NewDockerClientWithOpts`, so a deprecation path for consumers of the old method would be simply calling the new method including the context and Docker opts, if/when needed.

## Why is it important?
Initialisation involves fetching info from Docker which makes the ping unnecessary, as if fetching info worked then so will the ping. For reusing calls to the Docker engine, we are caching the calls to Info.

## Related issues

- Relates https://github.com/testcontainers/testcontainers-node/pull/635


docker.go
```diff
@@ -1191,9 +1187,9 @@ func (p *DockerProvider) attemptToPullImage(ctx context.Context, tag string, pul
 }
 
 // Health measure the healthiness of the provider. Right now we leverage the
-// docker-client ping endpoint to see if the daemon is reachable.
+// docker-client Info endpoint to see if the daemon is reachable.
 func (p *DockerProvider) Health(ctx context.Context) (err error) {
-	_, err = p.client.Ping(ctx)
+	_, err = p.client.Info(ctx)
```
mdelapenya, line 1192: Not sure if this has to be changed TBH

eddumelendez (dismissed):

logger.go
```diff
@@ -18,29 +17,10 @@ type Logging interface {
...
-		return
-	}
-
-	logger.Printf(infoMessage, packagePath,
-		info.ServerVersion, client.ClientVersion(),
-		info.OperatingSystem, info.MemTotal/1024/1024,
-		testcontainersdocker.ExtractDockerHost(ctx),
-		testcontainersdocker.ExtractDockerSocket(ctx),
-	)
+	// NOOP
```
mdelapenya, line 23: Should we keep the old code until the removal takes place? 🤔

PR #1491, opened 2023-08-10
feat: pass testcontainers-go version as HTTP header

mdelapenya:
## What does this PR do?
Pass the library version as HTTP header when calling Docker. We are going to use the `User-Agent` header for that.

## Why is it important?
Enrich the requests to the Docker engine with more information from the caller.

## Related issues

- Relates https://github.com/testcontainers/testcontainers-node/pull/637


internal/testcontainersdocker/client.go
```diff
@@ -31,7 +32,8 @@ func NewClient(ctx context.Context, ops ...client.Opt) (*client.Client, error) {
 
 	opts = append(opts, client.WithHTTPHeaders(
 		map[string]string{
-			"x-tc-sid": testcontainerssession.String(),
+			"x-tc-sid":     testcontainerssession.String(),
+			"X-User-Agent": "testcontainers-go@v" + internal.Version,
```
bsideup, line 36: WDYT about overriding the actual `User-Agent` header?
```suggestion
			"User-Agent": "tc-go/" + internal.Version,
```

mdelapenya: I'm OK with whatever name for the header name :)

We should agree in the format too:

- testcontainers-node/10.1.0
- testcontainers-node/v10.1.0
- testcontainers-go:v0.22.0 (gradle-like)
- testcontainers-go@v0.22.0
- github.com/testcontainers/testcontainers-go@v0.22.0 (as a Github import)

cristianrgreco (approved):

PR #1493, opened 2023-08-10
chore: Use generic db url for postgresql container

kevinrobayna:
## What does this PR do?

Use generic db url format with PG containers instead of PG specific one.

## Why is it important?

Before we where using the PG specific url format but most tools would expect the url on the generic one (i.e https://github.com/golang-migrate/migrate).

People could indeed write their own connString mapper but that would be unnecessary. (Ideally tools would not try to parse the url and instead just use the string but sometimes they might need to, thus using the generic format would work most of the times)

### DIFF
--- a/modules/postgres/postgres.go
+++ b/modules/postgres/postgres.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net"
 	"path/filepath"
+	"strings"
 
 	"github.com/testcontainers/testcontainers-go"
 )
@@ -36,14 +37,7 @@ func (c *PostgresContainer) ConnectionString(ctx context.Context, args ...string
 		return "", err
 	}
 
-	extraArgs := ""
-	if len(args) >= 1 {
-		extraArgs = args[0]
-	}
-	for _, arg := range args[1:] {
-		extraArgs += "&" + arg
-	}
-
+	extraArgs := strings.Join(args, "&")
 	connStr := fmt.Sprintf("postgres://%s:%s@%s/%s?%s", c.user, c.password, net.JoinHostPort(host, containerPort.Port()), c.dbName, extraArgs)
 	return connStr, nil
 }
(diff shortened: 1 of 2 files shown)


modules/postgres/postgres.go
```diff
@@ -36,11 +37,14 @@ func (c *PostgresContainer) ConnectionString(ctx context.Context, args ...string
 	}
 
 	extraArgs := ""
-	for _, arg := range args {
-		extraArgs += " " + arg
+	if len(args) >= 1 {
+		extraArgs = args[0]
+	}
+	for _, arg := range args[1:] {
```
kevinrobayna, line 43: this is to skip the first arg since we already added it to the string (to avoid ?&fo=var which is not valid)

mdelapenya: I think you can simplify it with strings.Join:

https://go.dev/play/p/6AEmqgAXfrq?v=goprev

modules/postgres/postgres.go
```diff
@@ -35,12 +37,8 @@ func (c *PostgresContainer) ConnectionString(ctx context.Context, args ...string
...
 	}
 
-	extraArgs := ""
-	for _, arg := range args {
-		extraArgs += " " + arg
-	}
-
-	connStr := fmt.Sprintf("host=%s port=%s user=%s password=%s dbname=%s %s", host, containerPort.Port(), c.user, c.password, c.dbName, extraArgs)
+	extraArgs := strings.Join(args, "&")
+	connStr := fmt.Sprintf("postgres://%s:%s@%s/%s?%s", c.user, c.password, net.JoinHostPort(host, containerPort.Port()), c.dbName, extraArgs)
```
mdelapenya, line 41: I'd like to double check with you: this connection string format will be compatible with pg, right?

mdelapenya: I said nothing: I understood PG as the Go client for Postgres, not Postgres itself 🙈

mdelapenya (approved):

PR #1496, opened 2023-08-11
fix: run tests for module generator in all platforms

mdelapenya:
## What does this PR do?
This PR adds a new input parameter to the reusable workflow, in order to "run the tests" step. Is true, it will execute the step.

With this approach the caller decides whether the tests are run or not passing true/false in the new `run-tests` parameter. In general, they will pass a `true` value, but for the core library, where we want to run the tests only for Linux, we must check that the platform is `ubuntu-latest`, only.

## Why is it important?
We noticed the value was hardcoded to run the tests only on Ubuntu, directly into the reusable workflow, but there are callers that need to run the tests on Mac and Windows too. For that reason we needed a way to tell the reusable workflow to skip the tests or to run them.

## Related issues

- Completes #1478

### DIFF
--- a/.github/workflows/ci-test-go.yml
+++ b/.github/workflows/ci-test-go.yml
@@ -22,6 +22,11 @@ on:
         type: boolean
         default: false
         description: "Run the test with rootless docker."
+      run-tests:
+        required: false
+        type: boolean
+        default: true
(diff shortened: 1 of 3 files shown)


mdelapenya:
The tests passed on Windows, so I think this is good to go! Merging

PR #1498, opened 2023-08-13
Golangci-lint for all go-modules

mmorel-35:
In the actual workflow only the root module is linted.
This adds the scanning of all modules.
The following modules need to pass the linter to be enabled (removed from the exclusion list):
* examples/cockroachdb
* examples/spanner
* examples/toxiproxy
* modules/compose
* modules/pulsar
* modules/redis

I also noticed that the workflow was not benefiting from setup-go cache capability for two reasons. The first was that the checkout was done after the setup. The second reason was that the go.sum used as the cache dependency was not defined.


mdelapenya (dismissed):
Great improvement, thanks!

LGMT

mmorel-35:
Sorry for the dismiss of the review, I was fixing localstack.

It's ready to be merge on my part.

.github/workflows/golangci-lint.yml
```diff
@@ -19,6 +19,34 @@ permissions:
 jobs:
   static-analysis:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        working-directory:
+          - "."
+          - examples/bigtable
```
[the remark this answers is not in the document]
mmorel-35: I agree. And also believe that every module/example shall be uncommented and fixed before adding it to the generation tool

mdelapenya (approved):
LGTM 🚀 

Let's include the YAML file to the code generation commands, but in a follow-up PR.

Thanks!

mdelapenya:
@mmorel-35 Mmmm let me step back a little bit and share this thought.

I've just realised about the number of workers that we are going to consume at the moment we merge this, putting it in front of any PR: it will request 1 worker per module + example, 26 workers allocated at the moment. Even if they run super fast, it could drain the number of workers for the organisation. How would you see adding the golangci-lint as a step or reusable job, which is called for each module? We could include the golangci call into the `ci-test-go.yml` file as the first step. Wdyt?

mdelapenya (dismissed):
Awesome, thanks!

image

.github/workflows/ci-test-go.yml
```diff
@@ -57,6 +57,21 @@ jobs:
       - name: Check out code into the Go module directory
         uses: actions/checkout@v3
 
+      - name: golangci-lint
+        # TODO: Remove each example/module once it passes the golangci-lint
+        if: ${{ inputs.platform == 'ubuntu-latest' && !contains(fromJSON('["examples/cockroachdb","examples/spanner","examples/toxiproxy","modules/compose","modules/pulsar","modules/redis"]'), inputs.project-directory) }}
```
mmorel-35, line 62: @mdelapenya , the condition for execution could be narrowed to enable less executions per module like on inputs.go-version or any other inputs.
What do you think ?

mmorel-35:
Hi @mdelapenya ,
I have updated the description of the PR ti explain what it fully does. There is a flaky test unrelated to my modifications.

mdelapenya:
> You might also need to update the PR rules

There is a repository `settings.yml` file including repo-as-code. You can update it yourself here in this PR: https://github.com/testcontainers/testcontainers-go/blob/71d58d8ec1ee541cb457ed40bdbfc287626aef24/.github/settings.yml#L186

I think removing the first string in the array would be enough

mmorel-35:
Anything else you want to add here @mdelapenya ?

.golangci.yml
```diff
@@ -1,16 +1,15 @@
 linters:
   enable:
     - gci
-    - gofmt
+    - gofumpt
```
mmorel-35, line 4: This is more strict linter than gofmt , it also ensure what has been talk [here](https://github.com/testcontainers/testcontainers-go/pull/1510#discussion_r1297170457)

.github/workflows/ci-test-go.yml
```diff
@@ -48,14 +53,30 @@ jobs:
...
-            go-version: ${{ inputs.go-version }}
+            go-version: '${{ inputs.go-version }}'
+            cache-dependency-path: '${{ inputs.project-directory }}/go.sum'
         id: go
 
-      - name: Check out code into the Go module directory
-        uses: actions/checkout@v3
+      - name: golangci-lint
+        # TODO: Remove each example/module once it passes the golangci-lint
+        if: ${{ inputs.platform == 'ubuntu-latest' && inputs.go-version == '1.20.x' && !contains(fromJSON('["examples/cockroachdb", "examples/spanner", "examples/toxiproxy", "modules/compose", "modules/pulsar", "modules/redis"]'), inputs.project-directory) }}
```
mdelapenya, line 68: What is the reason those modules need to be excluded? Are they hard to fix?

mdelapenya (approved):
LGTM, thanks for this huge work revisiting all files and updating them.

I wonder if we need a section in our docs regarding how to validate the format, e.g. running `golangci-lint run --verbose -c .golangci.yml`.

Do you think we can add it here in this PR, or in another one? No strong opinion on that, so whatever you prefer is fine.

In any case, thanks again for your work here, many times this kind of work is not visible at all, but super valuable, at least for me.

.github/workflows/ci-test-go.yml
mmorel-35: This PR has already several changes concerning the evolution of the workflow so I preferred not to include the fix of every linters for each module in there. That would be to much for me.

mmorel-35:
> I wonder if we need a section in our docs regarding how to validate the format, e.g. running golangci-lint run --verbose -c .golangci.yml.


I think something can be added in the documentation as well as a tooling in the makefiles but in my opinion this is for another PR.
IMHO it's better to have many small PR doing one thing and doing it well than having one big.

.github/workflows/ci-test-go.yml
mdelapenya: Indeed, no questions about that 😊 just wanted to double check with you

mdelapenya:
Merged, thanks!

image

PR #1503, opened 2023-08-15
modulegen: create internal/dependabot

mmorel-35:
Refactor dependabot code in modulegen


modulegen/dependabot.go
```diff
@@ -1,96 +1,20 @@
 package main
```
mdelapenya, line 1: I'd try to move all dependabot code from the main package to the new `pkg/dependabot`, so that the main.go file just consumes what the `dependabot` package exposes.

On the other hand, let's think if we want to expose a public API for handling dependabot files, using the publicly available `pkg/dependabot` package (which could be consumed in the wild), or on the contrary keep it internal (i.e. `./internal/dependabot`).

Before moving on with this PR and #1504 and #1505, let's discuss first about the APIs that we want to create, and the project layout for that. Let's iterate in one single command, and once we are happy with it, let's move to the others. I'm open to have a slack chat/huddle or zoom call for sync if needed

Does it sound good to you?

modulegen/internal/dependabot/main.go
```diff
@@ -0,0 +1,26 @@
...
+
+func GetUpdates(configFile string) (Updates, error) {
+	config, err := readConfig(configFile)
+	if err != nil {
+		return nil, err
+	}
+	return config.Updates, nil
+}
+
+func CopyDependabotConfig(configFile string, tmpFile string) error {
```
mdelapenya, line 20: The package namespace already provides the context :)
```suggestion
func CopyConfig(configFile string, tmpFile string) error {
```

mdelapenya (approved):
LGTM! Only a minor comment regarding a method name, but great job moving code around.

Thanks!

mdelapenya (approved):
LGTM, thanks!

Waiting for the CI, which will pass ✅

PR #1504, opened 2023-08-15
modulegen: create internal/mkdocs

mmorel-35:
Refactor mkdocs code in modulegen

### DIFF
--- a/modulegen/context.go
+++ b/modulegen/context.go
@@ -45,6 +45,25 @@ func (ctx *Context) getModulesByBaseDir(baseDir string) ([]string, error) {
 	return dirs, nil
 }
 
+func (ctx *Context) getMardownsByBaseDir(baseDir string) ([]string, error) {
+	dir := filepath.Join(ctx.DocsDir(), baseDir)
+
+	allFiles, err := os.ReadDir(dir)
+	if err != nil {
+		return nil, err
+	}
+
+	dirs := make([]string, 0)
+
+	for _, f := range allFiles {
+		if !f.IsDir() && filepath.Ext(f.Name()) == ".md" {
+			dirs = append(dirs, f.Name())
+		}
+	}
+	sort.Strings(dirs)
+	return dirs, nil
+}
+
 func (ctx *Context) GetExamples() ([]string, error) {
 	return ctx.getModulesByBaseDir("examples")
 }
@@ -53,6 +72,18 @@ func (ctx *Context) GetModules() ([]string, error) {
 	return ctx.getModulesByBaseDir("modules")
 }
 
+func (ctx *Context) GetExamplesDocs() ([]string, error) {
+	return ctx.getMardownsByBaseDir("examples")
+}
+
+func (ctx *Context) GetModulesDocs() ([]string, error) {
+	return ctx.getMardownsByBaseDir("modules")
+}
+
 func (ctx *Context) MkdocsConfigFile() string {
 	return filepath.Join(ctx.RootDir, "mkdocs.yml")
+}
+
+func NewContext(dir string) *Context {
+	return &Context{RootDir: dir}
 }
--- a/modulegen/context.go
+++ b/modulegen/context.go
@@ -45,7 +45,7 @@ func (ctx *Context) getModulesByBaseDir(baseDir string) ([]string, error) {
 	return dirs, nil
 }
 
-func (ctx *Context) getMardownsByBaseDir(baseDir string) ([]string, error) {
+func (ctx *Context) getMarkdownsFromDir(baseDir string) ([]string, error) {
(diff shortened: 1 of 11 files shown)


modulegen/main.go
```diff
@@ -375,3 +345,12 @@ func runGoCommand(cmdDir string, args ...string) error {
 	cmd.Dir = cmdDir
 	return cmd.Run()
 }
+
+func getRootDir() (string, error) {
```
mdelapenya, line 349: I think we can have a `NewContext(dir string) (Context, error)` function in the `context.go` file, instead of this one. In the end, we are always using it to get the dir, assert there is no error and create a context.

mdelapenya (dismissed):
LGTM!, just a comment, not a blocker, regarding one function. Other than that, great job extracting code! 💪

modulegen/main.go
mmorel-35: Do you want this in this PR ?
I can add this in the new one as it will be a small one.

mdelapenya: I think it's fine to have this change in this PR, as it's very narrowed to the PR itself. So I'd prefer having it here if possible 🙏

mmorel-35: I have created NewContext() but for the moment I cannot remove getRootDir as getModulesOrExamples still relies on it.
It shall be easier once getModulesOrExamples is replaced by ctx.GetModules()

mmorel-35: @mdelapenya ,
Can you please trigger a retry on the workflow ?

mmorel-35:
By the way, I noticed that there is no `docs/modules/compose.md` is it on purpose ?

mdelapenya:
> By the way, I noticed that there is no `docs/modules/compose.md` is it on purpose ?

Indeed, it's on purpose as we do have a separate page in https://golang.testcontainers.org/features/docker_compose/.

But I think it would be a good idea to add it to the modules and reduce the number of corner cases across code generation. I'll do it in a separate PR. Thanks for pointing this out!

modulegen/context.go
```diff
@@ -45,6 +45,25 @@ func (ctx *Context) getModulesByBaseDir(baseDir string) ([]string, error) {
 	return dirs, nil
 }
 
+func (ctx *Context) getMardownsByBaseDir(baseDir string) ([]string, error) {
```
mdelapenya, line 48:
```suggestion
func (ctx *Context) getMarkdownsFromDir(baseDir string) ([]string, error) {
```

modulegen/context.go
```diff
@@ -53,6 +72,18 @@ func (ctx *Context) GetModules() ([]string, error) {
 	return ctx.getModulesByBaseDir("modules")
 }
 
+func (ctx *Context) GetExamplesDocs() ([]string, error) {
+	return ctx.getMardownsByBaseDir("examples")
```
mdelapenya, line 76:
```suggestion
	return ctx.getMarkdownsFromDir("examples")
```

modulegen/context.go
```diff
@@ -53,6 +72,18 @@ func (ctx *Context) GetModules() ([]string, error) {
...
+}
+
+func (ctx *Context) GetModulesDocs() ([]string, error) {
+	return ctx.getMardownsByBaseDir("modules")
```
mdelapenya, line 80:
```suggestion
	return ctx.getMarkdownsFromDir("modules")
```

mdelapenya (approved):
LGTM! Only a final change in a method name: there was a typo and making the name a bit more consistent.

Other than that, once this is addressed, we are ready to merge

mmorel-35:
@mdelapenya ,
I just changed the method name.
The tests are running.

mmorel-35:
Hi @mdelapenya ,

The tests were passing before the rebase, something happened with the two last commits on the main branch ?

mdelapenya:
> The tests were passing before the rebase, something happened with the two last commits on the main branch ?

Not necessarily with the two commit :) I always take a look at the workflow:

Screenshot 2023-08-24 at 08 50 28

Because we run against the last two active versions of Go, I check both on Ubuntu: if one passes and the other not, then it seems certain flakiness to me. Same for the modules.

modulegen/main.go
```diff
@@ -170,14 +171,13 @@ func main() {
 	fmt.Println("Thanks!")
 }
 
-func generate(example Example, rootDir string) error {
+func generate(example Example, ctx *Context) error {
```
mdelapenya, line 174: I did not notice this: for consistency, let's pass the context as first arg.

But let's do this in a separate PR in order to merge this one as is :)

mdelapenya (approved):
LGTM, thanks for your hard work adding maintainability to the code generation tool 👏

PR #1510, opened 2023-08-17
doc fix: make file permission examples octal

martin-mfg:
Examples of file permissions should be in octal format. But in the code examples they are not. This PR fixes that.

## What does this PR do?
Change examples in docs.

## Why is it important?
Docs should be correct.

### DIFF
--- a/docs/features/copy_file.md
+++ b/docs/features/copy_file.md
@@ -14,7 +14,7 @@ nginxC, err := GenericContainer(ctx, GenericContainerRequest{
 		Started: true,
 	})
 
-nginxC.CopyFileToContainer(ctx, "./testdata/hello.sh", "/hello_copy.sh", 0700)
+nginxC.CopyFileToContainer(ctx, "./testdata/hello.sh", "/hello_copy.sh", 0o700)
 ```
 
 Or you can add a list of files in the `ContainerRequest` initialization, which can be copied before the container starts:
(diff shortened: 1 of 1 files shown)


docs/features/copy_file.md
```diff
@@ -14,7 +14,7 @@ nginxC, err := GenericContainer(ctx, GenericContainerRequest{
 		Started: true,
 	})
 
-nginxC.CopyFileToContainer(ctx, "./testdata/hello.sh", "/hello_copy.sh", 700)
+nginxC.CopyFileToContainer(ctx, "./testdata/hello.sh", "/hello_copy.sh", 0700)
```
mmorel-35, line 17: I think there is a linter for that provided by golangci but it usually suggest something like that 
```suggestion
nginxC.CopyFileToContainer(ctx, "./testdata/hello.sh", "/hello_copy.sh", 0o700)
```
If you find the linter it would worth adding it to golangci lint config with your PR , don't you think?

martin-mfg: I'm a newbie in golang, so maybe I'm wrong, but...
It seems golangci-lint doesn't have a check for octals: https://golangci-lint.run/usage/linters/
go-critic has such a check, but go-critic is not used in this repo: https://go-critic.com/overview.html#octalliteral

Do you want me to change the examle code to the suggested 0o... format anyway?

mmorel-35 (approved):

mdelapenya (approved):
@martin-mfg thanks for this super tiny and valuable change in our docs!

Because we are in the transition plan to migrate the inline snippets to executable Go example tests, I'm merging this as is, although at some point we will eliminate any inline snippet in favour of examples.

Before #1498, I'd have asked you to look for any other file permission usage in the library, so we can update it to the octal format, but I noticed @mmorel-35 already added a lint step for it in that PR.

Therefore, LGTM!

PR #1519, opened 2023-08-24
chore: properly render html and text templates

mdelapenya:
## What does this PR do?
This PR uses two different template engines for different template files: text-based and HTML-based ones.

The code is basically the same for both, but simplifying the logic for the target template file: only the markdown file for the new module will use the HTML template engine, as it needs to render HTML tags. The rest just needs the text engine.

At the same time, we are adding comments in the generated files explaining they are generated by the module generator tool.

Finally, the mkdocs file has a corner case with some entries that are not properly rendered: we are forcing the generation properly now.

## Why is it important?
When generating text files, the quotes and double quotes were interpolated to their HTML representations, which was incorrect.


mdelapenya:
@mmorel-35 could you take a look at this PR? It's very related to your work on the code generation tool 🙏

mmorel-35:
Would you mind keeping this as a draft, I can integrate it with the further PR of the refactoring of the modulegen.
In the meanwhile you can provide the header of the files with the comment header.

mdelapenya:
> Would you mind keeping this as a draft?

Sure, although I'm worried that anybody using the code generator could send a PR including wrong text values.

mmorel-35:
@mdelapenya ,
I created two PR that shall cover this one.
Please have look.
Notice that several other will come to cover the rest of the templates.

mdelapenya:
> @mdelapenya , I created two PR that shall cover this one. Please have look. Notice that several other will come to cover the rest of the templates.

Could you create an issue and add checkboxes with all the tasks you are planning? That would help me be prepared for new changes, not implementing things as duplicates, also making available for the community what you're working on. Something like:

- [ ] Extract mkdocs module
- [ ] Extract GH action module
- [ ] Sonarqube

Does it sound good to you?

mmorel-35:
Ok! I’m taking care of this today

PR #1520, opened 2023-08-24
modulegen: create internal/workflow

mmorel-35:
part 1 of #1519


modulegen/context_test.go
```diff
@@ -95,3 +95,9 @@ func TestModulesHasDependabotEntry(t *testing.T) {
 		assert.True(t, found, "module %s is not present in the dependabot updates", module)
 	}
 }
+
+func getTestRootContext(t *testing.T) *Context {
```
mdelapenya, line 99: We need this rename to avoid collisions, right?

mmorel-35: Exactly ! It could be avoided if the tests are migrated to a main_test package but this is a bigger work

mdelapenya (approved):
LGTM, thanks again for this contribution!!

PR #1537, opened 2023-08-25
modulegen: create internal/make

mmorel-35:
- extract Makefile file generation in /internal/make
- Extract the generation in a dedicated file for each internal domain.
- Isolate Example in a dedicated file `types.go`.


mdelapenya (approved):
LGTM, thanks!

PR #1538, opened 2023-08-26
chore: fix workflow

mmorel-35:
When dependabot is creating a PR to update dependencies, it generates 226 jobs while on a PR from a fork on 116 are generated.

Triggering tests only on push on main branch will avoid this.


mdelapenya (approved):
Good catch!

Issue #1536, opened 2023-08-25
[Enhancement]: Some ideas for testcontainers-go


mmorel-35:
### Proposal

There are several ideas I can see concerning testcontainers-go, some concerning modulegen, some concerning the CI/CD.

## Modulegen
- [ ] extract mkdocs example.md file generation in /internal/mkdocs
- [ ] extract Makefile file generation in /internal/make
- [ ] extract example.go and example_test.go file generation in /internal/module
- [ ] extract go.mod file generation in /internal/module and replace html/template with "golang.org/x/mod/modfile" to directly read from main go.mod

## CI/CD
- [ ] Provide code coverage to sonarcloud (starting with main module) #1518
- [ ] Generate sonar-project.properties per module/example
- [ ] Change dependabot updates schedule interval to "weekly" to address updates more frequently so it doesn't become an epic task every month

I'm uch more focus on modulegen for the moment

mdelapenya:
> Change dependabot updates schedule interval to "weekly" to address updates more frequently so it doesn't become an epic task every month ?

We already had that schedule, and there were lots of updates, exhausting the free tier of our GH actions workers across the entire testcontainers organisation.

mmorel-35:
I guess that at that time there was no groups for the updates. Would it helps smooth the process?

mdelapenya:
>  Move Image of the example/module to a Dockerfile to be able to receive updates from dependabot ?

Regarding this, I'd say no, as we want to keep modules stable, at least until we have a real support matrix (which we do not have). Also moving to a Dockerfile-based module will force users to build the image again and again.

Just in case, I'm exploring [updatecli](https://www.updatecli.io/) for certain updates

mdelapenya:
@mmorel-35 I have a working branch creating `cmd/devtools` as the primary way to run several things:
- the module generator
- the release shell script, migrated to Go

Would you like to take a look and review it if possible? I appreciate your valuable feedback about this kind of tasks

PR #1539, opened 2023-08-27
modulegen: create internal/module and internal/modfile

mmorel-35:
## What does this PR do?

This PR move the logic of go.mod construction from main package to internal/modfile package.
It also replaces the use of a template to the use of golang.org/x/mod library. This allows to read the go version and the main package informations from the main go.mod and use it to construct the new module.
As there is no more template for go.mod, the script for updating go version is updated too.

It also moves the construction of example.go and example_test.go to internal/module.

As the testcontainer-go version is directly retrieved from mkdocs.yml this property is not provided as an Example property.

## Why is it important?
This is separating code in there domain to improve the readability and maintanability.

## Related issues
- Relates #1536

## Follow-ups
- Relates #1543


mdelapenya (dismissed):
Cool, thanks! This LGTM and it's ready to be merged 👏  I myself understand the code but anybody else in the community could not understand the rationale behind the usage of `gomod` Vs keeping the template. BTW this is cool!

I'd appreciate if you use the PR template to include a description of what this PR does (technical implementation) and why is needed (product needs), because I see PR descriptions as a way of documentation for these two personas: a community member, and the future ourselves in the need to check why we did something in such a way in the past.

So I'd ask you for filling in the PR description before merging it 🙏

mmorel-35:
Ok @mdelapenya ,
Does this description is enough or do you need me to precise more on certain parts ?

mdelapenya:
> Ok @mdelapenya , Does this description is enough or do you need me to precise more on certain parts ?

Now it's perfect. Thanks for your hard work improving the module generator 🚀

mmorel-35:
There was a typo in some names "mode" instead of "mod" was used.

PR #1542, opened 2023-08-28
fix: use proper service name in Pulsar module

arjunmadan-toast:
## What does this PR do?

The pulsar module docs incorrectly mentioned the redis module. This PR updates the docs to correctly reference the pulsar module.

## Why is it important?

This is a pretty minor change that just updates some documentation.

## Related issues

-


mdelapenya (approved):
Good catch! Thank you for this PR, keeping docs up-to-date is important 🙇

PR #1543, opened 2023-08-28
modulegen: generate md file inside internal/mkdocs

mmorel-35:
## What does this PR do?
This removes codeinclude function and use text/template instead of html/template.

It also moves the logic of markdown example file creation to the internal/mkdocs package.

## Why is it important?
This is separating code in there domain to improve the readability and maintanability.

## Related issues
- Relates #1536

## Follow-ups
Remove Example functions and replace them with fields so they are calculated once.

### DIFF
--- a/modulegen/_template/example.md.tmpl
+++ b/modulegen/_template/example.md.tmpl
@@ -16,13 +16,13 @@ go get github.com/testcontainers/testcontainers-go/{{ ParentDir }}/{{ $lower }}
 
 ## Usage example
 
-{{ codeinclude "<!--codeinclude-->" }}
+<!--codeinclude-->
 [Creating a {{ $title }} container](../../{{ ParentDir }}/{{ $lower }}/{{ $lower }}.go)
-{{ codeinclude "<!--/codeinclude-->" }}
+<!--/codeinclude-->
 
-{{ codeinclude "<!--codeinclude-->" }}
+<!--codeinclude-->
 [Test for a {{ $title }} container](../../{{ ParentDir }}/{{ $lower }}/{{ $lower }}_test.go)
-{{ codeinclude "<!--/codeinclude-->" }}
+<!--/codeinclude-->
 
 ## Module reference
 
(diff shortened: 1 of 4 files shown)


mdelapenya:
Mmm it's interesting but I think we broke the generator for examples:

```
go run . --name questdb --image "questdb/questdb:latest" --title QuestDB
```

It wasn't in this PR but I'm currently bisecting where the example generator was broken.

mmorel-35:
I get it ! The {{ $entrypoint }} is not used in the example_test.go.tmpl ;)

mdelapenya:
Could you do a local verification, and generate an example of a module, and from the root for, run `mkdocs serve` to verify the docs site for the new module renders properly?

docs/examples/questdb.md
```diff
@@ -0,0 +1,69 @@
...
+[Test for a QuestDB container](../../examples/questdb/questdb_test.go)
+<!--/codeinclude-->
+
+## Module reference
+
+The QuestDB module exposes one entrypoint function to create the QuestDB container, and this function receives two parameters:
+
+```golang
+func RunContainer(ctx context.Context, opts ...testcontainers.ContainerCustomizer) (*QuestDBContainer, error)
+```
```
mmorel-35, line 33: @mdelapenya ,
This part of documentation seems to be erronous. Shall modules and example both have an entrypoint with `RunContainer` ?

mdelapenya: I have a plan to generate an `example_test.go` file in every module, in which we only add Go examples code, which will be rendered in the markdown files (please see the Artemis module). An example module will not have real API, only examples.

Since we are not there yet, the example must have a private function. We could even make it simpler and name that function `runContainer` so we just have to lowercase the initial upper char.

mmorel-35: I see !

mdelapenya (approved):
LGTM, thanks!

PR #1548, opened 2023-08-29
feat: add mariadb module

mdelapenya:
- chore: scaffolding for mariadb module
- feat: implement MariaDB module, using MySQL as reference
- fix: refine mysql tests and docs

## What does this PR do?
It uses the module generator to bootstrap the code for the MariaDB module, and also copies the MySQL one to implement the features we need for MariaDB.

The only differences I noticed:

- In the `my.cnf` file, I had to increase the `thread_stack` from 128K to 256K, otherwise an error was thrown by the container startup.
- When generating the connection string for non-TLS connections, MariaDB has security disabled by default, so the param `tls=false` is used in the tests.
- I started with the `mariadb:10.5.5` base image, but it does not accept the `MARIADB_` prefix environment variables, as expected. That image wants `MYSQL_` variables. For that reason, I moved to `mariadb:11.0.3` instead, which uses the `MARIADB_` prefix. Therefore, this module will copy all `MARIADB_` variables to the `MYSQL_` variables. @eddumelendez is this considered in the Java impl?

## Why is it important?
Add a new module!


mdelapenya:
Note that the sonarcloud warning is related to the existing duplication in the mariadb, which is basically shadowing what the mysql does. There subtle differences (see above), but the module is in practice the same.

eddumelendez:
* `thread_stack` in tc-java is 512k. See the [default config](https://github.com/testcontainers/testcontainers-java/blob/23478fac6f86c4dd8d0eb964386089144cb26f6e/modules/mariadb/src/main/resources/mariadb-default-conf/my.cnf)
* About TLS, I'll test using the same image
* From [docs](https://github.com/docker-library/docs/tree/master/mariadb#environment-variables): `From tag 10.2.38, 10.3.29, 10.4.19, 10.5.10 onwards, and all 10.6 and later tags, the MARIADB_* equivalent variables are provided. MARIADB_* variants will always be used in preference to MYSQL_* variants.` So, it's still safe using `MYSQL_` in tc-java

eddumelendez:
> I tested with 10.5.5 locally and te MARIADB_ prefix was not accepted (container failed to start because of no credentials were passed by the module container, which passed MARIADB-only variables.

According to the docs, `MARIADB_` in 10.5.x is supported from 10.5.10 onwards. So, it is expected to fail providing `MARIADB_` in version 10.5.5

> Do you think we should support that version, or on the contrary keep the lower limit in the 11.0.3?

11.0.3 is fairly new. In this case, I think it is ok to support previous images with `MYSQL_` for more coverage. Also, as of today, there will be compatibility with new versions due to the `MYSQL_` is not removed.

modules/mariadb/mariadb.go
```diff
@@ -0,0 +1,161 @@
...
+	username string
+	password string
+	database string
+}
+
+// WithDefaultCredentials applies the default credentials to the container request.
+// It will look up for MARIADB environment variables.
+func WithDefaultCredentials() testcontainers.CustomizeRequestOption {
+	return func(req *testcontainers.GenericContainerRequest) {
+		username := req.Env["MARIADB_USER"]
```
mdelapenya, line 37: @eddumelendez would you see it convenient to check for MYSQL_ variables here? I could imagine evaluating the vars in this order:

- check for MYSQL
- check for MARIADB (more precedence)

eddumelendez: interesting! However, I would rather use `withUsername`, `withPassword` and `withDatabase` rather than using the env vars directly. I don't think this is an issue because this is a new module, it would be if it is an existing one and just new utility fn were added but that's not the case. Those moving from GenericContainer to the module will also take that into account, I think. Historically, we haven't received issues about it in tc java. Of course, there is an assumption about using the module but if we suggest to use the module because of the pre configuration it brings then it is ok.

If we are going to plan to support `MARIADB_` env vars because `MYSQL_` can be deprecated then I would respect the default image's precedence.

PR #1549, opened 2023-08-29
feat(modulegen): print out VSCode workspace file if needed

mdelapenya:
- chore: move types to internal
- chore: delegate to the caller the representation of the template output
- Revert "chore: move types to internal"
- feat: support for printing out the vscode workspace file
- chore: define a flag to print the VSCode workspace if needed

## What does this PR do?
Following the recent refactors, it's now very easy to extend the module generator to generate more things. In this PR we are generating the VSCode workspace file, but instead of generating a file that could live whenever the user wants, it will print out the generated content to Stdout.

To achieve this, we have changed how our internal templating engine works: it accepted a string with the path where to write the file, but in this PR we are refactoring it to simply accept an `io.Writer` reference. Before this PR, there were two calls to this generation, so we have moved the creation of the files to those callers. In this PR we are adding a new caller of the template engine, and it will pass `os.Stdout`.

The generation of the VSCode template is optional, and for that reason is protected by a boolean flag: `-vscode-workspace`. If set, the string representation of the VSCode workspace file will be printed out to the stdout. By default it's false.

## Why is it important?
I see myself many times refreshing the VSCode workspace to see all modules in one single VSCode project, and I do this manually. With the current refactors (big kudos to @mmorel-35!!!) now it's easy to extend the generator with a few lines of code, and this improvement was very cheap in time and code.

### DIFF
--- a/modulegen/internal/make/main.go
+++ b/modulegen/internal/make/main.go
@@ -1,6 +1,7 @@
 package make
 
 import (
+	"os"
 	"path/filepath"
 	"text/template"
 
@@ -13,5 +14,19 @@ func Generate(exampleDir string, exampleName string) error {
 	if err != nil {
 		return err
 	}
-	return internal_template.Generate(t, filepath.Join(exampleDir, "Makefile"), name, exampleName)
+
+	exampleFilePath := filepath.Join(exampleDir, "Makefile")
+
+	err = os.MkdirAll(filepath.Dir(exampleFilePath), 0o755)
+	if err != nil {
+		return err
+	}
+
+	exampleFile, err := os.Create(exampleFilePath)
+	if err != nil {
+		return err
+	}
+	defer exampleFile.Close()
+
+	return internal_template.Generate(t, exampleFile, name, exampleName)
 }
--- a/modulegen/internal/template/main.go
+++ b/modulegen/internal/template/main.go
@@ -1,20 +1,12 @@
 package template
 
 import (
-	"os"
-	"path/filepath"
+	"io"
 	"text/template"
 )
 
-func Generate(t *template.Template, exampleFilePath string, name string, data any) error {
-	err := os.MkdirAll(filepath.Dir(exampleFilePath), 0o755)
-	if err != nil {
-		return err
-	}
-	exampleFile, _ := os.Create(exampleFilePath)
-	defer exampleFile.Close()
-
-	err = t.ExecuteTemplate(exampleFile, name, data)
+func Generate(t *template.Template, wr io.Writer, name string, data any) error {
+	err := t.ExecuteTemplate(wr, name, data)
 	if err != nil {
 		return err
 	}
--- a/docs/modules/index.md
+++ b/docs/modules/index.md
@@ -24,6 +24,7 @@ We have provided a command line tool to generate the scaffolding for the code of
 - a new Nav entry for the module in the docs site, adding it to the `mkdocs.yml` file located at the root directory of the project.
 - a GitHub workflow file in the .github/workflows directory to run the tests for the example.
 - an entry in Dependabot's configuration file, in order to receive dependency updates.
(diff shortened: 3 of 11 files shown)


modulegen/internal/make/main.go
```diff
@@ -13,5 +14,19 @@ func Generate(exampleDir string, exampleName string) error {
...
+	err = os.MkdirAll(filepath.Dir(exampleFilePath), 0o755)
+	if err != nil {
+		return err
+	}
+
+	exampleFile, err := os.Create(exampleFilePath)
+	if err != nil {
+		return err
+	}
+	defer exampleFile.Close()
```
mmorel-35, line 29: What is this ?

modulegen/internal/template/main.go
```diff
@@ -1,20 +1,12 @@
...
-	err := os.MkdirAll(filepath.Dir(exampleFilePath), 0o755)
-	if err != nil {
-		return err
-	}
-	exampleFile, _ := os.Create(exampleFilePath)
-	defer exampleFile.Close()
-
-	err = t.ExecuteTemplate(exampleFile, name, data)
+func Generate(t *template.Template, wr io.Writer, name string, data any) error {
+	err := t.ExecuteTemplate(wr, name, data)
```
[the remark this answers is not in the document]
mdelapenya: I like the idea. Let me rework on this PR

mmorel-35:
General question.
Why not generate it in a file in the project as there is already a folder for .devcontainer ?

mdelapenya:
> Why not generate it in a file in the project as there is already a folder for .devcontainer ?

I did not want to couple it with VSCode, but you are right: we already have a devcontainer file. Do you think generating it in the root folder with a hidden name is fine? Something like `.testcontainers-go.code-workspace`

mmorel-35:
> Something like `.testcontainers-go.code-workspace`

What are the practices recommended by vscode. From what I could see they usually define a dedicated folder in the repository.

mdelapenya:
> What do you expect to see ?

At some point I'd like to have one command per file that is not for the module itself, said in other words: a shared file: ci.yml, mkdocs, dependabot, vscode workspace... So I want to be able to refresh them and put it on sync

mmorel-35 (approved):

mdelapenya:
@mmorel-35 I'm thinking about the content of the VSCode file:

```
{
    "path": "src/github.com/testcontainers/testcontainers-go/examples/datastore"
},
```

The `src/github.com` part is specific to my machine and where I have the workspace. Don't you think that calculating that base path for all users would not be accurate? We could get the current execution dir and from there browse the parent file system until we get to github.com's parent. But could that be flaky?

mdelapenya:
We can keep it under the `.vscode` dir, which is already ignored. But I'm concerned about the motivation for generating an ignored file. Would it be useless?

mmorel-35:
The file is useful for anyone who wants to develop on this project. But as you said it, depending on the platform and the users setup the file path for the modules will be different. So modulegen can provide a way to generate the file but needs to ensure that you don't receive the config file of somebody else .

mdelapenya:
> The file is useful for anyone who wants to develop on this project. But as you said it, depending on the platform and the users setup the file path for the modules will be different. So modulegen can provide a way to generate the file but needs to ensure that you don't receive the config file of somebody else .

I think I have it: thanks to the existing building blocks, I can get the root dir, retrieve its absolute file path and pass it to the template, for all modules. Using `filepath` will guarantee the cross-platform thing.

If you like the implementation, I'll update the description to reflect the discussion in here

modulegen/main_test.go
mmorel-35, line 1: Why removing those tests ?

mmorel-35 (approved):
It LGTM !
You can update the description with what we discussed.

modulegen/main_test.go
mdelapenya: Because they were too coupled to the implementation, and having to update to update them with every change "smells". In fact, there are many of this type of tests in the module generator TBH

mdelapenya:
> It LGTM ! You can update the description with what we discussed.

Updated! Thanks for your review and guidance through this refactor, lots of learning!

PR #1550, opened 2023-08-29
modulegen: Use cobra instead of native golang flag

mmorel-35:
## What does this PR do?

This uses [cobra](https://github.com/spf13/cobra) instead of golang native flag .

## Why is it important?

This will help provide more commands to modulegen.

## Related issues

- Relates #1536


modulegen/main.go
```diff
@@ -1,76 +1,76 @@
...
-			os.Exit(2) // the same exit code flag.Parse uses
-		}
+	err := NewRootCmd().Execute()
+	if err != nil {
+		fmt.Printf("%v\n", err)
+		os.Exit(1)
 	}
+}
 
+func generateExample(exampleVar *ExampleVar, isModule bool) error {
```
mdelapenya, line 42: Let's be intentional here and call it:
```suggestion
func generateModule(exampleVar *ExampleVar, asExample bool) error {
```

We can rename the Example struct to `TCModule` (or similar, to avoid just `Module`) here or in a follow-up.

modulegen/module_cmd.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+func NewModuleCmd() *cobra.Command {
+	exampleVar := &ExampleVar{}
+	cmd := &cobra.Command{
+		Use:   "module",
+		Short: "Create a new Module",
+		Long:  "Create a new Module",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return generateExample(exampleVar, true)
```
mdelapenya, line 14:
```suggestion
			return generateModule(exampleVar, false)
```

modulegen/example_cmd.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+func NewExampleCmd() *cobra.Command {
+	exampleVar := &ExampleVar{}
+	cmd := &cobra.Command{
+		Use:   "example",
+		Short: "Create a new Example",
+		Long:  "Create a new Example",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return generateExample(exampleVar, false)
```
mdelapenya, line 14:
```suggestion
			return generateModule(exampleVar, true)
```

modulegen/main.go
```diff
@@ -1,76 +1,76 @@
...
+		return fmt.Errorf(">> could not get the root dir: %w", err)
 	}
 
 	example := Example{
-		Image:     imageVar,
-		IsModule:  asModuleVar,
-		Name:      nameVar,
-		TitleName: nameTitleVar,
+		Image:     exampleVar.image,
+		IsModule:  isModule,
```
mdelapenya, line 50:
```suggestion
		IsModule:  !asExample,
```

modulegen/main.go
mdelapenya: Also: do we need pointer semantics here for the `exampleVar`? We are not modifying its state in this method, as we only use it to instantiate the Example struct, right?

modulegen/example_cmd.go
```diff
@@ -0,0 +1,23 @@
+package main
```
mdelapenya, line 1: Wdyt about creating a `cmd` package with the commands?

mmorel-35: Yes but where shall the generateModule be placed as it can’t stay in the main package in that case ?

modulegen/example_cmd.go
mmorel-35: If you are calling the method generateModule() it’s more clear to have asModule provided true than a asExample false . WDT ?

mdelapenya: I thought about it when posting, but seeing the following method signatures:

```go
func generateModule(e ExampleVar, asModule bool) error
```

and 

```go
func generateModule(e ExampleVar, asExample bool) error
``` 

I prefer the second. Is that what you mean?

modulegen/main.go
mmorel-35: No modification indeed. I removed the pointers

modulegen/example_cmd.go
mmorel-35: I propose a more simple way : 
```go
func generate(exampleVar ExampleVar, isModule bool) error
```

modulegen/example_cmd.go
mmorel-35: This require to move a lot a code. Let's keep it for another PR ?

modulegen/main.go
mmorel-35: > We can rename the Example struct to TCModule (or similar, to avoid just Module) here or in a follow-up.

Let's do it in a follo-up, yes.

modulegen/example_cmd.go
mdelapenya: Mmm given this PR is adding cobra, I'd like to follow their project layout structure: https://github.com/spf13/cobra/blob/main/site/content/user_guide.md

```
  ▾ appName/
    ▾ cmd/
        add.go
        your.go
        commands.go
        here.go
      main.go
```

mmorel-35: I understand but this cannot be achieved without having first move out from main.go Context, Example, ExampleVar and every generate function . Otherwise you'll have a circular dependency.

mdelapenya: Let's do it in this PR,  the changes will be very narrowed to the module generator and with one single purpose: add cobra. I think it's fine if the PR gets a bit bigger

modulegen/cmd/new/example.go
```diff
@@ -0,0 +1,26 @@
...
+		Long:  "Create a new Example",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return internal.Generate(*exampleVar, false)
+		},
+	}
+	cmd.Flags().StringVarP(&exampleVar.Name, "name", "n", "", "Name of the example. Only alphabetical characters are allowed.")
+	cmd.Flags().StringVarP(&exampleVar.NameTitle, "title", "t", "", "(Optional) Title of the example name, used to override the name in the case of mixed casing (Mongodb -> MongoDB). Use camel-case when needed. Only alphabetical characters are allowed.")
+	cmd.Flags().StringVarP(&exampleVar.Image, "image", "i", "", "Fully-qualified name of the Docker image to be used by the example")
+	_ = cmd.MarkFlagRequired("image")
+	_ = cmd.MarkFlagRequired("name")
```
mdelapenya, line 24: Have you considered defining those flags, that are reused in the sibling `module` subcommand, into the `new` subcommand as PersistentFlags?

They will be shared downstream to all the subcommands of the command that defined them.

https://github.com/spf13/cobra/blob/main/site/content/user_guide.md#persistent-flags

modulegen/cmd/new/main.go
```diff
@@ -0,0 +1,16 @@
+package new
+
+import (
+	"github.com/spf13/cobra"
+)
+
+func NewCmd() *cobra.Command {
```
mdelapenya, line 7: I think the most accepted pattern for defining commands and subcommands is using the init function: https://github.com/spf13/cobra/blob/main/site/content/user_guide.md#organizing-subcommands

modulegen/internal/model/cmd.go
```diff
@@ -0,0 +1,7 @@
+package model
```
mdelapenya, line 1: I'm thinking about naming this package context instead of model, not sure about it. It could probably come with some renames: `model.GetRootContext()` >> `context.New()`

I'm fine having the discussion here and implement it probably in a follow-up?

modulegen/internal/module/main.go
```diff
@@ -1,13 +1,53 @@
...
 	"strings"
 	"text/template"
 
+	"github.com/testcontainers/testcontainers-go/modulegen/internal/mkdocs"
+	"github.com/testcontainers/testcontainers-go/modulegen/internal/model"
+	"github.com/testcontainers/testcontainers-go/modulegen/internal/modfile"
 	internal_template "github.com/testcontainers/testcontainers-go/modulegen/internal/template"
 )
 
+func GenerateGoModule(ctx *model.Context, example model.Example) error {
```
mdelapenya, line 15: At some point we will be writing our own `FileGenerator` interface, with the `Generate() error` function 😅

modulegen/internal/main.go
```diff
@@ -0,0 +1,87 @@
...
+	if err != nil {
+		return err
+	}
+	// update examples in dependabot
+	err = dependabot.GenerateDependabotUpdates(ctx, example)
+	if err != nil {
+		return err
+	}
+	// generate vscode workspace
+	err = vscode.GenerateVSCodeWorkspace()
```
mdelapenya, line 82: The idea of the interface takes more importance here. Again, probably as a follow-up :D

mdelapenya (commented):
@mmorel-35 looks very promising! I added a few comments, not blockers, but wanted to check with you first before approving the PR (which is already done in my mind)

modulegen/cmd/new/example.go
mmorel-35: I did but then I wouldn't be able to use a different description . I mean example or module depending on the subcommand

modulegen/cmd/new/main.go
mmorel-35: I see but it creates a dependency on the parent package while doing this way It's the parent package that knows it's children package

modulegen/internal/module/main.go
mmorel-35: That can indeed be an idea.

modulegen/cmd/new/main.go
mdelapenya: I mean within the same `new` package, not in the parent one. The new subcommand would mount the example and module subcommands in their init. For that, we would need a variable for each command.

Said that, we probably don't want a `new` package but a `modules` one, keeping same Go files. 🤔

modulegen/internal/model/cmd.go
mmorel-35: done

modulegen/cmd/new/main.go
mmorel-35: I have change the name of the package. 
Concerning of the use of the init method. This applies only when the command is a var. Those are functions and will be called only one time at the execution of the main() .
Concerning, the use of the init function there are linters that are forbidding it, `gochecknoinits` for example.

mdelapenya: I think we can live without `gochecknoinits` as we will be using the best-practices defined by Cobra authors. I understand your concern (I also see having multiple inits could lead to code that is difficult to follow) but we will have them very well located. Thanks for raising this concern btw

mmorel-35: OK ! This shall fit your expectations now ?

mdelapenya (approved):
LGTM, thanks for the contribution! Now the modulegen seems a proper dev tool!

PR #1551, opened 2023-08-29
modulegen: generate code-workspace with json marshal

mmorel-35:
## What does this PR do?

Replaces the template with the marshalling of a Config Object.
The path can be relative so it is relative to the .vs-code folder. As it is platform independant, it can be shared in the project.

## Why is it important?

This allows to define [multi-root-workspaces](https://code.visualstudio.com/docs/editor/multi-root-workspaces) in VSCode.

## Related issues

- Relates #1549

I tried it on windows plateform. @mdelapenya if you do have another platform, can you confirm, this works ?


mdelapenya (dismissed):
Hi @mmorel-35 I went ahead and pushed a commit simply adding the new file to the module docs: https://github.com/testcontainers/testcontainers-go/pull/1551/commits/6e4d7f7d07c2e6f80b0da8037cd17b4103239192

I think this PR LGTM! Thanks again!

mdelapenya:
I still need to get used to seeing the core module at the bottom, but I like the improvement. I guess you forced the commit of the workspace file, right?

mmorel-35:
Probably something like that as the .gitignore doesn’t say anything about .vscode.
For the ordering, it could be changed to follow the path order if you prefer ?

modulegen/internal/vscode/types.go
```diff
@@ -1,34 +1,43 @@
...
-		ModuleGeneratorDir: moduleGeneratorDirAbs,
-		Examples:           examples,
-		Modules:            modules,
+	for _, module := range modules {
+		config.Folders = append(config.Folders, Folder{
+			Path: "../modules/" + module,
+			Name: "modules/" + module,
+		})
 	}
+	sort.Slice(config.Folders, func(i, j int) bool { return config.Folders[i].Name < config.Folders[j].Name })
```
mmorel-35, line 41: You can replace Name by Path or define different names so they follow an order you like.

mdelapenya: In that case, let's have core library at the beginning. modulegen can be in its own place, but let's keep core with more visibility

mmorel-35: I'm  on it

.vscode/.testcontainers-go.code-workspace
```diff
@@ -0,0 +1,113 @@
+// This file is autogenerated by the 'modulegen' tool.
+{
+    "folders": [
+        {
+            "name": "core",
```
mdelapenya, line 5:
```suggestion
            "name": "testcontainers-go",
```

modulegen/internal/vscode/types.go
```diff
@@ -1,34 +1,43 @@
...
+}
 
-	for i, example := range examples {
-		examples[i] = filepath.Join(rootDirAbs, "examples", example)
+func newConfig(examples []string, modules []string) *Config {
+	config := Config{
+		Folders: []Folder{
+			{
+				Path: "../",
+				Name: "core",
```
mdelapenya, line 21:
```suggestion
				Name: "testcontainers-go",
```

mdelapenya (approved):
Super tiny change  in a name and we are done

modulegen/internal/vscode/types.go
mdelapenya: We probably need to hack the sort process

mmorel-35: OK, It's done ;)

mdelapenya (approved):
Cool, thanks! LGTM

PR #1555, opened 2023-08-30
chore: refine fail-fast strategy on CI

mdelapenya:
- chore: let all modules to finish their execution
- chore: save workers for examples

## What does this PR do?
This PR adds a new input parameter to the reusable workflow to run the Go tests, making it possible to call it with a fail-fast strategy.

At the same time, we are not running the Go examples (modules that do not expose any API, just code snippets) in one single version of Go: 1.20.x, and only in Ubuntu.

## Why is it important?
The first change tries to avoid the cancellation of the entire matrix for modules, which could happen if a flaky build fails: the rest of those 60 builds will be cancelled.

The second change is to save GH workers, as examples as less critical in terms of build conditions. We will reduce by 4 the number of user workers for the examples: Given there are 10 examples, instead of 40 builds, we will have just 10.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -113,7 +113,7 @@ jobs:
         module: [bigtable, cockroachdb, consul, datastore, firestore, nats, nginx, pubsub, spanner, toxiproxy]
     uses: ./.github/workflows/ci-test-go.yml
     with:
-      go-version: "1.x"
+      go-version: "1.20.x"
       fail-fast: true
       platform: ${{ matrix.platform }}
       project-directory: examples/${{ matrix.module }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -119,5 +119,5 @@ jobs:
       platform: 'ubuntu-latest'
       project-directory: examples/${{ matrix.module }}
       rootless-docker: false
-      run-tests: yes
+      run-tests: true
       ryuk-disabled: false
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,12 +109,11 @@ jobs:
     needs: test-modules
     strategy:
       matrix:
-        go-version: [1.20.x, 1.x]
         platform: [ubuntu-latest, macos-latest]
         module: [bigtable, cockroachdb, consul, datastore, firestore, nats, nginx, pubsub, spanner, toxiproxy]
     uses: ./.github/workflows/ci-test-go.yml
     with:
-      go-version: ${{ matrix.go-version }}
+      go-version: "1.x"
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
mmorel-35, line 1: Don't forget ci.yml.tmpl ;)

.github/workflows/ci.yml
```diff
@@ -104,12 +109,12 @@ jobs:
     needs: test-modules
     strategy:
       matrix:
-        go-version: [1.20.x, 1.x]
         platform: [ubuntu-latest, macos-latest]
         module: [bigtable, cockroachdb, consul, datastore, firestore, nats, nginx, pubsub, spanner, toxiproxy]
     uses: ./.github/workflows/ci-test-go.yml
     with:
-      go-version: ${{ matrix.go-version }}
+      go-version: "1.x"
```
mmorel-35, line 116: Shall the condition to run golangci-lint be changed to be applied on 1.x or use 1.20.x instead here ?

mdelapenya: Will use 1.20.x here, thanks!

.github/workflows/ci.yml
mdelapenya: Updating now!

.github/workflows/ci.yml
```diff
@@ -106,14 +111,13 @@ jobs:
...
     with:
-      go-version: ${{ matrix.go-version }}
-      platform: ${{ matrix.platform }}
+      go-version: "1.20.x"
+      fail-fast: true
+      platform: 'ubuntu-latest'
       project-directory: examples/${{ matrix.module }}
       rootless-docker: false
-      run-tests: ${{ matrix.platform == 'ubuntu-latest' }}
+      run-tests: yes
```
mmorel-35, line 122:
```suggestion
      run-tests: true
```

modulegen/_template/ci.yml.tmpl
```diff
@@ -106,14 +111,13 @@ jobs:
...
     with:
-      go-version: {{ "${{ matrix.go-version }}" }}
-      platform: {{ "${{ matrix.platform }}" }}
+      go-version: "1.20.x"
+      fail-fast: true
+      platform: 'ubuntu-latest'
       project-directory: {{ "examples/${{ matrix.module }}" }}
       rootless-docker: false
-      run-tests: {{ "${{ matrix.platform == 'ubuntu-latest' }}" }}
+      run-tests: yes
```
mmorel-35, line 122:
```suggestion
      run-tests: true
```

mdelapenya: 🤦

PR #1556, opened 2023-08-30
fix: remove wrong example from workspace

mdelapenya:
## What does this PR do?
Removes an entry in the VSCode workspace file.

## Why is it important?
Added by mistake from a dirty Git workspace

## Related issues

- Added by me by mistake in #1551


mdelapenya:
I think it's fine to merge this PR as is, the CI is passing

PR #1558, opened 2023-08-31
Update Pipenv to use Python 3.8

mmorel-35:
## What does this PR do?

This updates Python version to 3.8. It also synchronize pipenv.lock with requirements.txt.

## Why is it important?

Dependabot updates for pip are blocked because pipenv where using 3.7 version while runtime.txt was 3.8.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -6,10 +6,10 @@ verify_ssl = true
 [dev-packages]
 
 [packages]
-mkdocs = "*"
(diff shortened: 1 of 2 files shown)


mmorel-35:
> I think Netlify does not support Python 3.9... yet?

Yes indeed. I updated the Pipenv and Pipenv.lock only then.

PR #1559, opened 2023-08-31
feat: convert NATS example into a module

mdelapenya:
- chore: migrate NATS example to a module
- feat: support passing credentials to NATS
- chore: extract ports to constant
- chore: prepare for #1546

## What does this PR do?
This PR migrates the existing NATS example into a module, adding the ability to set up user, password and/or command line arguments that will be passed to the `nats-server` process in the container.

Code has been copied from the NATS implementation from testcontainers-node: https://github.com/testcontainers/testcontainers-node/blob/main/packages/modules/nats/src/nats-container.ts

## Why is it important?
More modules, as we already had the foundation as an example.


mdelapenya:
@adoublef please take a look, we converted the example into a module, with some more features, like setting user/pass.

adoublef:
Oh excellent, thank you. I will be taking a look. Similar to how the Postgres options includes `WithConfigFile` I could look into something similar for NATS.

PR #1560, opened 2023-08-31
feat: support waiting for logs with regexp

### LINKED ISSUE #1541: [Feature]: Support regex in ForLog wait strategy
### Problem

Sometimes it is necessary to check for log entries whose context varies, but follow a known pattern. Presently, the ForLog wait strategy only allows checking for fixed text.

For example, in the context of the k3s module, it would be required fo wait for a message with the pattern `".*Node controller sync successful.*` that indicates the cluster is ready, instead of the fixed text used now (`"k3s is up and running"`). 

See  [discussion in this thread](https://testcontainers.slack.com/archives/CBWBNR76G/p1693217479360819?thread_ts=1692981129.009009&cid=CBWBNR76G)

### Solution

Add a wait strategy that uses a regex instead of a fixed text.

### Benefit

* More flexibility in the validations.
* Consistency with for example the [java](https://github.com/testcontainers/testcontainers-java/blob/main/core/src/main/java/org/testcontainers/containers/wait/strategy/LogMessageWaitStrategy.java) implementation.

### Alternatives

Implementing this wait outside the container creation, reading all the log (for example using a {LogConsumer](https://golang.testcontainers.org/features/follow_logs/)). 

Yes

mdelapenya:
## What does this PR do?
This PR adds the ability to wait for logs using a regular expression. To avoid changing the current behaviour, we are adding a `AsRegexp` method to the fluent builder, which will be evaluated for the wait strategy: if regexp, search using the regexp, else, keep the old behaviour.

Therefore, it's optional for users to choose waiting for a fixed text or for a regular expression.

## Why is it important?
Allow more complex conditions thanks to regexp when needed, keeping the simplicity of the plain text when needed.

## Related issues

- Closes #1541


mdelapenya:
@pablochacin could you please take a look? 🙏

PR #1561, opened 2023-08-31
Update actions/github-script to 6.4.1

mmorel-35:
## What does this PR do?

Update actions/github-script to 6.4.1.
Adapt the script to use last version of github client.

## Why is it important?

Allow to be compatible with the new versions of actions/github-script

### DIFF
--- a/.github/workflows/ci-windows.yml
+++ b/.github/workflows/ci-windows.yml
@@ -13,11 +13,11 @@ jobs:
     timeout-minutes: 30
     steps:
       - name: Create pending status
-        uses: actions/github-script@0.9.0
+        uses: actions/github-script@6.4.1
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           script: |
-            github.repos.createStatus({
+            github.rest.repos.createCommitStatus({
               owner: context.repo.owner,
               repo: context.repo.repo,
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-windows.yml
```diff
@@ -55,12 +55,12 @@ jobs:
           gotestsum --format short-verbose --rerun-fails=5 --packages="./..."  --junitfile TEST-unit.xml -- -timeout=30m
```
mmorel-35, line 55: Does your Windows machine supports make ? 
You would be able to replace this with make test-unit command .

mdelapenya:
cc/ @cristianrgreco I remember we were looking for time and research about the bump. here you have it, kudos to @mmorel-35 !!!

.github/workflows/ci-windows.yml
mdelapenya: I do not think so, and the process to add packages to them is slow. I'd keep it as is, using go tooling.

At some point, I'm realising the makefiles are basically pass-through calling Go tools, and probably adding complexity instead of hiding it.

mmorel-35:
Can you trigger the ci-windows.yml workflow to be completly sure it works ?

PR #1568, opened 2023-09-01
chore: define interfaces for file and project generation

mdelapenya:
- chore: rename Example to TestcontainersModule
- chore: convert dependabot into a FileGenerator
- chore: convert makefile into a FileGenerator
- chore: convert go.mod into a FileGenerator
- chore: convert workflow into a ProjectGenerator
- chore: convert mkdocs into a FileGenerator
- chore: convert vscode into a ProjectGenerator

## What does this PR do?
This PRs renames the entities from Example to TestcontainersModule in order to be more intentional.

At the same time, creates two interfaces: FileGenerator and ProjectGenerator. The first one will generate files for a module when added, and the second one will inspect the state of the file system in order to generate the project files (CI workflow and VSCode workspace).

## Why is it important?
Better abstractions

## Related issues

- Continues with #1536

### DIFF
--- a/modulegen/internal/main.go
+++ b/modulegen/internal/main.go
@@ -75,9 +75,11 @@ func GenerateFiles(ctx *context.Context, m context.TestcontainersModule) error {
 	}
 
 	// they are based on the content of the modules in the project workspace,
-	// not in the new module to be added
+	// not in the new module to be added, that's why they happen after the actual
+	// module generation
 	projectGenerators := []ProjectGenerator{
 		workflow.Generator{}, // update github ci workflow
+		vscode.Generator{},   // update vscode workspace
 	}
 
 	for _, generator := range projectGenerators {
@@ -87,10 +89,5 @@ func GenerateFiles(ctx *context.Context, m context.TestcontainersModule) error {
 		}
 	}
 
-	// generate vscode workspace
-	err := vscode.GenerateVSCodeWorkspace(ctx)
-	if err != nil {
-		return err
-	}
 	return nil
 }
--- a/modulegen/internal/main.go
+++ b/modulegen/internal/main.go
@@ -14,25 +14,25 @@ import (
 	"github.com/testcontainers/testcontainers-go/modulegen/internal/workflow"
 )
 
-func Generate(exampleVar context.ExampleVar, isModule bool) error {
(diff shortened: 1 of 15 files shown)


modulegen/internal/context/context.go
```diff
@@ -68,38 +68,38 @@ func (ctx *Context) getMarkdownsFromDir(baseDir string) ([]string, error) {
...
+func New(dir string) Context {
+	return Context{RootDir: dir}
 }
 
-func GetRootContext() (*Context, error) {
+func GetRootContext() (Context, error) {
 	current, err := os.Getwd()
 	if err != nil {
-		return nil, err
+		return Context{}, err
```
mmorel-35, line 102: Why not return a pointer ?

modulegen/internal/main.go
```diff
@@ -14,25 +14,25 @@ import (
...
-func Generate(exampleVar context.ExampleVar, isModule bool) error {
+func Generate(moduleVar context.TestcontainersModuleVar, isModule bool) error {
 	ctx, err := context.GetRootContext()
 	if err != nil {
 		return fmt.Errorf(">> could not get the root dir: %w", err)
 	}
 
-	example := context.Example{
-		Image:     exampleVar.Image,
+	m := context.TestcontainersModule{
```
mmorel-35, line 23: rename it `module` to be explicit

modulegen/internal/context/context.go
mdelapenya: I'm removing the pointer semantics as we are not in the need of modifying the content of the struct

mmorel-35: OK

modulegen/internal/main.go
mdelapenya: I did not want to create a collision with the package. See https://github.com/testcontainers/testcontainers-go/pull/1568/files#diff-fa6baac23d1c3269416dc19f8301b7051795bb3bf9c98298d9e94c4870acfdbeR11

mmorel-35: Alright but it's a bit too short, don't you think ?

mdelapenya: I updated it to `tcModule`

mmorel-35 (approved):

Issue #809, opened 2023-02-06
[Docs]: Document how the "go test ./..." behaves when multiple test binaries are executed 


mdelapenya:
### Proposal

Context: https://testcontainers.slack.com/archives/C1SUBPZK6/p1674670070385259

`go test` works creating a test binary for each package if used with ./... . Therefore, each test binary (for each package to be tested) is isolated from the rest packages. Because of that, those test binaries do not share the state of the existing mutexes or any other thing that is synchronised: as it's in the case for the code that reuses or creates the reaper container, or the reuse container feature (we still have to revisit that, reusing the container by request hash).

This behaviour is how `go test` works, and we should stand with that, not allowing different go packages to access the state created by other packages.

Repro scenario: https://github.com/hahuang65/foo

Thanks to @hahuang65 for the use case and the repro.

HofmeisterAn:
From my understanding, this is the expected behavior. When tests are run in isolation, each test starts its own instance of the Resource Reaper. This is similar in .NET where, tests distributed across multiple test projects start their own instance of the Resource Reaper. Each test project is treated as a separate test session.

mdelapenya:
In the case of Go, they are not exactly different test projects but instead different test packages, although treated as different processes as each package will leave in different test binaries. That's why I think we should keep the behaviour and document it.

Apart from that, that has been the default behaviour in the past, and it was never an issue for our users 😅

hahuang65:
@mdelapenya first, thank you so much for looking into this, as well as your work on the project. It's enormously useful.

I've deleted my repo (my own OCD, but I'm happy to create it and store it somewhere if you have any suggestions, or think it's worth keeping the reproduction of this issue)I fully appreciate the limitations of how the `go test` will build separate binaries for each package.

I also fully appreciate and respect that, absolutely, it should not be allowed for different go packages to access state created by other packages.

I'm also speaking here, fully unaware of how the internals of `testcontainers-go` works with regards to the container `reuse` mechanism.

However, I will mention, from the point-of-view of a user, there is a disconnect. Reproduction of this "issue" ONLY happens if the container does not already exist. If you run the reproduction a second time, the issue is gone, because the first run stood up the container.

This suggests to me that the `reuse` mechanism CAN detect if a container is running in Docker. What I'm failing to understand is why it fails when two packages attempt to create the container at roughly the same time?

[...]

PR #593, opened 2022-10-27
fix: support parallel execution of reusable containers

mdelapenya:
## What does this PR do?
It adds a sync.Mutex lock/unlock when a reusable container is created, including a test that demonstrates the bug

In the original bug report (#592), a repro snippet was added, although I preferred a different test that leverages our `testcontainers.ParallelContainers` API, adding code coverage to the library. 

## Why is it important?
Thanks to @mhatch-nxcr, who reported this bug in #592, we demonstrated that the implementation of the reusable containers was not complete in terms of a parallel execution of tests creating containers with the library. The bug report found a race condition while checking the Docker client to find containers with the same name when called from a goroutine (i.e. running multiple Test functions including `t.Parallel()` or our `ParallelContainers` function).

Internally, when we are checking if we need to reuse a container by its name, the library calls the Docker client with a `findContainerByName` function of the DockerProvider. At the moment this `findContainerByName` function was called by any goroutine, no container was already created with the reusable container name, and in consequence the Docker client responded with a list of zero containers, which was expected. Therefore we had to look up the stack call to identify what code was calling that method, until we found that we needed synchronization at the moment the first container was created. That's why we are adding the `lock/unlock` only for the `req.Reuse` if/else branch.

**Before the changes**
```mermaid
graph TD
    B(Test execution)
    B --> C{Parallel creation}
    C -->|ReuseContainer| D[Name: reusable]
    C -->|ReuseContainer| E[Name: reusable]
    C -->|ReuseContainer| F[Name: reusable]
    D -->|findContainerByName| G[Docker client]
    E -->|findContainerByName| G[Docker client]
    F -->|findContainerByName| G[Docker client]
    G -->|filter:name=reusable| H["len(List)"]
```

**After the changes**
```mermaid
graph TD
    B(Test execution)
    B --> C{Parallel creation}
    C --> I[sync.Mutex]
    I -->|ReuseContainer| D[Name: reusable]
    I -->|ReuseContainer| E[Name: reusable]
    I -->|ReuseContainer| F[Name: reusable]
    D -->|findContainerByName| G[Docker client]
    E -->|findContainerByName| G[Docker client]
    F -->|findContainerByName| G[Docker client]
    G -->|filter:name=reusable| H["len(List)"]
```
## How to test this
Run the first commit of this PR with and without the second commit.

## Related issues
- Fixes #592

PR #1572, opened 2023-09-02
Allow use reuse containers in parrallel packages tests

### LINKED ISSUE #1277: [Bug]: Reuse ignores WaitingFor in container request
### Testcontainers version

v0.20.1

Yes

### Host OS

OS X

### Host arch

x86

### Go version

1.20.4

### Docker version

```shell
Client:
 Cloud integration: v1.0.29
 Version:           20.10.17
 API version:       1.41
 Go version:        go1.17.11
 Git commit:        100c701
 Context:           default
 Experimental:      true

 Engine:
  Version:          20.10.17
  Go version:       go1.17.11
  Git commit:       a89b842
  Built:            Mon Jun  6 23:01:23 2022
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.8
 runc:
  Version:          1.1.4
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### Docker info

```shell
Client:
 Context:    default
 Debug Mode: false
 Plugins:

Server:
 Containers: 7
  Running: 1
  Paused: 0
  Stopped: 6
 Images: 24
 Server Version: 20.10.17
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 init version: de40ad0
 Security Options:
  seccomp
   Profile: default
  cgroupns
 OSType: linux
 Architecture: x86_64
 CPUs: 8
 Total Memory: 15.64GiB
 Name: docker-desktop
 ID: NXUF:QSFY:ZCSL:ZGGF:MKTL:RTCA:OTKI:BUR6:PZWE:C654:Z4C4:NJBH
 Debug Mode: false
 Labels:
 Experimental: false
 Insecure Registries:
  hubproxy.docker.internal:5000
  127.0.0.0/8
 Live Restore Enabled: false
```

### What happened?

Running test suites for multiple packages (go test ./...) that use a kafka container started using testcontainers with Reuse enabled. Only the first test called actually waits for the WaitingFor conditions, the other processes skip waiting for and start executing before the container is ready (is running but services on the container aren't up).

### Relevant log output

_No response_

### Additional information

_No response_

prongq:
## What does this PR do?

When creating reusable container from different test binaries in parallel, you can get "409 - name conflict" from docker daemon, because different binaries don't have a shared mutex from #593.

Instead of exiting with error when receiving "409 - name conflict", we try to wait for the container creation to complete using the `ContainerList` method

## Why is it important?
This PR allows a reusable container to be used in parallel test binaries run from the "go test ./..." command.

## Related issues
- Closes #1277
- Relates  #809


generic_test.go
```diff
@@ -96,3 +102,77 @@ func TestGenericReusableContainer(t *testing.T) {
...
+	require.NoError(t, err)
+
+	// check is reuse container with WaitingFor work correctly.
+	resp, err := http.Get(origin)
+	require.NoError(t, err)
+	defer resp.Body.Close()
+
+	require.Equal(t, http.StatusOK, resp.StatusCode)
+
+	os.Exit(0)
```
mdelapenya, line 177: Could we add a comment here explaining why we need to call `os.Exit(0)` here? It could be weird to see an exit in the middle of a test, so an explanation would make sense 🙏

mdelapenya (commented):
@prongq thanks for working on this PR, I understand the use case and I see the potential value of it. If you mind, I'm postponing a final review until #1513 is merged (hopefully this week), as it comes with some synchronisation regarding the reaper container when running in the same test session (go test ./...), to be reused. I think it's very related to your work here.

In the mean time, it apparently looks good to me. Need to double check locally

generic_test.go
prongq: In fact, we can delete os.Exit(0). I just didn't delete it after copy-pasting it from the example

docker.go
```diff
@@ -1145,9 +1187,14 @@ func (p *DockerProvider) ReuseOrCreateContainer(ctx context.Context, req Contain
 		terminationSignal: termSignal,
 		stopProducer:      nil,
 		logger:            p.Logger,
-		isRunning:         c.State == "running",
 	}
 
+	if err := waitContainerReadyWithLog(ctx, dc); err != nil {
```
mdelapenya, line 1192: Why do we need to wait for the container here? The code should have already waited for it the first time it was created, right?

docker.go
```diff
@@ -829,6 +830,22 @@ func (p *DockerProvider) BuildImage(ctx context.Context, img ImageBuildInfo) (st
 	return repoTag, nil
 }
 
+func waitContainerReadyWithLog(ctx context.Context, container *DockerContainer) error {
```
mdelapenya, line 833: Why the `withLog` suffix to the name? Is it because it prints out the "Waiting for container..."??

Please also see my other comment about the need to wait twice, which could eliminate the need to extract the method into a reusable function. wdyt?

mdelapenya (commented):
I added a few comments before approving it.

FYI we are going to start an internal discussion around reusable containers, so in a short future, we could be visiting this code again for a refactor depending on the new design for reusable containers.

docker.go
prongq: > Why the withLog suffix to the name? Is it because it prints out the "Waiting for container..."??
Yes, I wanted to avoid duplicating the log message and put it in a separate function. If you would like another function name, feel free to ask.

docker.go
prongq: Not really. Waiting will only work correctly for the first caller. The other callers, if they start working before the first caller finishes waiting, will consider the container to be completely ready, although it is not. waiting in this place fixes this problem.

You can check this by removing the `waitContainerReadyWithLog` function call (and the log check in the test) and run the `TestGenericReusableContainerInSubprocess` test

prongq:
Hello @mdelapenya. Could you take a look at the pr please.

generic_test.go
```diff
@@ -137,3 +143,77 @@ func TestWithStartupCommand(t *testing.T) {
...
+	require.NoError(t, err, string(output))
+
+	return string(output)
+}
+
+// TestHelperContainerStarterProcess is a helper function
+// to start a container in a subprocess. It's not a real test.
+func TestHelperContainerStarterProcess(t *testing.T) {
+	if os.Getenv("GO_WANT_HELPER_PROCESS") != "1" {
+		return
```
mdelapenya, line 191: We can use t.Skip here to add an explanation on why this test method is skipped. Wdyt?

mdelapenya:
> Hello @mdelapenya. Could you take a look at the pr please.

Sure, let me do it tomorrow my morning. Sorry for the delay 🙏  conference stuff past weeks...

generic_test.go
prongq: Sure. Done

docker.go
mdelapenya: Ok, I think I now understand. Could we make the function `waitForContainerReady`? The withLog suffix could be an implementation detail, and there is no other alternative, so I think it's fine to not add it to the function name.

mdelapenya:
@prongq for some reason the helper test is not skipped. Could you please take a look at the errors in https://github.com/testcontainers/testcontainers-go/actions/runs/6466407444/job/17556657245?pr=1572?

prongq:
@mdelapenya Hello. The test shouldn't have been skipped. I don't understand why this error is happening. I have not been able to reproduce it locally. I think it has relation to keep-alive and I turned it off in the last commit. Could you please allow the tests to run.

Sorry for the long response, not enough time because of work

prongq:
I think the keep-alive problem is solved. I also removed the check that the same container is used with reuse, because tests can run for a long time and reaper can kill the container in the process.

prongq:
@mdelapenya can you run the tests, please?

PR #1574, opened 2023-09-05
feat: add Elasticsearch module

mdelapenya:
## What does this PR do?
This PR adds a Go module for Elasticsearch including the following features:

- support for Elasticsearch 6.x and above
- support for passing a custom password
- support for retrieving the CACert needed for configuring an HTTP connection over TLS
- support for getting container settings in order to create an HTTP connection

For Elasticsearch 7 and below:
- there is no ability to support TLS
- if no password is set, then unauthenticated HTTP requests can be done to the elasticsearch instance
- if a password is set, then only authenticated HTTP requests can be done to the elasticsearch instance (only over HTTP, no HTTPS support)

For Elasticsearch 8 and above:
- if no password is set, unauthenticated HTTPS requests won't be allowed (secure by default)
- if password is set, only authenticated HTTPS requests will be allowed, using the CACert internally extracted from the container (see below).
- the module will extract the default CACert file from the container, storing its bytes in the container instance as a setting. It will be possible to use it afterwards to build an HTTPS request.

As we mentioned, the container exposes setting to create an HTTP connection. It's responsibility of the consumer to build it properly. For demonstration purpose, we have included examples and tests showing how to create the connection with:
1. a plan HTTP client from the standard library,
2. the Elasticsearch Go client: https://github.com/elastic/go-elasticsearch

## Why is it important?
Adding modules is always great, but in particular Elasticsearch is a widely used technology.


internal/config/config.go
```diff
@@ -89,8 +89,6 @@ func read() Config {
 		return applyEnvironmentConfiguration(config)
 	}
 
-	fmt.Printf("Testcontainers properties file has been found: %s\n", tcProp)
```
mdelapenya, line 92: I noticed this print was here affecting the Go Examples, which matches the output of a test. That's why I'm removing it here

modules/elasticsearch/examples_test.go
```diff
@@ -0,0 +1,100 @@
+package elasticsearch_test
```
mdelapenya, line 1: Using Go examples for first time: I'm going to add them to all the modules, and the module generator, as they are true executable documentation that will land into the https://pkg.go.dev/github.com/testcontainers/testcontainers-go docs

eddumelendez (approved):

Issue #1362, opened 2023-07-17
[Enhancement]: incomplete documentation for postgres example


ChristianSch:
### Proposal

Hi there,

As [the postgres docs](https://golang.testcontainers.org/modules/postgres/) are based on the https://github.com/testcontainers/testcontainers-go/blob/7ae3f5ae950813ea3fe1b071f7a17012d222f2dd/modules/postgres/postgres_test.go, it produces code which is unusable. It is not clear what `tt` is from the context. You can't really know it, as it's rest related. Any change this could get fixed up?

It should really be something along the lines of:
```go
	"github.com/testcontainers/testcontainers-go"
	testpg "github.com/testcontainers/testcontainers-go/modules/postgres"
	"github.com/testcontainers/testcontainers-go/wait"

	container, err := testpg.RunContainer(ctx,
		testcontainers.WithImage("docker.io/postgres:15.2-alpine"),
		testpg.WithDatabase("test"),
		testpg.WithUsername("test"),
		testpg.WithPassword("test"),
		testcontainers.WithWaitStrategy(
			wait.ForLog("database system is ready to accept connections").
				WithOccurrence(2).
				WithStartupTimeout(5*time.Second)),
	)
```

Instead of:

![image](https://github.com/testcontainers/testcontainers-go/assets/6138133/0d265806-cb7a-462f-be8c-30f78dcd3124)

In theory I'm up for submitting a patch, but I really don't want to meddle with tests and doc generation processes.

mdelapenya:
Hi @ChristianSch , thanks for the issue report.

We are using a plugin to extract the test code from test files and put it into the markdown files, which is optimal to sync code and documentation. The downside of this is that the test code, as you pointed out, could have references to testing structures (in this case a table, or even the testing library).

We have this issue open: https://github.com/testcontainers/testcontainers-go/issues/611, and/but I'd like to also have Go examples, that end up in the pkg.dev documentation site.

Given we want to prioritise code snippets that don't get outdated, and are executed as real tests, I think it would be hard at the moment to unlink those code snippets from the test files.

sanan-go:
I think for documentation exports, you might consider testable Examples (https://go.dev/blog/examples)
They are also tests, but more readable and complete as snipped. Would it be possible?

At some point, I was going to give up using the postgres module and go with generic container API.
I found this issue after 30 minutes of digging - searching for proper wait strategies from random sources.

mdelapenya:
[...]

Issue #611, opened 2022-11-03
Migrate all code examples in the docs to "codeinclude" plugin


mdelapenya:
We have it documented in https://golang.testcontainers.org/contributing_docs/, but it's probably not clear enough.

Because code examples should be connected to the real code/tests, I suggest we should always use this when adding snippets to the docs.

PR #1600, opened 2023-09-06
chore: generate Go examples for new modules

mdelapenya:
- chore: rename templates to module
- chore: support generating Go Examples

## What does this PR do?
This PR adds a new Go template for including a dedicated Go file for Go examples. This examples file will include just an example on how to run the module, with the `RunContainer` function.

We have updated the tests, modifying the docs template to embed the code snippet to run a container from the examples, instead of from the test code.

## Why is it important?
It will open the gate to add executable code snippets as Go examples:
- are published in the Go docs (see pkg.dev)
- can be copy&pasted more easily than copying from a test code block

## Related issues

- Relates to #611
- Relates to #1362

## Follow-ups
Visit all modules and example modules, adding an `examples_test.go` file, and finally linking the code snippets in the docs to the new testable Examples code blocks


modulegen/_template/examples_test.go.tmpl
```diff
@@ -0,0 +1,36 @@
+{{ $entrypoint := Entrypoint }}{{ $lower := ToLower }}{{ $title := Title }}package {{ $lower }}_test
```
mdelapenya, line 1: The examples will live in the `_test` package so that the module has to be imported.

mdelapenya:
Merging, as the generator tests passed

PR #1603, opened 2023-09-08
docs: use Go testable examples in modules

### LINKED ISSUE #1362: [Enhancement]: incomplete documentation for postgres example

mdelapenya:
## What does this PR do?
This PR adds a `examples_test.go` file in all Go modules, adding there testable examples (`func ExampleRunContainer()`)

At the same time, we are adapting the code snippets that are embedded into the docs site to use this example functions instead of the test ones. In certain cases, we have refactored the test cases to follow a better structure that adapts to the testable examples.

We are aware that some Example functions could be duplication other existing tests, that's way we have removed those that collided.

During the refactor of the docs, we found useful to include docs on how to create a Docker network using the library, so docs have been added for that feature.

## Why is it important?
The code snippets in docs won't contain references to testing libraries such as `testing` or `testify`, and the examples will appear in pgk.go.dev, which is important for our users to see real, executable test examples.

## Related issues

- Relates #611
- Closes #1362


modules/pulsar/pulsar.go
```diff
@@ -107,15 +108,16 @@ func WithTransactions() testcontainers.CustomizeRequestOption {
 	return func(req *testcontainers.GenericContainerRequest) {
 		WithPulsarEnv("transactionCoordinatorEnabled", "true")(req)
 
-		// add the waiting strategy for the transaction topic
-		defaultWaitStrategies.Strategies = append(
-			defaultWaitStrategies.Strategies,
+		// clone defaultWaitStrategies
+		ss := []wait.Strategy{
```
mdelapenya, line 112: I discovered a bug here in the case the tests were run sequentially, as the default strategies variable was overridden, causing the `with_transactions` test case to fail because it was expecting the wait strategy for `with_functions_workers`. Fixing it here to avoid the test failures that were discovered on this PR

PR #1605, opened 2023-09-10
ci(lint): enable gocritic linter

mmorel-35:
## What does this PR do?

Enables [gocritic](https://golangci-lint.run/usage/linters/#gocritic) linter

## Why is it important?

Provides diagnostics that check for bugs, performance and style issues.
Extensible without recompilation through dynamic rules.
Dynamic rules are written declaratively with AST patterns, filters, report message and optional suggestion.


mdelapenya (approved):
LGTM!

PR #1608, opened 2023-09-11
ci(sonarcloud): run only when docker is not rootless

mmorel-35:
## What does this PR do?

* run sonarcloud only when docker is not rootless
* generate a checkstyle report from golangci-lint for sonarcloud,  see [documentation](https://docs.sonarcloud.io/enriching/external-analyzer-reports/) 
* correct path of test report as it was not correctly discovered


mdelapenya (approved):
LGTM, thanks!

PR #1034, opened 2023-04-04
Added Kafka (KRaft Mode) module

noctarius:
## What does this PR do?

This PR adds a new module for Kafka, running in KRaft (Zookeeper-less) mode

## Why is it important?

New modules are always important :-)

## Related issues

- Relates #1033 

## How to test this PR

Simple test for the module is included

PR #1610, opened 2023-09-12
feat: add kafka (KRaft mode only) module

### LINKED ISSUE #1033: [Enhancement]: Kafka module

mdelapenya:
## What does this PR do?
This PR adds a Kafka module, as a result of updating #1034 by @noctarius to the new design for modules.

The module comes with:

- a set of predefined env vars.
- a custom script that is copied right after the container is started, starting Kafka in KRaft mode.
- a wait strategy that is executed after the contianer is started and after the above file is copied.
- a container method to get the broker URL for kafka.

We copied the Java module implementation, but providing support for KRaft-only:
- validation of the lower version (7.0.0)
- same wait.ForLog message
- configure controller quorum voters based in the networks/aliases
- support for setting the cluster ID

## Why is it important?
A new module for a technology that is widely used

## Related issues

- Supersedes and Closes #1034
- Resolves #1033


mdelapenya:
@eddumelendez if you agree I0d like to proceed and merge this module, as it's clear that it uses kafka + KRaft. We can listen to the future feedback in the case we need to add support for Zookeeper mode. Thoughts?

ccampo133:
Why does `Brokers` return a slice when there is only ever a single element. Why not just return a string?

PR #1612, opened 2023-09-12
feat: support creating a network for localstack

### LINKED ISSUE #1602: [Localstack]: automatically create a network for Localstack
### Proposal

When running the Localstack container with a custom network defined, it is expected that all calls to the container will be **from other containers on that network**. `HOSTNAME_EXTERNAL` will be set to the *last* network alias that has been configured for the Localstack container.

A convenient `WithNetwork(name string)` would eventually create a network and populate the network and network aliases for the localstack container

mdelapenya:
## What does this PR do?
This PR adds a functional option to the localstack module, making it possible to create a Docker network for it and run localstack in that network with a given alias.

## Why is it important?
Support for running the localstack module in a network, as demo'ed in the existing tests.

## Related issues

- Closes #1602

### DIFF
--- a/modules/localstack/localstack.go
+++ b/modules/localstack/localstack.go
@@ -68,7 +68,11 @@ func WithNetwork(networkName string, alias string) testcontainers.CustomizeReque
 			},
 		})
 		if err != nil {
-			req.Logger.Printf("Failed to create network '%s'. Container won't be attached to this network: %v", networkName, err)
+			logger := req.Logger
+			if logger == nil {
+				logger = testcontainers.Logger
+			}
+			logger.Printf("Failed to create network '%s'. Container won't be attached to this network: %v", networkName, err)
 			return
 		}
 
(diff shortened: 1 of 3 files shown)


modules/localstack/localstack.go
```diff
@@ -58,6 +58,29 @@ func isVersion2(image string) bool {
...
+// on that network to the given alias.
+func WithNetwork(networkName string, alias string) testcontainers.CustomizeRequestOption {
+	return func(req *testcontainers.GenericContainerRequest) {
+		_, err := testcontainers.GenericNetwork(context.Background(), testcontainers.GenericNetworkRequest{
+			NetworkRequest: testcontainers.NetworkRequest{
+				Name: networkName,
+			},
+		})
+		if err != nil {
+			req.Logger.Printf("Failed to create network '%s'. Container won't be attached to this network: %v", networkName, err)
```
mdelapenya, line 71: May be of interest evaluate if the req.logger is nil 🤔

mdelapenya:
As a follow up, this new functional option could be pushed to the core, and available to all the modules 🤔

PR #1614, opened 2023-09-12
ci(sonarcloud): upload artifact

mmorel-35:
## What does this PR do?

Uploads unit-tests.xml and coverage.out in the sonarcloud artifact during the test phase. After the test-examples job the sonarcloud scan is executed with the artifacts uploaded.

## Why is it important?

This helps providing a correct code coverage on sonarcloud

### DIFF
--- a/.github/workflows/ci-test-go.yml
+++ b/.github/workflows/ci-test-go.yml
@@ -79,7 +79,7 @@ jobs:
           # Optional: working directory, useful for monorepos
           working-directory: ${{ inputs.project-directory }}
           # Optional: golangci-lint command line arguments.
-          args: --verbose --out-format=checkstyle:golangci-lint.report.xml
+          args: --verbose
           # Optional: if set to true then the all caching functionality will be complete disabled,
           #           takes precedence over all other caching options.
           skip-cache: true
--- a/.github/workflows/ci-test-go.yml
+++ b/.github/workflows/ci-test-go.yml
@@ -109,8 +109,7 @@ jobs:
             make test-unit
 
       - name: Upload SonarCloud files
-        if: ${{ inputs.run-tests && !inputs.rootless-docker }}
-        #if: ${{ github.ref_name == 'main' && github.repository_owner == 'testcontainers' && inputs.run-tests && !inputs.rootless-docker }}
+        if: ${{ github.ref_name == 'main' && github.repository_owner == 'testcontainers' && inputs.run-tests && !inputs.rootless-docker }}
         uses: actions/upload-artifact@v3
         with:
           name: sonarcloud
--- a/.github/workflows/ci-test-go.yml
+++ b/.github/workflows/ci-test-go.yml
@@ -107,15 +107,17 @@ jobs:
         run: |
             go install gotest.tools/gotestsum@latest
             make test-unit
-            
-      - name: Analyze with SonarCloud
-        if: ${{ github.ref_name == 'main' && github.repository_owner == 'testcontainers' && inputs.run-tests && !inputs.rootless-docker && inputs.project-directory == '.' }}
-        uses: sonarsource/sonarcloud-github-action@master
+
+      - name: Upload SonarCloud files
+        if: ${{ inputs.run-tests && !inputs.rootless-docker }}
+        #if: ${{ github.ref_name == 'main' && github.repository_owner == 'testcontainers' && inputs.run-tests && !inputs.rootless-docker }}
+        uses: actions/upload-artifact@v3
         with:
-          projectBaseDir: ${{ inputs.project-directory }}
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+          name: sonarcloud
+          path: |
+            ./.gitignore
+            ${{ inputs.project-directory }}/unit-tests.xml
+            ${{ inputs.project-directory }}/coverage.out
 
       - name: Run checker
         run: |
--- a/.github/workflows/ci-test-go.yml
+++ b/.github/workflows/ci-test-go.yml
@@ -116,7 +116,7 @@ jobs:
           name: sonarcloud
           path: |
             ./.gitignore
-            ${{ inputs.project-directory }}/unit-tests.xml
+            ${{ inputs.project-directory }}/TEST-unit.xml
             ${{ inputs.project-directory }}/coverage.out
 
       - name: Run checker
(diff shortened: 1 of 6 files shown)


sonar-project.properties
```diff
@@ -10,12 +10,12 @@ sonar.projectVersion=v0.23.0
...
 sonar.tests=.
 sonar.test.inclusions=**/*_test.go
 sonar.test.exclusions=**/vendor/**
 
-sonar.go.coverage.reportPaths=coverage.out
-sonar.go.golangci-lint.reportPaths=golangci-lint.report.xml
-sonar.go.tests.reportPaths=TEST-unit.xml
+sonar.go.coverage.reportPaths=**/coverage.out
+sonar.go.golangci-lint.reportPaths=**/golangci-lint.report.xml
+sonar.go.tests.reportPaths=**/TEST-unit.xml
```
mdelapenya, line 21: For modules, we are going to need updating this: https://github.com/testcontainers/testcontainers-go/blob/main/commons-test.mk#L19

```diff
- 		--junitfile TEST-$*.xml \
+ 		--junitfile TEST-unit.xml \
```

.github/workflows/ci.yml
```diff
@@ -139,3 +133,25 @@ jobs:
       rootless-docker: false
       run-tests: true
       ryuk-disabled: false
+
+  sonarcloud:
+    if: ${{ github.ref_name == 'main' && github.repository_owner == 'testcontainers' }}
+    needs: test-examples
```
mdelapenya, line 139: Do we need to wait until all modules&examples are run? I think we could `need` the main job.

In follow-up PRs, where we add support for the modules, we could move this out to the end (or have one sonarcloud job for each job type: core, modules, examples...) 🤔

mmorel-35: There is no obligation but I thought that having only one step to upload every report was better than having one per module.

sonar-project.properties
mmorel-35: Is the file not always named TEST-unit.xml ?

mdelapenya: I think not for modules :(

.github/workflows/ci.yml
mdelapenya: Yes, indeed. I meant one job per type: all the modules, all the examples, not for each module or example. I would not consider examples for code coverage, as they are just simple code snippets, not real modules exposing API

mmorel-35: Hum... that will make the workflow quite more complex as the upload-artifact will need to receive a dedicated name depending on module, modulegen or root .

.github/workflows/ci-test-go.yml
```diff
@@ -107,15 +107,18 @@ jobs:
...
+        #if: ${{ github.ref_name == 'main' && github.repository_owner == 'testcontainers' && inputs.run-tests && !inputs.rootless-docker }}
+        uses: actions/upload-artifact@v3
         with:
-          projectBaseDir: ${{ inputs.project-directory }}
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+          name: sonarcloud
+          path: |
+            ./.gitignore
```
mmorel-35, line 118: Just a trick to keep all folders

.github/workflows/ci-test-go.yml
```diff
@@ -107,15 +107,17 @@ jobs:
...
+            ${{ inputs.project-directory }}/unit-tests.xml
```
mdelapenya, line 119: Is the name of the XML file `unit-tests.xml` or `TEST-unit.xml`?

.github/workflows/ci.yml
mdelapenya: Ok, one last question: what if the build failed in any of the modules? this step won't be executed. i think we should make it mandatory (as a post-action) for the entire workflow

commons-test.mk
```diff
@@ -16,7 +16,7 @@ test-%:
 		--format short-verbose \
 		--rerun-fails=5 \
 		--packages="./..." \
-		--junitfile TEST-$*.xml \
+		--junitfile unit-tests.xml \
```
mdelapenya, line 19: I do prefer `TEST-unit.xml` instead.

.github/workflows/ci-test-go.yml
mdelapenya: Never mind, I just saw https://github.com/testcontainers/testcontainers-go/pull/1614/files#diff-f18f81daf1f8305a66be78cac5d122d165f39391f0348477f133858018585cbcR19

commons-test.mk
mmorel-35, line 19:
```suggestion
		--junitfile TEST-unit.xml \
```

.github/workflows/ci-test-go.yml
mmorel-35, line 119:
```suggestion
            ${{ inputs.project-directory }}/TEST-unit.xml
```

sonar-project.properties
```diff
@@ -10,12 +10,11 @@ sonar.projectVersion=v0.23.0
...
 
 sonar.tests=.
 sonar.test.inclusions=**/*_test.go
 sonar.test.exclusions=**/vendor/**
 
-sonar.go.coverage.reportPaths=coverage.out
-sonar.go.golangci-lint.reportPaths=golangci-lint.report.xml
-sonar.go.tests.reportPaths=TEST-unit.xml
+sonar.go.coverage.reportPaths=**/coverage.out
+sonar.go.tests.reportPaths=**/unit-tests.xml
```
mmorel-35, line 20:
```suggestion
sonar.go.tests.reportPaths=**/TEST-unit.xml
```

mdelapenya (commented):
This LGTM, although before approving I'd like to first discuss https://github.com/testcontainers/testcontainers-go/pull/1614/files#r1324001839

Right after that, I think this is ready to go.

Again, probably this is not adding features to the end-users, but adding lots of value to the maintainability of the project 🚀, which makes a lot for me. Thanks!!!

mmorel-35:
By the way, you’ll need to update the pre release script so the version is also updated in sonar properties

mdelapenya (approved):

mdelapenya:
> By the way, you’ll need to update the pre release script so the version is also updated in sonar properties

Will do it in a follow-up, thanks!

mdelapenya:
@mmorel-35 here you have it: https://github.com/testcontainers/testcontainers-go/pull/1615

PR #1618, opened 2023-09-13
feat: add RabbitMQ support

mdelapenya:
## What does this PR do?
This PR adds the RabbitMQ module, using Java module as reference implementation: all the options are implemented (see docs site)

It also fixes docs in postgres (a block that was removed) and the examples template, wording.

To implement the execution of startup commands, as demonstrated in the testable examples for this module, we are consuming the `testcontainers.WithStartupCommand` functional option from the core, which accepts a variadic argument of `testcontainers.Executable`. The latter is an interface with one single method: `AsCommand() []string`. Therefore, users of the RabbitMQ module (and eventually any other module) will be able to run arbitrary commands on the container right after it has started. The users will be responsible of creating valid commands.

NOTE: we are sharing some examples on how to create those commands. Please take a look at the `types_test.go` file.

All the options in the RabbitMQ module will be rendered into a `.conf` file, which will be the default configuration for the RabbitMQ container. We will leverage Go templates to render this conf file.

## Why is it important?
More modules!


modules/rabbitmq/rabbitmq.go
```diff
@@ -0,0 +1,294 @@
...
+// See withExecutable.
+func WithVirtualHostLimit(l VirtualHostLimit) testcontainers.CustomizeRequestOption {
+	return withExecutable(l)
+}
+
+// WithVirWithusertualHost declares the user on the RabbitMQ container, using "rabbitmqadmin".
+// See withExecutable.
+func WithUser(u User) testcontainers.CustomizeRequestOption {
+	return withExecutable(u)
+}
```
eddumelendez, line 218: In general, all of those fn helpers can be achieved by `container.Exec` but are coupling the container implementation to the tool cli and that's not intended. tc-java rabbitmq implementation should deprecate and remove those helper methods in future versions. At least, that was the last agreement on what the testcontainers implementation should provide.

modules/rabbitmq/rabbitmq.go
```diff
@@ -0,0 +1,294 @@
...
+	}
+}
+
+// WithSSL enables SSL on the RabbitMQ container, adding the necessary environment variables,
+// files and waiting conditions.
+// From https://hub.docker.com/_/rabbitmq: "As of RabbitMQ 3.9, all of the docker-specific variables
+// listed below are deprecated and no longer used. Please use a configuration file instead;
+// visit https://rabbitmq.com/configure to learn more about the configuration file. For a starting point,
+// the 3.8 images will print out the config file it generated from supplied environment variables.
+func WithSSL(sslSettings SSLSettings) testcontainers.CustomizeRequestOption {
```
eddumelendez, line 254: take into account the following [PR comment](https://github.com/testcontainers/testcontainers-java/pull/7338#pullrequestreview-1555502237)

eddumelendez (changes requested):

modules/rabbitmq/rabbitmq.go
mdelapenya: Ok, I see. But if we want to release a module in the same version as Java, then this implementation is fine, right?

We can evolve the module once we a new design for providing such capabilities is available. Or is the intention to remove them all without replacement?

kiview: > Or is the intention to remove them all without replacement?

That is currently the preferred case, yes. Like so often with modules, it comes down to the trade-off between _integration-surface_, _api-surface_ and _usability_. I think our current preference lies with minimizing the _integration-surface_.

Hence, I would prefer to remove the methods here in Go and merge the PR, independent from the methods still being present in Java.

mdelapenya: I'm going to proceed and remove the majority of the WithXXX functions, and keep the examples, but calling container.Exec instead. An approach could be keeping the types in the test code, for reference and not to maintain them as production code, and use them in the examples to produce the commands to be executed into the container

modules/rabbitmq/rabbitmq.go
mdelapenya: I've not implemented yet the SSL improvements. Should I proceed with the current state or rework the PR to remove the usage of the deprecated variables. It will define the lower version of RabbitMQ to use: +3.9

eddumelendez: IMO new Testcontainers implementations should take into account the most generic and compatible approach. In this case, using the configuration file is an option which will work for versions older than 3.9 and new ones. The implementation will differ from Testcontainers for Java but I think it will be a good one too

mdelapenya:
@eddumelendez I think this is ready to Go!

mdelapenya:
@eddumelendez we are going to merge this, and because it's not released yet, we can improve it in follow-up PRs.

PR #1644, opened 2023-09-18
modulegen: generate sonar configuration

mmorel-35:
## What does this PR do?

Generate `sonar.go.tests.reportPaths` in `sonar-project.properties`

## Why is it important?

`sonar.go.tests.reportPaths` doesn't supports wilcards so all the paths needs to be listed to avoid seeing this :

```
INFO: Sensor Go Unit Test Report [go]
ERROR: Test report can't be loaded, file not found: '/github/workspace/**/TEST-unit.xml', ignoring this file.
```


mdelapenya (approved):
LGTM, thanks!

At the same time it's weird that sonarqube does not accept glob expressions in the test reports path 🤷

PR #1648, opened 2023-09-20
docs: have common options in a common markdown file

mdelapenya:
## What does this PR do?
This PR adds support for including markdown files into others, leveraging the `mkdocs-include-markdown-plugin` plugin.

We will define a shared markdown file for all the available and common functional options, and will include that file into all existing modules and template.

## Why is it important?
When a new functional option appears, then just adding it to the file will work without having to add it to all the existing modules.


mdelapenya:
Merging, as this PR does not impact the core or the modules, just the docs

PR #1650, opened 2023-09-21
docs: remove shared docs in template

mdelapenya:
## What does this PR do?
It removes a block that is not needed anymore. Probably here after a wrong rebase

## Why is it important?
That block is already appended by a markdown include.

## Related issues

- Completes #1648


mdelapenya:
Merging, as it does not impact the tests

PR #1668, opened 2023-09-22
make ryuk timeouts configurable via properties file

srenatus:
## What does this PR do?

This adds two new properties, like

    ryuk.connection.timeout=2m
    ryuk.reconnection.timeout=3m

to control the timeouts used with the reaper container.

Their defaults are 1 minute (connection) and 10 seconds (reconnection), matching the defaults of the ryuk container's env vars.

## Why is it important?

In an E2E run, TC had problems re-creating the reaper container after it had stopped because the reconnection timeout was exceeded. With this, we can workaround that problem by increasing the timeout.

## Related issues

[🗨️ Discussed in Slack](https://testcontainers.slack.com/archives/CBWBNR76G/p1695319794716519)

## How to test this PR

Unit tests have been updated and added.

## Follow-ups

1. Some properties can also be controlled via env vars. We could do the same for these two new ones.
2. It might make sense to also add support for `ryuk.verbose=true`.
3. The actual underlying problem, being unable to start a second reaper instance after the first one went away, would also be good to fix, I think.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -6,6 +6,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"sync"
+	"time"
 
 	"github.com/magiconair/properties"
 )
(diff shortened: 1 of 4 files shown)


mdelapenya (approved):
This looks great! Thanks for being so quick with the implementation.

LGTM

mdelapenya:
> The actual underlying problem, being unable to start a second reaper instance after the first one went away, would also be good to fix, I think.

Could you elaborate a bit more, maybe in a different issue? 🙏

srenatus:
> Could you elaborate a bit more, maybe in a different issue? 🙏

Sure! I'll try to isolate the issue and share the test file -- in another issue -- early next week 🤞

srenatus:
✔️ https://github.com/testcontainers/testcontainers-go/issues/1671

PR #1692, opened 2023-09-27
fix: avoid checking for the reaper out of its mutex

mdelapenya:
## What does this PR do?
This PR removes redudant checks for getting the test session ID, simply delegating it to the sessionID global value, not reading it from the reaper (which already uses the very same value).

## Why is it important?
Having a nil-check for the `reaperInstance` to get the test sessionID introduced a race condition, as it's only set to not nil under a code block that is protected by a Mutex. Therefore, the check caused problems instead of benefits. I added that check intentionally, thinking that always reading from the reaper could be a good idea. But it was not.

## Related issues

- Caused by #1513

## How to test this PR

This should demonstrate the bug (before this fix) and the fix:

```go
go test -timeout 600s -run ^Test_BuildImageWithContexts$ . -race -count=1 -v
```

Also running all reaper tests:

```go
go test -timeout 600s -run "^(TestContainerStartsWithoutTheReaper|TestContainerStartsWithTheReaper|TestContainerStopWithReaper|TestContainerTerminationWithReaper|TestContainerTerminationWithoutReaper|Test_NewReaper|Test_ReaperForNetwork|Test_ReaperReusedIfHealthy|TestReaper_reuseItFromOtherTestProgramUsingDocker)$" github.com/testcontainers/testcontainers-go -count=1 -race -v
```

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -877,18 +877,13 @@ func (p *DockerProvider) CreateContainer(ctx context.Context, req ContainerReque
 		opt(&reaperOpts)
 	}
 
-	sessionID := testcontainerssession.SessionID()
-	if reaperInstance != nil {
-		sessionID = reaperInstance.SessionID
-	}
-
 	tcConfig := p.Config().Config
 
 	var termSignal chan bool
 	// the reaper does not need to start a reaper for itself
 	isReaperContainer := strings.EqualFold(req.Image, reaperImage(reaperOpts.ImageName))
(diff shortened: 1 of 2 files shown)


mdelapenya:
As a follow up, I think we could include the `-race` flag whenever running tests in the CI. I think it's a good practice that could help us identify race conditions the soonest.

I think I'll do it in a separate PR.

PR #1693, opened 2023-09-27
fix TestContainerRunningCheckingStatusCode to pass on arm platforms

gflarity:
## What does this PR do?

This PR uses the non-alpine influxdb image for the TestContainerRunningCheckingStatusCode test. There's no ARM alpine image for influx db. See https://github.com/influxdata/influxdata-docker/issues/335.

## Why is it important?

The test won't pass with Docker running under ARM. Example: Mac OS with Apply Silicon and Docker Desktop.  Seems like a simple fix to get the test passing more broadly.

## Related issues

N/A

<!-- Recommended
## How to test this PR

This should work under both x86 and ARM since it's just using the default non-alpine image (tested on ARM).

### DIFF
--- a/docker_test.go
+++ b/docker_test.go
@@ -1836,8 +1836,9 @@ func TestContainerCapAdd(t *testing.T) {
 func TestContainerRunningCheckingStatusCode(t *testing.T) {
 	ctx := context.Background()
 	req := ContainerRequest{
-		Image:        "influxdb:1.8.10", // influxdb doesn't provide an alpine+arm build (https://github.com/influxdata/influxdata-docker/issues/335)
-		ExposedPorts: []string{"8086/tcp"},
+		Image:         "influxdb:1.8.10-alpine",
+		ExposedPorts:  []string{"8086/tcp"},
+		ImagePlatform: "linux/amd64", // influxdb doesn't provide an alpine+arm build (https://github.com/influxdata/influxdata-docker/issues/335)
 		WaitingFor: wait.ForAll(
 			wait.ForHTTP("/ping").WithPort("8086/tcp").WithStatusCodeMatcher(
 				func(status int) bool {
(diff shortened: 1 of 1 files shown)


docker_test.go
```diff
@@ -1836,7 +1836,7 @@ func TestContainerCapAdd(t *testing.T) {
 func TestContainerRunningCheckingStatusCode(t *testing.T) {
 	ctx := context.Background()
 	req := ContainerRequest{
-		Image:        "influxdb:1.8.10-alpine",
+		Image:        "influxdb:1.8.10", // influxdb doesn't provide an alpine+arm build (https://github.com/influxdata/influxdata-docker/issues/335)
 		ExposedPorts: []string{"8086/tcp"},
```
mdelapenya, line 1840: I think this change is fine, and want to ask you about using the `ImagePlatform` field from the container request:
```suggestion
		Image:         "influxdb:1.8.10-alpine",
		ExposedPorts:  []string{"8086/tcp"},
		ImagePlatform: "linux/amd64", // influxdb doesn't provide an alpine+arm build (https://github.com/influxdata/influxdata-docker/issues/335)
```

The image size is practically irrelevant in bytes, although the non-alpine weights twice:

```table
influxdb                     1.8.10                     ba9488c478c5   7 days ago      295MB
influxdb                     1.8.10-alpine              3ab4fc5517d7   7 weeks ago     164MB
```


Whatever you prefer is fine.

gflarity: Hmmm, I'll give it a shot. I forogt you can run amd64 on Mac Docker Desktop .

gflarity: Confirmed, this runs fine. I'm just happy the test passes now, so whichever you prefer.

mdelapenya: Cool, I think it's more clear with the image platform being explicit. I'll approve it as is, if I may

mdelapenya (approved):

PR #1694, opened 2023-09-27
TestBuildContainerFromDockerfileWithDockerAuthConfig working on MacOS

gflarity:
## What does this PR do?

Port 5000 is used by "Control Center" on Mac OS Monterey and above. So it's essentially a reserved port on Macs. This PR simply uses port 5001 instead so that TestBuildContainerFromDockerfileWithDockerAuthConfig will pass on MacOS. 

## Why is it important?

MacOS is a fairly common developer platform. Seems like the tests should pass there as well as Linux. 

## Related issues

N/A

<!-- Recommended
## How to test this PR

Tests should just run as normal. Simple port change.


mdelapenya (approved):
LGTM! IIRC I've seen that port conflict on Mac in the past, but don't remember if I simply stopped the service or what.

This is cleaner, thanks for the fix

PR #1700, opened 2023-09-29
fix race condition in Test_StartStop

gflarity:
## What does this PR do?

Fixes an intermittent failure in Test_StartStop caused by a race condition with the TestLogConsumer and the test itself. The root of the issue is that on certain systems (like my slow M1 Macbook) it can take a while for the producer to spin up, mean while the old producer is still producing.  The cleanest way to fix this is just to update the TestLogConsumer to publish log lines and then wait for the expected lines before proceeding, in particular the ones after we call StartLogProducer for the second time. 

## Why is it important?

Fixes a test that can sometimes fail, particularly on slower systems. 

## Follow-ups

There's another potential race condition inside the Log Producer Start/Stop logic.  I'll submit it as a separate PR since it didn't actually solve this race condition when I tried it first, but it seems like it's a good idea to implement still.


gflarity:
FYI, all the tests seem to pass reliably on my system after this one :)

gflarity:
@mdelapenya can we try re-running these tests?

This change only just for log messages to come through before stopping the producer.  I skimmed some of the error messages and they seem pretty strange:

```
mount: permission denied (are you root?)
Could not mount /sys/kernel/security.
AppArmor detection and --privileged mode might break.
sed: write error
```

I also just re-ran make all-tests locally and it passed... very strange.

logconsumer_test.go
```diff
@@ -21,20 +21,32 @@ import (
...
-		g.Ack <- true
+		g.Done <- true
 		return
 	}
-
+	g.Accepted <- s
 	g.Msgs = append(g.Msgs, s)
 }
 
+// devNullAcceptorChan returns string chanel that essentially sends all strings to dev null
```
mdelapenya, line 39: Typo:
```suggestion
// devNullAcceptorChan returns string channel that essentially sends all strings to dev null
```

logconsumer_test.go
```diff
@@ -21,20 +21,32 @@ import (
 const lastMessage = "DONE"
 
 type TestLogConsumer struct {
-	Msgs []string
-	Ack  chan bool
+	Msgs     []string
+	Done     chan bool
+	Accepted chan string
```
mdelapenya, line 26: Could you add a comment explaining why we are adding the Accepted channel here, in combination with the Msgs slice? 🙏

mdelapenya (approved):
I left two comments to be addressed before the merge. They are not blockers in any case, but want to check with you first.

Said that, LGTM, thanks for this!

gflarity:
Updated commit with the request changes. 👍

gflarity:
These failures must unrelated, only test helper was updated in this PR.

mdelapenya (approved):
LGTM, thanks for this! Never had time myself to debug the flakiness in this test. Really appreciate your efforts!

PR #1703, opened 2023-09-30
feat: add module to support InfluxDB v1.x

JJCinAZ:
## What does this PR do?

Added support for InfluxDB v1.x containers.

## Why is it important?

Enhanced testcontainers-go project.


mmorel-35:
Hi @JJCinAZ ,

Have you read [this page](https://golang.testcontainers.org/modules/) ?

Some expected configurations configured by modulegen seems to be missing from your PR.

modules/influxdb/influxdb.go
```diff
@@ -0,0 +1,144 @@
+package influxdb
+
+import (
+	"context"
+	"fmt"
+	"github.com/testcontainers/testcontainers-go"
+	"github.com/testcontainers/testcontainers-go/wait"
```
mmorel-35, line 7: Please, don’t forget to lint your files 😉 !

modules/influxdb/influxdb.go
```diff
@@ -0,0 +1,144 @@
...
+const defaultImage = "influxdb:1.8"
+
+// }
+
+// InfluxDbContainer represents the MySQL container type used in the module
+type InfluxDbContainer struct {
+	testcontainers.Container
+}
+
+// RunContainer creates an instance of the MySQL container type
```
eddumelendez, line 22:
```suggestion
// RunContainer creates an instance of the InfluxDB container type
```

eddumelendez (commented):
Thanks for your contribution, @JJCinAZ ! I left some comments. Also, I think it should be influxdbv1 instead. Taking into account v2 will be supported anytime soon. WDYT @mdelapenya ?

mdelapenya:
Hey @JJCinAZ I'd recommend you using our module generator to bootstrap the module, as it will generate the right files also updating the descriptors files (GH actions, mkdocs, etc) 🙏

docs/modules/influxdb.md
```diff
@@ -0,0 +1,74 @@
...
+- `context.Context`, the Go context.
+- `testcontainers.ContainerCustomizer`, a variadic argument for passing options.
+
+### Container Options
+
+When starting the container, you can pass options in a variadic way to configure it.
+
+!!!tip
+
+    You can find configuration information for the InfluxDB image on [Docker Hub](https://hub.docker.com/_/mysql) and a list of possible
```
mdelapenya, line 40: Probably a wrong copy&paste:
```suggestion
    You can find configuration information for the InfluxDB image on [Docker Hub](https://hub.docker.com/_/influxdb) and a list of possible 
```

modules/influxdb/docker-entrypoint-initdb.d/test_v1_data.sh
```diff
@@ -0,0 +1 @@
+influx -import -path=/docker-entrypoint-initdb.d/NOAA_data.txt.gz -compressed -precision=s -database=NOAA_water_database
```
mdelapenya, line 1: Is this test script and its data needed by the module or only used in tests? If the latter, please add it to a `testdata` directory and copy it into the container. Otherwise all users of the module will receive the files in here

modules/influxdb/influxdb.go
```diff
@@ -0,0 +1,144 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return &InfluxDbContainer{container}, nil
+}
+
+func (c *InfluxDbContainer) MustHaveConnectionUrl(ctx context.Context, tls bool) string {
+	connectionString, err := c.ConnectionUrl(ctx, tls)
+	if err != nil {
+		panic(err)
```
mdelapenya, line 80: Do we want to panic here in the module code, or simply raise the error?

mdelapenya: Oh wait, this is a MustHave method, I guess to avoid the if err != nil, right? Forget about my previous comment.

I wonder if this pattern is common, as we do not usually use it, although I could see the value on it. If you think we can start adopting it, I'd appreciate more feedback around its usage 🙏

modules/influxdb/influxdb.go
```diff
@@ -0,0 +1,144 @@
...
+		cf := testcontainers.ContainerFile{
+			HostFilePath:      configFile,
+			ContainerFilePath: "/etc/influxdb/influxdb.conf",
+			FileMode:          0o755,
+		}
+		req.Files = append(req.Files, cf)
+	}
+}
+
+// WithInitDb will copy a 'docker-entrypoint-initdb.d' directory to the container.
```
mdelapenya, line 132: I think we cannot enforce users to create a `docker-entrypoint-initdb.d` dir into their projects, but let them know that the dir passed to this function will be copied to the container with that name.

I'd advocate for hardcoding `/docker-entrypoint-initdb.d` in https://github.com/testcontainers/testcontainers-go/pull/1703/files#diff-328ce3ff8f6404a82f6ade95861e87cb8e19ef7d45dd4579fad562d8dfde4f7fR139, removing the `docker-entrypoint-initdb.d` suffix from https://github.com/testcontainers/testcontainers-go/pull/1703/files#diff-328ce3ff8f6404a82f6ade95861e87cb8e19ef7d45dd4579fad562d8dfde4f7fR138

JJCinAZ:
@mdelapenya  I think that InfluxDB v2 is an abandoned product in so far as Influx has completely changed for v3 and abandoned the Flux language they introduced in v2.  At this point, I have little faith they will actually have an OSS v3, so many of us are stuck at v1.8 until we migrate completely.  If someone else has need for good v2 support, we can revisit how to make that happen.

modules/influxdb/influxdb.go
JJCinAZ: The Must pattern is quite common in Go.  For example, regexp.MustCompile or the template.Must() wrapper.  It's useful to avoid the temporary err and if-then-panic.  If the caller doesn't want the panic, they simply call ConnectionUrl and handle the error themselves.

mdelapenya: Thanks for the explanation. I like this pattern, so we can start adopting it more and more 👍

mdelapenya: In any case, lets add the proper method comment explaining the usage and the expected behaviours, as in regex: https://cs.opensource.google/go/go/+/refs/tags/go1.21.3:src/regexp/regexp.go;l=313-315

mdelapenya:
> If someone else has need for good v2 support, we can revisit how to make that happen.

So your intention is to create an InfluxDB v1-only module, right?

JJCinAZ:
Yes, v1 only.

On Oct 18, 2023, at 12:54 AM, Manuel de la Peña ***@***.***> wrote:



If someone else has need for good v2 support, we can revisit how to make that happen.

So your intention is to create an InfluxDB v1-only module, right?

mdelapenya:
@JJCinAZ sorry for not going on this PR since October 🙏 For that reason I did not want to cause you more trouble and went ahead and added some commits on top to update the code to the current state of the main branch.

If possible, I'd like you to review it and give your 👍 👎 

Thanks!

JJCinAZ:
@mdelapenya Thanks for adding the extra commits.   I'm not familiar with the documentation system you use and do not use external linters in my Go projects, so thanks for cleaning those up.

modules/influxdb/influxdb.go
```diff
@@ -0,0 +1,144 @@
...
+		Image:        defaultImage,
+		ExposedPorts: []string{"8086/tcp", "8088/tcp"},
+		Env: map[string]string{
+			"INFLUXDB_BIND_ADDRESS":          ":8088",
+			"INFLUXDB_HTTP_BIND_ADDRESS":     ":8086",
+			"INFLUXDB_REPORTING_DISABLED":    "true",
+			"INFLUXDB_MONITOR_STORE_ENABLED": "false",
+			"INFLUXDB_HTTP_HTTPS_ENABLED":    "false",
+			"INFLUXDB_HTTP_AUTH_ENABLED":     "false",
+		},
```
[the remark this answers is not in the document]
mdelapenya: They come from here: https://docs.influxdata.com/influxdb/v1/administration/config/

modules/influxdb/influxdb.go
```diff
@@ -0,0 +1,144 @@
...
+		return "", err
+	}
+	host, err := c.Host(ctx)
+	if err != nil {
+		return "", err
+	}
+	var connectionString string
+	if tls {
+		connectionString = fmt.Sprintf("https://%s:%s", host, containerPort.Port())
+	} else {
```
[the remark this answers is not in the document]
mdelapenya: For simplicity, I've removed the HTTPS part. We can add TLS support in the future.

modules/influxdb/influxdb.go
mdelapenya: After this time, I ended up doing exactly the same as in the original PR. SO let's keep it simple by the moment, and improve it if needed.

PR #1719, opened 2023-10-03
feat: support for replacing images with custom substitutions

### LINKED ISSUE #1715: [Enhancement]: Provide a way to customize extra docker containers used
### Proposal

In more locked down / secured environments, it can be problematic to pull images from docker hub and run them without additional precautions.

One need is to prefix images so that they can be pulled from a docker registry other than docker hub. This is the usual mechanism for using docker image proxies, caches, etc.

Another need is to pin images to specific hashes to protect against supply chain attacks. When images are not pinned to a specific hash, an attacker with control of the docker account that hosts the images can publish a new image with the same tag that contains malicious code.

The latter issue also exists for images referenced by this repo normally, so maybe it can be solved differently from the first. Maybe images used by testcontainers can always be pinned by hash by default?

mdelapenya:
## What does this PR do?
This PR adds a mechanism for modifying the Docker image used in the container request, and for that, an slice of ImageSubstitutors can be configured at the container request level.

At the moment we resolve the tag to pull/use the Docker image, it will run the defined substitutors in order to calculate the final image.

To facilitate users to provide their own substitutors, we have exposed an interface, so users of tc-go can implement their own substitutors. Examples could be:

- prepend a custom registry namespace: e.g. myregistry.com/docker.io/redis:latest
- massage the tag: e.g. docker.io/redis:6-MYCUSTOMHASH

The PR includes documentation for the substitutor interface, also adding them to each module's functional options.

## Why is it important?
In more locked down / secured environments, it can be problematic to pull images from Docker Hub and run them without additional precautions.

An image name substitutor converts a Docker image name, as may be specified in code, to an alternative name. This is intended to provide a way to override image names, for example to enforce pulling of images from a private registry.

## Related issues

- Closes #1715

## Follow-ups

We can support setting custom substitutions in the testcontainers.properties file. Please see https://java.testcontainers.org/features/image_name_substitution/#automatically-modifying-docker-hub-image-names to check requirements. cc/ @eddumelendez


docker.go
```diff
@@ -861,6 +861,8 @@ func (p *DockerProvider) CreateContainer(ctx context.Context, req ContainerReque
 		}
 	}
 
+	tag := req.Image
```
vikstrous2, line 864:
```suggestion
	imageName := req.Image
```

generic.go
```diff
@@ -60,6 +60,24 @@ func WithImage(image string) CustomizeRequestOption {
 	}
 }
 
+// imageSubstitutor {
+// ImageSubstitutor represents a way to substitute container image names
+type ImageSubstitutor interface {
+	// Description returns the name of the type and a short description of how it modifies the image.
+	// Useful to be printed in logs
+	Description() string
+	Substitute(image string) string
```
vikstrous2, line 69: I have 2 use cases: prefixing and pinning by hash.

For prefixing this works fine.

For pinning by hash, this is awkward because I'd have to write a substitutor that panics when it encounters an unknown image, right? Should I be even thinking about that use case when considering this interface or should I expect images to already be pinned by hash in the testcontainers code itself? I could add a verification in my image substitutor to ensure that images are pinned by hash and panic if they are not, so if something goes wrong upstream and an image is not pinned, it doesn't impact us.

mdelapenya: @vikstrous2 any feedback from this? I'd like to merge the image substitutor part, probably adding error handling to the substitution part, but want also to understand more on the pinned versions use case you mentioned 🙏

vikstrous2: Sorry for the delay. Pinning the images mitigates some of the supply chain risk. If your docker hub account is compromised, we wouldn't immediately pick up a malicious image. The code, without being changed, wouldn't change its behavior based on an external dependency.

A recent example of a high profile attack like this is the one that hijacked the circle ci install script: https://circleci.com/blog/jan-4-2023-incident-report/

Re: error handling, yeah, adding an error return parameter is better than having to panic.

mdelapenya: Ok, will add the error handling there.

Regarding the pinned images, will do it in a separate PR. I now understand the risk in supply chain scenarios. I appreciate for feedback 🙏

docker.go
mdelapenya: Done, thanks!

vikstrous2 (approved):
Love it! Thanks! I can't comment on if it works, but the API looks easy to use.

container_test.go
```diff
@@ -324,6 +325,113 @@ func Test_GetLogsFromFailedContainer(t *testing.T) {
...
+			expectedError: errSubstitution,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			ctx := context.Background()
+			req := ContainerRequest{
+				Image:             test.image,
+				ImageSubstitutors: test.substitutors,
```
eddumelendez, line 408: I wonder if `ImageSubstitutors` should apply to `Image` instead of the `ContainerRequest`.

mdelapenya: Image is a string. Is that what you mean?

There is a `WithImageSubstitutors` functional option that will be used to build the request. For modules, the API is more "elegant" nowadays, for the genetic container, before V1, it's part of the container request.

mdelapenya:
As discussed internally, and as a follow-up to this PR, we are going to provide a secondary way to substitute images at the properties level:

> You can then configure Testcontainers to apply the prefix registry.mycompany.com/mirror/ to every image that it tries to pull from Docker Hub. This can be done in one of two ways:
>
> - Setting environment variables TESTCONTAINERS_HUB_IMAGE_NAME_PREFIX=registry.mycompany.com/mirror/
> - Via config file, setting hub.image.name.prefix in the ~/.testcontainers.properties file in your user home directory
> Testcontainers will automatically apply the prefix to every image that it pulls from Docker Hub - please verify that all [the required images](https://java.testcontainers.org/supported_docker_environment/image_registry_rate_limiting/#which-images-are-used-by-testcontainers) exist in your registry.
>
> Testcontainers will not apply the prefix to:
>
> - non-Hub image names (e.g. where another registry is set)
> - Docker Hub image names where the hub registry is explicitly part of the name (i.e. anything with a docker.io or registry.hub.docker.com host part)

Please read more on this here: https://java.testcontainers.org/features/image_name_substitution/#automatically-modifying-docker-hub-image-names

PR #1721, opened 2023-10-03
K6 module

pablochacin:
## What does this PR do?

Implements a module for running k6.

## Why is it important?

K6 is a well-known tool for testing. This module allows the reuse of existing `k6` test scripts in tests using `TestContainers`. This is particularly interesting when implementing `e2e` tests.

## Note to reviewers

There are a few things I'm not totally satisfied with in its present form, and I would like to address in future iterations:

- The test script passed to `k6` is presently dfined as an option (`WithTestScript`), but it reality a "mandatory option" and if not specified, the test execution will fail. I considered adding it as an argument to the `RunContainer` function, but this seems not to be how other modules work.
```golanf
RunContainer(ctx context.Context, string script, options....)
```
- Specifying a build cache also requires running the container with the `AsUser` option to give it access to the cache directory. This is required because the image runs with its own user and mounting a local directory would fail with an access error. Maybe there's an alternative way to do this. For example, if instead of using a local directory, the cache is passed as a volume, this wouldn't be a problem. But requiring the user to create a docker volume before the test seems a little overkill.

~~**Edit:** I found a way to mount a regular directory in the image without requiring specifying an user.~~


.vscode/.testcontainers-go.code-workspace
mmorel-35, line 1: Are you providing a n example or a module ?
You have traces of both in your PR

.github/workflows/ci.yml
```diff
@@ -123,7 +123,7 @@ jobs:
     needs: test-modules
     strategy:
       matrix:
-        module: [cockroachdb, consul, nginx, toxiproxy]
+        module: [cockroachdb, consul, k6, nginx, toxiproxy]
```
mdelapenya, line 126:
```suggestion
        module: [cockroachdb, consul, nginx, toxiproxy]
```

.vscode/.testcontainers-go.code-workspace
```diff
@@ -13,6 +13,10 @@
             "name": "example / consul",
             "path": "../examples/consul"
         },
+        {
+            "name": "example / k6",
+            "path": "../examples/k6"
+        },
```
mdelapenya, line 19:
```suggestion
```

docs/examples/k6.md
```diff
@@ -0,0 +1,47 @@
+# k6
```
mdelapenya, line 1: This file is possibly a leftover of a previous run as an example, instead of a module.

Let's remove it :)

examples/k6/Makefile
```diff
@@ -0,0 +1,5 @@
+include ../../commons-test.mk
```
mdelapenya, line 1: Same, here. Let's remove this file and the entire `examples/k6` dir 🙏 .

modules/k6/examples_test.go
```diff
@@ -0,0 +1,26 @@
...
+		panic(err)
+	}
+
+	container, err := k6.RunContainer(ctx, k6.WithTestScript(absPath))
+	if err != nil {
+		panic(err)
+	}
+	//}
+
+	container.Terminate(ctx)
```
mdelapenya, line 25: Please checkout this an example of how to do it:

https://github.com/testcontainers/testcontainers-go/blob/main/modules/clickhouse/examples_test.go#L34-L39C2

```go
	defer func() {
		if err := clickHouseContainer.Terminate(ctx); err != nil {
			panic(err)
		}
	}()
	// }
```

Let's make sure the defer block is under the codeblock used by our docs

modules/k6/k6.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/testcontainers/testcontainers-go"
+	"github.com/testcontainers/testcontainers-go/wait"
+)
+
+// K6Container represents the K6 container type used in the module
+type K6Container struct {
+	testcontainers.Container
+}
+
+
```
mdelapenya, line 17: Running `make lint` from the root of your module will fix the lint errors 🤞

modules/k6/k6.go
```diff
@@ -0,0 +1,66 @@
...
+			},
+			Target: testcontainers.ContainerMountTarget(target),
+		}
+		req.Mounts = append(req.Mounts, mount)
+		req.Cmd = append(req.Cmd, target)
+	}
+}
+
+// WithOptions pass the given options to the k6 run command
+func WithOptions(options...string) testcontainers.CustomizeRequestOption {
```
mdelapenya, line 37: Wdyt if we rename this option to `WithCmdOptions`? I think it will be even more clear that the passed options will end up into the CMD of the image. Not a blocker though

modules/k6/k6.go
```diff
@@ -0,0 +1,66 @@
...
+	return func(req *testcontainers.GenericContainerRequest) {
+		req.Cmd = append(req.Cmd, options...)
+	}
+}
+
+// RunContainer creates an instance of the K6 container type
+func RunContainer(ctx context.Context, opts ...testcontainers.ContainerCustomizer) (*K6Container, error) {
+	req := testcontainers.ContainerRequest{
+		Image: "szkiba/k6x",
```
mdelapenya, line 46: Let's use a fixed tag instead of always `latest`, so we do not get surprises on eventual upgrades.

modules/k6/k6_test.go
```diff
@@ -0,0 +1,45 @@
...
+	"context"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+func TestK6(t *testing.T) {
+	ctx := context.Background()
+
+	absPath, err := filepath.Abs("./scripts/test.js")
```
mdelapenya, line 14: Let's build the path to be cross-OSs
```suggestion
	absPath, err := filepath.Abs(filepath.Join("scripts", "test.js"))
```

sonar-project.properties
```diff
@@ -18,4 +18,4 @@ sonar.test.inclusions=**/*_test.go
...
+sonar.go.tests.reportPaths=TEST-unit.xml,examples/cockroachdb/TEST-unit.xml,examples/consul/TEST-unit.xml,examples/k6/TEST-unit.xml,examples/nginx/TEST-unit.xml,examples/toxiproxy/TEST-unit.xml,modulegen/TEST-unit.xml,modules/artemis/TEST-unit.xml,modules/clickhouse/TEST-unit.xml,modules/compose/TEST-unit.xml,modules/couchbase/TEST-unit.xml,modules/elasticsearch/TEST-unit.xml,modules/gcloud/TEST-unit.xml,modules/k3s/TEST-unit.xml,modules/k6/TEST-unit.xml,modules/kafka/TEST-unit.xml,modules/localstack/TEST-unit.xml,modules/mariadb/TEST-unit.xml,modules/mongodb/TEST-unit.xml,modules/mysql/TEST-unit.xml,modules/nats/TEST-unit.xml,modules/neo4j/TEST-unit.xml,modules/postgres/TEST-unit.xml,modules/pulsar/TEST-unit.xml,modules/rabbitmq/TEST-unit.xml,modules/redis/TEST-unit.xml,modules/redpanda/TEST-unit.xml,modules/vault/TEST-unit.xml
```
mdelapenya, line 21:
```suggestion
sonar.go.tests.reportPaths=TEST-unit.xml,examples/cockroachdb/TEST-unit.xml,examples/consul/TEST-unit.xml,examples/nginx/TEST-unit.xml,examples/toxiproxy/TEST-unit.xml,modulegen/TEST-unit.xml,modules/artemis/TEST-unit.xml,modules/clickhouse/TEST-unit.xml,modules/compose/TEST-unit.xml,modules/couchbase/TEST-unit.xml,modules/elasticsearch/TEST-unit.xml,modules/gcloud/TEST-unit.xml,modules/k3s/TEST-unit.xml,modules/k6/TEST-unit.xml,modules/kafka/TEST-unit.xml,modules/localstack/TEST-unit.xml,modules/mariadb/TEST-unit.xml,modules/mongodb/TEST-unit.xml,modules/mysql/TEST-unit.xml,modules/nats/TEST-unit.xml,modules/neo4j/TEST-unit.xml,modules/postgres/TEST-unit.xml,modules/pulsar/TEST-unit.xml,modules/rabbitmq/TEST-unit.xml,modules/redis/TEST-unit.xml,modules/redpanda/TEST-unit.xml,modules/vault/TEST-unit.xml
```

mdelapenya (commented):
Hey @pablochacin great job with the module! Thank you so much!! 

I added a few comments regarding some leftovers from previous runs, as @mmorel-35 pointed out (thank you), so let's remove the `examples/k6` files and references.

I have one question regarding the test execution, probably caused by my total inexperience in the k6 project, so sorry about it if it's a dumb question: I can imagine the JS script will try to connect to an application to perform some load, right? Should the k6 module provide a way to connect to that app, or it's not needed? ATM I cannot picture how the script interacts with the app under test.

pablochacin:
Hi @mdelapenya @mmorel-35  I'm adding a module and the example for that module. 
Should I submit them separately? For me, it makes more sense to add both, as the example is part of the documentation.

pablochacin:
> I have one question regarding the test execution, probably caused by my total inexperience in the k6 project, so sorry about it if it's a dumb question: I can imagine the JS script will try to connect to an application to perform some load, right? Should the k6 module provide a way to connect to that app, or it's not needed? ATM I cannot picture how the script interacts with the app under test.

@mdelapenya yes, that is the case. In general, the scripts use an environment variable to get the address of the target application. I was planning to add an example for this.

mdelapenya:
> Hi @mdelapenya @mmorel-35 I'm adding a module and the example for that module. Should I submit them separately? For me, it makes more sense to add both, as the example is part of the documentation.

Oh I see now. Then the docs are not clear enough. Our bad.

Please add an examples_test.go file (IIRC the generator should create it) in the k6 module dir, including the `ExampleRunContainer` function.

TBH I'm thinking about removing everything under `examples`, as they are just plain code snippets to copy paste, probably causing more confusion than benefits. Since we added testable examples support I think they are not needed anymore.

pablochacin:
> Oh I see now. Then the docs are not clear enough. Our bad.

This PR is still a draft. The documentation and examples are still a work in progress.

> TBH I'm thinking about removing everything under examples, 

Then it makes sense to remove the examples. I will put all of them under the module's `examples_test.go`.

pablochacin:
@mdelapenya I don't really understand this linter message. Could you please help me point out the issue? :pray: 

>   Running [/home/runner/golangci-lint-1.54.1-linux-amd64/golangci-lint run --out-format=github-actions --path-prefix=modules/k6 --verbose] in [/home/runner/work/testcontainers-go/testcontainers-go/modules/k6] ...
  Error: File is not `gci`-ed with --skip-generated -s standard -s default -s prefix(github.com/testcontainers) (gci)
  Error: File is not `gci`-ed with --skip-generated -s standard -s default -s prefix(github.com/testcontainers) (gci)
  Error: File is not `gci`-ed with --skip-generated -s standard -s default -s prefix(github.com/testcontainers) (gci)

modules/k6/k6.go
mdelapenya: Could it be a double line here? 🤔

modules/k6/k6.go
```diff
@@ -0,0 +1,66 @@
...
+			},
+			Target: testcontainers.ContainerMountTarget(target),
+		}
+		req.Mounts = append(req.Mounts, mount)
+		req.Cmd = append(req.Cmd, target)
+	}
+}
+
+// WithCmdOptions pass the given options to the k6 run command
+func WithCmdOptions(options...string) testcontainers.CustomizeRequestOption {
```
mdelapenya:
```suggestion
func WithCmdOptions(options... string) testcontainers.CustomizeRequestOption {
```

modules/k6/k6.go
```diff
@@ -0,0 +1,66 @@
...
+		req.Cmd = append(req.Cmd, options...)
+	}
+}
+
+// RunContainer creates an instance of the K6 container type
+func RunContainer(ctx context.Context, opts ...testcontainers.ContainerCustomizer) (*K6Container, error) {
+	req := testcontainers.ContainerRequest{
+		Image: "szkiba/k6x:v0.3.1",
+		Cmd:   []string{"run"},
+		WaitingFor: wait.ForExit(),
```
mdelapenya:
```suggestion
		Image:      "szkiba/k6x:v0.3.1",
		Cmd:        []string{"run"},
		WaitingFor: wait.ForExit(),
```

modules/k6/k6.go
```diff
@@ -0,0 +1,66 @@
...
+// and passes it to k6 as the test to run.
+// The path to the script must be an absolute path
+func WithTestScript(scriptPath string) testcontainers.CustomizeRequestOption {
+	return func(req *testcontainers.GenericContainerRequest) {
+		script := filepath.Base(scriptPath)
+		target := fmt.Sprintf("/tests/%s", script)
+		mount := testcontainers.ContainerMount{
+			Source: testcontainers.GenericBindMountSource{
+			    HostPath: scriptPath,
+			},
```
mdelapenya:
```suggestion
			Source: testcontainers.GenericBindMountSource{
				HostPath: scriptPath,
			},
```

modules/k6/k6.go
mdelapenya:
```suggestion
```

mmorel-35:
When I try it on codespaces from artemis module's where I have created a deliberated issue

```bash
/workspaces/testcontainers-go/modules/artemis $ make lint
golangci-lint run --out-format=github-actions --path-prefix=. --verbose -c /workspaces/testcontainers-go/.golangci.yml --fix
INFO [config_reader] Used config file ../../.golangci.yml
INFO [lintersdb] Active 11 linters: [errcheck errorlint gci gocritic gofumpt gosimple govet ineffassign misspell staticcheck unused] 
INFO [loader] Go packages loading at mode 575 (types_sizes|files|imports|name|compiled_files|deps|exports_file) took 245.382898ms 
INFO [runner/filename_unadjuster] Pre-built 0 adjustments in 511.104µs 
INFO [linters_context/goanalysis] analyzers took 0s with no stages 
INFO [runner] Fix issue &result.Issue{FromLinter:"gci", Text:"File is not `gci`-ed with --skip-generated -s standard -s default -s prefix(github.com/testcontainers)", Severity:"", SourceLines:[]string{"\t\"github.com/docker/go-connections/nat\""}, Replacement:(*result.Replacement)(0xc00114a6f0), Pkg:(*packages.Package)(0xc000f9f500), LineRange:(*result.Range)(nil), Pos:token.Position{Filename:"artemis.go", Offset:0, Line:7, Column:0}, HunkPos:0, ExpectNoLint:false, ExpectedNoLintLinter:""} with range {7 7} 
INFO [runner] fixer took 1.232051ms with stages: all: 1.232051ms 
INFO [runner] Issues before processing: 3, after processing: 0 
INFO [runner] Processors filtering stat (out/in): source_code: 1/1, path_shortener: 1/1, filename_unadjuster: 3/3, autogenerated_exclude: 3/3, exclude: 3/3, nolint: 1/1, uniq_by_line: 1/1, diff: 1/1, identifier_marker: 3/3, max_same_issues: 1/1, severity-rules: 1/1, cgo: 3/3, path_prettifier: 3/3, skip_files: 3/3, skip_dirs: 3/3, max_from_linter: 1/1, fixer: 0/1, exclude-rules: 1/3, max_per_file_from_linter: 1/1 
INFO [runner] processing took 1.745278ms with stages: fixer: 1.258429ms, nolint: 198.761µs, identifier_marker: 92.634µs, exclude-rules: 66.093µs, path_prettifier: 52.589µs, autogenerated_exclude: 43.16µs, source_code: 15.779µs, skip_dirs: 11.432µs, path_shortener: 1.483µs, cgo: 1.172µs, max_same_issues: 772ns, filename_unadjuster: 561ns, uniq_by_line: 391ns, exclude: 371ns, sort_results: 301ns, max_from_linter: 280ns, severity-rules: 280ns, skip_files: 260ns, max_per_file_from_linter: 230ns, diff: 170ns, path_prefixer: 130ns 
INFO [runner] linters took 457.480433ms with stages: goanalysis_metalinter: 455.644814ms 
INFO File cache stats: 1 entries of total size 3.0KiB 
INFO Memory: 9 samples, avg is 52.1MB, max is 84.7MB 
INFO Execution took 707.979266ms
```

See the relative path to the "Used config file ../../.golangci.yml" is different from your output. Any reason ?

mdelapenya:
When I fetched the PR before [420041c](https://github.com/testcontainers/testcontainers-go/pull/1721/commits/420041c98045ce3af95b86dc6bdf396a0947549e), I saw errors in the command output with the same lines that GH added as comments in the PR.

Basically saving the k6.go file on my VSCode fixed it, as it ran the automatic format for Go.

docs/modules/k6.md
```diff
@@ -74,4 +74,12 @@ The `AsUser` sets the user id and group id to be used when running the `k6` cont
 
 ```golang
 k6.RunContainer(ctx, AsUser(os.Getuid(), os.Getgid()), k6.WithTestScript("/tests/test.js"))
+```
+
+#### SetEvVar
```
mmorel-35, line 79:
```suggestion
#### SetEnvVar
```

pablochacin:
Hi I'm submitting this first version of the `k6` module. Thanks @mdelapenya and @mmorel-35 for your reviews of the draft.

pablochacin:
**Update**: Just after submitting the PR, I realized there is a way to specify the cache dir without requiring the user. I would prefer to make the changes in this PR because it really improves the user experience. What do you think?

mmorel-35 (approved):

pablochacin:
Hi. What is the preferred way to update a PR when it is out-of-date with the base branch? merge commit or rebase?

Screenshot at 2023-10-09 09-37-34

mmorel-35:
I would recommend with rebase when it's possible.

modules/k6/k6.go
```diff
@@ -0,0 +1,95 @@
...
+	return func(req *testcontainers.GenericContainerRequest) {
+		script := filepath.Base(scriptPath)
+		target := fmt.Sprintf("/tests/%s", script)
+		mount := testcontainers.ContainerMount{
+			Source: testcontainers.GenericBindMountSource{
+				HostPath: scriptPath,
+			},
+			Target: testcontainers.ContainerMountTarget(target),
+		}
+		req.Mounts = append(req.Mounts, mount)
```
mdelapenya, line 30: Let's use req.Files instead of Mounts. Why? Because Files are copied to the container, and mounts not. That will make it work with remote Docker engines.

modules/k6/k6.go
```diff
@@ -0,0 +1,95 @@
...
+	return func(req *testcontainers.GenericContainerRequest) {
+		req.Cmd = append(req.Cmd, "--env", fmt.Sprintf("%s=%s", variable, value))
+	}
+}
+
+// WithCache uses the given directory as a cache directory building the k6 binary.
+// The path to the directory must be an absolute path
+// Note: The container must run using an user that
+// has access to the directory. See AsUser option
+func WithCache(cacheDir string) testcontainers.CustomizeRequestOption {
```
mdelapenya, line 53: I wonder if we need to be mounted in the host. Do we need to access the cache from outside the container for any reason? Asking as a newbie in k6

modules/k6/examples_test.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+func ExampleRunContainer() {
+	// runK6Container {
+	ctx := context.Background()
+
+	gcr := testcontainers.GenericContainerRequest{
+		ProviderType: testcontainers.ProviderDocker,
+		ContainerRequest: testcontainers.ContainerRequest{
+			Image: "kennethreitz/httpbin",
```
mdelapenya, line 20: Let's use a fixed tag for this image if possible.

modules/k6/k6.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+// WithCmdOptions pass the given options to the k6 run command
+func WithCmdOptions(options ...string) testcontainers.CustomizeRequestOption {
+	return func(req *testcontainers.GenericContainerRequest) {
+		req.Cmd = append(req.Cmd, options...)
+	}
+}
+
+// SetEnvVar sets an environment variable for the test script
```
mdelapenya, line 42: Let's be more expressive in the comment, as it's not modifying the env vars of the container request, as a casual reader could understand, but instead the real implementation. Something like 
```suggestion
// SetEnvVar sets an environment variable for the test script, that will end up as a command-line flag of the main program in the K6 container
```
You probably know better than me :)

modules/k6/examples_test.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/testcontainers/testcontainers-go"
+	"github.com/testcontainers/testcontainers-go/modules/k6"
+	"github.com/testcontainers/testcontainers-go/wait"
+)
+
+func ExampleRunContainer() {
+	// runK6Container {
+	ctx := context.Background()
+
+	gcr := testcontainers.GenericContainerRequest{
```
mdelapenya, line 17: Could you include some comments to the method signature including why we need the httpbin application? I guess it's for https://github.com/testcontainers/testcontainers-go/pull/1721/files#diff-d82d09415608143910841e29e49f3b7e85d98c3d6e7def8a44468a5ffb1732abR5, but want to be more explicit in the comments for this testable example 🙏

mdelapenya (commented):
I think we are super close to be ready. Only concerned about the usage of Mounts Vs Files (see comments), which we should explain in our docs in a better way.

Other than that, once addressed, we are ready to go

Thank you for your time here!

modules/k6/k6.go
pablochacin: Maybe I'm not understanding something about mount vs copy, but the idea is that the cache will be modified by the container and those changes will be available for the next container execution. This is important when running multiple tests, for example in a table-driven test.

modules/k6/k6.go
pablochacin: In reality we don't need access to the cache from outside the container. 

The main reason for mounting is that mounting a local directory makes it easier for the test to create a temporary cache dir and then clean it up when the tests are finished. Please notice that as mentioned in another comment, the cache is intended to be used not only by one test but across multiple tests in a test suite. 

The alternative I considered was using a volume but as far as I can see, TestContainers do not offer any API to clean up this volume once the tests are finished.

modules/k6/examples_test.go
pablochacin: As far as I know, there are no tags for this image, which is widely used for testing.

https://hub.docker.com/r/kennethreitz/httpbin/tags

mdelapenya: Cool than. Let's be specific about using `latest` then 🙏

modules/k6/k6.go
mdelapenya: Ok, now I get the reasons why we need the cache: to speed up "successive" executions, not the current one. Cool.

Have you considered using a kind of a singleton container for this? Where we start the container once and then use it across all the tests in a suite?

modules/k6/k6.go
mdelapenya: I think the script files must be copied. The cache path, this is the one I'd like to understand better (see https://github.com/testcontainers/testcontainers-go/pull/1721#discussion_r1350115875)

modules/k6/examples_test.go
```diff
@@ -0,0 +1,84 @@
...
+	if err != nil {
+		panic(err)
+	}
+	if state.ExitCode != 0 {
+		logs, err := k6.Logs(ctx)
+		if err != nil {
+			panic(err)
+		}
+
+		fmt.Println("k6 test failed:")
```
mdelapenya, line 80: IIRC, testable examples need fmt prints and comments with the output right?

I was more on fmt-printing the exit code instead of the panic.

modules/k6/k6.go
pablochacin: >Have you considered using a kind of a singleton container for this? Where we start the container once and then use it across all the tests in a suite?

yes, but I think that makes the interface more complex because each successive execution may require passing a different test script.  

I think providing a way of having a cache across executions has better usability. 

The only drawback of the current implementation is the need to set the container user. 

That is why lean towards using a volume, but I don't see an easy way to remove it when the test suite ends.

mdelapenya: Have you also taken a look to reusable containers? https://golang.testcontainers.org/features/creating_container/#reusable-container

modules/k6/examples_test.go
pablochacin: I'm not familiar with testable examples and maybe I misunderstood your previous comment.

Could you please indicate what is exactly required in this case? All examples from other modules simply panic on any error.

modules/k6/k6.go
pablochacin: My bad. I though the comment was about the cache. :pray:

modules/k6/examples_test.go
mdelapenya: Please look at https://github.com/testcontainers/testcontainers-go/blob/3079ce0890d15d82d45540ae476aec2b040d64e6/modules/clickhouse/examples_test.go#L46-L62

- L46 prints if the state of the container is Running, matched in L61
- L58 prints if the connection string includes "clickhouse-go" in L62

We are panicking every time we are not performing an assertion. For the assertions, we just print a value and add a comment under `// Output:`. Please see https://go.dev/blog/examples

So for this case, I'd print out the container state with fmt, and in the comment the expected state code. Or if the container log is deterministic, then print out the logs (very similar to what you already did)

modules/k6/k6.go
pablochacin: I think that would require a complete redesign of the module, as now the module executes a command and ends. There's no way to keep it waiting to execute other scripts. 

If passing the cache so problematic, maybe I should simply remove the feature at the expense of a significant penalty in the test execution time.

pablochacin: > We are considering adding volume support for an eventual v1 release, where you could manage volumes with testcontainers APIs. That would simplify the thing.

That would be an interesting addition

> In the mean time, I'd do the following:
>
1. create a named volume using Docker cli, directly.

That makes sense. I will change the implementation for using a volume as cache, add some examples and remove the `AsUser` option which is no longer needed. 

> please change how the test scripts are copied: instead of Mounts, please use Files instead.

This is already done

modules/k6/k6.go
```diff
@@ -20,14 +20,17 @@ type K6Container struct {
 func WithTestScript(scriptPath string) testcontainers.CustomizeRequestOption {
 	return func(req *testcontainers.GenericContainerRequest) {
 		script := filepath.Base(scriptPath)
-		target := fmt.Sprintf("/tests/%s", script)
-		mount := testcontainers.ContainerMount{
-			Source: testcontainers.GenericBindMountSource{
-				HostPath: scriptPath,
+		target := filepath.Join("/home/k6x", script)
```
mdelapenya, line 23: I guess this is the expected path from where k6 will read the test scripts, right?

```diff
- filepath.Join("/home/k6x", script)
+ "/home/k6x/" + script
```

As we know there is a Linux inside that container

modules/k6/k6.go
pablochacin: By the way, the problem is not creating the volume. The volume is created if it doesn't exist. The real problem is to clean it up after the test suite ends.

mdelapenya: If you manually set the right labels in the new volume, then Ryuk will remove it: https://github.com/testcontainers/moby-ryuk/blob/main/main.go#L334

You can get the labels from `testcontainersdocker.DefaultLabels(testcontainerssession.SessionID())` (_Just realised how friendly is that 🤦_)

pablochacin: I think we should add this as a helper function in a follow-up PR (I volunteer for that). Meanwhile, I will simply document that it is the responsibility of the test to cleanup the volume when not longer needed.

modules/k6/k6_test.go
```diff
@@ -0,0 +1,58 @@
...
+		tc := tc
+		t.Run(tc.title, func(t *testing.T) {
+			ctx := context.Background()
+
+			absPath, err := filepath.Abs(filepath.Join("scripts", tc.script))
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			container, err := RunContainer(ctx, WithTestScript(absPath))
```
mdelapenya, line 37: Could you add the cache here? It will demonstrate the benefits of using it, also speeding up the build (I did try locally)
```suggestion
			container, err := RunContainer(ctx, WithTestScript(absPath), WithCache("k6-cache"))
```

modules/k6/examples_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+	absPath, err := filepath.Abs(filepath.Join("scripts", "httpbin.js"))
+	if err != nil {
+		panic(fmt.Errorf("failed to get path to test script: %w", err))
+	}
+
+	// run the httpbin.js test scripts passing the IP address the httpbin container
+	k6, err := k6.RunContainer(
+		ctx,
+		k6.WithTestScript(absPath),
```
mdelapenya, line 54: Let's include the cache here too, for others to see it in action since the first time:
```suggestion
		k6.WithTestScript(absPath),
		k6.WithCache("k6-cache"),
```

pablochacin:
@mdelapenya The reason I've been hesitating in adding the `WithCache` in the examples is that we don't provide a way for cleaning them after the test.

modules/k6/examples_test.go
pablochacin: I don't see the value here, as it runs only one test.

mdelapenya: Indeed, but the docs page serve as reference and many people go there to copy&paste code

modules/k6/k6.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+// WithCache uses the given volume as a cache for building the k6 binary.
+// If the volume does not exists, it is created.
+func WithCache(cache string) testcontainers.CustomizeRequestOption {
+	return func(req *testcontainers.GenericContainerRequest) {
+		mount := testcontainers.ContainerMount{
+			Source: testcontainers.DockerVolumeMountSource{
+				Name: cache,
+			},
```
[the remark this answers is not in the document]
pablochacin: @mdelapenya  I like the idea of somehow providing a way to automatically delete the cache volume, but not sure about introducing these changes in the API before carefully considering how it affects the use cases for the module.

My main problem specifically with the suggestion of suggestion of creating the volume with the name of the test session is that I'm not familiar with the `SessionID`. 

Could you please point me to more information regarding the life-cycle of this session? 

I'm particularly concerned about this comment ` to avoid collisions across multiple runs (maybe in parallel)`. Does this mean that if each test in a table-driven test runs in parallel each one will have its own cache? If so, it would be of no benefit.

modules/k6/examples_test.go
pablochacin: I agree, but there's also the documentation and I don't think all features can be demonstrated in one example without either making it too complex or abusing the options, like in this case the `WithCache` that would be of no benefit at all.

Another option would be to offer more than one example.

modules/k6/k6.go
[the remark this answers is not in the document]
mdelapenya: We introduced sessionID semantics here: https://github.com/testcontainers/testcontainers-go/pull/1513

Basically:

By test session/execution we mean:

- a single "go test" invocation (including flags)
- a single "go test ./..." invocation (including flags)
- the execution of a single test or a set of tests using the IDE

As a consequence, with the sole goal of aggregating test execution across multiple packages, the new code will use the parent process ID (pid) of the current process and use it to generate a unique session ID. We are using the parent pid because the current process will be a child process of:

- the process that is running the tests, e.g.: "go test";
- the process that is running the application in development mode, e.g. "go run main.go -tags dev";
- the process that is running the tests in the IDE, e.g.: "go test ./...".

At the same time, one Ryuk instance will control that test session, killing anything that is related to it: volumes, images, containers and networks.

To your question: multiple parallel() calls will share one single parent process ID, therefore they'll share the same session ID.

modules/k6/examples_test.go
mdelapenya: Agreed. Whatever you prefer is fine then 👍

modules/k6/k6.go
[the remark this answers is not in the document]
pablochacin: Thanks for the detailed explanation.

My initial thought is that, as far as I can see  (it is quite a large PR and maybe I'm missing something) these changes are not user-facing. Therefore, making the semantics of the `WithCache` depend on a concept that the user may not be aware of, requires more documentation and can introduce some unnecessary cognitive overload.

For example
```
WithCache creates an ephemeral volume to be used as a build cache for all the tests in a test session.
This volume will be automatically deleted at the end of the session.
```

As a user, I will ask, what a session is?

Also, I think there are two clearly different scenarios for using this feature:
1. When running tests locally. I probably don't want to clean this cache ever. 
2 When running in a CI/CD. I very likely will want to clean it after each execution (a session as you describe it) to prevent volumes from leaking.

I don't have a strong opinion about which option is better: making the auto cleanup the default behavior with an option to disable it, or the other way around. But that would affect the API. 

In any case, I find changing the semantics of the API on the fly on this PR to be a little bit forced. 

I would prefer to either drop the feature until we come up with a better design or introduce it as it is (with no auto-cleanup) and implement the auto-cleanup in a follow-up PR.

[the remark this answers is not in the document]
pablochacin: > When running tests locally. I probably don't want to clean this cache ever.
Just to clarify, this cache should not affect the test themselves as it is only used for speeding up the build process, in the same way that go has a cache for the modules. So keeping the cache across tests should not be considered a form of side-effect, like a database may be.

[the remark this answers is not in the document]
mdelapenya: Remember that with https://github.com/testcontainers/testcontainers-go/pull/1721#discussion_r1350350051 we will have clean-up for free.

About the cognitive load, I'm taking a note to add the test session semantics to our docs. Good catch!

To this point, and given the module is 100%, I'm fine with anything you consider moving to a follow-up, or to keep it at this stage. You did a good job with the code and we have a fully working module for k6! Let's improve it in follow-ups if/when needed.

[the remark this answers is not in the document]
pablochacin: > Remember that with https://github.com/testcontainers/testcontainers-go/pull/1721#discussion_r1350350051 we will have clean-up for free.

I agree. As I mentioned before, I think that is the way to go! 

However, there are a couple of details about the API that I think need more careful consideration.

I prefer to merge this PR and open a follow-up to discuss how to handle the two scenarios described above.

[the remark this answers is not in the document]
mdelapenya: Ok, I'm resolving the comment and approving the PR, let's work on the follow-up comments then 🚀

PR #1725, opened 2023-10-08
ignore patterns defined in dockerignore

### LINKED ISSUE #1486: [Enhancement]: Respect `.dockerignore` in `FromDockerfile` builds
### Proposal

as much fun as it is to have a 6gb image on a developer machine ;) 

Currently it can be achieved with something like:

```go
import (
	"archive/tar"
	"io"
	"io/fs"
	"os"
	"path/filepath"
	"strings"

	"github.com/sabhiram/go-gitignore"
)

func createDockerContext() (*os.File, error) {
	ignoreFile, err := ignore.CompileIgnoreFile(".dockerignore")
	if err != nil {
		return nil, err
	}

	// Store the tar file in a temp file because it can get a bit large for in-memory
	tmpF, err := os.CreateTemp(os.TempDir(), "mmr-docker-context")
	tarContext := tar.NewWriter(tmpF)

	err = filepath.WalkDir(".", func(path string, d fs.DirEntry, err error) error {
		if err != nil {
			return err
		}
		if d.IsDir() {
			return nil // skip directories
		}
		info, err := d.Info()
		if err != nil {
			return err
		}

		if match := ignoreFile.MatchesPath(path); match {
			return nil // skip ignored files
		}
		//fmt.Println("[Image Build] Including file: ", path)
		err = tarContext.WriteHeader(&tar.Header{
			Name:    strings.ReplaceAll(path, "\\", "/"), // fix windows pathing
			Mode:    int64(info.Mode()),
			ModTime: info.ModTime(),
			Size:    info.Size(),
		})
		if err != nil {
			return err
		}
		f2, err := os.Open(path)
		if err != nil {
			return err
		}
		defer f2.Close()
		_, err = io.Copy(tarContext, f2)
		if err != nil {
			return err
		}
		return nil
	})
	if err != nil {
		return nil, err
	}
	err = tarContext.Close()
	if err != nil {
		return nil, err
	}

	err = tmpF.Close()
	if err != nil {
		return nil, err
	}

	// Reopen the temp file for a useful read (otherwise it ends up closed for some reason)
	return os.Open(tmpF.Name())
}
```

Luukvdm:
## What does this PR do?

Respect `.dockerignore` files similar to how the Docker CLI [handles them](https://github.com/docker/cli/blob/master/cli/command/image/build/dockerignore.go#L14).

## Why is it important?

Prevents unnecessary large docker images and mirrors behavior of the Docker CLI.

## Related issues

- Closes #1486


container.go
```diff
@@ -188,7 +190,17 @@ func (c *ContainerRequest) GetContext() (io.Reader, error) {
 	}
 	c.Context = abs
 
-	buildContext, err := archive.TarWithOptions(c.Context, &archive.TarOptions{})
+	// based on https://github.com/docker/cli/blob/master/cli/command/image/build/dockerignore.go#L14
+	var excluded []string
+	if f, err := os.Open(filepath.Join(abs, ".dockerignore")); err != nil {
+		var err error
+		excluded, err = ignorefile.ReadAll(f)
```
mdelapenya, line 197: To avoid a new dependency, I'd also copy the function in here: https://github.com/moby/patternmatcher/blob/main/ignorefile/ignorefile.go

Probably into the `internal/testcontainersdocker` package.

What are your thoughts on this?

mdelapenya: Exactly, that's what I think too: it's basically parsing the dockerignore file. Only if you agree, let's copy that, but as you said, it's part of the indirect deps already. Whatever is your preference I'm fine with it.

Luukvdm: I think I prefer leaving it as is. Because the dependency is small and its easier to track upstream changes this way.

container.go
```diff
@@ -188,7 +190,17 @@ func (c *ContainerRequest) GetContext() (io.Reader, error) {
...
+		if err != nil {
+			return nil, fmt.Errorf("error reading .dockerignore: %w", err)
```
mdelapenya, line 199: Do we want to raise an error here and fail our logic, or simply log the error and continue?

Luukvdm: At this point in the code the dockerignore file is read, and any errors are coming from the parsing of the file. I feel like if this ever fails its better to return the error and let the user fix the problem.

mdelapenya:
@Luukvdm there is a few tests failing. Please see https://github.com/testcontainers/testcontainers-go/actions/runs/6448106348/job/17580097884?pr=1725

They are all related to the `.dockerignore` new feature we added in this PR, very likely because the file does not exist for the tests. I think we are missing a check for the file existence 🤔 

Thankfully we have tests! 💪

Luukvdm:
> @Luukvdm there is a few tests failing. Please see https://github.com/testcontainers/testcontainers-go/actions/runs/6448106348/job/17580097884?pr=1725
> 
> They are all related to the `.dockerignore` new feature we added in this PR, very likely because the file does not exist for the tests. I think we are missing a check for the file existence 🤔
> 
> Thankfully we have tests! 💪

Thank god for automated testing 😄
I made a silly mistake in the error handling. I fixed it and added a regression test just in case. Do you agree with the changes? 
Also wasn't sure about code styling/ order and added the function under the `GetContext` function.

mdelapenya (approved):
LGTM, thanks for your contribution!

PR #1726, opened 2023-10-08
feat: Add Cassandra module

anilsenay:
## What does this PR do?

This PR adds Cassandra module for go-testcontainers

## Why is it important?

It is important as part of a task of https://github.com/testcontainers/testcontainers-go/issues/636 to support Cassandra modules too.

## Related issues

- Relates to https://github.com/testcontainers/testcontainers-go/issues/636

## Follow-ups

Cassandra docker image does not provide a configuration from environment variables for authentication. It allows all connections by default without auth. To enable authentication and set a user, you should provide it in config yaml file or it may be ran by `cqlsh` command in container to execute query as an alternative. As a follow-up improvement, authentication options could be added somehow.


docs/modules/postgres.md
```diff
@@ -51,7 +51,7 @@ If you need to set a different database, and its credentials, you can use the `W
 
 #### Init Scripts
 
-If you would like to do additional initialization in the Postgres container, add one or more `*.sql`, `*.sql.gz`, or `*.sh` scripts to the container request with the `WithInitScript` function.
+If you would like to do additional initialization in the Postgres container, add one or more `*.sql`, `*.sql.gz`, or `*.sh` scripts to the container request with the `WithInitScripts` function.
```
anilsenay, line 54: When I saw typo in postgres readme, I wanted to fix this typo in this pr also.

modules/cassandra/cassandra.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+// WithInitScripts sets the init cassandra queries to be run when the container starts
+func WithInitScripts(scripts ...string) testcontainers.CustomizeRequestOption {
+	return func(req *testcontainers.GenericContainerRequest) {
+		var initScripts []testcontainers.ContainerFile
+		for _, script := range scripts {
+			cf := testcontainers.ContainerFile{
+				HostFilePath:      script,
+				ContainerFilePath: "/" + filepath.Base(script),
```
anilsenay, line 60: cassandra docker image does not support `/docker-entrypoint-initdb.d/` so I put script files in root directory and in line 105 they will be executed via `cqlsh` or `sh` command by file extension. Is this a valid solution? What do you think?

docs/modules/postgres.md
mmorel-35: Please do it in another PR as it is outside the current scope.

modules/cassandra/cassandra.go
```diff
@@ -0,0 +1,116 @@
+package cassandra
+
+import (
+	"context"
+	"io"
+	"path/filepath"
+	"strings"
+
+	"github.com/docker/go-connections/nat"
```
mmorel-35, line 9: Gci shall require a new line here. 
Which IDE are you using ?

modules/cassandra/examples_test.go
```diff
@@ -0,0 +1,64 @@
+package cassandra_test
+
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+
+	"github.com/gocql/gocql"
```
mmorel-35, line 8: There shall be a new line here too

mmorel-35 (changes requested):

modules/cassandra/cassandra.go
anilsenay: Im using vscode but also i checked in GoLand, it did not do anything or warn about it. But I added new line manually. Thnak you

mmorel-35: @mdelapenya , shall an issue be created ? This is the second module where the linting is not applied.
@anilsenay , Is golangci-lint installed on your machine ?
In vscode, are you using the workspace configuration from [here](https://github.com/testcontainers/testcontainers-go/blob/main/.vscode/.testcontainers-go.code-workspace) ?
In your settings, do you have this: 
```
"go.lintTool": "golangci-lint"
```

mdelapenya: I wonder if we could run golangci-lint when a module is generated, but instead of using the binary (see Makefile) calling the right Docker image to avoid having to install it on every developer's machine.

mmorel-35: Not every machine has docker, on windows for exemple, it's not installed by default.

modules/cassandra/cassandra.go
```diff
@@ -0,0 +1,117 @@
...
+// RunContainer creates an instance of the Cassandra container type
+func RunContainer(ctx context.Context, opts ...testcontainers.ContainerCustomizer) (*CassandraContainer, error) {
+	req := testcontainers.ContainerRequest{
+		Image:        "cassandra:4.1.3",
+		ExposedPorts: []string{string(port)},
+		Env: map[string]string{
+			"CASSANDRA_SNITCH":          "GossipingPropertyFileSnitch",
+			"JVM_OPTS":                  "-Dcassandra.skip_wait_for_gossip_to_settle=0 -Dcassandra.initial_token=0",
+			"HEAP_NEWSIZE":              "128M",
+			"MAX_HEAP_SIZE":             "1024M",
```
mdelapenya, line 79: Do you think these two should be configurable with options?

modules/cassandra/cassandra.go
```diff
@@ -0,0 +1,117 @@
...
+			"CASSANDRA_ENDPOINT_SNITCH": "GossipingPropertyFileSnitch",
+			"CASSANDRA_DC":              "datacenter1",
+		},
+		WaitingFor: wait.ForAll(
+			wait.ForListeningPort(port),
+			wait.ForExec([]string{"cqlsh", "-e", "SELECT release_version FROM system.local"}).WithResponseMatcher(func(body io.Reader) bool {
+				data, _ := io.ReadAll(body)
+				return strings.Contains(string(data), "4.1.3")
```
mdelapenya, line 87: I think this wait condition is tricky, as it will be broken any time users change the image using `WithImage`. Do you think we can simplify it?

modules/cassandra/cassandra.go
```diff
@@ -0,0 +1,117 @@
...
+
+	if len(genericContainerReq.Files) > 0 {
+		for _, file := range genericContainerReq.Files {
+			if strings.HasSuffix(file.ContainerFilePath, ".cql") {
+				container.Exec(ctx, []string{"cqlsh", "-f", file.ContainerFilePath})
+			} else if strings.HasSuffix(file.ContainerFilePath, ".sh") {
+				container.Exec(ctx, []string{"/bin/sh", file.ContainerFilePath})
+			}
+		}
+	}
```
mdelapenya, line 114: Probably removed after my comments above.

modules/cassandra/cassandra_test.go
```diff
@@ -0,0 +1,147 @@
...
+	var test Test
+	err = session.Query("SELECT id, name FROM test_keyspace.test_table WHERE id=1").Scan(&test.Id, &test.Name)
+	assert.NoError(t, err)
+	assert.Equal(t, Test{Id: 1, Name: "NAME"}, test)
+}
+
+func TestCassandraWithConfigFile(t *testing.T) {
+	ctx := context.Background()
+
+	container, err := RunContainer(ctx, WithConfigFile("testdata/config.yaml"))
```
mdelapenya, line 59: Let's use filepath to build paths and be friendly with Windows users

modules/cassandra/cassandra_test.go
```diff
@@ -0,0 +1,147 @@
...
+	assert.NoError(t, err)
+	assert.Equal(t, "My Cluster", result)
+}
+
+func TestCassandraWithInitScripts(t *testing.T) {
+	t.Run("with init cql script", func(t *testing.T) {
+		ctx := context.Background()
+
+		// withInitScripts {
+		container, err := RunContainer(ctx, WithInitScripts("testdata/init.cql"))
```
mdelapenya, line 90: Same here about using filepath 🙏

mdelapenya (commented):
Great job with the new module! Thanks for contributing it!

I left a few comments, the most important for me is the one leveraging the container lifecycle hooks using startup commands.

Other than than, feels look good to me. Once we address them, I think it will be ready to be merged

Thanks!

modules/cassandra/cassandra.go
mdelapenya: Yeah, but installing docker is easier than installing a binary for linting. In any case, I'm biased from my past experience in a release engineering team, where we dockerised everything to run smoothly on the CI 😅

modules/cassandra/cassandra.go
```diff
@@ -0,0 +1,117 @@
...
+		var initScripts []testcontainers.ContainerFile
+		for _, script := range scripts {
+			cf := testcontainers.ContainerFile{
+				HostFilePath:      script,
+				ContainerFilePath: "/" + filepath.Base(script),
+				FileMode:          0o755,
+			}
+			initScripts = append(initScripts, cf)
+		}
+		req.Files = append(req.Files, initScripts...)
```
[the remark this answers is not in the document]
anilsenay: Oh got it. So should I add them to `ContainerRequest` in `RunContainer` function like `LifecycleHooks: []testcontainers.ContainerLifecycleHooks{ ... }` or should i leave it to user to add them one by one with `testcontainers.WithStartupCommand`? Because as a user I expect from`WithInitScripts` function to run scripts its own.

Another thing, why the file name is `types_test.go` in RabbitMQ module? I thought it was a test file 😅

modules/cassandra/cassandra.go
anilsenay: I don't know actually, I took this default configs in cassandra module in testcontainers-java. I thought they add it for a purpose. Because while I was testing it, container cration time increased without these configs.

modules/cassandra/cassandra.go
anilsenay: You so right, i totally missed this. I will refactor it.

modules/cassandra/cassandra.go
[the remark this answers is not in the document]
mdelapenya: > Why the file name is types_test.go in RabbitMQ module? I thought it was a test file 😅

Because I intentionally did not want to expose those types, but use them as reference for others to see how it works. Doing so allowed us not to couple with RabbitMQ implementation. Probably poor-old-man way to achieve it though! 😅 

I'd just keep the `WithInitScripts` function as you have, adding the `WithStartupCommand` internally to that method: with each new file, add its command. Does it make sense to you?

modules/cassandra/cassandra.go
mdelapenya: Cool then, let's keep it as is. I asked because I could imagine environments with more or less RAM where this setup could be different.

modules/cassandra/cassandra.go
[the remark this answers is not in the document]
anilsenay: It makes sense when knowing it, but I thought as a user, I could think that `WithInitScripts` will do all the trick and I don't need anything more, like in other modules. 

But I will stick what you said. So after I refactor it, I think I should add information about it to Readme 👍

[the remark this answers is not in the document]
mdelapenya: yes, that's what we'll achieve: the user will only know about `WithInitScripts`. We'll internally call the `WithStartupCommand` for every file.

Please check this code snippet locally before commiting:

```golang

type initScript struct {
	File     string
}

func (i initScript) AsCommand() []string {
	if strings.HasSuffix(i.File, ".cql") {
		container.Exec(ctx, []string{"cqlsh", "-f", i.File})
	} else if strings.HasSuffix(i.File, ".sh") {
		container.Exec(ctx, []string{"/bin/sh", i.File})
	}

	// what do we return here if the file extension is incorrect? an empty array, an incorrect file to raise an error later on?
}

func WithInitScripts(scripts ...string) testcontainers.CustomizeRequestOption {
		var initScripts []testcontainers.ContainerFile
		for _, script := range scripts {
			cf := testcontainers.ContainerFile{
				HostFilePath:      script,
				ContainerFilePath: "/" + filepath.Base(script),
				FileMode:          0o755,
			}
			initScripts = append(initScripts, cf)

			testcontainers.WithStartupCommand(initScript{File: "/" + filepath.Base(script)})
		}
```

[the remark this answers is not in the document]
anilsenay: Ah alright. This makes more sense to me. I initially thought like this, but when I saw it in RabbitMQ tests, I thought the user would add `WithStartupCommand`.
It's more clear now. Thank you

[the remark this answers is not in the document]
mdelapenya: Yeah, sure I can understand that :D textual communications could be hard! Hopefully we got to it very fast. Thanks for your patience!

modules/cassandra/executable.go
```diff
@@ -0,0 +1,18 @@
...
+}
+
+func (i initScript) AsCommand() []string {
+	if strings.HasSuffix(i.File, ".cql") {
+		return []string{"cqlsh", "-f", i.File}
+	} else if strings.HasSuffix(i.File, ".sh") {
+		return []string{"/bin/sh", i.File}
+	}
+	return nil
+}
```
anilsenay, line 18: @mdelapenya hey, can you review this again? I add this to the separated file as "executable" but I need help with the naming convention because there is no example except `rabbitmq/types_test.go`. I think we should decide on good naming for both file names and structs for this executable usage for the sake of example to other modules

anilsenay, line 18: @mdelapenya hey, can you review this again? I add this to the separated file as "executable" but I need help with the naming convention because there is no example except rabbitmq/types_test.go. I think we should decide on good naming for both file names and structs for this executable usage for the sake of example to other modules

mdelapenya: I think `initScript`, not exported, is fine. I thought about the following names:
- `cassandraCommand`: no because it includes the package namespace, and no again because calling the .AsCommand() method will come with duplication.
- `script`: no because it's very simple
- `executable`: no because it's very abstract
- `initScript`:  in the context of Cassandra, I think this is the one that makes more sense.

mdelapenya (approved):
Let's rename the executable.go file, and once there. we are good to go.

In any case, LGTM!

modules/cassandra/executable.go
```diff
@@ -0,0 +1,18 @@
+package cassandra
```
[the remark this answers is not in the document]
mdelapenya: Mmmm I see your point. Probably `command` is a better name, although the struct in it is not names command. Or it's probably a good name...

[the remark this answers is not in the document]
mdelapenya: Let's keep executable then. We can remember it later if needed.

mdelapenya (approved):
LGTM! I took the license to fix a typo in a comment to not load you again with more rework.

Thanks again for your had work here! Much appreciated! 💪

mdelapenya:
Merging, as last commit is a change in a comment.

PR #1743, opened 2023-10-11
DOCKER_HOST var typo

udf2457:
Missed a forward-slash: unix:/ -> unix://


mdelapenya (approved):
Good catch! Thanks for this contribution :)

LGTM

PR #1744, opened 2023-10-11
docs: add default options to k6 module

mdelapenya:
## What does this PR do?
This PR adds the core options to the k6 module, updating the markdown file for it.

At the same time, it's splitting the usage example in different parts: create the httpbin app using a generic container, getting its IP, the JS script to be used, and the k6 container.

## Why is it important?
We forgot adding them in previous PRs.

## Related issues

- Completes #1721

### DIFF
--- a/docs/modules/k6.md
+++ b/docs/modules/k6.md
@@ -80,3 +80,7 @@ Use the `WithTestScript` option to specify the test script to run. The path to t
 ```golang
 k6.RunContainer(ctx, k6.WithTestScript("/tests/test.js"))
 ```
+
+### Container Methods
+
+The K6 container does not expose any method.
--- a/docs/modules/k6.md
+++ b/docs/modules/k6.md
@@ -45,7 +45,7 @@ When starting the K6 container, you can pass options in a variadic way to config
 #### Image
 
 !!! warning
-    The K6 module uses a custom Docker image to build the `k6` binary with all the required extensions. This image is based on the official `k6` image and it is available on [Docker Hub](https://hub.docker.com/r/szkiba/k6x). Therefore, only the `szkiba/k6x` image should be used with this module.
+    The K6 module uses a `k6x` image to build a `k6` binary with all the required extensions. Therefore, only the [szkiba/k6x](https://hub.docker.com/r/szkiba/k6x) image should be used with this module.
 
 If you need to set a different K6 Docker image, you can use `testcontainers.WithImage` with a valid Docker image
 for k6x. E.g. `testcontainers.WithImage("szkiba/k6x:v0.3.1")`.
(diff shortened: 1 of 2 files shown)


mdelapenya:
@pablochacin could you please take a look? 🙏

docs/modules/k6.md
```diff
@@ -40,6 +42,13 @@ func RunContainer(ctx context.Context, opts ...testcontainers.ContainerCustomize
 
 When starting the K6 container, you can pass options in a variadic way to configure it.
 
+#### Image
+
+If you need to set a different K6 Docker image, you can use `testcontainers.WithImage` with a valid Docker image
+for Cassandra. E.g. `testcontainers.WithImage("szkiba/k6x:v0.3.1")`.
```
pablochacin, line 48:
```suggestion
for k6x. E.g. `testcontainers.WithImage("szkiba/k6x:v0.3.1")`.
```

mdelapenya: Could it be like this instead?
```suggestion
for K6. E.g. `testcontainers.WithImage("szkiba/k6x:v0.3.1")`.
```

Ot making it explicit for k6x is better?

mdelapenya: Probably worth it to add it as an info/warning in the `Image` section?

pablochacin: I would prefer to keep it simple and assume only `k6x` images are used. Introducing other variants will just make the usage of the module more complex.

mdelapenya: Ok, let's include a message in the docs regarding the usage of k6x.

I'm on PTO until Monday, could you elaborate a text that we can add to the docs explaining that? 🙏

docs/modules/k6.md
```diff
@@ -40,6 +42,16 @@ func RunContainer(ctx context.Context, opts ...testcontainers.ContainerCustomize
 
 When starting the K6 container, you can pass options in a variadic way to configure it.
 
+#### Image
+
+!!! warning
+    The K6 module uses a custom Docker image to build the `k6` binary with all the required extensions. This image is based on the official `k6` image and it is available on [Docker Hub](https://hub.docker.com/r/szkiba/k6x). Therefore, only the `szkiba/k6x` image should be used with this module.
```
pablochacin, line 48: I find this warning repetitive as this is explained in the description of the module. Also, there are some inaccuracies in the description of k6x and k6 in the context of this module. I would be more concise: 

```suggestion
    The K6 module uses a `k6x` image to build a `k6` binary with all the required extensions. Therefore, only the [szkiba/k6x](https://hub.docker.com/r/szkiba/k6x) image should be used with this module.
```

pablochacin (changes requested):

docs/modules/k6.md
mdelapenya: Do we want to remove the entire warning. I'm fine doing it if you consider it redundant

pablochacin: I think a short warning here is convenient.

pablochacin (approved):

PR #1775, opened 2023-10-13
fix: Add HTTPStrategy WithForcedIPv4LocalHost To Fix Docker Port Map

tateexon:
## What does this PR do?

Associated docker port mapping bugs:
https://github.com/moby/moby/issues/42442
https://github.com/moby/moby/issues/42375
If ipv6 is enabled in docker then these bugs affect this library. This is even if we build the docker network with ipv6 disabled since the ipv6 ports are still forwarded. This creates the potential for localhost for a container to be mapped to two different ports between ipv4 and ipv6. This is fine if you only have one container but once you have multiple containers spun up these ports can overlap where one containers ipv4 port is the same as another containers ipv6 port, at which point if you use localhost you are not guaranteed the ipv4 address and thus can end up calling into the wrong container.

## Why is it important?

When waiting for a container to be ready it is important that we are waiting for the container we expect when ipv6 is enabled in docker.

## Related issues

- https://github.com/moby/moby/issues/42442
- https://github.com/moby/moby/issues/42375

Note: It may make sense to do this for the Container.Endpoint and Container.Host methods in some way as well. These are nasty docker bugs and are very hard to triage since they only happen when docker happens to forward the ports out of sync.


mdelapenya (commented):
Hi @tateexon thanks for submitting this PR, and sorry for the late response, I had a few conference trips since October and did not pay attention to this one 😞

I added a comment regarding the testable example, and I'd like to change the bug into a feature: we are adding new features to the wait strategy to resolve a bug elsewhere. So for me this is providing support for a non-covered use case. But I'm open to listen to your thoughts on this.

Nevertheless, the changes LGTM! I'm waiting for the CI to pass, also the changes in my comments.

Thanks for your time here!

mdelapenya:
@tateexon  I went ahead and merged the branch with the current state of main, to not cause you more trouble with it. Please let me know if the added changes (ultra minor changes, not feature related) are fine with your initial implementation.

Cheers!

PR #1278, opened 2023-06-08
fix: don't panic when logs waits for more than 5 seconds (v2)

martin-sucha:
This should fix https://github.com/testcontainers/testcontainers-go/issues/946

## What does this PR do?

There are two commits.

The first commit fixes synchronization issues in producer tests.

The second commit removes panic when logs endpoint takes more than 5 seconds to respond. The panic happened at least with podman when no new logs appear when using follow and since parameters.

We keep retrying until the context is canceled (the retry request would fail anyway) or the producer is stopped, whichever comes first. This makes the retry behavior consistent. The HTTP calls to fetch logs are interrupted immediately (previously the producer waited for in-progress calls to complete).

## Why is it important?

Panicking the whole test suite breaks all my tests if the container ends early or there is some temporary error fetching the logs.

## Related issues

- Closes #946 
- Closes #606 
- Supersedes #947

PR #1701, opened 2023-09-29
Log producer start stop sync

gflarity:
## What does this PR do?

As part of trying to fix the issues discuss in #1700, I tried this first. Turns out  the lack of sychronization in StartLogProducer and StopLogProducer was actually red herring, however it seems like is still a useful addition so I thought I'd send out the PR for review. 

## Why is it important?

Makes (Start/Stop)LogProducer  more deterministic and perhaps will prevent future race condition bugs.

PR #1777, opened 2023-10-16
fix: don't panic when logs waits for more than 5 seconds

### LINKED ISSUE #606: [Bug]: can not get logs from container which is dead or marked for removal
### Testcontainers version

0.15.0

Yes

### Host OS

Linux

### Host Arch

amd64

### Go Version

1.19

### Docker version

```shell
 Version:           20.10.21
 API version:       1.41
 Go version:        go1.18.7
 Git commit:        baeda1f
 Built:             Tue Oct 25 18:04:24 2022
 Context:           default
 Experimental:      true

 Engine:
  Version:          20.10.21
  Go version:       go1.18.7
  Git commit:       3056208
  Built:            Tue Oct 25 18:02:38 2022
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.9
  GitCommit:        1c90a442489720eec95342e1789ee8a5e1b9536f
 runc:
  Version:          1.1.4
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### Docker info

```shell
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Docker Buildx (Docker Inc., v0.9.1-docker)
  scan: Docker Scan (Docker Inc., v0.21.0)

Server:
 Containers: 1
  Running: 1
  Paused: 0
  Stopped: 0
 Images: 5
 Server Version: 20.10.21
 Storage Driver: overlay2
  Backing Filesystem: xfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 1c90a442489720eec95342e1789ee8a5e1b9536f
 init version: de40ad0
 Security Options:
  seccomp
   Profile: default
 Kernel Version: 3.10.0-1160.71.1.el7.x86_64
 Operating System: CentOS Linux 7 (Core)
 OSType: linux
 Architecture: x86_64
 CPUs: 4
 Total Memory: 15.51GiB
 Name: dev-bgr-test1-lnx7.iv.local
 ID: JCOB:B6QD:MFHY:4B4Y:Z7OA:H7LA:Z5NM:TDOA:7DD7:AMMB:ZBSJ:4KIU
 Debug Mode: true
  File Descriptors: 34
  Goroutines: 41
  System Time: 2022-11-02T17:54:20.539106249+01:00
  EventsListeners: 0
```

### What happened?

I'm starting a container and waiting for the end of its execution (it is the expected behavior). I also have a log consumer to display what is going on.

```
	c, err := testcontainers.GenericContainer(context.Background(), testcontainers.GenericContainerRequest{
			Image:    image,
			Networks: []string{network},
		},
	})
	if err != nil {
		return err
	}
	defer c.Terminate(context.Background())

	c.FollowOutput(&stdoutLogConsumer{})

	err = c.Start(context.Background())
	if err != nil {
		return err
	}

	err = c.StartLogProducer(context.Background())
	if err != nil {
		return err
	}

	err = wait.ForExit().WithExitTimeout(180*time.Second).WaitUntilReady(context.Background(), c)
	if err != nil {
		return err
	}

	err = c.StopLogProducer()
	if err != nil {
		return err
	}
```

Sometimes the test fails with this error:

```
     | panic: Error response from daemon: can not get logs from container which is dead or marked for removal
     | goroutine 51 [running]:
     | github.com/testcontainers/testcontainers-go.(*DockerContainer).StartLogProducer.func1()
     | 	/go/pkg/mod/github.com/testcontainers/testcontainers-go@v0.15.0/docker.go:576 +0x5f6
     | created by github.com/testcontainers/testcontainers-go.(*DockerContainer).StartLogProducer
     | 	/go/pkg/mod/github.com/testcontainers/testcontainers-go@v0.15.0/docker.go:558 +0x8a
```

I think there is a problem with the way the log producer is handling a container that terminates: 

```
		r, err := c.provider.client.ContainerLogs(ctx, c.GetContainerID(), options)
		if err != nil {
			// if we can't get the logs, panic, we can't return an error to anything
			// from within this goroutine
			panic(err)
		}
```

I suggest to log a warning and stop log producer instead of calling panic.

### Relevant log output

_No response_

### Additional Information

_No response_

### LINKED ISSUE #946: [Bug]: log producer panics if no logs are produced when using podman
### Testcontainers version

v0.19.0

Yes

### Host OS

Linux

### Host arch

x86_64

### Go version

go1.20

### Docker version

```shell
podman version
Client:       Podman Engine
Version:      4.4.2
API Version:  4.4.2
Go Version:   go1.19.6
Built:        Wed Mar  1 12:22:59 2023
OS/Arch:      linux/amd64
```
```


### Docker info

```shell
podman info:

host:
  arch: amd64
  buildahVersion: 1.29.0
  cgroupControllers:
  - cpu
  - io
  - memory
  - pids
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: conmon-2.1.7-2.fc37.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.1.7, commit: '
  cpuUtilization:
    idlePercent: 94.41
    systemPercent: 0.97
    userPercent: 4.62
  cpus: 12
  distribution:
    distribution: fedora
    variant: workstation
    version: "37"
  eventLogger: journald
  hostname: fedora
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 6.1.18-200.fc37.x86_64
  linkmode: dynamic
  logDriver: journald
  memFree: 1141927936
  memTotal: 33228570624
  networkBackend: netavark
  ociRuntime:
    name: crun
    package: crun-1.8.1-1.fc37.x86_64
    path: /usr/bin/crun
    version: |-
      crun version 1.8.1
      commit: f8a096be060b22ccd3d5f3ebe44108517fbf6c30
      rundir: /run/user/1000/crun
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +CRIU +LIBKRUN +WASM:wasmedge +YAJL
  os: linux
  remoteSocket:
    exists: true
    path: /run/user/1000/podman/podman.sock
  security:
    apparmorEnabled: false
    capabilities: CAP_CHOWN,CAP_DAC_OVERRIDE,CAP_FOWNER,CAP_FSETID,CAP_KILL,CAP_NET_BIND_SERVICE,CAP_SETFCAP,CAP_SETGID,CAP_SETPCAP,CAP_SETUID
    rootless: true
    seccompEnabled: true
    seccompProfilePath: /usr/share/containers/seccomp.json
    selinuxEnabled: true
  serviceIsRemote: false
  slirp4netns:
    executable: /usr/bin/slirp4netns
    package: slirp4netns-1.2.0-8.fc37.x86_64
    version: |-
      slirp4netns version 1.2.0
      commit: 656041d45cfca7a4176f6b7eed9e4fe6c11e8383
      libslirp: 4.7.0
      SLIRP_CONFIG_VERSION_MAX: 4
      libseccomp: 2.5.3
  swapFree: 8272998400
  swapTotal: 8589930496
  uptime: 5h 36m 47.00s (Approximately 0.21 days)
plugins:
  authorization: null
  log:
  - k8s-file
  - none
  - passthrough
  - journald
  network:
  - bridge
  - macvlan
  volume:
  - local
registries:
  search:
  - registry.fedoraproject.org
  - registry.access.redhat.com
  - docker.io
  - quay.io
store:
  configFile: /home/martin/.config/containers/storage.conf
  containerStore:
    number: 4
    paused: 0
    running: 4
    stopped: 0
  graphDriverName: overlay
  graphOptions: {}
  graphRoot: /home/martin/.local/share/containers/storage
  graphRootAllocated: 510389125120
  graphRootUsed: 55164096512
  graphStatus:
    Backing Filesystem: btrfs
    Native Overlay Diff: "true"
    Supports d_type: "true"
    Using metacopy: "false"
  imageCopyTmpDir: /var/tmp
  imageStore:
    number: 182
  runRoot: /run/user/1000/containers
  transientStore: false
  volumePath: /home/martin/.local/share/containers/storage/volumes
version:
  APIVersion: 4.4.2
  Built: 1677669779
  BuiltTime: Wed Mar  1 12:22:59 2023
  GitCommit: ""
  GoVersion: go1.19.6
  Os: linux
  OsArch: linux/amd64
  Version: 4.4.2
```

### What happened?

The worker goroutine in DockerContainer.StartLogProducer panics:

```
panic: Get "http://%2Frun%2Fuser%2F1000%2Fpodman%2Fpodman.sock/v1.41/containers/17131d3a1a9090c17437423165cbaf4324140e875b5f69c0ff5c4ee0a78711c5/logs?follow=1&since=1679409204.942642010&stderr=1&stdout=1&tail=": context deadline exceeded

goroutine 66 [running]:
github.com/testcontainers/testcontainers-go.(*DockerContainer).StartLogProducer.func1()
        /home/martin/go/pkg/mod/github.com/testcontainers/testcontainers-go@v0.19.0/docker.go:604 +0x617
created by github.com/testcontainers/testcontainers-go.(*DockerContainer).StartLogProducer
        /home/martin/go/pkg/mod/github.com/testcontainers/testcontainers-go@v0.19.0/docker.go:586 +0x8a

```

This is because:
* there is a harcoded 5 second timeout
* `since` parameter is current time
* the container does not produce log output after the since parameter, there are logs only before that time
* `podman` does not return HTTP response headers until some log output is available

So the context deadline expires and the code panics.

It seems that instead of a hardcoded 5 second timeout, the context should be cancelled when the producer is stopped.

Unless this is a bug in podman, in that case we can open a bug report there.

### Relevant log output

_No response_

### Additional information

_No response_

mdelapenya:
## Disclaimer

Credits to @martin-sucha, as the original author of this PR.  🙌 

I resolved the conflicts manually, reverting the changes on the `logconsumer_test.go` file. I sent a different PR because I could not push to the upstream of #1278. But all the commits have been preserved to keep history and authorship untouched.

## What does this PR do?

This should fix https://github.com/testcontainers/testcontainers-go/issues/946

There are two commits.

The first commit fixes synchronization issues in producer tests.

The second commit removes panic when logs endpoint takes more than 5 seconds to respond. The panic happened at least with podman when no new logs appear when using follow and since parameters.

We keep retrying until the context is canceled (the retry request would fail anyway) or the producer is stopped, whichever comes first. This makes the retry behavior consistent. The HTTP calls to fetch logs are interrupted immediately (previously the producer waited for in-progress calls to complete).

## Why is it important?

Panicking the whole test suite breaks all my tests if the container ends early or there is some temporary error fetching the logs.

## Related issues

- Closes #946 
- Closes #606 
- Supersedes #947 
- Relates #1700 
- Supersedes and Closes #1278
- Closes #1783


docker.go
```diff
@@ -628,26 +628,30 @@ func (c *DockerContainer) StartLogProducer(ctx context.Context) error {
...
-			// if we can't get the logs, panic, we can't return an error to anything
-			// from within this goroutine
-			panic(err)
+			// if we can't get the logs, retry in one second.
+			if ctx.Err() != nil {
+				// context done.
+				return
+			}
+			c.logger.Printf("cannot get logs for container %q: %v", c.ID, err)
+			time.Sleep(1 * time.Second)
```
pmalek, line 639: Better use `time.After()` to allow cancellation when sleeping

```
			// if we can't get the logs, retry in one second.
			c.logger.Printf("cannot get logs for container %q: %v", c.ID, err)
			select {
				case <-ctx.Done():
					// context is done
					return
				case <-time.After(time.Second):
			}
```

If you really want to log when `err` is `nil` then you might want to add 1 more select case without the sleep above and put the  log in between those 1 selects.

docker.go
```diff
@@ -686,13 +692,16 @@ func (c *DockerContainer) StartLogProducer(ctx context.Context) error {
...
+						// If the outer context is done, loop will exit in the next iteration.
 						continue
 					}
-					// we can not continue here as the next read most likely will not be the next header
-					_, _ = fmt.Fprintln(os.Stderr, logStoppedForOutOfSyncMessage)
-					return
+					// if we would continue here, the next header-read will result into random data...
+					// we need to restart the whole request.
+					_, _ = fmt.Fprintf(os.Stderr, "read log message: %+v. %s", err, logRestartedForOutOfSyncMessage)
+					time.Sleep(1 * time.Second)
```
pmalek, line 703: Ditto about `time.Sleep()` -> `time.After()`

docker.go
mdelapenya: For the second log if `err == nil`, you mean simply this?
```golang
select {
default:
	c.logger.Printf("cannot get logs for container %q: %v", c.ID, err)
}
```

pmalek, line 639: Something like this would work

```suggestion
			select {
			case <-ctx.Done():
				return	
			default:
				c.logger.Printf("cannot get logs for container %q: %v", c.ID, err)
			}
			select {
			case <-ctx.Done():
				return
			case <-time.After(time.Second):
			}
```

mdelapenya: Done in [9e4b5a1](https://github.com/testcontainers/testcontainers-go/pull/1777/commits/9e4b5a1ca76c94f4918755abcea36197bc8b3e89), thanks!

docker.go
```diff
@@ -684,15 +722,34 @@ func (c *DockerContainer) StartLogProducer(ctx context.Context) error {
...
-					if errors.Is(err, context.DeadlineExceeded) {
-						// Probably safe to continue here
+					select {
+					case <-ctx.Done():
+						return
+					default:
+						c.logger.Printf("error occurred reading log with known length %s", err.Error())
+					}
+
+					if errors.Is(err, context.DeadlineExceeded) || errors.Is(err, context.Canceled) {
```
pmalek, line 732: The `err` here is from the `io.ReadFull()` so we shouldn't really be concerned about the context cancellation here.

docker.go
```diff
@@ -658,13 +676,27 @@ func (c *DockerContainer) StartLogProducer(ctx context.Context) error {
...
 						continue
 					}
-					_, _ = fmt.Fprintf(os.Stderr, "container log error: %+v. %s", err, logStoppedForOutOfSyncMessage)
+
+					select {
+					case <-ctx.Done():
+						return
+					default:
+						c.logger.Printf("read log header: %+v. %s", err, logRestartedForOutOfSyncMessage)
+					}
```
pmalek, line 689: You'd have to excuse me for my previous comments. While this is totally OK, the `select` statement isn't really necessary here. There's a very slim chance of hitting the `case <-ctx.Done()` branch here so we might as well remove the whole `select` statement here.

docker.go
```diff
@@ -658,13 +676,27 @@ func (c *DockerContainer) StartLogProducer(ctx context.Context) error {
 						since = fmt.Sprintf("%d.%09d", now.Unix(), int64(now.Nanosecond()))
 						goto BEGIN
 					}
-					if errors.Is(err, context.DeadlineExceeded) {
-						// Probably safe to continue here
+					if errors.Is(err, context.DeadlineExceeded) || errors.Is(err, context.Canceled) {
+						// If the outer context is done, loop will exit in the next iteration.
 						continue
 					}
```
pmalek, line 682: Ditto here about `err` coming from `io.ReadFull()` (which is not context aware).

docker.go
```diff
@@ -628,26 +628,44 @@ func (c *DockerContainer) StartLogProducer(ctx context.Context) error {
...
-			select {
-			case <-stop:
+			if ctx.Err() != nil {
 				err := r.Close()
 				if err != nil {
 					// we can't close the read closer, this should never happen
 					panic(err)
 				}
-				return
+			}
```
pmalek, line 665: You might move this whole `if` into the `select` case in lines 667-668 where you check if the context is done.

docker.go
```diff
@@ -628,26 +628,44 @@ func (c *DockerContainer) StartLogProducer(ctx context.Context) error {
...
+
+			// we need a separate select here in order to be able to print the error
+			// and this cannot be done in the same select that checks for the timeout.
+			select {
+			case <-ctx.Done():
+				return
+			default:
+				c.logger.Printf("cannot get logs for container %q: %v", c.ID, err)
+			}
```
pmalek, line 646: Ditto as elsewhere, no point in the separate `select`, there's a very slim chance the context gets done between check in line 634 and here.

docker.go
```diff
@@ -684,15 +722,34 @@ func (c *DockerContainer) StartLogProducer(ctx context.Context) error {
...
-					return
+					// if we would continue here, the next header-read will result into random data...
+					// we need to restart the whole request.
+
+					select {
+					case <-ctx.Done():
+						return
+					default:
+						c.logger.Printf("read log message: %+v. %s", err, logRestartedForOutOfSyncMessage)
+					}
```
pmalek, line 744: ditto

docker.go
mdelapenya: Do you think we should not override the outer err and still check for the outer context cancellation?

pmalek: It's not overridden, is it? Between line 723 and here - 732 - nothing writes to `err`.

mdelapenya: Oh you are right, the initial err comes from another read: `_, err := io.ReadFull(r, h)` Sorry, my bad

docker.go
mdelapenya: Do you want the default case to be moved to the second select, or simply remove it completely?

pmalek: We can basically do this:

```suggestion
					c.logger.Printf("read log header: %+v. %s", err, logRestartedForOutOfSyncMessage)
```

mdelapenya: Ok cool! we are going to roll it back to the initial state where the print was out of the select 😅 

In any case, thanks for your assistance in this part of the code, some time you find it difficult to control all of them

docker.go
mdelapenya: I guess we can extract the log out and remove the select, right?

pmalek: Yup

mdelapenya:
@pmalek we merged https://github.com/testcontainers/testcontainers-go/pull/1701 yesterday, which is causing the conflicts here. Do you think that this PR is still needed after that?

pmalek:
> @pmalek we merged #1701 yesterday, which is causing the conflicts here. Do you think that this PR is still needed after that?

I'm not sure, haven't compared #1701 with this one in detail. Was just commenting out of good will on this PR :)

mdelapenya:
> I'm not sure, haven't compared #1701 with this one in detail. Was just commenting out of good will on this PR :)

Sure, and I do appreciate your help here. Many times, debugging concurrent tasks is extremely difficult, so having more eyes on it is really helpful 🙇

docker.go
```diff
@@ -628,26 +628,38 @@ func (c *DockerContainer) StartLogProducer(ctx context.Context) error {
...
-				err := r.Close()
-				if err != nil {
-					// we can't close the read closer, this should never happen
-					panic(err)
+			case <-ctx.Done():
+				if ctx.Err() != nil {
+					err := r.Close()
+					if err != nil {
+						// we can't close the read closer, this should never happen
+						panic(err)
```
pckhoi, line 658: Why are we panicking here? If you just want to log the error and return then isn't it better to consistently do it in the same way? Regarding logging, I see both usage of `c.logger.Printf` and `fmt.Fprintf(os.Stderr,` which seems strange.

If the purpose is to print the stack trace then maybe panic is best, but why here?

pckhoi:
I would say this change is still absolutely needed after #1701. I'm still observing `StopLogProducer` hanging when the go routine panic. We should replace the `stopProducer chan bool` with the `stopProducer context.CancelFunc` sooner or later, if not in this PR then in a separate PR.

pmalek:
> I would say this change is still absolutely needed after #1701. I'm still observing `StopLogProducer` hanging when the go routine panic. We should replace the `stopProducer chan bool` with the `stopProducer context.CancelFunc` sooner or later, if not in this PR then in a separate PR.

Hey @pckhoi 👋 

I've pushed #1911 which should fix this very issue. Still, as you've suggested some refactor of internal structure of `DockerContainer` might be desirable.

mdelapenya:
I think #1971 supersedes this one. Closing it

PR #1780, opened 2023-10-16
fix: use correct formatting directive for errors in lifecycle logs

robherley:
## What does this PR do?

This is a very tiny fix for the lifecycle logs so that they correctly print the error.

## Why is it important?

The actual error content is more useful than `%!w(...)` 😄 

## Related issues

none


mdelapenya (approved):
Good catch! LGTM

Thanks for contributing this tiny but valuable fix 👏

PR #1788, opened 2023-10-20
Auto-cleanup of k6 build cache

### LINKED ISSUE #1740: [Feature]: Provide mechanism for cleaning up build cache in k6 module
### Problem

The k6 module is built around the [k6x](https://github.com/szkiba/k6x) project, which allows the dynamic builds of a k6 executable with all the [k6 extensions](https://k6.io/docs/extensions/get-started/explore/) required by a test. 

In order to improve the speed of this build process when multiple tests are executed in a test suite, the k6 module's `RunContainer` method offers the `WithCache` option, which mounts a volume as a [cache](https://github.com/szkiba/k6x#cache) for the build process. 

Notice that this cache only contains build artifacts and therefore keeping it across tests will not have side effects in the test (as would for example reusing a test database).

This cache volume is created automatically by docker if it doesn't exist, but it is the responsibility of the test to delete it when no longer required. This creates a usability problem as `TestContainers` presently does not offer any mechanism for managing docker volumes.

### Solution

One alternative, [suggested](https://github.com/testcontainers/testcontainers-go/pull/1721#discussion_r1350350051) by @mdelapenya is to add a label to the cache volume with the test session ID and let the [garbage collection process](https://golang.testcontainers.org/features/garbage_collector/#ryuk) automatically delete it when no longer used.

A test session is defined as:
- a single "go test" invocation (including flags)
- a single "go test ./..." invocation (including flags)
- the execution of a single test or a set of tests using the IDE

Notice that if the cache volume is created and deleted automatically, the k6 module can generate a name automatically based on the session ID. Therefore the `WithCache` option does not require any argument:

```golang
k6.RunContainer(ctx, WithCache(), k6.WithTestScript("/tests/test.js"))
```

There is, however, another scenario that is not properly covered by the solution described above.

In a local environment, the developer may want to keep the build cache across multiple test sessions. Remember that each execution of `go test` is a new test session. 

Therefore the automatic cleanup of a test after each session will severely affect the execution time of the tests and the developer's productivity.

We consider that there should be a mechanism for overriding the default behavior of auto-cleanup of the build cache volume.

One possibility is to add an environment variable, such as `k6_BUILD_CACHE_VOLUME`. When this variable is set, the `WithCache` option will use the volume named in the variable as a build cache and will not add the session ID label, preventing the auto-cleanup at the end of the test session.

### Benefit

The lifecycle of the cache volume will be aligned with the life cycle of the test session, allowing reuse in a test suite but preventing the cache volume from being leaked in a CI environment. 

Also, using the `k6_BUILD_CACHE_VOLUME` the developer can reuse the test cache in their local environments and reuse.

### Alternatives

Leave the `WithCache` option as it is presently and add to `TestContainers` an API for managing docker volumes, to allow the test code to delete the cache volume when no longer required. 

Yes

pablochacin:
## What does this PR do?

Changes the behavior of the `WithCache` option to implement an automatic cleanup of the build cache.

## Why is it important?

Improves the usability of the module as the developer doesn't need to take care of deleting cache volume.

## Related issues
- Closes #1740 

## How to test this PR
 Running twice the tests in `k6_test.go` with the environment variable `TC_K6_BUILD_CACHE=k6-cache` the second run should last significantly less than the first (in the order of `30s` vs `3s`)


modules/k6/k6.go
```diff
@@ -49,13 +54,27 @@ func SetEnvVar(variable string, value string) testcontainers.CustomizeRequestOpt
...
+func WithCache() testcontainers.CustomizeRequestOption {
+	var volOptions *mount.VolumeOptions
+
+	cacheVol := os.Getenv("TC_K6_BUILD_CACHE")
+	// if no volume is provided, create one and ensure add labels for garbage collection
+	if cacheVol == "" {
+		sessionID := testcontainerssession.SessionID()
+		cacheVol = fmt.Sprintf("k6-cache-%s", sessionID)
+		volOptions = &mount.VolumeOptions{
+			Labels: testcontainersdocker.DefaultLabels(sessionID),
```
mdelapenya, line 69: I'm going to merge this today: https://github.com/testcontainers/testcontainers-go/pull/1790/files#diff-e9b78e9b34e14f076ba250ebd1042b315e63673569f82372a8064cece5f034bfR211

It includes a function that exposes the labels out of the internal package. Could you update this PR once the other is merged 🙏 ?

modules/k6/k6.go
```diff
@@ -49,13 +54,27 @@ func SetEnvVar(variable string, value string) testcontainers.CustomizeRequestOpt
 	}
 }
 
-// WithCache uses the given volume as a cache for building the k6 binary.
-// If the volume does not exists, it is created.
-func WithCache(cache string) testcontainers.CustomizeRequestOption {
+// WithCache sets a volume as a cache for building the k6 binary
+// If a volume name is provided in the TC_K6_BUILD_CACHE, this volume is used.
```
mdelapenya, line 58: I think is more clear. Wdyt?
```suggestion
// If a volume name is provided in the TC_K6_BUILD_CACHE, this volume is used and it will persist across test sessions.
```

modules/k6/k6.go
```diff
@@ -49,13 +54,27 @@ func SetEnvVar(variable string, value string) testcontainers.CustomizeRequestOpt
...
+// If no value is provided, a volume is created and removed when the test session ends.
+func WithCache() testcontainers.CustomizeRequestOption {
+	var volOptions *mount.VolumeOptions
+
+	cacheVol := os.Getenv("TC_K6_BUILD_CACHE")
```
mdelapenya, line 63: Out of curiosity: it does not collide on two different test sessions, right? The first one creates the volume, and the second simply uses it. At least this is how `docker volume create` works.

mdelapenya (approved):
LGTM! Added a few comments, not blockers.

Thanks for your time!

modules/k6/k6.go
pablochacin: I'm not sure what you mean by colliding on two test sessions. Effectively, the idea is that the first session (`go test` command)  using this environment variable will create the volume and the subsequent invocations will use it.

modules/k6/k6.go
pablochacin: Sure please ping me when it is merged.

mdelapenya: Done!

modules/k6/k6.go
mdelapenya: Yeah, this case:

1. the first `go test` creates the volume as it was passed as an env var, and it persists
2. the second `go test`. It will reuse the volume, right? Just want to double check that the second one does not fail because the volume already exists. I know it's handled by the docker client (we simply pass the volume mount to it), but want to verify it.

modules/k6/k6.go
pablochacin: @mdelapenya Something I don't see is how to get the `SessionID` if I don't use the internal package. I need it to create a unique name for the cache volume for each session. I cannot use a simple `UUID` because I need to ensure all tests in a session use the same volume name.

I could get it from the `sessionId` label in the `GenericLabels` map, but the name of this label is defined in the `testcontainersdocker` internal package. 

Maybe it should be exposed in the `testcontainers` package? Notice that at this point the container is not created and therefore I don't have access to `Container.SessionID`

mdelapenya: Isn't it enough with this? The sessionID is an implementation detail of the internals of the library.

```go
Labels: testcontainers.GenericLabels()
```
But I see you point regarding the cacheVol variable. You could get a hash of the labels map, so we do not need to expose anything else at the core. I would see the value in exposing the SessionID, but at the moment there is a way to get a unique identifier for that volume, formed by the hash of:
```go
return map[string]string{
		LabelBase:      "true",
		LabelLang:      "go",
		LabelSessionID: sessionID,
		LabelVersion:   internal.Version,
	}
```
Probably more work calculating the hash at the consumer side, that's true, at the cost of saving another method in the interface. If we see that consuming the SessionID is useful in more than one use case, then we can for sure add it back, not affecting this code, as how the vol is names is another implementation detail.

In any case, I'm not opposed to expose it in a separate PR so that this consumes it, but I'd like to get a clear value out of it. Thoughts?

pablochacin: I find it a little bit obscure and fragile to use a hash of the labels. I need this hash to be the same for all tests in a session. Therefore, all elements of the map should be the same for all tests, which seems to be the case presently, but this is an implementation detail that can change at any moment.

I'm curious about your reluctance to expose the SessionID when it is already exposed in `Container` interface. The problem is that I don't have access to it until the container is created and this logic is executed before this.

mdelapenya: I see.. I did not remember the `container.SessionID` method tbh, thanks for pointing out to that. Let's do it as you propose: I'll send a PR today exposing the SessionID as part of the core so that consumers can read it.

mdelapenya (approved):
LGTM, thanks for your implication in making this PR move forward including changes in the upstream.

PR #1790, opened 2023-10-23
chore: add an example of using localstack alongside AWS lambdas

mdelapenya:
- feat: expose the default labels of the library
- fix: do not use deprecated APIs
- chore: add an example of running localstack with Lambdas

## What does this PR do?
This PR adds a testable example demonstrating how to use the localstack module to test lambdas.

To allow that, we had to improve how the core exposes the default labels, in order to add those labels to the containers used as backend of the lambdas, so Ryuk can kill them at the end of the test session.

## Why is it important?
More use cases on localstack with testcontainers-go


modules/localstack/examples_test.go
```diff
@@ -113,3 +116,83 @@ func ExampleRunContainer_legacyMode() {
...
+	flagsFn := func() string {
+		labels := testcontainers.GenericLabels()
+
+		flags := ""
+		for k, v := range labels {
+			flags = fmt.Sprintf("%s -l %s=%s", flags, k, v)
+		}
+
+		return flags
+	}
```
mdelapenya, line 132: See https://github.com/localstack/localstack/issues/8616#issuecomment-1741278016

I noticed this does not work with localstack v1.4.0, that's why I moved the example to v2.3.0

modules/localstack/examples_test.go
```diff
@@ -113,3 +116,83 @@ func ExampleRunContainer_legacyMode() {
...
+	lambdaName := "localstack-lambda-url-example"
+
+	container, err := localstack.RunContainer(ctx,
+		testcontainers.WithImage("localstack/localstack:2.3.0"),
+		testcontainers.CustomizeRequest(testcontainers.GenericContainerRequest{
+			ContainerRequest: testcontainers.ContainerRequest{
+				Env: map[string]string{
+					"SERVICES":                 "lambda",
+					"LAMBDA_EXECUTOR":          "docker",
+					"PROVIDER_OVERRIDE_LAMBDA": "asf",
```
eddumelendez, line 143: I think those are not required in LS 2.3.0

modules/localstack/examples_test.go
```diff
@@ -113,3 +116,83 @@ func ExampleRunContainer_legacyMode() {
...
+	_, reader, err := container.Exec(ctx, []string{"cat", "output.txt"}, exec.Multiplexed())
+	if err != nil {
+		panic(err)
+	}
+
+	content, err := io.ReadAll(reader)
+	if err != nil {
+		panic(err)
+	}
+	fmt.Println(string(content))
```
eddumelendez, line 194: What about asserting the text returned by the function?

eddumelendez (approved):
Just few notes. 👍🏽

modules/localstack/examples_test.go
mdelapenya: In testable examples, there are no assertions but an internal comparison of the output of stdout against [the](https://github.com/testcontainers/testcontainers-go/pull/1790/files#diff-2a7f31c70cbb0499942309ff81dbc15898810fb8734f0278eaa52f4da525b16dR196-R197)

modules/localstack/examples_test.go
mdelapenya: Cool! I tried locally without LAMBDA_EXECUTOR and PROVIDER_OVERRIDE_LAMBDA and it works! Thanks!

PR #1825, opened 2023-10-27
docs: add some Vault module examples

kuisathaverat:
## What does this PR do?

some of the examples in doce are not shown at https://pkg.go.dev/github.com/testcontainers/testcontainers-go, This PR add the `//output` to make them visible.
It add some examples for the Vault module.

## Why is it important?

Examples in the documentation allow to save time to onboarding a project in Testcontainers.


docker_test.go
```diff
@@ -1134,6 +1164,10 @@ func ExampleContainer_Stop() {
 	}()
 	timeout := 10 * time.Second
 	_ = nginxC.Stop(ctx, &timeout)
+
+	fmt.Print("you have a running and stop the container")
+
+	// Output: you have a running and stop the container
```
mdelapenya, line 1170: Could you apply this to https://github.com/testcontainers/testcontainers-go/pull/1825/files#diff-037b83426e676aa3786c08b334b0db7bb70235792961cbe4ed3340bc0a60a7a8R1157? 🙏  
```
	if err != nil {
		panic(err)
	}
```

I think we can improve the output like this:
```suggestion
	fmt.Println("Container has been started")
	timeout := 10 * time.Second
	err = nginxC.Stop(ctx, &timeout)
	if err != nil {
		panic(err)
	}

	fmt.Println("Container has been stopped")

	// Output:
	// Container has been started
	// Container has been stopped
```

mdelapenya (approved):
I added one comment regarding one of the outputs, but other than that LGTM!

Thanks for taking care of the docs 🙏 

PS: At some point I'll go back to more examples trying to improve them even more.

mdelapenya (approved):
LGTM, thanks!

modules/vault/examples_test.go
```diff
@@ -34,3 +35,76 @@ func ExampleRunContainer() {
 	// Output:
 	// true
 }
+
+func ExampleRunContainer_withToken() {
```
mdelapenya, line 39: @kuisathaverat I fixed the testable example name, as it does not find a `RunContainerWithToken` function in the code. The testable examples need to follow this pattern: `Example${FunctionName}`. Please see https://go.dev/blog/examples#example-function-names

mdelapenya (approved):
LGTM, thanks!!

As I said earlier in the comments, I'll work in the short term in refining all the testable examples. Thanks for pointing out! 🙏

Issue #1898, opened 2023-11-03
[Enhancement]: reduce flakiness in CLickHouse tests


mdelapenya:
### Proposal

Current failure on tests: https://github.com/testcontainers/testcontainers-go/actions/runs/6749314222/job/18349567515
> clickhouse_test.go:214: 
          	Error Trace:	/home/runner/work/testcontainers-go/testcontainers-go/modules/clickhouse/clickhouse_test.go:214
          	Error:      	Received unexpected error:
          	            	[handshake] unexpected packet [72] from server
          	Test:       	TestClickHouseWithConfigFile/YAML_Config
      clickhouse_test.go:215: 
          	Error Trace:	/home/runner/work/testcontainers-go/testcontainers-go/modules/clickhouse/clickhouse_test.go:215
          	Error:      	"[]" should have 1 item(s), but has 0
          	Test:       	TestClickHouseWithConfigFile/YAML_Config

Probably improved with a retry 🤔

rafiramadhana:
@mdelapenya can i work on this? thx

PR #1902, opened 2023-11-04
Reduce flakiness in ClickHouse tests

rafiramadhana:
## What does this PR do?

Reduce flakiness in ClickHouse tests by replacing `performCRUD` with `performCRUDWithRetry`.

## Related issues

- Closes https://github.com/testcontainers/testcontainers-go/issues/1898

### DIFF
--- a/modules/clickhouse/clickhouse_test.go
+++ b/modules/clickhouse/clickhouse_test.go
@@ -4,11 +4,11 @@ import (
 	"context"
 	"path/filepath"
 	"testing"
+	"time"
 
 	ch "github.com/ClickHouse/clickhouse-go/v2"
 	"github.com/ClickHouse/clickhouse-go/v2/lib/driver"
 	"github.com/stretchr/testify/assert"
-
 	"github.com/testcontainers/testcontainers-go"
 )
 
@@ -89,7 +89,7 @@ func TestClickHouseConnectionHost(t *testing.T) {
 	defer conn.Close()
 
 	// perform assertions
-	data, err := performCRUD(conn)
+	data, err := performCRUDWithRetry(conn)
 	assert.NoError(t, err)
 	assert.Len(t, data, 1)
 }
@@ -121,7 +121,7 @@ func TestClickHouseDSN(t *testing.T) {
 	defer conn.Close()
 
 	// perform assertions
-	data, err := performCRUD(conn)
+	data, err := performCRUDWithRetry(conn)
 	assert.NoError(t, err)
 	assert.Len(t, data, 1)
 }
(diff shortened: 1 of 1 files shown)


modules/clickhouse/clickhouse_test.go
```diff
@@ -210,13 +210,29 @@ func TestClickHouseWithConfigFile(t *testing.T) {
...
 			// perform assertions
-			data, err := performCRUD(conn)
+			data, err := performCRUDWithRetry(conn)
 			assert.NoError(t, err)
 			assert.Len(t, data, 1)
 		})
 	}
 }
 
+func performCRUDWithRetry(conn driver.Conn) (tt []Test, err error) {
```
mdelapenya, line 220: Could we avoid named return parameters? There is no specific contributing rule for it, although I do prefer not using them if possible (we do have it in a few places, planned to refactor it at some point)

mdelapenya: I'd also merge both methods into just one, as we are not going to use `performCRUD` by itself.

mdelapenya (commented):
This LGTM, although I'd like to know your thoughts on the comments I left.

Pending my final approval until then.

Thank you very much for taking the time to contribute this!

modules/clickhouse/clickhouse_test.go
rafiramadhana: > Could we avoid named return parameters?

ok, i will update the method

> I'd also merge both methods into just one, as we are not going to use performCRUD by itself.

i see. i was avoiding to delete methods because im new to this code base

but, if you say so, ok i'll merge them

mdelapenya: No worries, as the methods are used only within the tests, so nobody will be affected by the change.

modules/clickhouse/clickhouse_test.go
```diff
@@ -218,22 +218,35 @@ func TestClickHouseWithConfigFile(t *testing.T) {
...
+		delay    = 100 * time.Millisecond
+		err      error
+		rows     []Test
+	)
 
-	err = conn.Exec(context.Background(), "INSERT INTO test_table (id) VALUES (1);")
-	if err != nil {
-		return nil, err
-	}
+	for i := 0; i < maxRetry; i++ {
```
mdelapenya, line 228: Given we already use backoff package, let's use it here too. Not a blocker, though

rafiramadhana: ok, on it

rafiramadhana: updated @mdelapenya

mdelapenya: Thanks! The CI is running. Will approve it after the CI passes, thanks!

mdelapenya:
Could you run `make lint` in the clickhouse directory? It seems there is something wrong with the format. 🙏

rafiramadhana:
> Could you run `make lint` in the clickhouse directory? It seems there is something wrong with the format. 🙏

updated https://github.com/testcontainers/testcontainers-go/pull/1902/commits/a941bb212ddb46fe62aa127a1a46f6338c6df284

mdelapenya (approved):
LGTM, thanks for making the test suite more reliable!!!

PR #1904, opened 2023-11-05
fix(reaper): fix race condition when reusing reapers

lefinal:
Reaper reuse/creation logic has been adjusted to facilitate the reuse of already running reapers. This includes using a specific naming convention based on the session id, and handling failed container creation attempts due to name conflicts by retrieving the already running reaper container. This fixes a race condition when tests are run in parallel in multiple packages which renders global locks ineffective.

## What does this PR do?

Reaper container now uses a deterministic name based on the session id.
If container creation fails due to name conflict, we reuse the existing  container.

## Why is it important?

If an application uses testcontainers in multiple packages, `go test` will run them isolated from each other.
This will render global locks as in _reaper.go_  ineffective causing a race condition as multiple reapers are created.
However, they all use the same session id, because of the parent process being the same.
This may lead to unwanted cancellation of containers of still running tests in other packages.

There is already logic for reducing the risk of race conditions in _reaper.go_ by introducing a small random delay when searching for any running reaper container.
However, this does not fully avoid the race condition as there is a small delay after container creation and it being visible from `ContainerList`.

## How to test this PR

Tests have been added to _reaper_test.go_.
You can test the functionality by simply discarding the changes in _reaper.go_ and running the tests.

### DIFF
--- a/reaper.go
+++ b/reaper.go
@@ -283,7 +283,7 @@ func newReaper(ctx context.Context, sessionID string, provider ReaperProvider, o
 				}
 			}
 			if err != nil {
-				return nil, fmt.Errorf("look up reaper container because creation failed due to name conflict: %w", err)
+				return nil, fmt.Errorf("look up reaper container due to name conflict failed: %w", err)
 			}
 			// If the reaper container was not found, it is most likely to have died in
 			// between as we can exclude any client errors because of the previous error
--- a/reaper.go
+++ b/reaper.go
@@ -283,7 +283,7 @@ func newReaper(ctx context.Context, sessionID string, provider ReaperProvider, o
 			if reaperContainer == nil {
 				return nil, fmt.Errorf("look up reaper container returned nil although creation failed due to name conflict")
 			}
-			Logger.Printf("🔥 Canceling creation - Reaper obtained from Docker for this test session %s", reaperContainer.ID)
+			Logger.Printf("🔥 Reaper obtained from Docker for this test session %s", reaperContainer.ID)
 			reaper, err := reuseReaperContainer(ctx, sessionID, provider, reaperContainer)
 			if err != nil {
 				return nil, err
--- a/reaper.go
+++ b/reaper.go
@@ -261,8 +261,13 @@ func newReaper(ctx context.Context, sessionID string, provider ReaperProvider, o
 		// already existing due to race conditions. We manually match the error message
 		// as we do not have any error types to check against.
 		if createContainerFailDueToNameConflictRegex.MatchString(err.Error()) {
-			// Manually retrieve the already running reaper container. As it may take a while
-			// to observe the container, despite creation failure, we retry a few times.
+			// Manually retrieve the already running reaper container. However, we need to
+			// use retries here as there are two possible race conditions that might lead to
+			// errors: In most cases, there is a small delay between container creation and
+			// actually being visible in list-requests. This means that creation might fail
+			// due to name conflicts, but when we list containers with this name, we do not
+			// get any results. In another case, the container might have simply died in the
+			// meantime and therefore cannot be found.
 			const timeout = 5 * time.Second
 			const cooldown = 100 * time.Millisecond
 			start := time.Now()
@@ -280,6 +285,10 @@ func newReaper(ctx context.Context, sessionID string, provider ReaperProvider, o
 			if err != nil {
 				return nil, fmt.Errorf("look up reaper container because creation failed due to name conflict: %w", err)
 			}
+			// If the reaper container was not found, it is most likely to have died in
+			// between as we can exclude any client errors because of the previous error
+			// check. Because the reaper should only die if it performed clean-ups, we can
+			// fail here as the reaper timeout needs to be increased, anyway.
 			if reaperContainer == nil {
 				return nil, fmt.Errorf("look up reaper container returned nil although creation failed due to name conflict")
 			}
--- a/reaper.go
+++ b/reaper.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"math/rand"
 	"net"
+	"regexp"
 	"strings"
 	"sync"
 	"time"
@@ -50,6 +51,14 @@ func NewReaper(ctx context.Context, sessionID string, provider ReaperProvider, r
 	return reuseOrCreateReaper(ctx, sessionID, provider, WithImageName(reaperImageName))
 }
 
+// reaperContainerNameFromSessionID returns the container name that uniquely
+// identifies the container based on the session id.
+func reaperContainerNameFromSessionID(sessionID string) string {
+	// The session id is 64 characters, so we will not hit the limit of 128
+	// characters for container names.
+	return fmt.Sprintf("reaper_%s", sessionID)
+}
+
 // lookUpReaperContainer returns a DockerContainer type with the reaper container in the case
 // it's found in the running state, and including the labels for sessionID, reaper, and ryuk.
 // It will perform a retry with exponential backoff to allow for the container to be started and
@@ -67,7 +76,7 @@ func lookUpReaperContainer(ctx context.Context, sessionID string) (*DockerContai
 
 	// we want random intervals between 100ms and 500ms for concurrent executions
 	// to not be synchronized: it could be the case that multiple executions of this
-	// function happen at the same time (specially when called from a different test
+	// function happen at the same time (specifically when called from a different test
 	// process execution), and we want to avoid that they all try to find the reaper
 	// container at the same time.
 	exp.InitialInterval = time.Duration(rand.Intn(5)*100) * time.Millisecond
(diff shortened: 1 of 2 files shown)


reaper.go
```diff
@@ -237,6 +257,39 @@ func newReaper(ctx context.Context, sessionID string, provider ReaperProvider, o
 
 	c, err := provider.RunContainer(ctx, req)
 	if err != nil {
+		// We need to check whether the error is caused by a container with the same name
+		// already existing due to race conditions. We manually match the error message
+		// as we do not have any error types to check against.
```
mdelapenya, line 262: Do you think that creating our own error here would make sense? Given the description, it seems it's needed and I think the reaper is part of our own infra, which could deserve its own error type. Thoughts?

reaper.go
```diff
@@ -146,19 +156,11 @@ func reuseOrCreateReaper(ctx context.Context, sessionID string, provider ReaperP
 	reaperContainer, err := lookUpReaperContainer(context.Background(), sessionID)
 	if err == nil && reaperContainer != nil {
 		// The reaper container exists as a Docker container: re-use it
-		endpoint, err := reaperContainer.PortEndpoint(ctx, "8080", "")
+		Logger.Printf("🔥 Reaper obtained from Docker for this test session %s", reaperContainer.ID)
```
mdelapenya, line 159: I think this log entry should be printed just once, instead of: https://github.com/testcontainers/testcontainers-go/pull/1904/files#diff-e8ab4ec17e978db949cd15726ba0efa9c89fd714e7e6e0aa537ed7fc1bbde7fdR286

mdelapenya (dismissed):
Hey @lefinal thanks for this fix. You're doing a great job investigating the potential race conditions in the project, which I enormously appreciate it.

I took the test code and put it into the main branch and it failed, so it demonstrated the bug. Same test in this branch does pass 👏 

In any case, there a few comments to be addressed, but I'm approving this PR already. LGTM

reaper.go
lefinal: What do you mean exactly? The regex matching is because the returned error from the Docker client is not typed. We only get a string representation from the SDK, no error codes, etc.

mdelapenya: Indeed. I wonder if i't worth it to create our own error for that case, doing the regexp when asserting the error type. But probably is overkill. I wanted to check with you first

reaper.go
lefinal: The "original" entry was logged when lookup succeeded and we would not try to create. The added (slightly changed) one is logged if we already concluded that the reaper is not up yet but it turned out that we were "wrong". This is why the entry is logged inside of `newReaper` as we usually would not expect it from outside.
The main reason why I would add it is that we get this log entry in `newReaper` right before creation  fails:
```
2023/11/06 17:03:02 🐳 Creating container for image docker.io/testcontainers/ryuk:0.5.1
```
I thought that it would be confusing to see this message multiple times, expecting Ryuk to be started despite it _not_ starting because of recovery.
If you still want to remove the added log entry, just tell me :)

reaper.go
lefinal: Okay, I get it. Would definitely be more clean but if I add it for the `DockerProvider` and only check against this error type, other implementations of `ContainerProvider` that people might use, will definitely not work with this as they would need to update their implementations to return the correct error, right?

reaper.go
```diff
@@ -237,6 +257,39 @@ func newReaper(ctx context.Context, sessionID string, provider ReaperProvider, o
...
+				if err == nil && reaperContainer != nil {
+					break
+				}
+				select {
+				case <-ctx.Done():
+				case <-time.After(cooldown):
+				}
+			}
+			if err != nil {
+				return nil, fmt.Errorf("look up reaper container because creation failed due to name conflict: %w", err)
```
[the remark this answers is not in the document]
lefinal: Uhm, sounds like some words twisted, right? If you  want to shorten it, we can replace `because` with `as` or `due to` :)

reaper.go
mdelapenya: For me the most confusing part was to see this in the logs:

> 🔥 Canceling creation - Reaper obtained from Docker for this test session 06ed5f976d4d87db94954a7505a3b7cef8e5e22afe511dabd27bfc373d99afc8

I like this PR, so I'm not blocking it at all, just trying to make the log a bit more readable. We can simply keep the PR as is updating the above log message to the original one:

> 🔥 Reaper obtained from Docker for this test session 06ed5f976d4d87db94954a7505a3b7cef8e5e22afe511dabd27bfc373d99afc8

Regarding the log you mention that appears before the container is created:

> 2023/11/06 17:03:02 🐳 Creating container for image docker.io/testcontainers/ryuk:0.5.1

it comes from the container lifecycle, in the beforeCreates hook. We could (in a follow-up) move part of this code detecting the existence of the reaper to the reaper's container lifecycle hook, trying to be smarter on what we print out. But that would be totally secondary, and not related to this PR.

reaper.go
mdelapenya: I had in mind a custom error type, something like `ErrReuseEmptyName = errors.New("with reuse option a container name mustn't be empty")`, not added to the Provider.

But in any case, let's not block this PR and keep it simple :)

reaper.go
```diff
@@ -237,6 +257,39 @@ func newReaper(ctx context.Context, sessionID string, provider ReaperProvider, o
...
+			}
+			if reaperContainer == nil {
```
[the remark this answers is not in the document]
lefinal: The only reason for this is that the reaper might have died in the meantime. There are two possible options of handling this:
1. We simply fail (current).
2. We perform a recursive call. I'd actually prefer this as this should rarely be the case. This would result in these changes (improved documentation as well):
```go
if createContainerFailDueToNameConflictRegex.MatchString(err.Error()) {
			// Manually retrieve the already running reaper container. However, we need to
			// use retries here as there are two possible race conditions that might lead to
			// errors: In most cases, there is a small delay between container creation and
			// actually being visible in list-requests. This means that creation might fail
			// due to name conflicts, but when we list containers with this name, we do not
			// get any results. In another case, the container might have simply died in the
			// meantime and therefore cannot be found.
			const timeout = 5 * time.Second
			const cooldown = 100 * time.Millisecond
			start := time.Now()
			var reaperContainer *DockerContainer
			for time.Since(start) < timeout {
				reaperContainer, err = lookUpReaperContainer(ctx, sessionID)
				if err == nil && reaperContainer != nil {
					break
				}
				select {
				case <-ctx.Done():
				case <-time.After(cooldown):
				}
			}
			if err != nil {
				return nil, fmt.Errorf("look up reaper container because creation failed due to name conflict: %w", err)
			}
			// If the reaper container was not found, it is most likely to have died in
			// between as we can exclude any client errors because of the previous error
			// check. We simply try creating a new reaper again. As this should rarely be the
			// case, we can assume that performing a recursive call is safe.
			if reaperContainer == nil {
				return newReaper(ctx, sessionID, provider, opts...)
			}
			Logger.Printf("🔥 Canceling creation - Reaper obtained from Docker for this test session %s", reaperContainer.ID)
			reaper, err := reuseReaperContainer(ctx, sessionID, provider, reaperContainer)
			if err != nil {
				return nil, err
			}
			return reaper, nil
		}
```

reaper.go
lefinal: Mhm, but for what would the error be? I assumed you mean a custom error that is returned by `RunContainer` (or underlying  methods). This would allow replacing the regex check in _reaper.go_ with `errors.Is`. But then we would expect the Provider to actually return this custom error. And custom implementations of `ContainerProvider` would need to return the correct type as well for this to work.

reaper.go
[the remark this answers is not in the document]
mdelapenya: Oh I think I expressed myself incorrectly and incompletely 🙏 I want this explanation as code comments 🙏 

If possible, I'd try to avoid recursive calls.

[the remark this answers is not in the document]
lefinal: Okay, I get it. Pushed  the desired changes.

reaper.go
mdelapenya: I was referring defining our own error type, and creating a type assertion for the error, something like:

```go
type duplicatedReaperError struct
   Message string
}

func NewDuplicatedReaperError(message string) error {
	return &duplicatedReaperError{message}
}

func (err *duplicatedReaperError) Error() string {
   return err.Message
}

// IsDuplicatedReaperError checks if an error of type duplicatedReaperError exists.
func IsDuplicatedReaperError(err error) bool {
	var e duplicatedReaperError
	// do the regex here
	return errors.As(err, &e)
}
```

but in any case, I'm not forcing it :D It's fine as it is now.

reaper.go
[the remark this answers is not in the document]
mdelapenya: You're right. I need to see the output message with more clarity. It's me with this part of the sentence:

> look up reaper container creation failed

Can we find a simpler one to describe that the reaper was not found after a name conflict?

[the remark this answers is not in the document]
lefinal: See added commit: `look up reaper container due to name conflict failed: ..`. I usually prefer imperative style but I saw that you use the `x failed` style throughout the code base.

mdelapenya (approved):
LGTM, thanks for your patience during the review 🙇

PR #1907, opened 2023-11-07
chore: deprecate BindMount APIs

mdelapenya:
## What does this PR do?
This PR deprecates all the BindMount types and functions/methods, removing their internal usage from the library (test code, basically).

The internal changes will modify the usage of BindMounts in this manner:

- using ContainerRequest's Files field.
- using ContainerRequest's HostCOnfigModifier field, using the closure to update the Binds of the Docker container's HostConfig type.
- using the CopyFileToContainer/CopyDirToContainer methods of a running container.

It's also merging the docs regarding copy files and volumes, as both topics are closely related.

## Why is it important?
BindMounts will make code not be portable cross docker environments, e.g. in remote Docker hosts, as the file binding won't exists in the remote.

For that reason we advocate for using the Files attribute of the container request in order to copy files/dirs to a container after it's created but before it's started, or the HostCOnfigModifier to define container binds, or CopyFileToContainer/CopyDirToContainer container methods.

## Related issues

- Relates https://github.com/testcontainers/testcontainers-java/pull/7652


container.go
```diff
@@ -305,6 +305,8 @@ func (c *ContainerRequest) validateContextOrImageIsSpecified() error {
 	return nil
 }
 
+// validateMounts ensures that the mounts do not have duplicate targets.
+// It will check the Mounts and HostConfigModifier.Binds fields.
 func (c *ContainerRequest) validateMounts() error {
```
mdelapenya, line 310: TBH I think this validation function is very little useful, as it's does really too few checks.

We probably need a follow up discussion to remove it or rethink what to check here

docker_test.go
```diff
@@ -1187,43 +1196,29 @@ func ExampleContainer_MappedPort() {
...
 
-func TestContainerCreationWithBindAndVolume(t *testing.T) {
+func TestContainerCreationWithVolumeAndFileWritingToIt(t *testing.T) {
 	absPath, err := filepath.Abs(filepath.Join(".", "testdata", "hello.sh"))
 	if err != nil {
 		t.Fatal(err)
 	}
 	ctx, cnl := context.WithTimeout(context.Background(), 30*time.Second)
 	defer cnl()
-	// Create a Docker client.
```
mdelapenya, line 1197: No need to instantiate a client for the volume, as Docker will handle the mounts correctly.

PR #1910, opened 2023-11-10
Add username and password functions to mongodb

### LINKED ISSUE #1908: [Enhancement]: adding more customization capability to MongoDB module
### Proposal

Currently, there is no functions in MongoDB module to add username and password to the container. (There is this capability for postgres https://github.com/testcontainers/testcontainers-go/blob/555cb764ee8dbd9b8ce54349d230a402aa439b8f/modules/postgres/postgres.go#L91C36-L91C73 )

If this seems useful, I would like to work on this.

ugrkm:
## What does this PR do?

This will add username and password functions to mongodb module. I removed mongodb_test module and put the tests in mongodb module. Because this was the pattern for other modules like postgres and mariadb.

## Why is it important?

This enables users to be able to configure username and password. Also, it creates feature parity with other modules.

## Related issues

- Closes #1908

## How to test this PR

I already created a test for the change in mongodb_test.go file.

### DIFF
--- a/modules/mongodb/mongodb.go
+++ b/modules/mongodb/mongodb.go
@@ -2,6 +2,7 @@ package mongodb
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/testcontainers/testcontainers-go"
 	"github.com/testcontainers/testcontainers-go/wait"
@@ -13,6 +14,8 @@ const defaultImage = "mongo:6"
 // MongoDBContainer represents the MongoDB container type used in the module
 type MongoDBContainer struct {
 	testcontainers.Container
+	username string
+	password string
 }
 
 // RunContainer creates an instance of the MongoDB container type
@@ -24,6 +27,7 @@ func RunContainer(ctx context.Context, opts ...testcontainers.ContainerCustomize
 			wait.ForLog("Waiting for connections"),
 			wait.ForListeningPort("27017/tcp"),
 		),
+		Env: map[string]string{},
 	}
 
 	genericContainerReq := testcontainers.GenericContainerRequest{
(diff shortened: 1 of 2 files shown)


modules/mongodb/mongodb_test.go
```diff
@@ -1,21 +1,22 @@
-package mongodb_test
+package mongodb
```
mdelapenya, line 1: Even tough there could be packages with the tests in the same package than the module, I'd not change this, as this is the preferred way to test the module.

I can open an issue to update those tests that live in the same package as its module.

modules/mongodb/mongodb.go
```diff
@@ -40,10 +44,44 @@ func RunContainer(ctx context.Context, opts ...testcontainers.ContainerCustomize
...
+// WithPassword sets the initial password of the user to be created when the container starts
+// It is used in conjunction with WithUsername to set a username and its password.
+// This environment variable sets the superuser password for MongoDB.
+func WithPassword(password string) testcontainers.CustomizeRequestOption {
+	return func(req *testcontainers.GenericContainerRequest) {
+		req.Env["MONGO_INITDB_ROOT_PASSWORD"] = password
+	}
+}
+
 // ConnectionString returns the connection string for the MongoDB container
```
mdelapenya, line 73: Could you update this method comment to reflect the username and the password?

modules/mongodb/mongodb.go
```diff
@@ -40,10 +44,44 @@ func RunContainer(ctx context.Context, opts ...testcontainers.ContainerCustomize
 		return nil, err
 	}
 
+	username := req.Env["MONGO_INITDB_ROOT_USERNAME"]
+	password := req.Env["MONGO_INITDB_ROOT_PASSWORD"]
+	if username != "" && password != "" {
+		return &MongoDBContainer{Container: container, username: username, password: password}, nil
+	}
```
mdelapenya, line 51: Do we want to add some validation if one of the pair is empty? PTAL to MariaDB module for that.

mdelapenya (commented):
Hi @ugrkm thanks for this! I added some comments before accepting the PR.

Could you take a look at them 🙏 ?

modules/mongodb/mongodb_test.go
ugrkm: I reverted it to mongodb_test. Out of curiosity, why do you want test to be in a separate package? Is to make sure importers of the package shouldn't use the test methods ?

modules/mongodb/mongodb.go
ugrkm: Updated it.

modules/mongodb/mongodb.go
ugrkm: Added validation to return an error if one of the pair is empty. This was the behavior of MongoDB though. It fails to run if one of the pair is missing. But I think it is better to be explicit about it.

modules/mongodb/mongodb_test.go
mdelapenya: >  Is to make sure importers of the package shouldn't use the test methods ?

Indeed, and also for "feeling" as a real consumer of the module, having to consume the public APIs as a downstream user.

modules/mongodb/mongodb_test.go
```diff
@@ -76,3 +78,38 @@ func ExampleRunContainer_connect() {
...
+
+	t.Cleanup(func() {
+		if err := container.Terminate(ctx); err != nil {
+			t.Fatalf("failed to terminate container: %s", err)
+		}
+	})
+
+	connStr, err := container.ConnectionString(ctx)
+	if err != nil {
+		panic(err)
```
mdelapenya, line 103: Given this is a regular test, not a testable example, let's use the testing library to perform the assertions.
```suggestion
		t.Fatal(err)
```

modules/mongodb/mongodb_test.go
```diff
@@ -76,3 +78,38 @@ func ExampleRunContainer_connect() {
...
+	}
+
+	mongoClient, err := mongo.Connect(ctx, options.Client().ApplyURI(connStr))
+	if err != nil {
+		panic(err)
```
mdelapenya, line 108:
```suggestion
		t.Fatal(err)
```

modules/mongodb/mongodb_test.go
```diff
@@ -76,3 +78,38 @@ func ExampleRunContainer_connect() {
...
+	}
+
+	err = mongoClient.Ping(ctx, nil)
+	if err != nil {
+		panic(err)
```
mdelapenya, line 113:
```suggestion
		t.Fatal(err)
```

modules/mongodb/mongodb_test.go
```diff
@@ -76,3 +78,38 @@ func ExampleRunContainer_connect() {
 	// Output:
 	// test
 }
+
+func TestMongoDB(t *testing.T) {
```
mdelapenya, line 82: One last thing before merging this PR. The code in this file is having only testable examples. Are you OK having your test as a regular test or do you prefer to have it as a testable example? If so, you'd need to rename the test method to `ExampleRunContainer_withCredentials`, printing messages and checking them in an output comment. Please see above examples

mdelapenya: One of the benefits for testable examples is they appear in the pkg.go.dev site

modules/mongodb/mongodb_test.go
```diff
@@ -76,3 +78,38 @@ func ExampleRunContainer_connect() {
...
+
+	mongoClient, err := mongo.Connect(ctx, options.Client().ApplyURI(connStr))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	err = mongoClient.Ping(ctx, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
```
mdelapenya, line 114: A testable example needs an output comment for something already printed with fmt. Please see https://github.com/testcontainers/testcontainers-go/pull/1910/files#diff-cbb99f9ae72ba7f7b2a251421a76419f12d724152c813a69fa40b2c4d4b730cfR76 and https://github.com/testcontainers/testcontainers-go/pull/1910/files#diff-cbb99f9ae72ba7f7b2a251421a76419f12d724152c813a69fa40b2c4d4b730cfR78

Therefore this example needs both to be runnable at the Ci

ugrkm: Oh, I missed that. I updated it. I also removed the testing.T argument since testable examplea do not take  arguments.

mdelapenya (approved):
LGTM, thanks for contributing to the MongoDB module!

I'll add a follow-up PR including docs for the functions, so we see this merged and you see how to contribute to the docs (not very well documented yet).

Thanks!

mdelapenya:
@ugrkm PTAL at #1930, thanks!

PR #1912, opened 2023-11-11
feat: get container image

FedericoAntoniazzi:
## What does this PR do?

As per the `testcontainers.Container.Name` method, allow retrieving the image from a `testcontainers.Container` interface.

The function has been named `ContainerImage` since `Image` clashes with the Container struct field.

## Why is it important?

I'm using it on a project to retrieve images from containers running in a docker host and I use testcontainers-go to create containers for a testing environment.

## Related issues

No related issues.

## How to test this PR
This PR works exactly like `Container.Name()` method, which hasn't been tested, probably because it mainly uses external dependencies.

### DIFF
--- a/container.go
+++ b/container.go
@@ -42,6 +42,7 @@ type Container interface {
 	MappedPort(context.Context, nat.Port) (nat.Port, error)         // get externally mapped port for a container port
 	Ports(context.Context) (nat.PortMap, error)                     // get all exposed ports
 	SessionID() string                                              // get session id
+	ContainerImage(context.Context) (string, error)                 // get the container image
 	IsRunning() bool
 	Start(context.Context) error                 // start the container
 	Stop(context.Context, *time.Duration) error  // stop the container
--- a/docker.go
+++ b/docker.go
@@ -378,6 +378,15 @@ func (c *DockerContainer) Name(ctx context.Context) (string, error) {
 	return inspect.Name, nil
 }
 
+// ContainerImage gets the image of the container.
+func (c *DockerContainer) ContainerImage(ctx context.Context) (string, error) {
+	inspect, err := c.inspectContainer(ctx)
+	if err != nil {
+		return "", err
+	}
+	return inspect.Image, nil
+}
+
 // State returns container's running state
 func (c *DockerContainer) State(ctx context.Context) (*types.ContainerState, error) {
 	inspect, err := c.inspectRawContainer(ctx)


mdelapenya:
Hi @FedericoAntoniazzi, thanks for taking the time for your contribution. I'd like to understand more about your use case, as the Image of a container can be obtained from the container request that originated it (unless there is a process modifying it). Could you elaborate a little bit more if possible?

FedericoAntoniazzi:
Thank you for your response @mdelapenya, let me try to explain my use case:
I'm developing a tool that queries the docker engine to retrieve the image from running containers, and I'm using testcontainers-go as a wrapper for creating N containers for each test case. 
Basically, I want to test different image names and formats (e.g. `nginx`, `library/nginx`, and `docker.io/library/nginx`) so I created a function for creating those containers by calling it in each test case. This function returns the list of testcontainers that have been spawned, in order to have information such as the ID or the name (which is modified from the request, I notice a `/` added by the docker daemon), and retrieving the image directly from the testcontainers-go requires less custom code.

mdelapenya:
Hi @FedericoAntoniazzi thanks for the explanation. We are exposing a Docker client in order to do this kind of things:

```go
ctx := context.Background()
client, err := testcontainers.NewDockerClientWithOpts(ctx)
if err != nil {
	// ...
}
defer client.Close()

inspect, err := client.ContainerInspect(ctx, c.ID)
if err != nil {
	// ...
}
// Do whatever you need with the ContainerJSON here
```

I'd advise against adding a new method to the interface given the simplicity of the above solution, thanks to the client. But I'd like to listen to your thoughts before closing this one.

FedericoAntoniazzi:
Hi @mdelapenya, I apologize for my late reply.

Basically I've already implemented the proposed solution but I noticed the implementation on the library is way simpler both for implementation code and the interface for developers. 

I also acknowledge the feature I implemented is an edge case or even out of scope for the library, so we can close this PR

mdelapenya:
> Hi @mdelapenya, I apologize for my late reply.
> 
> Basically I've already implemented the proposed solution but I noticed the implementation on the library is way simpler both for implementation code and the interface for developers.
> 
> I also acknowledge the feature I implemented is an edge case or even out of scope for the library, so we can close this PR

Thanks for your sensitivity, please feel free to send more contributions if you see them feasible 🙏  I'm closing this PR now

Cheers!

PR #1922, opened 2023-11-16
chore: define proper execution order for container lifecycle hooks

### LINKED ISSUE #1919: [Enhancement]: Properly define the execution order for container lifecycle hooks
### Proposal

At this moment, user-defined hooks are enqueued after the library-defined hooks (default logging, file copying). This is fine until we added the ability to add startup commands to the modules API, to run code right after the container is started.

It could be the case a user wants to add a startup command, and wait for something caused by that command, but because wait strategies are executed before the user-defined hooks (see https://github.com/testcontainers/testcontainers-go/blob/181897116febbf95410560096466daa0045df85e/docker.go#L1031-L1056) then it will be impossible for users to achieve that.

We propose define the following schema:

- make a distinction between user-defined hooks, and library hooks, where:
  - library hooks are mandatory (default logging, file copying)
  - user-defined hooks are optional, defined in the container request
  - **Pre** hooks: always run library hooks **before** user-defined hooks
  - **Post** hooks: always run library hooks **after** user-defined hooks

mdelapenya:
## What does this PR do?
This PR internally changes how the container lifecycle hooks are executed. Before these changes:

- default hooks (logging, config modifiers, copy files and wait strategies) were executed first
- user-defined hooks were executed after the default ones.

It could be OK, but it is not allowing users to perform tasks related to the warm-up of a container:

- create a dir so that copied files are able to live in a directory that the underlying Docker image does not have.
- execute code in a startup command, and wait for it as part of the wait strategies

With the changes in this PR, we are updating the execution order to:

1. execute default PRE hooks first
2. execute user-defined PRE hooks second
3. execute user-defined POST hooks first
4. execute default POST hooks second

With this order, we make sure that users can define their stuff at the right place in the container life cycle.

In order to achieve this order, we are combining default and user-defined hooks into one single hook, and that will be the one finally added to the container request. We have added unit tests to verify that the order is honoured.

To enrich the experience, we are defining a new container lifecycle event: **Ready**, which happens right after the PostStart event. This `Ready` event defines `PostReadies` hooks only. This new event will represent the execution of the wait strategies, which is the Testcontainers representation for Readiness.

Finally, given the PR touches the container lifecycles, we are refining the log messages in the default logging hooks, always using ✅ for the default post hooks.

## Why is it important?
Define a consistent container life cycl, allowing users to act before the default hooks.

## Related issues

- Closes #1919


pablochacin:
>execute default PRE hooks first
execute user-defined PRE hooks second
execute user-defined POST hooks first
execute default POST hooks second 
With this order, we make sure that users can define their stuff at the right place in the container life cycle.

I think this still doesn't make clear how this hook execution order related to the container's readiness. I just found it by looking at the code and realizing the `waitfor` is executed as a post-start hook.

mdelapenya:
> I suggest we make this more explicit in some way.

You mean in the docs and in this PR description? PTAL at https://deploy-preview-1922--testcontainers-go.netlify.app/features/creating_container/#lifecycle-hooks, the Info section

mdelapenya:
> I think this still doesn't make clear how this hook execution order related to the container's readiness.

In that sense I believe we need to always run the default wait strategies before the user-defined post-start hooks. Otherwise we can see a module trying to run commands even before the readyness (defined by the wait strategies) is not complete.

I'm tempted to always prepend the wait strategies to the post-start hooks. Wdyt? @kiview we have already discussed about this today

pablochacin:
>> I suggest we make this more explicit in some way.
>
>You mean in the docs and in this PR description? PTAL at https://deploy-preview-1922--testcontainers-go.netlify.app/features/creating_container/#lifecycle-hooks, the Info section

Documentation. In the proposed changes you mention this in a info note regarding the default hooks, but I think the definition of Post-start hook should be more specific, because "started" and "ready" (post wait condition meets) are in my opinion two different things and make create confusion.

mdelapenya:
@pablochacin I've updated the PR and description to reflect the new `Ready` lifecycle event, with its own Pre/Post hooks. PTAL 🙏

docs/features/creating_container.md
```diff
@@ -91,19 +91,32 @@ func TestIntegrationNginxLatestReturn(t *testing.T) {
...
+
+1. default `pre` hooks.
+2. user-defined `pre` hooks.
+3. user-defined `post` hooks.
+4. default `post` hooks.
+
+Inside each group, the hooks will be executed in the order they were defined.
+
+!!!info
+	The default hooks are for logging (applies to all hooks), customising the Docker config (applies to the pre-create hook) and copying files in to the container (applies to the post-create hook).
```
pablochacin, line 116: I'm not sure what this means: "applies to all hooks" or "applies to pre-create hook". 
Neither do I see why it is relevant to the developers.

docs/features/creating_container.md
```diff
@@ -91,19 +91,32 @@ func TestIntegrationNginxLatestReturn(t *testing.T) {
...
 * `PostCreates` - hooks that are executed after the container is created
 * `PreStarts` - hooks that are executed before the container is started
 * `PostStarts` - hooks that are executed after the container is started
+* `PostReadies` - hooks that are executed after the container is ready
 * `PreStops` - hooks that are executed before the container is stopped
 * `PostStops` - hooks that are executed after the container is stopped
 * `PreTerminates` - hooks that are executed before the container is terminated
 * `PostTerminates` - hooks that are executed after the container is terminated
 
+_Testcontainers for Go_ defines some default lifecycle hooks that cannot be skipped, but you can add your own hooks to the default ones. The execution order will be the following:
```
pablochacin, line 106:
```suggestion
_Testcontainers for Go_ defines some default lifecycle hooks that are executed in a specific order with respect to the user-defined hooks:
```

options.go
```diff
@@ -177,6 +177,28 @@ func WithStartupCommand(execs ...Executable) CustomizeRequestOption {
 	}
 }
 
+// WithAfterReadyCommand will execute the command representation of each Executable into the container.
+// It will leverage the container lifecycle hooks to call the command right after the container
+// is ready.
+func WithReadyCommand(execs ...Executable) CustomizeRequestOption {
```
pablochacin, line 183:
```suggestion
func WithAfterReadyCommand(execs ...Executable) CustomizeRequestOption {
```

options.go
```diff
@@ -177,6 +177,28 @@ func WithStartupCommand(execs ...Executable) CustomizeRequestOption {
 	}
 }
 
+// WithAfterReadyCommand will execute the command representation of each Executable into the container.
```
pablochacin, line 180: I prefer `WithAfterReadyCommand`
```suggestion
// WithReadyCommand will execute the command representation of each Executable into the container.
```

docs/features/creating_container.md
```diff
@@ -91,19 +91,32 @@ func TestIntegrationNginxLatestReturn(t *testing.T) {
...
-You'll be able to pass multiple lifecycle hooks at the `ContainerRequest` as an array of `testcontainers.ContainerLifecycleHooks`, which will be processed one by one in the order they are passed.
-
-The `testcontainers.ContainerLifecycleHooks` struct defines the following lifecycle hooks, each of them backed by an array of functions representing the hooks:
+You'll be able to pass multiple lifecycle hooks at the `ContainerRequest` as an array of `testcontainers.ContainerLifecycleHooks`. The `testcontainers.ContainerLifecycleHooks` struct defines the following lifecycle hooks, each of them backed by an array of functions representing the hooks:
 
 * `PreCreates` - hooks that are executed before the container is created
 * `PostCreates` - hooks that are executed after the container is created
 * `PreStarts` - hooks that are executed before the container is started
 * `PostStarts` - hooks that are executed after the container is started
+* `PostReadies` - hooks that are executed after the container is ready
```
pablochacin, line 100: I'm not sure about the name "PostReadies". Sound strange to me, but maybe just because I'm not a native English speaker.

mdelapenya: I'm following the same pattern, using the plural of the event removing the hook noun. So the plural of a Ready (as noun) is "Readies".

docs/features/creating_container.md
mdelapenya: I added where those hooks are applied, as documentation. I think it'd be clearer using passive style here:

```diff
- The default hooks are for logging (applies to all hooks), customising the Docker config (applies to the pre-create hook) and copying files in to the container (applies to the post-create hook).
+ The default hooks are for logging (applied on all hooks), customising the Docker config (applied on the pre-create hook) and copying files in to the container (applied to the post-create hook).
```

docs/features/creating_container.md
mdelapenya: Applied, thanks!

options.go
mdelapenya: Changed. I initially had doubts and opted for the shorter version. If it's not clear enough, let's be explicit

docs/features/common_functional_options.md
```diff
@@ -40,6 +40,19 @@ It also exports an `Executable` interface, defining the following methods:
...
+- Not available until the next release of testcontainers-go <a href="https://github.com/testcontainers/testcontainers-go"><span class="tc-version">:material-tag: main</span></a>
+
+Testcontainers exposes the `WithAfterReadyCommand(e ...Executable)` option to run arbitrary commands in the container right after it's ready, which happens when the defined wait strategies have finished with success.
+
+!!!info
+    To better understand how this feature works, please read the [Create containers: Lifecycle Hooks](/features/creating_container/#lifecycle-hooks) documentation.
+
+It leverages the `Executable` interface to represent the command and positional arguments to be executed in the container.
+
+You could use this feature to run a custom script, or to run a command that is not supported by the module right after the container is ready   .
```
mdelapenya, line 54:
```suggestion
You could use this feature to run a custom script, or to run a command that is not supported by the module right after the container is ready.
```

pablochacin (approved):
LGTM.

Issue #1921, opened 2023-11-15
[Bug]: kafka uses MaxInt64, prevents 32-bit usage


powersj:
### Testcontainers version

0.26.0

### Using the latest Testcontainers version?

Yes

### Host OS

linux

### Host arch

i386

### Go version

1.21.4

### Docker version

```shell
n/a
```

### Docker info

```shell
n/a
```

### What happened?

Running `go vet` with program that imports `github.com/testcontainers/testcontainers-go/modules/kafka` on 32-bit environment.

See: https://app.circleci.com/pipelines/github/influxdata/telegraf/18935/workflows/552535d3-8378-43eb-a37d-781756d50128/jobs/290322

### Relevant log output

```shell
# github.com/testcontainers/testcontainers-go/modules/kafka
/go/pkg/mod/github.com/testcontainers/testcontainers-go/modules/kafka@v0.26.0/kafka.go:55:72: cannot use math.MaxInt64 (untyped int constant 9223372036854775807) as int value in argument to fmt.Sprintf (overflows)
```

### Additional information

Could either switch to `math.MaxInt` or tell me no 32-bit support and we can use build tags.

Thanks!

PR #1923, opened 2023-11-16
fix(modules.kafka): Switch to MaxInt for 32-bit support

powersj:
## What does this PR do?

Avoid using MaxInt64, which only works on 64-bit systems.

## Why is it important?

Telegraf is a project that runs on a variety of architectures, and we do run our tests on a 32-bit system to ensure compatibility.

```s
❯ GOARCH=386 go vet ./modules/kafka/kafka.go 
# command-line-arguments
vet: modules/kafka/kafka.go:55:72: cannot use math.MaxInt64 (untyped int constant 9223372036854775807) as int value in argument to fmt.Sprintf (overflows)
```

## Related issues

fixes: #1921


mdelapenya (approved):
LGTM, thanks for the quick change!

PR #1927, opened 2023-11-28
fix: fallback matching of registry authentication config

sermio-te:
## What does this PR do?

This PR will fix an issue when config.json contains `auths` with keys set to the a registry URL such as `https://artifactory.com`. This way the authentication configuration for the registry will be matched with the configuration file and used.

## Why is it important?

In case where a registry is configured with an `https://` prefix, the authentication to pull images from that registry fails.

## How to test this PR
Manually add a registry and its authentication config in the `~/.docker/config.json` and run a test that requires authentication to pull images and start containers.

### DIFF
--- a/docker_auth_test.go
+++ b/docker_auth_test.go
@@ -30,14 +30,14 @@ func TestGetDockerConfig(t *testing.T) {
 	// Then, we can safely run the tests that rely on it.
 	defaultCfg, err := dockercfg.LoadDefaultConfig()
 	require.Nil(t, err)
-	require.NotNil(t, defaultCfg)
+	require.NotEmpty(t, defaultCfg)
 
 	t.Run("without DOCKER_CONFIG env var retrieves default", func(t *testing.T) {
 		t.Setenv("DOCKER_CONFIG", "")
 
 		cfg, err := getDockerConfig()
 		require.Nil(t, err)
-		require.NotNil(t, cfg)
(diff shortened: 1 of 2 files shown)


mdelapenya (commented):
Thank you for submitting this PR, could you please add tests demonstrating the bug? 🙏 

From your PR description I totally understand, but I'd like to have automated tests for the fix, thanks!

docker_auth_test.go
```diff
@@ -30,14 +30,14 @@ func TestGetDockerConfig(t *testing.T) {
 	// Then, we can safely run the tests that rely on it.
 	defaultCfg, err := dockercfg.LoadDefaultConfig()
 	require.Nil(t, err)
-	require.NotNil(t, defaultCfg)
+	require.NotEmpty(t, defaultCfg)
```
sermio-te, line 33: `NotEmpty` properly handles empty structs, but `NotNil` does not for the non-pointer types.

sermio-te:
> Thank you for submitting this PR, could you please add tests demonstrating the bug? 🙏
> 
> From your PR description I totally understand, but I'd like to have automated tests for the fix, thanks!

Thanks for you prompt response. I added unit tests to cover/verify the new code.

mdelapenya (approved):
LGTM! Thanks for this contribution, much appreciated!!

PR #1928, opened 2023-11-28
feat: support for custom registry prefixes at the configuration level

### LINKED ISSUE #1893: [Bug]: ImageSubstitutor not applied to ryuk image
### Testcontainers version

v0.26.0

Yes

### Host OS

Linux

### Host arch

x86

### Go version

1.21

### Docker version

```shell
24.0.7 (dind + client)
```

### Docker info

```shell
24.0.7
```

### What happened?

When I use [image name substitution](https://golang.testcontainers.org/features/image_name_substitution/), the substitutor will not be applied to the ryuk image. Hence, I see logs like

```
Waiting for container id 4887f0360761 image: docker.io/testcontainers/ryuk:0.5.1.
```

Now, this looks like a bug. In an environment that requires you to use a private registry, you need to be able to adapt ryuk, as well. Or am I missing something?

My test code looks like this:

```go
type MyImageSubstitutor struct{}

func (s MyImageSubstitutor) Description() string {
	return "MyImageSubstitutor: prepending my.private.registry.com/"
}

func (s MyImageSubstitutor) Substitute(image string) (string, error) {
        # these lines will never appear for the ryuk image, just for user-defined images
	fmt.Println("############################")
	fmt.Println(image)

	prefix := "my.private.registry.com/"
	if !strings.Contains(image, prefix) {
		return prefix + image, nil
	}

	return image, nil
}

        Image: "hashicorp/consul:latest," # this will be replaced, as expected
        ...
        ImageSubstitutors: []testcontainers.ImageSubstitutor{MyImageSubstitutor{}},
}
```

### Relevant log output

_No response_

### Additional information

_No response_

mdelapenya:
## What does this PR do?
This PR does two major things:

1. It tries to simplify how the reaper is internally configured, and for that:
  - deprecates the reaper options defined in the container and network requests. This is needed because we do not want to configure the reaper per container, but instead, do that in one single place: the testcontainers configuration
  - moving the constant representing the Ryuk version to internal, as we do not want users to modify it but instead make it possible to define a custom prefix (see later)
  - removing ocurrences of the deprecated code at the test level, which they won't do anything to not collide with the new features.
2. It adds support for configuring a Docker Hub prefix to be prepended to any Docker Hub image, with:
  - the `TESTCONTAINERS_HUB_IMAGE_NAME_PREFIX` env var, or
  - the `hub.image.name.prefix` property

_Testcontainers for Go_ will automatically apply the prefix to every image that it pulls from Docker Hub.

It will not apply the prefix to:
* non-Hub image names (e.g. where another registry is set)
* Docker Hub image names where the hub registry is explicitly part of the name (i.e. anything with a `docker.io` or `registry.hub.docker.com` host part)

The PR is adding a custom internal ImageSubstitutor, which is always appended to the user-defined image substitutor functions for a container request, so make sure it's always applied This substitutor will read the configuration to get the prefix, and will append it, when needed, to the image. Unit tests for it has been included.

The PR also improves the docs regarding image substitutors, listing all the images used in the tests.

Finally, the shell script to bump the reaper version has been updated, as the Go file for the Ryuk version has changed.

## Why is it important?
First, alignment with the Java behavior. Second, create a consistent experience while pulling images from custom registries.

## Related issues

- Relates #1719
- Closes #1893

### DIFF
--- a/container_test.go
+++ b/container_test.go
@@ -306,11 +306,13 @@ func Test_BuildImageWithContexts(t *testing.T) {
 
 func Test_GetLogsFromFailedContainer(t *testing.T) {
 	ctx := context.Background()
+	// directDockerHubReference {
 	req := ContainerRequest{
 		Image:      "docker.io/alpine",
 		Cmd:        []string{"echo", "-n", "I was not expecting this"},
 		WaitingFor: wait.ForLog("I was expecting this").WithStartupTimeout(5 * time.Second),
 	}
+	// }
 
 	c, err := GenericContainer(ctx, GenericContainerRequest{
 		ContainerRequest: req,
@@ -340,6 +342,7 @@ func Test_GetLogsFromFailedContainer(t *testing.T) {
 	}
 }
 
+// dockerImageSubstitutor {
 type dockerImageSubstitutor struct{}
 
 func (s dockerImageSubstitutor) Description() string {
@@ -349,6 +352,8 @@ func (s dockerImageSubstitutor) Description() string {
 func (s dockerImageSubstitutor) Substitute(image string) (string, error) {
 	return "docker.io/" + image, nil
 }
+
+// }
 
 // noopImageSubstitutor {
 type NoopImageSubstitutor struct{}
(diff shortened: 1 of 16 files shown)


container_test.go
```diff
@@ -506,3 +511,35 @@ func TestParseDockerIgnore(t *testing.T) {
...
 }
+
+func ExampleGenericContainer_withSubstitutors() {
+	ctx := context.Background()
+
+	// applyImageSubstitutors {
+	container, err := GenericContainer(ctx, GenericContainerRequest{
+		ContainerRequest: ContainerRequest{
+			Image:             "alpine:latest",
+			ImageSubstitutors: []ImageSubstitutor{dockerImageSubstitutor{}},
```
eddumelendez, line 522: I missed this in a previous PR, but wonder in which uses cases do we need more than one substitute. If a substitute is added then the new image name should be pull from the registry, in this case, what happens when more than one is added. If it is not intended then should we deprecate it in a next PR and only accept one?

eddumelendez (approved):
LGTM, just left a comment that we can address later.

container_test.go
mdelapenya: Mmmm I see your point, now with more perspective given the context from this PR. Yeah, makes sense. We just want one function plus the prefix prepender one, in the case it exists.

PR #1930, opened 2023-11-29
docs: include MongoDB's username and password options into the docs

mdelapenya:
- docs: include mongodb options into the docs
- chore: apply introduced-when string to all markdown files from the docs

## What does this PR do?
It adds the new two functional options for MongoDB to the module's docs.

It also modifies the shell script for the pre-release to locate all the markdown files to replace the introduced-at string.

## Why is it important?
We are seeing more places where the introduced-at string is required, so instead of applying to modules, we apply it to all the markdown files.

## Related issues

- Completes #1910

### DIFF
--- a/docs/modules/mongodb.md
+++ b/docs/modules/mongodb.md
@@ -37,8 +37,26 @@ When starting the MongoDB container, you can pass options in a variadic way to c
 
 #### Image
 
-If you need to set a different MongoDB Docker image, you can use `testcontainers.WithImage` with a valid Docker image
+If you need to set a different MongoDB Docker image, you can use `testcontainers.WithUsername` with a valid Docker image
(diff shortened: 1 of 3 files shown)


docs/modules/mongodb.md
```diff
@@ -37,9 +37,27 @@ When starting the MongoDB container, you can pass options in a variadic way to c
 
 #### Image
 
-If you need to set a different MongoDB Docker image, you can use `testcontainers.WithImage` with a valid Docker image
+If you need to set a different MongoDB Docker image, you can use `testcontainers.WithUsername` with a valid Docker image
```
mdelapenya, line 40:
```suggestion
If you need to set a different MongoDB Docker image, you can use `testcontainers.WithImage` with a valid Docker image
```

PR #1931, opened 2023-11-29
feat: support customizing the Docker build command

### LINKED ISSUE #1926: [Enhancement]: Add Target field to FromDockerfile struct
### Proposal

I've been trying to add `testcontainers` to my employer's CI, so my teammates can run integration tests. At some point, I needed to build a container image based off an existing `Dockerfile`, but the problem is that I need to specify a target stage to run the container in a test mode, where test dependencies are included.

I found out that [buildOptions](https://github.com/testcontainers/testcontainers-go/blob/480cadd123c9bd58506bbe34cf49592886d577f5/docker.go#L791C2-L791C14) is omitting the `Target` field, which is present in the [ImageBuildOptions](https://github.com/moby/moby/blob/2249db0c73433b5e3f34feabd9412e9948b955b7/api/types/client.go#L84) struct. Is there a reason to not expose this field?

mdelapenya:
- break: add BuildOptions method in order to support advanced customisations at build time
- chore: add a test for not finding the given target in the Dockerfile
- chore: make sure the auth configs are always added
- docs: document the advanced usage

## What does this PR do?
This PR adds a new field to the `FromDockerfile` struct, in order to hold a customizer function that will be executed when building a Docker image. This customizer will have access to the `ImageBuildOptions` Docker type, allowing users to build the image with as many configuration as they need. Of course, under their own risk.

We were forced to add a new method to the `ImageBuildInfo` interface, `BuildOptions`, in order to call the customizer defined by the user.

This customizer causes the deprecation of the `GetAuthConfigs` method from the same interface, although it probably needed to be deprecated before, when the code added support to discover the registry auths automatically. In any case, we are doing it so, creating a private method that automatically extracts the auths from the Dockerfile. These auths will be appended after the build options modifier is applied, to avoid undesired mistakes from the user.

The same applies for the Build args, the Image tag, the Dockerfile path and the build context, which are applied after the modifier for the same reasons.

## Why is it important?
Instead of adding fields for the build image options type, we prefer having a small reduced of fields/methods to it, and one single modifier function that allows the user to modify what they need. Of course, with power comes responsibility, so it's on their own risk to modify the build options with all the possible modifications.

Testcontainers for Go will keep the build args, the tag for the image, the Dockerfile and build context paths, automatically discovering the registry auths for the Dockerfile.

## Related issues

- Closes #1926
- Relates #869

### DIFF
--- a/container.go
+++ b/container.go
@@ -270,6 +270,7 @@ func (c *ContainerRequest) GetAuthConfigs() map[string]registry.AuthConfig {
 	return getAuthConfigsFromDockerfile(c)
 }
 
+// getAuthConfigsFromDockerfile returns the auth configs to be able to pull from an authenticated docker registry
 func getAuthConfigsFromDockerfile(c *ContainerRequest) map[string]registry.AuthConfig {
 	images, err := testcontainersdocker.ExtractImagesFromDockerfile(filepath.Join(c.Context, c.GetDockerfile()), c.GetBuildArgs())
 	if err != nil {
@@ -308,7 +309,6 @@ func (c *ContainerRequest) BuildOptions() (types.ImageBuildOptions, error) {
 	buildOptions := types.ImageBuildOptions{
 		BuildArgs:   c.GetBuildArgs(),
 		Dockerfile:  c.GetDockerfile(),
-		AuthConfigs: getAuthConfigsFromDockerfile(c),
 		Tags:        []string{fmt.Sprintf("%s:%s", c.GetRepo(), c.GetTag())},
 		Remove:      true,
 		ForceRemove: true,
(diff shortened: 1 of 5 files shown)


eddumelendez (approved):

danvergara:
Just for clarification. I can do this:

```go
				FromDockerfile: FromDockerfile{
					Context:       ".",
					PrintBuildLog: true,
					KeepImage:     false,
					BuildOptionsModifier: func(buildOptions *types.ImageBuildOptions) {
						buildOptions.Target = "development"
					},
				},
			},
			Started: true,
		})
``` 

and it'll immediately pick up the target stage, right? 

If so, that's awesome! That's a way more elegant solution than my proposal. I bit riskier as well.

mdelapenya:
> Just for clarification. I can do this:
> 
> ```go
> c, err := GenericContainer(ctx, GenericContainerRequest{
> 			ContainerRequest: ContainerRequest{
> 				FromDockerfile: FromDockerfile{
> 					Context:       ".",
> 					PrintBuildLog: true,
> 					KeepImage:     false,
> 					BuildOptionsModifier: func(buildOptions *types.ImageBuildOptions) {
> 						buildOptions.Target = "development"
> 					},
> 				},
> 			},
> 			Started: true,
> 		})
> ```
> 
> and it'll immediately pick up the target stage, right?
> 
> If so, that's awesome! That's a way more elegant solution than my proposal. I bit riskier as well.

Exactly, PTAL at the tests we added: https://github.com/testcontainers/testcontainers-go/pull/1931/files#diff-fb8725b2e45d4dd45709211d85444408a4af558390996f8a54c5806f61f35a5aR169-R171

And indeed riskier: we transfer the full control of the Docker type to the user, and if the Docker folks change that type, breaking changes are coming! But this will give more flexibility to our users, as I explained in the description.

Thanks for your feedback!

danvergara:
Awesome!!

Thanks @mdelapenya for the quick response and action!!!

PR #1911, opened 2023-11-11
fix: fix panic when getting container logs

pmalek:
## What does this PR do?

This PR fixes a `panic` when testcontainers cannot get container logs.

Appropriate tests have been added in `docker_test.go`

## Related issues

Should fix https://github.com/testcontainers/testcontainers-go/issues/946

## How to test this PR

```
go test -v -run TestLogs .
```

## Follow-ups

Consider changing the types used in `DockerContainer` as suggested in https://github.com/testcontainers/testcontainers-go/pull/1777#issuecomment-1803084148

Issue #525, opened 2022-09-15
[Enhancement]: improve the experience using log producers


mdelapenya:
### Proposal

When needing to check logs for a container, we have to:
- define a log producer
- implement the accept function

This is fairly simple, but not very developer friendly.

We could provide a default log producer that outputs container logs to console, which is enabled/disabled in the container request, default true.

Then we would be able to see container logs at the moment they are produced.

We could start with a simple `LogProducer bool`, or an array of log producers for the case we want to send the logs to any other place, i.e. to elasticsearch 🤷

mritunjaysharma394:
Is it still required yes, if yes then I would love to contribute to this, since I also needed it while debugging while I was adding k3s modules, I think this can be good feature to have

PR #1971, opened 2023-12-05
[breaking] Add err chan to log producer and don't panic on error

Tofel:
## What does this PR do?

As per [conversation](https://github.com/testcontainers/testcontainers-go/pull/1278#issuecomment-1784715136) I changed the way starting of log producer works. Now it won't panic if there's an error and instead send it over an error channel. When that happens producer state is cleaned, so that it can be started again. Callers can listen to to the error channel and if there's an error decide how to handle.

Also, it's now required to pass context timeout value, when starting log producer. I've added some boundaries there, it has to be within `<5,60>` seconds. If it's too low the http connection to get the logs is too short and some logs might get missed randomly (happened, when I was setting it to 1s), if it's too long the connection might timeout.

Now... I was thinking about not breaking the method signature and instead adding a new field to the `Container` struct that would store the timeout by default set to 5 seconds and then having a setter method to modify it. If you think that's better I could look into it.

Last, but not least. Tests are very simple ones. In our downstream project I have more complex ones which actually focus on testing a detached goroutine that's listening to errors and restarting log producer if it's needed. If current approach is accepted I could play around with Mockery and add some tests here as well.

## Why is it important?

We don't want unrecoverable panics if there's an error, when reading logs. It should be up to client to decide what should be done if there's an issue.

## Related issues

https://github.com/testcontainers/testcontainers-go/pull/1278
https://github.com/testcontainers/testcontainers-go/pull/1911

## How to test this PR

No straight-forward way. In our downstream project we are using only a limited set of methods from `Container` interface, so I narrowed it even further, put them in a separate interface and in tests created a mocked container with very stripped-down log-related logic that allowed me inject errors at various points. Without it, though... it's not trivial since you'd need to either disrupt/pause the network connection with Docker container so that reading logs fails due to timeout. Or use a bogus container that makes the connection timeout.


mdelapenya:
Before starting the review of this PR, I'd like to let you know that I'd like to plan a total refactor of the log producers feature: https://github.com/testcontainers/testcontainers-go/issues/525, so would like to know if we could work on this as a solid feature, instead of patching something that is not as usable as desired

Tofel:
@mdelapenya gotcha. Regarding what you wanna do there, we get the logs from log producer and then our log component has different log targets: in-memory, file, loki. It buffers all logs and then flushes then, when you want to.

Anyway, I will leave the PR open, because what you wanna do doesn't sound like something easy and fast to do :-)

docker.go
```diff
@@ -613,17 +616,39 @@ func (c *DockerContainer) CopyToContainer(ctx context.Context, fileContent []byt
 
 // StartLogProducer will start a concurrent process that will continuously read logs
 // from the container and will send them to each added LogConsumer
-func (c *DockerContainer) StartLogProducer(ctx context.Context) error {
-	if c.stopProducer != nil {
-		return errors.New("log producer already started")
+// timeout accepts values between 5 and 60s and will be used to set the context timeout
+// which means that each log-reading loop will last at least the specified timeout
+// and that it cannot be cancelled earlier
+func (c *DockerContainer) StartLogProducer(ctx context.Context, timeout time.Duration) error {
```
mdelapenya, line 622: Wdyt about this: we could probably create functional options for the StartLogProduce method:

```go
func (c *DockerContainer) StartLogProducer(ctx context.Context, opts ...LogProduceOptions) error
```

Being the timeout one of them. That would reduce the breaking change (BC) surface (only people assigning the function to a variable will get impacted). 

I'd try to avoid adding a BC here, but for this upcoming release we already have two (https://github.com/testcontainers/testcontainers-go/releases). So in the end I'm fine adding it if you consider it's fine too :)

mdelapenya (commented):
@Tofel I left a few comments before merging this. Please let me know what you think

docker.go
Tofel: Yeah, I was already thinking about that. I think it's a good idea. Applied.

docker.go
```diff
@@ -709,24 +729,31 @@ func (c *DockerContainer) StartLogProducer(ctx context.Context) error {
...
 		// block until the producer is actually done in order to avoid strange races
 		<-c.producerDone
 		c.stopProducer = nil
 		c.producerDone = nil
+		return <-c.producerError
 	}
 	return nil
 }
 
+func (c *DockerContainer) GetLogProducerErrorChannel() <-chan error {
```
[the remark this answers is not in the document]
Tofel: Well, that's the only way for the consumer to be able to listen to errors and react to them. I can point you to our code where we use it, although it might not be the clearest example as there we retry starting/listening in a detached goroutine. 

When I have a second I'd like to use Mockery or something similar to mock
```
r, err := c.provider.client.ContainerLogs(ctx, c.GetContainerID(), options)
```

so that we can cause timeouts or other read errors to make sure the channel works as expected and provide an example of how to use it. WDYT?

docker.go
```diff
@@ -613,17 +616,39 @@ func (c *DockerContainer) CopyToContainer(ctx context.Context, fileContent []byt
...
+	{
+		c.producerMutex.Lock()
+		defer c.producerMutex.Unlock()
+
+		if c.stopProducer != nil {
+			return errors.New("log producer already started")
+		}
+	}
+
+	if timeout < time.Duration(5*time.Second) || timeout > time.Duration(60*time.Second) {
```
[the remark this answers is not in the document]
Tofel: Why not! Done.

docker.go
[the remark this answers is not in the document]
mdelapenya: You already have a `c.Logs(ctx)` method in the container interface. Probably not the simplest thing to have your own mock for that, but it could work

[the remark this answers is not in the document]
mdelapenya: It's the other way around: `c.Logs` calls `c.provider.client.ContainerLogs`: https://github.com/testcontainers/testcontainers-go/blob/9c97abfa233364d792ff7723606bebc242237c19/docker.go#L323

Tofel:
@mdelapenya is there anything left for me to do before we can merge it? I checked the failure in the pipeline, but it doesn't seem related to my changes at all:
```
=== Failed
=== FAIL: . TestPrependHubRegistrySubstitutor/should_prepend_the_hub_registry_to_images_from_Docker_Hub/plain_image (0.00s)
    image_substitutors_test.go:27: expected my-registry/foo, got foo:latest

=== FAIL: . TestPrependHubRegistrySubstitutor/should_prepend_the_hub_registry_to_images_from_Docker_Hub (0.00s)

=== FAIL: . TestPrependHubRegistrySubstitutor (0.00s)
```

logconsumer_test.go
```diff
@@ -453,3 +453,65 @@ func TestContainerLogsShouldBeWithoutStreamHeader(t *testing.T) {
 	}
 	assert.Equal(t, "0", strings.TrimSpace(string(b)))
 }
+
+func Test_StartLogProducerStillStartssWithTooLowTimeout(t *testing.T) {
```
mdelapenya, line 457: Typo 🙏 
```suggestion
func Test_StartLogProducerStillStartsWithTooLowTimeout(t *testing.T) {
```

mdelapenya (dismissed):
LGTM, thanks!

I added a NIT comment regarding a test method name, but I will merge it as soon it's changed, without waiting for the CI

mdelapenya (approved):
LGTM! Thanks!

mdelapenya:
Hey @Tofel I think the PR is fine, although I see a consistent failure in the tests when Ryuk is disabled. Please see https://github.com/testcontainers/testcontainers-go/actions/runs/7299330173/job/19892388225?pr=1971 There are two containers leaked. I thought it was caused for something else, but I see this consistent error build after build.

Could you take a look please?

OTOH, I was planning to include this PR in the next release (today), but finding this makes me postpone it to the following one. Hope you understand.

Tofel:
@mdelapenya sure, totally understand, I'd do the same. Will have a look and fix, but probably won't be able to do it this year :/

mdelapenya:
> but probably won't be able to do it this year

Thankfully the rest of this year is really short 😂 . Thanks for understanding, see you next year 🎄

Tofel:
@mdelapenya should be fixed now, I didn't call `terminateContainerOnEnd()` in my tests, which made containers run indefinitely 

Happy new year 🥂

docker.go
```diff
@@ -710,24 +756,31 @@ func (c *DockerContainer) StartLogProducer(ctx context.Context) error {
...
 		// block until the producer is actually done in order to avoid strange races
 		<-c.producerDone
 		c.stopProducer = nil
 		c.producerDone = nil
+		return <-c.producerError
 	}
 	return nil
 }
 
+func (c *DockerContainer) GetLogProducerErrorChannel() <-chan error {
```
mdelapenya, line 780:
```suggestion
// GetLogProducerErrorChannel exposes the only way for the consumer to be able to listen to errors and react to them.
func (c *DockerContainer) GetLogProducerErrorChannel() <-chan error {
```

mdelapenya (commented):
@Tofel This LGTM, thanks for your time here 👏👏👏

I added a comment regarding the docs for the changes, in two places: the comments for the methods and our docs website. But I'd be fine with including the docs website in a separate PR if you prefer it.

Thanks!

Tofel:
@mdelapenya done! by website did you mean the markdown file you mentioned or something yet different?

docs/features/follow_logs.md
```diff
@@ -39,4 +39,58 @@ if err != nil {
...
+}
+
+// usage
+err := c.StartLogProducer(ctx, WithLogProducerTimeout(10*time.Second))
+if err != nil {
+	// do something with err
+}
+```
+
+If no parameter is passed a default time out of 5 seconds will be used. Values below 5 seconds and above 60 seconds will
```
mdelapenya, line 61:
```suggestion
If no parameter is passed a default timeout of 5 seconds will be used. Values below 5 seconds and above 60 seconds will
```

docs/features/follow_logs.md
```diff
@@ -39,4 +39,58 @@ if err != nil {
...
+
+When log producer fails to start within given timeout (causing a context deadline) or there's an error returned while closing the reader it will no longer panic, but instead will return an error over a channel. You can listen to it using `DockerContainer.GetLogProducerErrorChannel()` method:
+```golang
+func (c *DockerContainer) GetLogProducerErrorChannel() <-chan error {
+	return c.producerError
+}
+```
+
+This allows you to, for example, retry restarting log producer if it fails to start the first time. For example:
+```golang
```
mdelapenya, line 74:
```suggestion

```golang
```

docs/features/follow_logs.md
```diff
@@ -39,4 +39,58 @@ if err != nil {
...
+					startErr := container.StartLogProducer(ctx, timeout)
+					if startErr != nil {
+						return 
+					}
+			case <-done:
+				return
+			}
+		}
+	}(cons.logListeningDone, time.Duration(10*time.Second))
+	```
```
mdelapenya, line 96: I cannot suggest here to remove the leading whitespaces (Github UI failing this use case). Could you remove them for L96?

mdelapenya (commented):
Thanks for adding the docs, I think it's now clear how to use the new features. I added two comments regarding formatting in the docs. Once fixed, I think this is ready to be merged.

Tofel:
done!

docs/features/follow_logs.md
```diff
@@ -39,4 +39,59 @@ if err != nil {
...
+err := c.StartLogProducer(ctx, WithLogProducerTimeout(10*time.Second))
+if err != nil {
+	// do something with err
+}
+```
+
+If no parameter is passed a default timeout of 5 seconds will be used. Values below 5 seconds and above 60 seconds will
+be coerced to these boundary values.
+
+# Listening to errors
```
mdelapenya, line 64: Use H2 instead:
```suggestion
## Listening to errors
```

mdelapenya (approved):
LGTM, thanks for your work here!

Would you mind checking if the linked issues/PRs are potentially closed?

Cheers!

Tofel:
* https://github.com/testcontainers/testcontainers-go/pull/1278 -- I think we can close this one as it was retrying conecting again in a loop instead of panicking, but currently the same behaviour can be achieved using approach similar to what what I put in the example (listening in a detached goroutine to the error channels, sleeping for 1s if there's an error and starting log producer again)
* https://github.com/testcontainers/testcontainers-go/pull/1911 -- as well as instead of panicking it would go back to beginning immediately and start over 

@mdelapenya: do you have another release planned?

PR #1972, opened 2023-12-06
fix(modules.gcloud): pass as ptr to allow request customization

jboyens:
## What does this PR do?

Fixes the GCloud modules to correctly pass the `GenericContainerRequest` into `applyOptions` as a pointer instead of copying the struct.

Currently, the modifications are done, but are thrown away because `req` is passed as a copy.

## Why is it important?

Currently, any container customizations will completely be wiped out as the `req` param is copied into `applyOptions`

## Related issues

- None found.

## How to test this PR

1. Start on the current `main` branch
2. Make an image modification (using `WithImage()`) verify that the image is NOT modified
3. Apply this PR
4. Redo the test from step 2 and verify that the image is now correctly modified

### DIFF
--- a/modules/gcloud/bigquery.go
+++ b/modules/gcloud/bigquery.go
@@ -20,7 +20,7 @@ func RunBigQueryContainer(ctx context.Context, opts ...testcontainers.ContainerC
 		Started: true,
 	}
 
-	settings := applyOptions(req, opts)
+	settings := applyOptions(&req, opts)
 
 	req.Cmd = []string{"--project", settings.ProjectID}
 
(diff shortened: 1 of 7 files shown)


mdelapenya (commented):
Good catch! I'm so sorry for not having seen that before 🤦 

Could you add a test demonstrating the bug (which does exist but want to be covered against potential future regressions)?

mdelapenya (approved):
Don't mind, I'm going to merge this as is, as the existing tests are enough for the purpose.

Thanks for resolving this bug!

PR #1989, opened 2023-12-08
fix(modules.kafka): Use broker container IP instead of host IP for advertised broker listener

khartld:
## What does this PR do?

This PR changes the advertised broker ip from being the host ip to container ip. Port 9092 is not exposed to the host, so we can not connect to it via host IP, but need to use the broker container IP.
The current setup works with a normal desktop setup, because there the returned host IP is "localhost".

Alternatively, it should also be sufficient to hardcode the advertised broker as "localhost", because there is only 1 broker that needs to be able to connect to itself.

## Why is it important?

This makes it possible to use the Kafka module inside another Docker container or other setups where Host() does not return "localhost". This enables the use of this module in CI/CD pipelines that run in Docker.

Currently it leads to errors in the TransactionCoordinator:
`WARN [TransactionCoordinator id=1] Connection to node 1 (/172.17.0.1:9092) could not be established. Broker may not be available. (org.apache.kafka.clients.NetworkClient)`

## Related issues

- Closes #1786
This PR tackles the same problem, but using the container name instead of ip breaks the normal Desktop setup, as the system doesn't know the container name as a hostname.

## How to test this PR

After starting a Kafka container using the Kafka module, the KAFKA_ADVERTISED_LISTENERS should contain the kafka container ip.

E.g.
the /usr/sbin/testcontainers_start.sh script should contain the following value:

export KAFKA_ADVERTISED_LISTENERS=PLAINTEXT://localhost:51702,BROKER://172.17.0.6:9092


mdelapenya (dismissed):
LGTM! Although I'd like to have @eddumelendez 's eyes on this one too before merging.

Thanks for your time in any case, will merge it once the review and the CI pass.

modules/kafka/kafka.go
```diff
@@ -72,12 +72,17 @@ func RunContainer(ctx context.Context, opts ...testcontainers.ContainerCustomize
 							return err
 						}
 
+						ip, err := c.ContainerIP(ctx)
```
eddumelendez, line 75: [this message](https://github.com/testcontainers/testcontainers-go/pull/1786/files#r1367487506) is still valid in this PR and align implementations. It will make things easier to use the hostname rather the ip when getting access using a container network.

khartld: I tested your PR with the hostname and in my case it broke running the tests via `go test`. I'm not sure if there is some (easy) way for the host to resolve containers via hostname

eddumelendez: about the error in the description

> Currently it leads to errors in the TransactionCoordinator:
WARN [TransactionCoordinator id=1] Connection to node 1 (/172.17.0.1:9092) could not be established. Broker may not be available. (org.apache.kafka.clients.NetworkClient)

The client should connect to the mapped port in 9093 instead. I think I am missing something in your setup. Can you elaborate more, please?

khartld: The error is thrown from inside the kafka container. The TransactionCoordinator is a Kafka service running on the broker and it tries to reach the broker it is running on. It for some reason can't resolve the container name as hostname (this I also find a bit strange, because it should run in the docker network)

khartld:
@mdelapenya @eddumelendez
what do you think about merging this? :)

mdelapenya:
@eddumelendez, as our kafka expert, what are your thoughts on this?

container.go
```diff
@@ -62,6 +62,7 @@ type Container interface {
 	CopyFileToContainer(ctx context.Context, hostFilePath string, containerFilePath string, fileMode int64) error
 	CopyFileFromContainer(ctx context.Context, filePath string) (io.ReadCloser, error)
 	GetLogProductionErrorChannel() <-chan error
+	Hostname(ctx context.Context) (string, error)
```
mdelapenya, line 65: Introducing a new method in the interface could be seen as a breaking change, so not sure if we should do this here.

Of course, if it's needed it's needed

modules/kafka/kafka.go
khartld: nice, that works out :)

container.go
eddumelendez: I think we should just return the whole inspect struct 🤔

mdelapenya: Then we should rename the method to `Config`

mdelapenya:
@khartld I went ahead and rebased your PR with main, as we already merged #2534 

Please let us know what you think

Thanks!

khartld:
@mdelapenya  looks good, thanks :)

PR #982, opened 2023-03-24
feat: Auto detect the use of Podman from DOCKER_HOST

jtwatson:
## What does this PR do?

When running `testcontainers-go` with **Podman**, it seems that `ProviderType: ProviderPodman,` must **always** be specified.

The documentation [here](https://github.com/testcontainers/testcontainers-go/blame/main/docs/system_requirements/using_podman.md#L9) seems to contradict my experience.

Since **Podman 4** was released with a new network stack, it appears that the `default` network for **Podman** uses the `bridge` driver and is named `podman`.  As such, the library must be configured diffrently when using **Podman**.

This PR attempts to auto-detect when `ProviderType` has not been explicitly set and **Podman** is in use, using the following logic:

If `ProviderType` has not been set, and `DOCKER_HOST` contains `podman.sock`, then use `ProviderPodman`

## Why is it important?

When attempting to use `testcontainers-go` with **Podman**, it appears that it is always the case that 
```
	ProviderType: testcontainers.ProviderPodman,
```
must be set. When it is not set, the resulting error is quite difficult to troubleshoot. 

```
testcontainers.GenericContainer(): Error response from daemon: container create: unable to find network with name or ID bridge: network not found: creating reaper failed: failed to create container
```

If the library can reliably detect when Podman is in use, the developer experience would be much better.

## Related issues

- Closes #604

Issue #604, opened 2022-11-02
[Bug]: GenericContainer fails when used with podman, unless SkipReaper is true


awagner-iq:
### Testcontainers version

0.15.0

### Using the latest Testcontainers version?

Yes

### Host OS

Linux

### Host Arch

x86_64

### Go Version

1.19

### Docker version

```shell
mero@dibbler ~/tmp/x$ podman version
Version:      3.4.7
API Version:  3.4.7
Go Version:   go1.16.15
Built:        Thu Apr 21 19:38:09 2022
OS/Arch:      linux/amd64
```

### Docker info

```shell
mero@dibbler ~/tmp/x$ podman info   
host:
  arch: amd64
  buildahVersion: 1.23.1
  cgroupControllers:
  - cpu
  - io
  - memory
  - pids
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: conmon-2.0.32-1.fc34.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.32, commit: '
  cpus: 12
  distribution:
    distribution: fedora
    variant: workstation
    version: "34"
  eventLogger: journald
  hostname: dibbler
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 5.17.12-100.fc34.x86_64
  linkmode: dynamic
  logDriver: k8s-file
  memFree: 2583973888
  memTotal: 16543584256
  ociRuntime:
    name: crun
    package: crun-1.4.4-1.fc34.x86_64
    path: /usr/bin/crun
    version: |-
      crun version 1.4.4
      commit: 6521fcc5806f20f6187eb933f9f45130c86da230
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +CRIU +YAJL
  os: linux
  remoteSocket:
    exists: true
    path: /run/user/1000/podman/podman.sock
  security:
    apparmorEnabled: false
[...]

PR #1990, opened 2023-12-12
feat: allow explicitly specifying provider type

austince:
## What does this PR do?

Allows explicitly setting the provider type via an environment variable or the properties file in addition to the autodetection introduced in https://github.com/testcontainers/testcontainers-go/pull/982

## Why is it important?

Autodetection does not work in all scenarios, like when using the [Podman Desktop compatibility mode](https://podman-desktop.io/docs/migrating-from-docker/using-podman-mac-helper) since the docker host still points to the docker unix socket (which has been soft linked). 

## Related issues

Mentioned in https://github.com/testcontainers/testcontainers-go/issues/604#issuecomment-1843656397

### DIFF
--- a/docs/system_requirements/using_podman.md
+++ b/docs/system_requirements/using_podman.md
@@ -35,6 +35,15 @@ func TestSomething(t *testing.T) {
 }
 ```
 
+If using the [Podman Desktop Docker compatibility mode](https://podman-desktop.io/docs/migrating-from-docker/using-podman-mac-helper),
+where the `DOCKER_HOST` still points to the Docker socket,
+you can also configure the provider explicitly in a `.testcontainers.properties` file
+or the `TESTCONTAINERS_PROVIDER_TYPE` env variable.
+```properties
+provider.type=podman
+```
+
+
 The `ProviderPodman` configures the `DockerProvider` with the correct default network for Podman to ensure complex network scenarios are working as with Docker.
 
 ## Podman socket activation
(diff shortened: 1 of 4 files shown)


mdelapenya:
Hey @austince, as I commented in https://github.com/testcontainers/testcontainers-go/discussions/1991#discussioncomment-7887917, I think we should disregard the manual configuration of the provider but instead make the existing auto-discovery mechanism to detect the provider in a more robust manner.

If you agree, I'd close this PR in favour of another one that makes the mechanism more robust, deprecating the existing non-default providers. Wdyt?

austince:
> If you agree, I'd close this PR in favour of another one that makes the mechanism more robust, deprecating the existing non-default providers. Wdyt?

Would seem to make sense, though a proposal of how to do so would be good to understand what it would take\ -- what are you thinking for this one?

godart:
As mentiones in #2264: Maybe using '/info/' '/version' requests to determine if Podman is used.

```
~> curl  --unix-socket /var/run/docker.sock "http://v1.41/version"  -s | jq '.Components'
[
  {
    "Name": "Podman Engine",
    "Version": "5.0.1",
    "Details": {
         <... snip />
    }
  },
  {
    "Name": "Conmon",
     <... snip />
  },
  {
    "Name": "OCI Runtime (crun)",
     <... snip />
  }
]
```

mdelapenya:
I'm working on a `v1` release candidate (see https://github.com/testcontainers/testcontainers-go/tree/v1) and in there I'm removing the Provider abstraction. So I think we should close this PR.

mdelapenya:
Closing as per https://github.com/testcontainers/testcontainers-go/pull/1990#issuecomment-2162395284

Thank you so much!

PR #1993, opened 2023-12-14
feat: create networks with random names

### LINKED ISSUE #1992: [Enhancement]: create networks using random names
### Proposal

The networks API is fairly simple at the moment, but it's creating networks with a given name, which could not be the most valuable thing when running tests in parallel.

Ideally, the network name should be a random string, even thought it's an alias for Docker:

```go
// Network name is strictly a user-friendly alias to the network
// which is uniquely identified using ID.
// And there is no guaranteed way to check for duplicates.
```

mdelapenya:
- chore: handle error in test
- feat: add a first-order package for creating networks

## What does this PR do?
This PR creates a new `network` package that will allow Go users to use the `package.New` idiom when creating a Docker network with Testcontainers for Go.

This new package exposes functional options representing the former `NetworkRequest` struct, so that users of the old struct are still able to convert their code to the new format using this deprecation path: instead of creating a Network request, please pass functional options with the needed settings.

As a consequence, the old code to create a network has been deprecated, and will eventually removed in the next major version (v1).

To continue, it's important to mention that the new network will be created **with a random name**: instead of passing the name as before, we want names to be non-deterministic. And this comes with two benefits:

1. alignment with other Testcontainers libraries
2. users' code won't need to "remember" the network name, simply get the struct returned by the `New` function, and read the name from there.

Finally, the new `network` package will have this:

- the `New` function will return a struct, not an interface as in the previous `GenericNetwork`: we want to adhere to the Go idiom of "accept interface, return struct" so the consumer of the method can work with the expected, concrete, returned type, not dealing with an abstraction. In other words: producers should produce concrete types to consumers.
- it defines some functional options to configure the creation of the network, also exposing the `NetworkCustomizer` interface in the case consumers want to define their own options to create the network customising more Docker types. Implementing the `Customize` function will be enough.
- The `WithNetwork` functional option used in modules, not released yet, has been moved here and refactored a bit to allow receiving an already created Network struct, and use it in downstream containers.
- A `WithNewNetwork` functional option to be used in modules has been added. It will create a new network using the functions and options in the `network` package.

## Why is it important?
We want users to create networks with random names. We have stepped-up here to create a new package, which will be a PoC for an eventual v1 release, in which we could have first-order packages for the different entities we manage: networks, images, volumes and containers.

## Related issues

- Closes #1992

## How to test this PR

Please read the docs: https://deploy-preview-1993--testcontainers-go.netlify.app/features/creating_networks/ there we explain how the new APIs are used.

All the modules will receive two new options:
- https://deploy-preview-1993--testcontainers-go.netlify.app/modules/clickhouse/#withnetwork
- https://deploy-preview-1993--testcontainers-go.netlify.app/modules/clickhouse/#withnewnetwork


network/network.go
```diff
@@ -0,0 +1,122 @@
...
+// implement the new network APIs when they will be available.
+// By default, the network is created with the following options:
+// - Driver: bridge
+// - Attachable: true
+// - Labels: the Testcontainers for Go generic labels, to be managed by Ryuk. Please see the GenericLabels() function
+// And those options can be modified by the user, using the CreateModifier function field.
+func New(ctx context.Context, opts ...NetworkCustomizer) (*testcontainers.DockerNetwork, error) {
+	nc := types.NetworkCreate{
+		Driver:     "bridge",
+		Attachable: true,
```
eddumelendez, line 24: Can you elaborate more about the having it with `true` value, please? Looking at the docs and the java implementation, seems like is not needed.

Docs

> Globally scoped network is manually attachable by regular containers from workers in swarm mode.

network/network.go
```diff
@@ -0,0 +1,122 @@
...
+	//nolint:staticcheck
+	netReq := testcontainers.NetworkRequest{
+		Driver:         nc.Driver,
+		CheckDuplicate: nc.CheckDuplicate,
+		Internal:       nc.Internal,
+		EnableIPv6:     nc.EnableIPv6,
+		Name:           uuid.NewString(),
+		Labels:         nc.Labels,
+		Attachable:     nc.Attachable,
+		IPAM:           nc.IPAM,
```
eddumelendez, line 41: Looking at the networks configuration, `checkDuplicate` looks like something we can use in java and make it default in both. But, wonder if I missed something from the first implementation. The rest of them offer much flexibility when creating networks and, at least, in Java there is no request to add all those configurations.

network/network.go
mdelapenya: Mmmm I missed the official docs connect this option with Docker swarm. You're totally right and it's not needed it. Removing it

network/network.go
mdelapenya: The previous impl provided a way to modify any of those values, that's why I added a functional option for each of them.

eddumelendez: My suggestion is to start supporting the same as in Java and then see if support for the rest of them are needed. Maybe a networkModifier would be used for the rest of them.

mdelapenya: The thing is that in Go we already supported all those fields (there are more in the Docker type). That's why I kept them

network/network.go
```diff
@@ -0,0 +1,162 @@
...
+		if req.NetworkAliases == nil {
+			req.NetworkAliases = make(map[string][]string)
+		}
+		req.NetworkAliases[networkName] = []string{alias}
+	}
+}
+
+// WithNewNetwork creates a new network with random name and customizers, and attaches the container to it.
+// Finally it sets the network alias on that network to the given alias.
+func WithNewNetwork(ctx context.Context, alias string, opts ...NetworkCustomizer) testcontainers.CustomizeRequestOption {
```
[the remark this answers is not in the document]
mdelapenya: Done in https://github.com/testcontainers/testcontainers-go/pull/1993/commits/b02232c135129d367765930a5f8ed8c2bfa33cc5

PR #1994, opened 2023-12-15
feat: expose Redpanda's listener in the docker network

sago2k8:
Add a new `WithListener` function to allow adding new listeners to the advertised listeners so connection within the docker networks is possible

## What does this PR do? 

- Added a method to add new listener to the containers
- Validate the listener
- Iterate over the networks and add the new listeners as alias so it can be easily accessible.

## Why is it important?

Currently there is no way to access the Kafka Api from another container in the same docker network, at least not an obvious one, so as [Redpanda docs suggests](https://docs.redpanda.com/current/manage/security/listener-configuration/#advertise-a-listener) adding an extra advertise listener allows access to the Kafka endpoint from within the docker network so it can be consume from another test-container. 
## Related issues

-

### DIFF
--- a/modules/redpanda/mounts/redpanda.yaml.tpl
+++ b/modules/redpanda/mounts/redpanda.yaml.tpl
@@ -19,6 +19,13 @@ redpanda:
       port: 9093
       authentication_method: {{ if .KafkaAPI.EnableAuthorization }}sasl{{ else }}none{{ end }}
 
+  {{ range .KafkaAPI.Listeners }}
+    - address: 0.0.0.0
+      name: {{ .Address }}
+      port: {{ .Port }}
+      authentication_method: {{ .AuthenticationMethod }}
+  {{ end }}
+
   advertised_kafka_api:
     - address: {{ .KafkaAPI.AdvertisedHost }}
       name: external
@@ -26,6 +33,12 @@ redpanda:
     - address: 127.0.0.1
       name: internal
       port: 9093
+ {{ range .KafkaAPI.Listeners }} 
+    - address: {{ .Address }}
+      name: {{ .Address }}
+      port: {{ .Port }}
+  {{ end }} 
+  
 
 {{ if .EnableTLS }}
   admin_api_tls:
--- a/modules/redpanda/options.go
+++ b/modules/redpanda/options.go
@@ -1,6 +1,11 @@
 package redpanda
 
-import "github.com/testcontainers/testcontainers-go"
+import (
+	"net"
+	"strconv"
+
+	"github.com/testcontainers/testcontainers-go"
+)
 
 type options struct {
 	// Superusers is a list of service account names.
@@ -29,7 +34,12 @@ type options struct {
 
 	// EnableTLS is a flag to enable TLS.
 	EnableTLS bool
+
 	cert, key []byte
+
+	// Listeners is a list of custom listeners that can be provided to access the
+	// containers form within docker networks
+	Listeners []listener
 }
 
 func defaultOptions() options {
@@ -41,6 +51,7 @@ func defaultOptions() options {
 		ServiceAccounts:                    make(map[string]string, 0),
 		AutoCreateTopics:                   false,
 		EnableTLS:                          false,
+		Listeners:                          make([]listener, 0),
 	}
 }
 
@@ -86,6 +97,8 @@ func WithEnableKafkaAuthorization() Option {
 	}
 }
 
+// WithEnableSchemaRegistryHTTPBasicAuth enables HTTP basic authentication for
+// Schema Registry.
 func WithEnableSchemaRegistryHTTPBasicAuth() Option {
 	return func(o *options) {
 		o.SchemaRegistryAuthenticationMethod = "http_basic"
@@ -104,5 +117,29 @@ func WithTLS(cert, key []byte) Option {
 		o.EnableTLS = true
 		o.cert = cert
 		o.key = key
+	}
+}
+
+// WithListener adds a custom listener to the Redpanda containers. Listener
+// will be aliases to all networks, so they can be accessed from within docker
+// networks. At leas one network must be attached to the container, if not an
+// error will be thrown when starting the container.
+func WithListener(lis string) Option {
+	host, port, err := net.SplitHostPort(lis)
+	if err != nil {
+		return func(o *options) {}
+	}
+
+	portInt, err := strconv.Atoi(port)
+	if err != nil {
+		return func(o *options) {}
+	}
+
+	return func(o *options) {
+		o.Listeners = append(o.Listeners, listener{
+			Address:              host,
+			Port:                 portInt,
+			AuthenticationMethod: o.KafkaAuthenticationMethod,
+		})
 	}
 }
--- a/modules/redpanda/redpanda.go
+++ b/modules/redpanda/redpanda.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	_ "embed"
 	"fmt"
+	"math"
 	"os"
 	"path/filepath"
 	"text/template"
@@ -31,6 +32,7 @@ const (
 	defaultKafkaAPIPort       = "9092/tcp"
 	defaultAdminAPIPort       = "9644/tcp"
 	defaultSchemaRegistryPort = "8081/tcp"
+	defaultDockerKafkaApiPort = "29092"
 
 	redpandaDir         = "/etc/redpanda"
 	entrypointFile      = "/entrypoint-tc.sh"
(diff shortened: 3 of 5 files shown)


modules/redpanda/mounts/redpanda.yaml.tpl
```diff
@@ -10,6 +10,13 @@ redpanda:
       port: 9092
       authentication_method: {{ .KafkaAPI.AuthenticationMethod }}
 
+    # This listener is required to enable connection from within the Docker
+    # network
+    - address: 0.0.0.0
+      name: docker
+      port: 29092
+      authentication_method: {{ .KafkaAPI.AuthenticationMethod }}
```
eddumelendez, line 18: In java, [adding multiple listeners is allowed](https://github.com/testcontainers/testcontainers-java/blob/main/modules/redpanda/src/main/resources/testcontainers/redpanda.yaml.ftl#L24-L29) because it can be shared through the docker network and via a proxy such as toxiproxy. I think the same approach can be offered in go. For the test, look at this [one](https://github.com/testcontainers/testcontainers-java/blob/main/modules/redpanda/src/test/java/org/testcontainers/redpanda/RedpandaContainerTest.java#L114-L142) using kcat container.

eddumelendez (commented):
Thanks for your contribution! I've left a suggestion regarding the implementation and how to test it.

modules/redpanda/mounts/redpanda.yaml.tpl
sago2k8: @eddumelendez I have update the PR to be more aligned with the implementation you linked, mind to check it out ?, thanks

modules/redpanda/options.go
```diff
@@ -106,3 +119,28 @@ func WithTLS(cert, key []byte) Option {
 		o.key = key
 	}
 }
+
+// WithListener adds a custom listener to the Redpanda containers. Listener
+// will be aliases to all networks, so they can be accessed from within docker
+// networks
+func WithListener(lis string) Option {
+	host, port, err := net.SplitHostPort(lis)
+
```
mdelapenya, line 128:
```suggestion
```

modules/redpanda/redpanda.go
```diff
@@ -252,6 +260,29 @@ func renderBootstrapConfig(settings options) ([]byte, error) {
 	return bootstrapConfig.Bytes(), nil
 }
 
+// validateListeners validates that the provided listeners are valid and set network alieases for the provided address.
```
mdelapenya, line 263:
```suggestion
// validateListeners validates that the provided listeners are valid and set network aliases for the provided address.
```

modules/redpanda/redpanda.go
```diff
@@ -98,10 +100,15 @@ func RunContainer(ctx context.Context, opts ...testcontainers.ContainerCustomize
...
 
+	if err := validateListeners(ctx, settings, req); err != nil {
+		return nil, fmt.Errorf("failed validate listeners: %w", err)
+	}
+
 	// Bootstrap config file contains cluster configurations which will only be considered
 	// the very first time you start a cluster.
 	bootstrapConfigPath := filepath.Join(tmpDir, bootstrapConfigFile)
 	bootstrapConfig, err := renderBootstrapConfig(settings)
+
```
mdelapenya, line 111: Let's have errors next to their evaluation
```suggestion
```

mdelapenya (changes requested):
Hi @sago2k8, thanks for submitting this feature. It LGTM, although I've added a few comments regarding format (errors and its evaluation) and comments.

Regarding documentation, I'd like to see this new functional option added to the docs here: `docs/modules/redpanda.md`. Please take a look at the `codeinclude` blocks in order to embed existing code snippets into the docs.

Once this is added, I think we can merge.

Cheers!

sago2k8:
> Hi @sago2k8, thanks for submitting this feature. It LGTM, although I've added a few comments regarding format (errors and its evaluation) and comments (not blockers).
> 
> Regarding documentation, I'd like to see this new functional option added to the docs here: `docs/modules/redpanda.md`. Please take a look at the `codeinclude` blocks in order to embed existing code snippets into the docs.
> 
> Once this is added, I think we can merge.
> 
> Cheers!

Thanks for the review @mdelapenya, I think I addressed all your comments and added the documentation. could you give it another look please :)

modules/redpanda/redpanda_test.go
```diff
@@ -278,6 +293,117 @@ func TestRedpandaWithTLS(t *testing.T) {
 	require.Error(t, results.FirstErr(), kerr.UnknownTopicOrPartition)
 }
 
+func TestRedpandaListener_Simple(t *testing.T) {
+	ctx := context.Background()
+
+  	// 1. Create network 
+	RPNetwork, err := network.New(ctx, network.WithCheckDuplicate())
```
mdelapenya, line 300: Nit: let's use lowercase here. Please apply it everywhere
```suggestion
	rpNetwork, err := network.New(ctx, network.WithCheckDuplicate())
```

modules/redpanda/redpanda_test.go
```diff
@@ -278,6 +293,117 @@ func TestRedpandaWithTLS(t *testing.T) {
...
+  	// withListenerExec {
+	_, _, err = kcat.Exec(ctx, []string{"kcat", "-b", "redpanda:29092", "-t", "msgs", "-P", "-l", "/tmp/msgs.txt"})
+  	// }
+
+	require.NoError(t, err)
+
+	// 6. Consume message from Redpanda
+	_, stdout, err := kcat.Exec(ctx, []string{"kcat", "-b", "redpanda:29092", "-C", "-t", "msgs", "-c", "1"})
+	require.NoError(t, err)
+	out, err := io.ReadAll(stdout)
```
mdelapenya, line 349: Nit: separate concerns
```suggestion

	out, err := io.ReadAll(stdout)
```

modules/redpanda/redpanda_test.go
```diff
@@ -5,19 +5,34 @@ import (
...
+type TestContainersLogger struct {
+	// LogPrefix is a string that is prefixed for every individual log line.
+	LogPrefix string
+}
+
+// Accept prints the log to stdout
+func (lc TestContainersLogger) Accept(l testcontainers.Log) {
+	fmt.Print(lc.LogPrefix + string(l.Content))
+}
+
```
mdelapenya, line 35: Is this used elsewhere?

modules/redpanda/redpanda.go
```diff
@@ -252,6 +261,29 @@ func renderBootstrapConfig(settings options) ([]byte, error) {
 	return bootstrapConfig.Bytes(), nil
 }
 
+// validateListeners validates that the provided listeners are valid and set network aliases for the provided address.
+// The container must be attached to at least one network.
+func validateListeners(ctx context.Context, settings options, req testcontainers.GenericContainerRequest) error {
```
mdelapenya, line 266: Not a blocker:

I think this function it's correct, although for me it's doing two things in one: validate the listeners and define the network aliases.

Given the function name is `validate`, I could expect that the second part for setting the networks aliases is applied elsewhere.

I'm not sure how to improve this: a new name for the function (applyListeners??). PTAL at the comment in https://github.com/testcontainers/testcontainers-go/pull/1994/files#diff-ea8e8772fe279375ed6f3310607a32f3a18780f6ebfe9b3f817699b6afbdda4cR24-R35:

> // 4. Validate listeners and add network aliases to every network... 

If we rename the function, I'd change this too to

> // 4. Apply listeners to network aliases if they are valid... 

Wdyt?

mdelapenya (commented):
I think we are almost there, please check my comments 🙏

modules/redpanda/redpanda_test.go
sago2k8: Good catch, this is a leftover I used for debugging, gonna remove it.

modules/redpanda/redpanda.go
sago2k8: I agree with the new name

mdelapenya: I like it, thanks for understanding

sago2k8:
Just noticed there was a merge conflict in one of my commits, addressed now :)

mdelapenya (approved):
This LGTM, thanks for your time!

@eddumelendez if you see anything weird, please let's do it in a follow up PR 🙏

modules/redpanda/redpanda.go
```diff
@@ -252,6 +260,29 @@ func renderBootstrapConfig(settings options) ([]byte, error) {
...
+	}
+
+	for _, listener := range settings.Listeners {
+		if listener.Port < 0 || listener.Port > math.MaxUint16 {
+			return fmt.Errorf("invalid port on listener %s:%d (must be between 0 and 65535)", listener.Address, listener.Port)
+		}
+
+		for _, network := range req.Networks {
+			req.NetworkAliases[network] = append(req.NetworkAliases[network], listener.Address)
+		}
```
eddumelendez, line 281: one issue with this approach that Java implementation still have is that when adding Toxiproxy upfront Kafka container and using `0.0.0.0:<random-toxiproxy-port>` as a listener the registration will fail because will try to add 0.0.0.0 as an alias. I forgot to mention that when suggesting the approach initially.

mdelapenya: @sago2k8 @eddumelendez would you like to collaborate in a follow-up PR for this?

sago2k8: Hey, thanks for your comments, yes I could work on it, do you happen to have a suggestion of a expected behaviour for that specific case

PR #1995, opened 2023-12-15
feat(modules.clickhouse): Add zookeeper for clickhouse clusterization

laskoviymishka:
If you would like to perform queries against clusterized ClickHouse you need to start zookeeper near your cluster and pass link to it via `config.xml`

API example:

```
	zkhost, zkport, err := RunZookeeper(ctx)
	if err != nil {
		t.Fatal(err)
	}

	container, err := RunContainer(ctx,
		WithUsername(user),
		WithPassword(password),
		WithDatabase(dbname),
		WithZookeeper(zkhost, zkport),
	)
	if err != nil {
		t.Fatal(err)
	}
```

## What does this PR do?

Add option: `WithZookeeper` to clickhouse module.

## Why is it important?

ClickHouse queries (especially DDLs) vary a lot between single node and cluster. Having it a clusterized allows you to apply exactly same code to real prod clusters and in your tests.


modules/clickhouse/zookeeper.go
```diff
@@ -0,0 +1,129 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	ipaddr, err := zkcontainer.ContainerIP(ctx)
+	if err != nil {
+		return nil, err
+	}
+	return &ZookeeperContainer{Container: zkcontainer, exposedPort: zkExposedPort, ipaddr: ipaddr}, nil
+}
```
eddumelendez, line 104: I don't think the clickhouse module should have a zookeeper implementation but rather just receive the connection string to be used in clickhouse

modules/clickhouse/zookeeper.go
```diff
@@ -0,0 +1,129 @@
...
+        <replica>replica</replica>
+    </macros>
+
+    <distributed_ddl>
+        <path>/clickhouse/task_queue/ddl</path>
+    </distributed_ddl>
+
+    <format_schema_path>/var/lib/clickhouse/format_schemas/</format_schema_path>
+</clickhouse>
+`, c.IP())
```
eddumelendez, line 78: this would be a limitation for those who need other configurations.

eddumelendez (commented):
Thanks for your contribution. I've left some comments but in general I would rather prefer to add an example about how to do this instead of hardcoding the configuration file. Also, wonder if it is possible to configure the zookeeper section to do not assume other config values. If so, then using templates would be great, redpanda module is doing so [here](https://github.com/testcontainers/testcontainers-go/tree/main/modules/redpanda/mounts).

With Testcontainers, containers can be composed programmatically and with the upcoming network support in https://github.com/testcontainers/testcontainers-go/pull/1993 modules would improve this experience. That's why I don't see having zookeeper as part of clickhouse.

modules/clickhouse/zookeeper.go
laskoviymishka: I can move zk container into a separate module, but it's like 20 lines of code, not sure regards it worth a hassle.

laskoviymishka: I can move this code just in examples

modules/clickhouse/zookeeper.go
laskoviymishka: actually, it's possible - https://clickhouse.com/docs/en/operations/configuration-files
I will put this zookeeper config in a separate file.

modules/clickhouse/mounts/zk_config.xml.tpl
```diff
@@ -0,0 +1,31 @@
...
+        <cluster>default</cluster>
+        <shard>shard</shard>
+        <replica>replica</replica>
+    </macros>
+
+    <distributed_ddl>
+        <path>/clickhouse/task_queue/ddl</path>
+    </distributed_ddl>
+
+    <format_schema_path>/var/lib/clickhouse/format_schemas/</format_schema_path>
```
eddumelendez, line 30: are those default values? Can we just keep the zookeeper section?

modules/clickhouse/clickhouse.go
```diff
@@ -72,6 +81,53 @@ func (c *ClickHouseContainer) ConnectionString(ctx context.Context, args ...stri
...
+// ZookeeperOptions arguments for zookeeper in clickhouse
+type ZookeeperOptions struct {
+	Host, Port string
+}
+
+// renderZookeeperConfig generate default zookeeper configuration for clickhouse
+func renderZookeeperConfig(settings ZookeeperOptions) ([]byte, error) {
+	tpl, err := template.New("bootstrap.yaml").Parse(zookeeperConfigTpl)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse redpanda config file template: %w", err)
```
eddumelendez, line 93:
```suggestion
		return nil, fmt.Errorf("failed to parse zookeeper config file template: %w", err)
```

modules/clickhouse/clickhouse.go
```diff
@@ -72,6 +81,53 @@ func (c *ClickHouseContainer) ConnectionString(ctx context.Context, args ...stri
...
+	}
+
+	var bootstrapConfig bytes.Buffer
+	if err := tpl.Execute(&bootstrapConfig, settings); err != nil {
+		return nil, fmt.Errorf("failed to render redpanda bootstrap config template: %w", err)
```
eddumelendez, line 98:
```suggestion
		return nil, fmt.Errorf("failed to render zookeeper bootstrap config template: %w", err)
```

docs/modules/clickhouse.md
```diff
@@ -73,6 +73,15 @@ initialization before starting the service.
 [Init script content](../../modules/clickhouse/testdata/init-db.sh)
 <!--/codeinclude-->
 
+#### Zookeeper
+
+If you would like to perform queries against clusterized ClickHouse you need to start zookeeper near your cluster and pass link to it via `config.xml`
```
eddumelendez, line 78:
```suggestion
Clusterized ClickHouse requires to start Zookeeper and pass link to it via `config.xml`.
```

modules/clickhouse/mounts/zk_config.xml.tpl
laskoviymishka: all this is a part of zookeeper default configuration, without it cluster still considered as not replicated one.

laskoviymishka:
@eddumelendez gentle ping on this one <3

eddumelendez (dismissed):
LGTM

modules/clickhouse/clickhouse.go
```diff
@@ -72,6 +79,53 @@ func (c *ClickHouseContainer) ConnectionString(ctx context.Context, args ...stri
...
+
+	return bootstrapConfig.Bytes(), nil
+}
+
+// WithZookeeper pass a config to connect clickhouse with zookeeper and make clickhouse as cluster
+func WithZookeeper(host, port string) testcontainers.CustomizeRequestOption {
+	return func(req *testcontainers.GenericContainerRequest) {
+		f, err := os.CreateTemp("", "clickhouse-tc-config-")
+		if err != nil {
+			panic(err)
```
mdelapenya, line 107: I think we should document that this function could panic under this condition. Would you mind adding it to the docs?

modules/clickhouse/clickhouse.go
```diff
@@ -72,6 +79,53 @@ func (c *ClickHouseContainer) ConnectionString(ctx context.Context, args ...stri
...
+
+		defer f.Close()
+
+		// write data to the temporary file
+		data, err := renderZookeeperConfig(ZookeeperOptions{Host: host, Port: port})
+		if err != nil {
+			panic(err)
+		}
+		if _, err := f.Write(data); err != nil {
+			panic(err)
```
mdelapenya, line 118: Same here

mdelapenya (commented):
LGTM, will approve it once the comments for the docs are addressed.

Thank you so much for your work here! And @eddumelendez, thanks for the review too!

modules/clickhouse/clickhouse.go
mdelapenya: I went ahead and added the docs 🙏  Once the CI passes, I think we can merge this one. Thanks!

modules/clickhouse/clickhouse_test.go
```diff
@@ -219,6 +221,105 @@ func TestClickHouseWithConfigFile(t *testing.T) {
...
+	ipaddr, err := zkcontainer.ContainerIP(ctx)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	container, err := RunContainer(ctx,
+		WithUsername(user),
+		WithPassword(password),
+		WithDatabase(dbname),
+		WithZookeeper(ipaddr, zkPort.Port()),
```
mdelapenya, line 251: I'm about to merge this one, but @eddumelendez: how would you see if this `WithZookeeper` option automatically starts the container for you? Would this be overkill? Or is it preferred this way, that allows setting up an external zookeeper passing the IP/port to the clickhouse container? For the "embedded" one, it can be done as a follow-up PR

mdelapenya: As discussed internally, having the `WithZookeper` options could be implemented, but it's not a priority at all, as the supporting containers (the Zookeeper) can be managed from outside the module.

mdelapenya:
Thanks @laskoviymishka for submitting this new feature! Much appreciated 🙇

PR #2035, opened 2023-12-19
[WiP] feat: Add DependsOn support

### LINKED ISSUE #1791: [Enhancement]: dependsOn() implementation
### Proposal

As a test developer, I would be interested to provision a cluster of services depending on each other. For that the Java implementation offers the `dependsOn()` implementation: https://java.testcontainers.org/features/startup_and_waits/#depending-on-another-container . It would be nice to have a similar feature in Testcontainers for Go.

**Why not Docker Compose?**. As discussed on Slack, at the moment Docker Compose in Testcontainers for Go is considered rather temporary and not stable. So it is better. My experiments in [Testcontainers for C/C++](https://github.com/testcontainers/testcontainers-c) show that it rather does not work at all with recent API versions. So it is preferred to have a native implementation instead.

## References

- [Slack convo](https://testcontainers.slack.com/archives/CBWBNR76G/p1697780785263819)

Mathew-Estafanous:
## What does this PR do?
This PR adds a `DependsOn` field to `ContainerRequest` which can be used to supply a list of containers that must be running prior to starting the current container. On startup of the container, all containers it depends on will be started (if they are not already running).

Dependencies are started within the `PreStarts` container hook and respects waiting strategies of the dependency. As such, end-users can deem when a dependant container is 'ready' (ie. health, log, exec...).

While likely uncommon, dependencies can be chained together in the form of sub-dependencies.
```go
containerA, err := GenericContainer(ctx, .........)

containerB, err := GenericContainer(ctx, GenericContainerRequest{
		// other fields
		DependsOn:    []Container{containerA},
	},
	Started: false
})
crC := ContainerRequest{
        // other fields
	DependsOn:    []Container{containerB},
},
```
In the example above, container C depends on B which then depends on A.

## Why is it important?
* This feature would allow for **explicit** descriptions of container dependencies in cases where a cluster of containers are required and depend on each other.
* Supply feature parity between testcontainer-java's own [dependsOn](https://java.testcontainers.org/features/startup_and_waits/#depending-on-another-container) implementation.

## Related issues

- Closes #1791 

## How to test this PR
A test suite for this new field accompanies the PR. See [Test_ContainerDependsOn](https://github.com/Mathew-Estafanous/testcontainers-go/blob/ffac9c970b7b24e05dd073ce013769ec51c763aa/container_test.go#L307)


mdelapenya:
Hey @Mathew-Estafanous thanks for submitting this PR. Wdyt if we first create a GH discussion about the design of this feature? As an example, I had in mind having a way to detect cycles in the dependsOn: i.e. srv1 depends on srv2, but what happens if srv2 depends on srv1 (users doing crazy things?) how would we handle that error?

In that sense, feel free to create the discussion so we can elaborate a great design. Please take a look at https://github.com/testcontainers/testcontainers-go/discussions/559 as an example

Mathew-Estafanous:
Thanks for reviewing! I appreciate it.

> @ preventing circular dependencies, sounds reasonable but not sure it should be a blocker, can always followup with such in a different PR
> 
> circular dependencies are inherently un-resolvable so imho it wouldn't change the contract API
Come to think of it, given the current API design, I'm not sure how someone could create a circular dependency. You'd already need to have created both containers. Am I missing something? How could someone define a circular dependency?

Mathew-Estafanous:
> ....if they're knowledgeable about the semantics required to do so they should be knowledgeable enough to realize it probably wouldn't work as expected.

Agreed, it would be quite the feat to shoe-horn such a blatant anti-pattern in the current design.

But you do bring up a good point on maybe generalizing the api to take in something less concrete than a Container. I think referencing just a name/container ID might not be enough information. We need to respect each container's `wait.strategy` and I'm not aware of a way to get that with just a name as the reference.

> Looks like all their dependsOn are of type [Startable](https://github.com/testcontainers/testcontainers-java/blob/36c872749862918ddbf449f224aeadd8c87c0261/core/src/test/java/org/testcontainers/junit/DependenciesTest.java#L29) which aligns with a container instance and not containerrequest instance.

For sure. I'm still trying to stick with Go idioms as opposed to copy-pasting the Java design where it won't fit.

Mathew-Estafanous:
> Perhaps we can make the type of DependsOn a custom type/interface (pseudo-type union) we could then implement it as-is and extend in the future?

Not sure what you mean by make `DependsOn` a custom type. Are you referring to having a custom interface (similar to `Startable` in the java impl)?

mdelapenya:
Hi folks, sorry for the radio silence during this month but Xmas holidays and company trips made it difficult to me to jump in without all the context.

First thank you all for the great community work you've been doing here with your proposals 🙇 I've enjoyed a lot seeing you collaborating here.

And now to the topic. The library defines two different concepts for a container:
- a ContainerRequest struct, which represents the definition for the future container, used by developers to define and configure the containers they need. This struct will be used to start a container.
- a Container, an interface representing an already created/started/stopped container, created from a ContainerRequest struct.

So, coming back to the `DependsOn` thing, when defining my container, I'd like to declare all the containers I need to be running before a given `containerA` is started. But because we are declaring `containerA` with a ContainerRequest_A, as a developer, I would have two options:

1. define and create the containers that are needed by `containerA`, but without starting them (not passing the `Started: true` field to the container request)
2. define multiple container requests and set them as dependencies for `containerA` at this declaration level.

The 1) option would be simpler, but less developer friendly, as I'd be forcing developers to not start the supporting containers; then the library, using the lifecycle hooks, would start them before containerA.

The 2) option seems more developer friendly, as I'd declare all the container requests, would set the DependsOn dependencies, and would start the containerA. Then the library, using the lifecycle hooks, would start all the dependencies.

I see this PR more aligned with option 1, which fairly works, although I see it less easy to use. At least at the moment.

For options 2, we would need to depend on another container request, but still relying on the users not calling the GenericContainer(...) method on the dependent containers before calling `containerA`.

So, what would you prefer?

Mathew-Estafanous:
I agree that the implemented solution isn't as easy to use and I think it would be be better to work on a solution similar to your proposed 2nd solution.

However, I'm concerned that with the use of `ContainerRequest` we would be left without a `Container` for each of the dependant containers.  `GenericContainer(..)` would only return the 'parent' container, leaving out the other containers. If maintaining access to dependant to containers isn't necessary, then option 2 would suffice, I believe.

We could introduce a `Dependency` struct as a wrapper for `ContainerRequest`s with a callback function that is triggered when the container is started.
```go
type Dependency struct {
	ContainerRequest ContainerRequest
	OnStarted func(Container) error
}
```
It's not the most elegant solution, but it offers a workaround in gaining access to the `Container` of dependant containers.

Alternatively, (if not already possible) we could consider introducing a 'find' container mechanism, possibly similar to the internal `findContainerByName(..)` method. 

What are your thoughts?

Mathew-Estafanous:
Hey @mdelapenya 

I just want to bump this conversation. Have you had time to review my comments? I understand if this on hold due to scheduling. Thanks!

pablochacin:
Hi. As a potential user of this feature, first at all thanks @Mathew-Estafanous for moving this forward.

I tend to agree with @mdelapenya that using containers is not natural in this context, mostly because the feature will automatically start dependencies if they are not already started. 

> What are your thoughts? Is having access to dependant containers necessary?

I think in general it is needed. For instance, an application (container `App` ) needs the IP addresses of the databases (MySQL, Redis)  to setup db connections. 

What I don't see is how this could work, as we need this address before starting the container `App`, to pass it as Environment variables for example.

The way Docker compose solves this is by injecting the dependencies as environment variables. In my example above, the `App` container would have environment variables with the addresses of the `Mysql` and `Redis` containers. 

> Alternatively, (if not already possible) we could consider introducing a 'find' container mechanism, possibly similar to the internal findContainerByName(..) method

I think this could work better, but still, I don't see how this would work in the case I described above,

Mathew-Estafanous:
Hey @pablochacin, thanks for your input! And sorry for not getting back earlier.

> What I don't see is how this could work, as we need this address before starting the container `App`, to pass it as Environment variables for example.
> 
> The way Docker compose solves this is by injecting the dependencies as environment variables. In my example above, the `App` container would have environment variables with the addresses of the `Mysql` and `Redis` containers.

You make a good point. We can enable a similar behaviour by automatically adding all dependant containers and the parent to the same network. This should allow the parent (or dependant) containers to reference each other's IP addresses through the use of docker's built-in DNS service.

The parent container could then reference other containers using calls like `http://my-redis:6379`.

What do you think? Would that suffice in this case?

pablochacin:
@Mathew-Estafanous 

I think putting the containers in the same network and using a fixed name for the containers could work, but with that many requirements, the solution looks fragile and somehow forced.

I think the callback you proposed could work as a more robust mechanism.  if it is called after the dependency is running but before the container with the dependencies is created, we can inject any parameter we need in the container request, host names, environment variables, etc.  

I lean towards this initial approach and enhance developer experience by providing some options that implement frequent use cases like:
-  `WithDNSDependency(depdency ContainerRequest, hostname string)`  injects the IP address of the dependency with the given hostname
-  `WithEnvVarDependency(depdency ContainerRequest, var string)` injects the IP address of the dependency as an environment variable

Mathew-Estafanous:
@pablochacin

I don't believe we'd necessitate the use of fixed names. 

We could have a callback similar to `WithDNSDependency(dependency ContainerRequest, hostname string)`. In this case, the containers would be in the same network then we'd inject an environment variable with the same name as `hostnameEnv` that supplies the name (hostname) of the container it depends on.

---
For example. `caching_service` might depend on a `redis` instance. 
```go
...WithDNSDependency(myRedisService, "REDIS_HOSTNAME")
```
On creation of the `redis` container, it might be assigned some arbitrary name which we then inject to the parent container as an env variable `REDIS_HOSTNAME=<name>`. 

Internally, the containers would be created in a way similar to this:
```sh
> docker network create testcontainers

# start dependent redis container
> docker run -itd --network=testcontainers redis:latest
# image named 'youthful_elgamal' is created

> docker run -itd --network=testcontainers -e  REDIS_HOSTNAME=youthful_elgamal service-A:latest
# creates `service-A` container which can rely on DNS resolution based on the environment variable passed to it.
```

What do you think? Does this combination of the two ideas work?

pablochacin:
> I don't believe we'd necessitate the use of fixed names.

@Mathew-Estafanous I explained poorly. I meant that if we don't introduce a mechanism like the one I suggested, we would need fixed names.  With the options I suggested, that is not necessary, as you explained. 

So, in summary, I think implementing the dependency with a callback is a good first step and the the options for automatically setting DNS or ENV with the container's IP can be built on top of it.

dependency_test.go
```diff
@@ -0,0 +1,119 @@
...
+			configureDependants: func(ctx context.Context, t *testing.T) []*ContainerDependency {
+				nginxReq := ContainerRequest{
+					Image:        nginxAlpineImage,
+					ExposedPorts: []string{nginxDefaultPort},
+					WaitingFor:   wait.ForListeningPort(nginxDefaultPort),
+				}
+
+				return []*ContainerDependency{
+					NewContainerDependency(nginxReq, "FIRST_DEPENDENCY"),
+					NewContainerDependency(nginxReq, "SECOND_DEPENDENCY"),
```
Mathew-Estafanous, line 33: Here's the interface that I came up with for dependencies. I still have to add more tests and work on edge cases, but thought I'd push this up to get some eyes on it.

dependency.go
```diff
@@ -0,0 +1,133 @@
...
+
+				net, err := GenericNetwork(ctx, GenericNetworkRequest{
+					NetworkRequest: NetworkRequest{
+						Driver:   Bridge,
+						Labels:   GenericLabels(),
+						Name:     fmt.Sprintf("testcontainer-dependency-%v", uuid.NewString()),
+						Internal: false,
+					},
+				})
+				depNetwork = net.(*DockerNetwork)
```
Mathew-Estafanous, line 96: I was hoping to use `network.New` for this, but that causes a cyclical dependency - which is why I'm using these depreciated functions.

dependency.go
```diff
@@ -0,0 +1,171 @@
...
+)
+
+// ContainerDependency represents a reliance that a container has on another container.
+type ContainerDependency struct {
+	Request ContainerRequest
+	EnvKey  string
+	// CallbackFunc is called after the dependency container is started.
+	CallbackFunc func(Container)
+	// KeepAlive determines whether the dependency should be kept alive after the parent container is terminated.
+	KeepAlive bool
```
Mathew-Estafanous, line 19: I've added a `KeepAlive` flag since I thought it'd be helpful to define whether or not the dependency should terminate alongside its parent.

I set the default to `true` in `NewContainerDependency` but I'd be interested to hear other's opinions of the flag itself and if `true` is an appropriate default.

dependency.go
```diff
@@ -0,0 +1,171 @@
...
+	"github.com/docker/docker/api/types/container"
+	"github.com/google/uuid"
+	"github.com/pkg/errors"
+	"golang.org/x/exp/slices"
+)
+
+// ContainerDependency represents a reliance that a container has on another container.
+type ContainerDependency struct {
+	Request ContainerRequest
+	EnvKey  string
```
pablochacin, line 15: The environment variable can be injected by the callback function and not all containers can take advantage of it, so I'm not convinced this should be implemented in the dependency struct as a default behavior.

pablochacin:
@Mathew-Estafanous Overall I like the approach. My only observation is that I think injecting the environment variable by default is not necessary. The callback function can do this.  I would remove that from the interface.

dependency.go
Mathew-Estafanous: Sorry, i'm not sure what you mean by the env variable can be injected by the callback func. The container would be already running by the time the callback function is called and I'm not aware if injecting env vars into a live container would be possible at that point.

Mathew-Estafanous: Do you mean that we should make Env injection optional through the use of something like `WithEnvName(envKey string)` ?

dependency.go
```diff
@@ -0,0 +1,171 @@
...
+	// CallbackFunc is called after the dependency container is started.
+	CallbackFunc func(Container)
```
pablochacin, line 17: I think the callback should receive also the container request of the container that has the dependency. In this way the callback can modify the request (e.g. add env variables, setup the hosts file, etc)

Mathew-Estafanous:
@pablochacin I discussed this issue at length with Manuel and wrote a quick explanation of what was decided in the [DependsOn discussion page.](https://github.com/testcontainers/testcontainers-go/discussions/2042#discussioncomment-9279550)

dependency.go
stevenh: Is there any reason why the env would be a requirement?

dependency.go
```diff
@@ -0,0 +1,171 @@
...
+	var depNetwork *DockerNetwork
+	depContainers := make(map[Container]bool)
+	return ContainerLifecycleHooks{
+		PreCreates: []ContainerRequestHook{
+			func(ctx context.Context, req ContainerRequest) (err error) {
+				if len(req.DependsOn) == 0 {
+					return nil
+				}
+				defer func() {
+					if err != nil {
```
stevenh, line 115: suggestion: invert the condition and return early reducing indentation and making it easier grok.

dependency.go
```diff
@@ -0,0 +1,171 @@
...
+						// clean up dependencies that were created if an error occurred.
+						cleanupErr := cleanupDependencyNetwork(ctx, depContainers, depNetwork)
+						if cleanupErr != nil {
+							Logger.Printf("Could not cleanup dependencies after an error occured: %v", cleanupErr)
+						}
+					}
+				}()
+
+				net, err := GenericNetwork(ctx, GenericNetworkRequest{
```
stevenh, line 124: question: Do we need to enforce a separate network?

dependency.go
```diff
@@ -0,0 +1,171 @@
...
+			},
+		},
+		PostCreates: []ContainerHook{
+			func(ctx context.Context, container Container) error {
+				if depNetwork == nil {
+					return nil
+				}
+				err := depNetwork.provider.client.NetworkConnect(ctx, depNetwork.ID, container.GetContainerID(), nil)
+				defer depNetwork.provider.Close()
+				return err
```
stevenh, line 162: suggestion: check and wrap
```suggestion
				defer depNetwork.provider.Close()
				if err := depNetwork.provider.client.NetworkConnect(ctx, depNetwork.ID, container.GetContainerID(), nil); err != nil {
					return fmt.Errorf("connect network: %w", err)
				}

				return nil
```

dependency_test.go
```diff
@@ -0,0 +1,260 @@
...
+
+func Test_ContainerDependency(t *testing.T) {
+	type TestCase struct {
+		name                string
+		configureDependants func(ctx context.Context, t *testing.T) []ContainerDependency
+		containerRequest    ContainerRequest
+		expectedEnv         []string
+		expectedError       string
+	}
+	testCases := []TestCase{
```
stevenh, line 21: suggestion: There's lots of complexity here in the actual table and more in the main flow because of the conditional behaviour which actually results in more code than splitting into dedicated sub sets with `t.Run(...)`

We should always use succinct test names, avoiding spaces as they are replaced on the output making failures harder to track down.

Here's an example from the last item in the table.
```go
	t.Run("empty-key", func(t *testing.T) {
			nginxReq := ContainerRequest{
				Image:        nginxAlpineImage,
				ExposedPorts: []string{nginxDefaultPort},
				WaitingFor:   wait.ForListeningPort(nginxDefaultPort),
			}
			dependency := NewContainerDependency(nginxReq, "")
				ContainerRequest: tc.containerRequest,
				Started:          true,
				DependsOn:  []ContainerDependency{dependency},
			})
			require.EqualError(t, err, "cannot create dependency with empty environment key")
        })
````

If there is a block of the validation logic that's repeated extract that to a helper which continues to keep things easy to follow, reducing duplication.

File: dependency_test.go [hunk 10]
```diff
@@ -0,0 +1,260 @@
...
+		WaitingFor:   wait.ForListeningPort(nginxDefaultPort),
+	}
+
+	testCases := []struct {
+		name          string
+		keepAlive     bool
+		expectRunning bool
+	}{
+		{"dependency is terminated when KeepAlive is set to false", false, false},
+		{"dependency is still running when KeepAlive is set to true", true, true},
```
stevenh, line 224: suggestion: succinct test names.

docs/features/common_functional_options.md
```diff
@@ -152,3 +152,32 @@ The above example is updating the predefined command of the image, **appending**
...
+```golang
+redisDep := testcontainers.NewContainerDependency(testcontainers.ContainerRequest{
+    Image: "docker.io/redis:latest",
+    /* Other options */
+}, "REDIS_NAME")
+
+container, err := testcontainers.GenericContainer(ctx, testcontainers.GenericContainerRequest{
+    ContainerRequest: testcontainers.ContainerRequest{
+        /* Other options */
+		Name: "my-web-app"
```
stevenh, line 173: suggestion: miss indented.
```suggestion
	Name: "my-web-app"
```

stevenh (changes requested):
This is clearly well on the way and would be a great addition, so with the goal of supporting to get this over the line, I've done a pass with some feedback.

I've been doing a bunch of PRs, but still I'm new around here and not everyone will have encountered them, so calling out I use [conventional comments](https://conventionalcomments.org/) to try and ensure that the intent is consistent and well understood.

On of the key things we should try to do is keep the functionality minimal, seems there's been a bit a feature creep for example, keepalive which could delay getting it over the line so might want to keep that for a follow one after more discussion.

Hope this helps.

Mathew-Estafanous:
> This is clearly well on the way and would be a great addition, so with the goal of supporting to get this over the line, I've done a pass with some feedback.

Thanks for the feedback @stevenh. Admittedly I've been swamped with work and other that I haven't had the chance to seriously sit down and work on this.

I'll try to find some time these next few weeks to go through your feedback and implement them - however, feel free to make your own changes if you'd like to get the ball rolling again.

dependency.go
```diff
@@ -0,0 +1,171 @@
...
+	// KeepAlive determines whether the dependency should be kept alive after the parent container is terminated.
+	KeepAlive bool
+}
+
+// NewContainerDependency can be used to define a dependency and the environment variable that
+// will be used to pass the DNS name to the parent container.
+func NewContainerDependency(containerRequest ContainerRequest, envKey string) ContainerDependency {
+	return ContainerDependency{
+		Request:      containerRequest,
+		EnvKey:       envKey,
```
pablochacin, line 27: I think this is specific to an use case (setting an environment variable with the IP address of the dependency) and can be implemented by the call-back function, so in think shouldn't be part of the generic API.

dependency.go
```diff
@@ -0,0 +1,171 @@
...
+	c.KeepAlive = keepAlive
+	return c
+}
+
+func (c ContainerDependency) WithCallback(callbackFunc func(Container)) ContainerDependency {
+	c.CallbackFunc = callbackFunc
+	return c
+}
+
+func (c ContainerDependency) StartDependency(ctx context.Context, network string) (Container, error) {
```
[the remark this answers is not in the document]
pablochacin: I agree.

docs/features/common_functional_options.md
```diff
@@ -152,3 +152,32 @@ The above example is updating the predefined command of the image, **appending**
...
+dependent container together. Thereby making the dependent container accessible within the parent container. All
+dependencies listed in `ContainerRequest#DependsOn` will be started prior to the parent container.
+
+Seen below is a parent container that depends on a redis container. The environment variable `REDIS_NAME` will be
+injected into the parent container with the address of the redis container.
+```golang
+redisDep := testcontainers.NewContainerDependency(testcontainers.ContainerRequest{
+    Image: "docker.io/redis:latest",
+    /* Other options */
+}, "REDIS_NAME")
```
pablochacin, line 168: I think options like the environment variable should be implemented following the functional option patter:

```
redisDep := testcontainers.NewContainerDependency(
                Image: "docker.io/redis:latest",
        }, 
        /* Other options */
       SetEnvVariable( "REDIS_NAME"),
)
```

PR #2039, opened 2023-12-21
chore: lookup exposed ports in the image from Config, not from ContainerConfig

mdelapenya:
## What does this PR do?
This PR uses the Docker type `image.Config` to the lookup of the exposed ports from the underlying image. 

## Why is it important?
If there are no exposed ports in the container request, the code  previously checked the Docker type `image.ContainerConfig`. There are recent errors in the GH actions regarding this: exposed ports being moved from `image.ContainerConfig` to `image.Config`, because it has been deprecated and depending on how the images were built it could be the case that the `ContainerConfig` field comes empty (e.g. images built with buildkit).

## Related issues

- Relates to https://github.com/moby/moby/pull/46939

### DIFF
--- a/lifecycle.go
+++ b/lifecycle.go
@@ -333,6 +333,10 @@ func (p *DockerProvider) preCreateContainerHook(ctx context.Context, req Contain
 		for p := range image.ContainerConfig.ExposedPorts {
 			exposedPorts = append(exposedPorts, string(p))
 		}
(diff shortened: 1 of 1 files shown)


mdelapenya:
Core tests passed, merging now to rebase existing PRs

vvoland (approved):

PR #2043, opened 2023-12-22
adds gotestsum to install tools

hughesjj:
## What does this PR do?

Adds installation for `gotestsum` as required by make target

## Why is it important?
Minor change to help new contributors contribute

1. checkout this PR
2. set your `GOPATH` envvar to a new directory
3. run `make test-all` or `make test-tools`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 include ./commons-test.mk
 
 .PHONY: test-all
-test-all: tools test-unit
+test-all: tools test-tools test-unit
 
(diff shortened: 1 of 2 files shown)


mdelapenya (approved):
Thanks for this tiny refinement to the "build", but valuable for contributions, as you said

I recall having this in the past, probably removed when we delegated the usage of gotestsum to the CI, but some leftovers were missing. Thanks in any case for taking care of the dev experience 👏 

LGTM

PR #2047, opened 2023-12-22
chore: do not read config but instead pass the hub prefix to the prependHub modifier

mdelapenya:
## What does this PR do?
It removes the need to read the config from the prependHubRegistry image substitutor, and instead receives the prefix from the conf.

## Why is it important?
It makes the tests not depend on the singleton config, but instead in the prefix to be used. As a consequence, the unit tests are not flaky anymore.

## Related issues

- Caused by #1928


mdelapenya:
Core test passing. Merging

PR #2048, opened 2023-12-22
[redpanda] sasl test for wrong mechanism

rwaweber:
## What does this PR do?

- Adds a test that ensures that when using a different sasl mechanism from the broker, unsupported error is returned

## Why is it important?

Would like to get another PR out to support SASL-PLAIN auth for redpanda's broker component. Figured I'd use this as a starting point for the discussion. Happy to file a separate issue for that if you'd prefer!

In the interest of not breaking existing calls by changing the name of the `WithEnableSASL` function, I was thinking that the SASL-PLAIN auth mechanism could be added by incorporating a `WithEnablePlainSASL` function for an option at the same level. What do you all think?

If  amenable to that idea, you think adding a `WithEnableScramSASL` function could help clear up the distinction between them? I certainly don't feel strongly about this approach at all. 

## How to test this PR
- Should be able to be picked up by the existing test harness without doing anything differently.


mdelapenya:
Hi @rwaweber thanks for creating this PR. I was on Xmas PTO until this week, that's why I did not get to you before.

I've shared this PR, and your proposal for the SALS PLAIN mechanism, with the Redpanda folks, so they eventually jump in here with more insights.

Again, thanks for your time improving the module!

rwaweber:
No worries at all, hope you had a restful PTO!

Sounds great to me, and thank you for doing the hard work of maintaining it! It's certainly saved my bacon a few times, so I'm happy to help where I can.

modules/redpanda/redpanda_test.go
```diff
@@ -162,6 +163,23 @@ func TestRedpandaWithAuthentication(t *testing.T) {
...
+				User: "no-superuser",
+				Pass: "test",
+			}.AsMechanism()),
+		)
+		require.NoError(t, err)
+
+		kafkaAdmCl := kadm.NewClient(kafkaCl)
+		_, err = kafkaAdmCl.Metadata(ctx)
+		require.Error(t, err)
+		assert.ErrorContains(t, err, "UNSUPPORTED_SASL_MECHANISM")
```
mdelapenya, line 180:
```suggestion
		require.ErrorContains(t, err, "UNSUPPORTED_SASL_MECHANISM")
```

mdelapenya (approved):
LGTM, thanks for this addition! Do you think we could add a warning in the module docs or waiting for the follow-up PR with the PLAIN mechanism is fine?

mdelapenya:
Thanks for contributing this test! Much appreciated 👏

PR #2073, opened 2024-01-02
Adding LogConsumers and LogProducer start as part of the GenericRequest

### LINKED ISSUE #525: [Enhancement]: improve the experience using log producers

jespino:
## What does this PR do?

This PR adds the posibility of configuring the LogProducer and the LogConsumers directly in the GenericRequest config of the container.

## Why is it important?

This provides better ergnomoics in the library giving a better understanding of the expected container behavior looking only at the container request, and also provides an easier way to initialize the logging process, without the need of calling the FollowLogs or StartLogProducer methods as separate calls after the creation of the container.

## Related issues

- Closes #525

## How to test this PR

I included a test for this in the code. But if you want to test it manually you need to create a new GenericRequest that contains the LogConsumers and LogProducer settings, and verify that the LogConsumers are behaving as expected, that means that the log are process since the beginning of the container execution.

### DIFF
--- a/container.go
+++ b/container.go
@@ -142,6 +142,8 @@ type ContainerRequest struct {
 	HostConfigModifier      func(*container.HostConfig)                // Modifier for the host config before container creation
 	EnpointSettingsModifier func(map[string]*network.EndpointSettings) // Modifier for the network settings before container creation
 	LifecycleHooks          []ContainerLifecycleHooks                  // define hooks to be executed during container lifecycle
+	LogProducer             bool                                       // enable the log producer
+	LogConsumers            []LogConsumer                              // define log consumers to follow the logs
 }
 
 // containerOptions functional options for a container
--- a/docker.go
+++ b/docker.go
@@ -1020,6 +1020,16 @@ func (p *DockerProvider) CreateContainer(ctx context.Context, req ContainerReque
 				},
 			},
 			PostStarts: []ContainerHook{
+				func(ctx context.Context, c Container) error {
+					for _, consumer := range req.LogConsumers {
+						c.FollowOutput(consumer)
+					}
+
+					if req.LogProducer {
+						return c.StartLogProducer(ctx)
+					}
+					return nil
+				},
 				// first post-start hook is to wait for the container to be ready
 				func(ctx context.Context, c Container) error {
 					dockerContainer := c.(*DockerContainer)
--- a/logconsumer_test.go
+++ b/logconsumer_test.go
@@ -458,3 +458,52 @@ func TestContainerLogsShouldBeWithoutStreamHeader(t *testing.T) {
 	}
 	assert.Equal(t, "0", strings.TrimSpace(string(b)))
 }
+
+func TestContainerLogsEnableAtStart(t *testing.T) {
+	ctx := context.Background()
+	g := TestLogConsumer{
+		Msgs:     []string{},
+		Done:     make(chan bool),
+		Accepted: devNullAcceptorChan(),
+	}
+
+	req := ContainerRequest{
+		FromDockerfile: FromDockerfile{
+			Context:    "./testdata/",
+			Dockerfile: "echoserver.Dockerfile",
+		},
+		ExposedPorts: []string{"8080/tcp"},
+		WaitingFor:   wait.ForLog("ready"),
+		LogProducer:  true,
+		LogConsumers: []LogConsumer{&g},
+	}
+
+	gReq := GenericContainerRequest{
+		ContainerRequest: req,
+		Started:          true,
+	}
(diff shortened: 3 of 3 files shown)


container.go
```diff
@@ -142,6 +142,8 @@ type ContainerRequest struct {
 	HostConfigModifier      func(*container.HostConfig)                // Modifier for the host config before container creation
 	EnpointSettingsModifier func(map[string]*network.EndpointSettings) // Modifier for the network settings before container creation
 	LifecycleHooks          []ContainerLifecycleHooks                  // define hooks to be executed during container lifecycle
+	LogProducer             bool                                       // enable the log producer
```
mdelapenya, line 145: IIUC this bool is a flag to start producing logs, right? Wdyt if we remove it and simply check that the LogConsumers slice has elements? We can assume that if the users provided consumers it's because they want to produce logs.

docker.go
```diff
@@ -1020,6 +1020,16 @@ func (p *DockerProvider) CreateContainer(ctx context.Context, req ContainerReque
 				},
 			},
 			PostStarts: []ContainerHook{
+				func(ctx context.Context, c Container) error {
```
mdelapenya, line 1023:
```suggestion
				// first post-start hook is to produce logs and start log consumers
				func(ctx context.Context, c Container) error {
```

docker.go
```diff
@@ -1020,6 +1020,16 @@ func (p *DockerProvider) CreateContainer(ctx context.Context, req ContainerReque
...
+					for _, consumer := range req.LogConsumers {
+						c.FollowOutput(consumer)
+					}
+
+					if req.LogProducer {
+						return c.StartLogProducer(ctx)
+					}
+					return nil
+				},
 				// first post-start hook is to wait for the container to be ready
```
mdelapenya, line 1033:
```suggestion
				// second post-start hook is to wait for the container to be ready
```

container.go
jespino: Done

mdelapenya:
@Tofel I'd like you to take a look at this one if possible, as you already contributed the log producer options 🙏

container.go
```diff
@@ -43,25 +43,25 @@ type Container interface {
...
+	NetworkAliases(context.Context) (map[string][]string, error)    // get container network aliases for a network
 	Exec(ctx context.Context, cmd []string, options ...tcexec.ProcessOption) (int, io.Reader, error)
 	ContainerIP(context.Context) (string, error)    // get container ip
 	ContainerIPs(context.Context) ([]string, error) // get all container IPs
 	CopyToContainer(ctx context.Context, fileContent []byte, containerFilePath string, fileMode int64) error
 	CopyDirToContainer(ctx context.Context, hostDirPath string, containerParentPath string, fileMode int64) error
 	CopyFileToContainer(ctx context.Context, hostFilePath string, containerFilePath string, fileMode int64) error
 	CopyFileFromContainer(ctx context.Context, filePath string) (io.ReadCloser, error)
-	GetLogProducerErrorChannel() <-chan error
+	GetLogProductionErrorChannel() <-chan error
```
mdelapenya, line 64: Changing the name is fine, not a breaking change, as it was added for this release.

docker.go
```diff
@@ -615,66 +615,71 @@ func (c *DockerContainer) CopyToContainer(ctx context.Context, fileContent []byt
...
 }
 
-type LogProducerOption func(*DockerContainer)
+type LogProductionOption func(*DockerContainer)
 
-// WithLogProducerTimeout is a functional option that sets the timeout for the log producer.
+// WithLogProductionTimeout is a functional option that sets the timeout for the log production.
 // If the timeout is lower than 5s or greater than 60s it will be set to 5s or 60s respectively.
-func WithLogProducerTimeout(timeout time.Duration) LogProducerOption {
+func WithLogProductionTimeout(timeout time.Duration) LogProductionOption {
```
mdelapenya, line 622: Same here, the functions and types were added for this release

mdelapenya (approved):
LGTM, thanks for this!

As mentioned in the issue, as a follow-up, we can update the log consumer pattern and simply use an `io.Writer` interface, which will allow to reduce the amount of custom types and also leverage the `io.MultiWriter` type for having multiple consumers/writers.

PR #2081, opened 2024-01-09
fix: align codeql versions in GH workflow

mdelapenya:
## What does this PR do?
It uses the same version for all the codeql actions.

## Why is it important?
After #2056, the versions where not aligned, causing a typescript error in the codeql analysis.

## Related issues

- Caused by #2056
- Discussed with the codeql team in https://github.com/github/codeql-action/issues/2075


mdelapenya:
Merging now, as it only applies to codeql actions

PR #2082, opened 2024-01-09
GenericContainer: in case of error: return a reference to the failed container

JordanP:
## What does this PR do?
`GenericContainer` function: in case of error: return a reference to the failed container. This way, the caller has an opportunity to clean things up and call Destroy on the failed container.

## Why is it important?

If a `WaitStrategy` fails (timeouts for instance, context canceled, etc.), we may still want to return a reference to the running container so that the caller can Destroy() it and not leave half-healthy containers around.

### DIFF
--- a/generic.go
+++ b/generic.go
@@ -74,7 +74,8 @@ func GenericContainer(ctx context.Context, req GenericContainerRequest) (Contain
 		c, err = provider.CreateContainer(ctx, req.ContainerRequest)
 	}
 	if err != nil {
-		return nil, fmt.Errorf("%w: failed to create container", err)
+		// At this point `c` might not be nil. Give the caller an opportunity to call Destroy on the container.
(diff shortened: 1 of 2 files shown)


mdelapenya:
Hi @JordanP thanks for submitting this enhancement. I see the value if it although I'd like to confirm with you that this

> not leave half-healthy containers around.

will happen if Ryuk is disabled, right? If Ryuk is enabled then there is no risk of leaking containers.

JordanP:
> will happen if Ryuk is disabled, right? 

Yes, that's correct. 

More generally, I would say it's fair to let the caller (that tried to create the container) deal with the failed container, if it wants to. But, to give the caller a chance to do that, it needs a reference to the container.

mdelapenya (approved):
LGTM, thanks @JordanP for changing this. Many times, tiny things are the hardest to discover. Thank you

PR #2085, opened 2024-01-10
Adding mockserver module

jespino:
## What does this PR do?

This PR adds support for the mockserver module.

## Why is it important?

This is part of the current modules supported in testcontainers-java, having this in go is an good improvement.

## Related issues

Relates #636

## How to test this PR

I already added 2 test examples, but you can test manually based on that examples to run a mock server and add some expectations


docs/modules/mockserver.md
```diff
@@ -0,0 +1,58 @@
+# MockServer
+
+Not available until the next release of testcontainers-go <a href="https://github.com/testcontainers/testcontainers-go"><span class="tc-version">:material-tag: main</span></a>
+
+## Introduction
+
+The Testcontainers module for MockServer.
```
mdelapenya, line 7:
```suggestion
The Testcontainers module for MockServer. MockServer can be used to mock HTTP services by matching requests against user-defined expectations.
```

modules/mockserver/coverage.out
```diff
@@ -0,0 +1 @@
+mode: set
```
mdelapenya, line 1: Do we need this coverage file here? If so what would be the rationale for it, in the case we need to spread it to the rest of the modules.

modules/mockserver/mockserver.go
```diff
@@ -0,0 +1,59 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/testcontainers/testcontainers-go"
+	"github.com/testcontainers/testcontainers-go/wait"
+)
+
+// defaultImage is the default MockServer container image
+const defaultImage = "mockserver/mockserver"
```
mdelapenya, line 12: I'd choose a fixed version instead: `mockserver/mockserver:5.15.0`, to make it reliable and not get undesired updates if latest image changed. Please apply it to the examples too 🙏

modules/mockserver/mockserver_test.go
```diff
@@ -0,0 +1,26 @@
+package mockserver
+
+import (
+	"context"
+	"testing"
+
+	"github.com/testcontainers/testcontainers-go"
+)
+
+func TestMockServer(t *testing.T) {
```
mdelapenya, line 10: We can remove this test, as it's not doing any assertion.

I have opposed sentiments when writing both testable examples and tests: what is a testable example and what is a test? In general, I like testable examples because they end up into the Go docs. But I do prefer not adding too many testable examples and expand the test suite as regular tests. Wdyt?

modules/mockserver/coverage.out
jespino: Nope, sorry, that was generated somehow, let me remove it.

modules/mockserver/examples_test.go
```diff
@@ -0,0 +1,100 @@
+package mockserver_test
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+
+	client "github.com/BraspagDevelopers/mock-server-client"
```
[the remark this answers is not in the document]
jespino: Yes, looks like is the most recent one.

modules/mockserver/mockserver.go
jespino: :+1:

modules/mockserver/mockserver_test.go
jespino: I see no value in this test specifically because the example already covers this use case. In general, I prefer to avoid using examples as testing because you can end up sacrificing test reliability or example readability. So I prefer to have them separated, but for this specific case where this is so simple and so general, there is no point in adding this test instead of just relying on the example.

mdelapenya (approved):
LGTM, thanks for adding a new module! 🚀

PR #2117, opened 2024-01-12
Openldap module support

jespino:
## What does this PR do?

This PR adds the OpenLDAP module.

## Why is it important?

LDAP is a widely used technology that is a nice to have for testing purposes.

## How to test this PR

This PR includes its own unit tests


mdelapenya (approved):
LGTM, thank you @jespino for this contribution!

Are you willing to add more features/options in this PR to it? If not, I think we can merge it

modules/openldap/openldap.go
```diff
@@ -0,0 +1,123 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	connStr := fmt.Sprintf("ldap://%s", net.JoinHostPort(host, containerPort.Port()))
+	return connStr, nil
+}
+
+// LoadLdif loads an ldif file into the OpenLDAP container
+func (c *OpenLDAPContainer) LoadLdif(ctx context.Context, ldif []byte) error {
```
mdelapenya, line 46: Do we want this API method here, or as a functional option to be passed to the `RunContainer` function? If the latter, I'd go with adding a post-starts hook doing what you already did in this method

modules/openldap/go.mod
```diff
@@ -0,0 +1,57 @@
+module github.com/testcontainers/testcontainers-go/modules/openldap
```
mdelapenya, line 1: I think you need to run mod tidy here with the recent bumps. I know, we are in a monorepo and dependabot should automate the bumps across the repo 😞

jespino: Done

modules/openldap/openldap.go
mdelapenya: @jespino I think you want to add this method also as part of the functional options. Is it fine if we merge the PR as is and that option comes as a follow-up? I'd like to have it as part of the PR, but I also see it important to value your work merging the PR now.

jespino: I'm actually working on that right now, I'm struggling to find the way of doing it in a way that I feel comfortable with.

mdelapenya: Great then! Let me know if I can be of any help

mdelapenya (approved):
LGTM, thanks!

PR #2125, opened 2024-01-17
improve OSSF score

mmorel-35:
## What does this PR do?

Pin github-actions dependencies and define permissions on workflows and jobs when necessary

## Why is it important?

This help improve OSSF score

Signed-off-by: Matthieu MOREL <matthieu.morel35@gmail.com>


mdelapenya (approved):
LGTM, thanks!

PR #2127, opened 2024-01-18
Go install gotestsum and golangci-lint 

mmorel-35:
## What does this PR do?

Define Makefile targets to install golangci-lint and gotestsum 

## Why is it important?

Helps having the right requirements to make different targets work.

### DIFF
--- a/.github/workflows/ci-test-go.yml
+++ b/.github/workflows/ci-test-go.yml
@@ -104,9 +104,7 @@ jobs:
         if: ${{ inputs.run-tests }}
         working-directory: ./${{ inputs.project-directory }}
         timeout-minutes: 30
-        run: |
-            go install gotest.tools/gotestsum@latest
-            make test-unit
+        run: make test-unit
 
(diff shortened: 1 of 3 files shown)


mmorel-35:
Hi @mdelapenya ,
I'm wondering if I shall add a PHONY for the two new target, any idea?

mdelapenya (approved):
LGTM! The makefile define is cool!

mdelapenya:
> Hi @mdelapenya , I'm wondering if I shall add a PHONY for the two new targets, any idea?

From Make docs: https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html

> A GNU target is one that is not really the name of a file; rather it is just a name for a recipe to be executed when you make an explicit request. There are two reasons to use a phony target: to avoid a conflict with a file of the same name, and to improve performance.

So I think it's fine to keep those two without PHONY, as they are indeed files.

Thanks for raising the concern 🙇

PR #2131, opened 2024-01-21
feat(modules.cockroachdb) Adds cockroachdb module

rcrowe:
## What does this PR do?

Adds [CockroachDB](http://cockroachdb.com/) as a module

- TLS support via `cockroachdb.WithTLS` & the helper `cockroachdb.NewTLSConfig`
- `COCKROACH_DATABASE` support via `cockroachdb.WithDatabase`
- `COCKROACH_USER` support via `cockroachdb.WithUser`
- `COCKROACH_PASSWORD` support via `cockroachdb.WithPassword`

## Why is it important?

Previously an example only, a module lets us develop it over time to include features like TLS.


docs/modules/cockroachdb.md
```diff
@@ -0,0 +1,68 @@
...
+When starting the CockroachDB container, you can pass options in a variadic way to configure it.
+
+#### Image
+
+If you need to set a different CockroachDB Docker image, you can use `testcontainers.WithImage` with a valid Docker image
+for CockroachDB. E.g. `testcontainers.WithImage("cockroachdb/cockroach:latest-v23.1")`.
+
+{% include "../features/common_functional_options.md" %}
+
+#### Image tag
```
mdelapenya, line 45: I'd offer just one way to change the image, to make it difficult to commit a mistake: what will happen if both `WithImage` and `WithImageTag` are passed? Reading the code I think the tag will win, but it will be less obvious than simply providing the core option for replacing the full image.

Said that, I could have missed any other value with the tag. What are your thoughts?

mdelapenya (commented):
LGTM, thanks for this new module! I only left a comment regarding the `WithImageTag` option. Once discussed, I think this PR is ready to be merged, thanks!

docs/modules/cockroachdb.md
rcrowe: My thinking at the time was that we change the version (tag) but not the image, but that's a fair point & consideration, so I'll remove it to keep things simple 👍🏻

mdelapenya:
@rcrowe the errors in the rootless-mode pipeline are not related to this PR but to an automatic bump in the Github worker, which is installing Docker v25.0.0 since a few days ago. I'm talking to the Moby team to check if that's a regression or the expected behaviour, because the failures are consistent:

1. One of the failures is related to an error message that seems to have changed, when build target is not found, from failed to reach build target target-foo in Dockerfile to target stage 'target-foo' could not be found  [#46754](https://github.com/moby/moby/pull/46754)
    - I think we can simply check for the error and avoid asserting for a specific text in the error message
2. Second error is related to the number of network aliases for the the bridge network, which in rootless mode seems to have changed from 0 to 1: `Expected number of aliases for 'bridge' network 0. Got 1.`
    - I'm not sure about this one, as in non-rootless mode, the same test passes

rcrowe:
Thanks @mdelapenya 

Gives me some time to finish TLS support & push that up as a separate PR

modules/cockroachdb/cockroachdb.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+// MustConnectionString panics if the address cannot be determined.
+func (c *CockroachDBContainer) MustConnectionString(ctx context.Context) string {
+	addr, err := c.ConnectionString(ctx)
+	if err != nil {
+		panic(err)
+	}
+	return addr
+}
```
eddumelendez, line 36: @mdelapenya is this ok?

modules/cockroachdb/cockroachdb.go
```diff
@@ -0,0 +1,111 @@
...
+	if req.Image != "" {
+		return req.Image
+	}
+	return fmt.Sprintf("%s:%s", defaultImage, opts.ImageTag)
+}
+
+func cmd(opts Options) []string {
+	return []string{
+		"start-single-node",
+		"--insecure",
```
eddumelendez, line 108: Given support for USERNAME, PASSWORD, DATABASE env vars are already available in recent versions, those should be supported for the module. See [implementation](https://github.com/testcontainers/testcontainers-java/pull/6448) in Java.

modules/cockroachdb/cockroachdb.go
eddumelendez: Thanks! Also, asking because I don't see this for other database modules

mdelapenya: No, it's a patter that we should add as a consequence, as I like having both methods

modules/cockroachdb/cockroachdb.go
rcrowe: Tags on Cockroach images have various versions that aren't going to be semver compatible. I'm wondering if you'd be happy if we always set the env vars, they just won't take effect with versions `<22.1`. That seems easier than trying to support tags like:

- `latest-v23.1`
- `v23.1.13-fips`

WDYT?

rcrowe: Cockroach seems happier when I drop `--insecure` & add `--accept-sql-without-tls`. Will see about landing a change here

rcrowe: Pushed the a change to include changing user & password that will set the necessary env vars & connection string

rcrowe:
Given I've added user & password as part of the review, I'll push TLS support to this branch tomorrow

modules/cockroachdb/cockroachdb.go
```diff
@@ -0,0 +1,254 @@
...
+		wait.ForSQL(defaultSQLPort, "pgx/v5", func(host string, port nat.Port) string {
+			connStr := connString(opts, host, port)
+			if tlsConfig == nil {
+				return connStr
+			}
+
+			// register TLS config with pgx driver
+			connCfg, err := pgx.ParseConfig(connStr)
+			if err != nil {
+				panic(err)
```
rcrowe, line 176: I wasn't sure how best to go about this as we need the host & port to build the connection string before adding TLS config.

One option is creating a custom `wait.Strategy` but that felt heavy-handed. Thoughts?

rcrowe:
@mdelapenya I'm assuming docker rootless failing here because it's flaky, notice on `main` it passed, rather than anything in my changeset?

Have addressed the feedback of `COCKROACH_DATABASE`, `COCKROACH_USER` & `COCKROACH_PASSWORD`, along with TLS support. The PR grew a little because of that, so apologies for the larger diff.

docs/modules/cockroachdb.md
```diff
@@ -0,0 +1,78 @@
...
+
+Disable insecure mode and connect with password authentication by setting `cockroachdb.WithUser` and `cockroachdb.WithPassword`.
+
+#### Store size
+
+Control the maximum amount of memory used for storage, by default this is 100% but can be changed by provided a valid option to `WithStoreSize`. Checkout https://www.cockroachlabs.com/docs/stable/cockroach-start#store for the full range of options available.
+
+#### TLS authentication
+
+`cockroachdb.WithTLS` lets you provde the CA certificate along with the certicate and key for the node & clients to connect with.
```
mdelapenya, line 59:
```suggestion
`cockroachdb.WithTLS` lets you provide the CA certificate along with the certicate and key for the node & clients to connect with.
```

modules/cockroachdb/cockroachdb.go
mdelapenya: We could explicitly mention the panic in the docs, in the TLS section.

mdelapenya (commented):
Thanks for your hard work adding this new module. I left a few comments that are non-blocking, so I think we can merge the module once discussed.

In any case, great job!

modules/cockroachdb/cockroachdb.go
```diff
@@ -116,16 +163,44 @@ func addEnvs(req *testcontainers.GenericContainerRequest, opts options) {
...
+	})
+	files := map[string][]byte{
+		"ca.crt":          caBytes,
+		"node.crt":        opts.TLS.NodeCert,
+		"node.key":        opts.TLS.NodeKey,
+		"client.root.crt": opts.TLS.ClientCert,
+		"client.root.key": opts.TLS.ClientKey,
+	}
+	for filename, contents := range files {
+		if err := container.CopyToContainer(ctx, contents, filepath.Join(certsDir, filename), 0o600); err != nil {
```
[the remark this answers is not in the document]
rcrowe: I'll have a better look at that. I had tried that first, but it required local files whereas I had `[]byte`; guess I can always create temporary files.

Secondly, at least on my first attempt, I couldn't start Cockroach without the files being in place first.

It's likely I just need another go at it & go a little deeper. Thanks for the pointer.
Is that generally the approach to take rather than the one I've taken, is there something wrong with what's here now, so I can learn?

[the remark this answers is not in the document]
mdelapenya: No, I think this could be merged as is, and if you want, create a follow-up PR with some refinements.

There is no mandatory rule for that, some modules uses the Options pattern, some not.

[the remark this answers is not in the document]
rcrowe: Moved to a lifecycle which is a much cleaner way to access `testcontainers.Container`, thanks for the tip

rcrowe:
@mdelapenya 🤞🏻 this is in a shape to get merged & iterate once it's in the hands of people 🤞🏻

modules/cockroachdb/cockroachdb.go
mdelapenya: Although not included in other modules, I'd like to star having this kind of mentions in the docs. Could you please add that to the docs?

modules/cockroachdb/cockroachdb.go
```diff
@@ -0,0 +1,259 @@
...
+	host, err := c.Host(ctx)
+	if err != nil {
+		return "", err
+	}
+
+	return connString(c.opts, host, port), nil
+}
+
+// TLSConfig returns config necessary to connect to CockroachDB over TLS.
+func (c *CockroachDBContainer) ConnectionTLS() (*tls.Config, error) {
```
mdelapenya, line 67: I think this name would be even clearer. If you agree on that (the comment indicates you already thought about that 😅), let's update the docs too.
```suggestion
func (c *CockroachDBContainer) TLSConfig() (*tls.Config, error) {
```

mdelapenya (commented):
LGTM, thank you for your contribution!

I added a few minor comments, once they are addressed, we are good to go! 🚀

docs/modules/cockroachdb.md
```diff
@@ -0,0 +1,80 @@
...
+The CockroachDB module exposes one entrypoint function to create the CockroachDB container, and this function receives two parameters:
+
+```golang
+func RunContainer(ctx context.Context, opts ...testcontainers.ContainerCustomizer) (*CockroachDBContainer, error)
+```
+
+- `context.Context`, the Go context.
+- `testcontainers.ContainerCustomizer`, a variadic argument for passing options.
+
+When TLS is enabled there's a very small, unlikely chance that the underlying driver can panic when registering the driver as part of waiting for CockroachDB to be ready to accept connections. If this is repeatadly happening please open an issue.
```
mdelapenya, line 34: I'd move this section under the TLS function section.

Also please indent it with:
```suggestion
!!!warning
    When TLS is enabled there's a very small, unlikely chance that the underlying driver can panic when registering the driver as part of waiting for CockroachDB to be ready to accept connections. If this is repeatedly happening please open an issue.
```

mdelapenya (approved):
LGTM, thanks for this contribution and for your patience while reviewing the code.

Merging it now

PR #2135, opened 2024-01-22
chore: skip assertions for Docker Rootless

mdelapenya:
## What does this PR do?
In this PR, we are removing an assertion by the error message, as it changed in upstream moby (see https://github.com/moby/moby/pull/46754/files and https://github.com/moby/buildkit/pull/4394).

At the same time, we are skipping an assertion only for Rootless mode, as it's causing a failure on Docker v25.0.0, but not on lower versions.

Finally, and for the sake of reducing the total build time of the pipelines, we are moving the no-reaper and the rootless pipelines from the second stage to the first one, so that the ~5min they take happen at the same time as the main `test` pipeline.

## Why is it important?
We noticed that, since 3 days ago, the Github workers are including Docker v25.0.0 for rootless Docker, which causes the above error to appear.

The regular workers are still in v24.0.7, which will eventually change causing the errors for the entire test suite.

## Related issues

- Relates to Buildkit: https://github.com/moby/buildkit/pull/4394
- Relates to Moby: https://github.com/moby/moby/pull/46754

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,12 +41,11 @@ jobs:
       ryuk-disabled: false
 
   # The job below is a copy of the job above, but with ryuk disabled.
-  # It's executed in a secondary stage to avoid concurrency issues.
+  # It's executed in the first stage to avoid concurrency issues.
   test-reaper-off:
     # do not run this job if it's a PR from dependabot that is not approved yet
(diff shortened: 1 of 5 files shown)


mdelapenya:
Merging, as the Mysql error is resolved in #2130

Issue #2138, opened 2024-01-23
[Feature]: MongoDB ReplicaSet


trungdlp-wolffun:
### Problem

Create test with MongoDB transaction and it require MongoDB ReplicaSet

### Solution

Setup MongoDB ReplicaSet

### Benefit

Transaction can testable

### Alternatives

https://github.com/testcontainers/testcontainers-java/blob/main/modules/mongodb/src/main/java/org/testcontainers/containers/MongoDBContainer.java

### Would you like to help contributing this feature?

Yes

PR #2139, opened 2024-01-23
feat(modules.mongodb): setup mongodb replicaset

trungdlp-wolffun:
## What does this PR do?
Fix issue #2138

## Related issues

-

### DIFF
--- a/modules/mongodb/mongodb.go
+++ b/modules/mongodb/mongodb.go
@@ -23,12 +23,50 @@ func RunContainer(ctx context.Context, opts ...testcontainers.ContainerCustomize
 	req := testcontainers.ContainerRequest{
 		Image:        defaultImage,
 		ExposedPorts: []string{"27017/tcp"},
+		Cmd:          []string{"--replSet", "rs0"},
 		WaitingFor: wait.ForAll(
 			wait.ForLog("Waiting for connections"),
 			wait.ForListeningPort("27017/tcp"),
 		),
 		Env: map[string]string{},
 	}
+
+	// TODO: should I used fixed array?
+	req.LifecycleHooks = make([]testcontainers.ContainerLifecycleHooks, 1)
+	req.LifecycleHooks[0].PostStarts = append(req.LifecycleHooks[0].PostStarts,
+		func(ctx context.Context, container testcontainers.Container) error {
+			exitCode, _, err := container.Exec(ctx, []string{"mongosh", "--eval", "rs.initiate();"})
+			if err != nil {
+				return err
+			}
+
+			if exitCode != 0 {
+				return fmt.Errorf("init replicaset error")
+			}
+
+			// TODO: should move to const or func?
+			waitCmd := fmt.Sprintf("var attempt = 0; "+
+				"while"+
+				"(%s) "+
+				"{ "+
+				"if (attempt > %d) {quit(1);} "+
+				"print('%s ' + attempt); sleep(100);  attempt++; "+
+				" }",
+				"db.runCommand( { isMaster: 1 } ).ismaster==false",
+				60,
+				"An attempt to await for a single node replica set initialization:")
+
+			exitCode, _, err = container.Exec(ctx, []string{"mongosh", "--eval", waitCmd})
+			if err != nil {
+				return err
+			}
+
+			if exitCode != 0 {
+				return fmt.Errorf("await replica error")
+			}
+
+			return nil
+		})
 
 	genericContainerReq := testcontainers.GenericContainerRequest{
 		ContainerRequest: req,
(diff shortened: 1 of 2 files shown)


eddumelendez:
Thanks for your contribution, @trungdlp-wolffun. Currently, the module support standalone mode by default. Adding ReplicaSet is really nice, however, IMO, this should be an opt-in feature. In the future, sharding can be added as well as a opt-in feature too.

modules/mongodb/mongodb.go
```diff
@@ -122,16 +105,66 @@ func (c *MongoDBContainer) ConnectionString(ctx context.Context) (string, error)
...
+	if c.replicaSet != "" {
+		connStr.WriteString(fmt.Sprintf("/test?replicaSet=%s&directConnection=true", c.replicaSet))
+	}
+
+	return connStr.String(), nil
+}
+
+func replicaSetLifecycleHooks() testcontainers.ContainerLifecycleHooks {
+	return testcontainers.ContainerLifecycleHooks{
+		PostStarts: []testcontainers.ContainerHook{
```
mdelapenya, line 128: I think it would be clearer if we append the container hook functions directly to the PostStarts hook, instead of appending a complete lifecycle hook struct. In the end, the result is the same, but having them all together in the same lifecycle hook seems better aligned, as they all belong to the same hook's post-start events. Wdyt?

trungdlp-wolffun: I move `replicaSetPostStart` code into `replicaSetLifecycleHooks`, right? @mdelapenya

mdelapenya: You could leverage the `Options` pattern we are seen in other modules, such as Redpanda, where the Options struct holds all the values to be applied and all the functional options to configure the module are setting the values into that options struct.

Finally, I'd define a container lifecycle hook in that options, and with each call to the `WithReplicaSet` function, I'd be appending to the PostStarts field of the lifecycle hook. Last thing would be to append the mongo lifecycle hook to the container request before starting the container, right after all the options have been applied.

Does it make sense to you?

mdelapenya: @trungdlp-wolffun please let us know if you need anything from us 🙏

modules/mongodb/mongodb.go
```diff
@@ -73,6 +85,15 @@ func WithPassword(password string) testcontainers.CustomizeRequestOption {
 	}
 }
 
+// WithReplicaSet TODO: help me fill this func comment
+func WithReplicaSet(rsName string) testcontainers.CustomizeRequestOption {
+	return func(req *testcontainers.GenericContainerRequest) {
+		req.Env["MONGO_REPLICASET_NAME"] = rsName
```
eddumelendez, line 91: why is setting the env var also required?

modules/mongodb/mongodb_test.go
```diff
@@ -118,3 +117,44 @@ func ExampleRunContainer_withCredentials() {
...
+
+func ExampleRunContainer_withReplicaSet() {
+	ctx := context.Background()
+
+	container, err := mongodb.RunContainer(ctx,
+		testcontainers.WithImage("mongo:6"),
+		mongodb.WithUsername("root"),
+		mongodb.WithPassword("password"),
+		mongodb.WithReplicaSet("rs0"),
+		testcontainers.WithWaitStrategy(wait.ForLog("Waiting for connections")),
```
eddumelendez, line 129: IMO, the replicaSet opt-in should be just `WithReplicaSet()` and it should be stored in a flag used to set the right wait strategy and exec the commands in the container lifecycle. However, not sure how important is setting the replica set name but having a default could be just enough?

eddumelendez (commented):
Thanks for the update and sorry for the delay. Just left a few more comments to improve experience.

modules/mongodb/mongodb.go
trungdlp-wolffun: @mdelapenya Thanks for your help, I will back soon.

modules/mongodb/mongodb.go
```diff
@@ -122,16 +105,66 @@ func (c *MongoDBContainer) ConnectionString(ctx context.Context) (string, error)
 	if err != nil {
 		return "", err
 	}
+
+	connStr := strings.Builder{}
+	connStr.WriteString("mongodb://")
+
 	if c.username != "" && c.password != "" {
+		connStr.WriteString(fmt.Sprintf("%s:%s@", c.username, c.password))
 		return fmt.Sprintf("mongodb://%s:%s@%s:%s", c.username, c.password, host, port.Port()), nil
```
mdelapenya, line 114: IIUC from this block, if there is an username and password, the connection string won't have anything about the replicaSet, even though it was passed as an option, right?

If not sure, you can extract the url-builder part to a function and add unit tests for it

mdelapenya:
Hi @trungdlp-wolffun I think https://github.com/testcontainers/testcontainers-go/pull/2469 has better readability so I think we are going to build on top of that. This PR will be closed once that one is merged.

I'm sorry if the review took so long 🙏 

Thanks!

PR #2141, opened 2024-01-24
chore: bring back assertion for network aliases for bridge in rootless mode

mdelapenya:
## What does this PR do?
This PR adds a partial revert of #2135, restoring the original assertion for the length of the network aliases for the bridge network.

## Why is it important?
The assertion was skipped for rootless mode, as we detected a regression in Docker v25.0.0, which has been addressed for v25.0.1

## Related issues

- Resolved in upstream Moby: https://github.com/moby/moby/pull/47182


docker_client.go
```diff
@@ -37,7 +38,7 @@ func (c *DockerClient) Events(ctx context.Context, options types.EventsOptions)
 // Info returns information about the docker server. The result of Info is cached
 // and reused every time Info is called.
 // It will also print out the docker server info, and the resolved Docker paths, to the default logger.
-func (c *DockerClient) Info(ctx context.Context) (types.Info, error) {
+func (c *DockerClient) Info(ctx context.Context) (system.Info, error) {
```
mdelapenya, line 41: I believed this is a breaking change for those users of our docker client's Info method, but the returned type is compatible so the compiler accepts it:

```golang
	t.Run("deprecated type", func(t *testing.T) {
		ctx := context.Background()
		c, err := NewDockerClientWithOpts(ctx)
		require.NoError(t, err)

		info, _ := c.Info(ctx)
		require.NotNil(t, info)

		// is of deprecated type
		require.IsType(t, types.Info{}, info)
		// is of the new type
		require.IsType(t, system.Info{}, info)
	})
```

PR #2142, opened 2024-01-24
Adding inbucket module

jespino:
## What does this PR do?

This PR adds the inbucket module

## Why is it important?

An important part of testing application usually involves sending emails,
inbucket helps a lot in this task acting like an MTA that you can inspect later
to do assertions.

## How to test this PR

I included unit tests that verifies the correct behavior


modules/inbucket/inbucket.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+// InbucketContainer represents the Inbucket container type used in the module
+type InbucketContainer struct {
+	testcontainers.Container
+}
+
+// Smtp returns the connection string for the smtp server, using the default
+// 2500 port, and obtaining the host and exposed port from the container.
+func (c *InbucketContainer) Smtp(ctx context.Context) (string, error) {
```
mdelapenya, line 19: Wdyt about this name?
```suggestion
func (c *InbucketContainer) SmtpConnection(ctx context.Context) (string, error) {
```

modules/inbucket/inbucket.go
```diff
@@ -0,0 +1,73 @@
...
+		return "", err
+	}
+
+	return net.JoinHostPort(host, containerPort.Port()), nil
+}
+
+// WebInterface returns the connection string for the web interface server,
+// using the default 9000 port, and obtaining the host and exposed port from
+// the container.
+func (c *InbucketContainer) WebInterface(ctx context.Context) (string, error) {
```
mdelapenya, line 36: In an automated test suite, what would be the use case for this URL? is this something you plan to use in dev-environment, where all the backend services are running?

mdelapenya: Ok, I've read the tests now, and this URL is exposing REST APIs. 👌

modules/inbucket/inbucket_test.go
```diff
@@ -0,0 +1,71 @@
...
+		t.Fatal(err)
+	}
+	if len(headers) != 0 {
+		t.Fatalf("expected 0 messages, got %d", len(headers))
+	}
+
+	msg := []byte("To: to@example.org\r\n" +
+		"Subject: testcontainers test!\r\n" +
+		"\r\n" +
+		"This is a testcontainer test.\r\n")
```
mdelapenya, line 57: Proper naming is important 😄 
```suggestion
		"This is a Testcontainers test.\r\n")
```

mdelapenya (commented):
I left a few comments, not blockers, regarding names.

modules/inbucket/inbucket_test.go
jespino: Done

modules/inbucket/inbucket.go
jespino: Done

modules/inbucket/inbucket_test.go
```diff
@@ -0,0 +1,71 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	restClient, err := client.New(webInterfaceUrl)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	headers, err := restClient.ListMailbox("to@example.org")
```
mdelapenya, line 46: We could make the container provide a way to retrieve the restClient, so the webInterface is not needed 🤔

Not an strong opinion on that, you are most familiar with the technology, so please consider keeping this as is.

mdelapenya: Mmmm forget about that. I've just realised that if the container retrieves a `github.com/inbucket/inbucket/pkg/rest/client` type, then we'll be adding that dependency to the consumers (at the moment it's only present at test time).

mdelapenya (approved):
LGTM, thanks again for your hard work contributing modules!

mdelapenya (approved):
Great job again! Thanks for this new module, I'm pretty sure the community will thank having a way to test email interactions! 👏👏👏 

LGTM!

PR #2143, opened 2024-01-25
testifylint: enable compares rule

mmorel-35:
## What does this PR do?

testifylint: enable compares rule

## Why is it important?

Helps applying the best practices with testify


mdelapenya (approved):
LGTM! Great to see this new lint added, thank you so much!!!

PR #2178, opened 2024-01-30
chore(docker_mounts): stop doing misleading logging

srabraham:
As things are, the logger statement here always triggers for normal use cases. For example, if I start a container with this `mount: testcontainers.VolumeMount("some-vol", "/some/dir")`

then I get this weird message in my logs
`Mount type %!s(testcontainers.MountType=1) is not supported by Testcontainers for Go`

because a GenericVolumeMountSource doesn't have an implementation of `GetVolumeOptions() *mount.VolumeOptions`

I suggest removing that logging, as in this PR, because it's going to be purely confusing for 99% of users.

## What does this PR do?

removes bad logging

## Why is it important?

because it was confusing before

### DIFF
--- a/docker_mounts.go
+++ b/docker_mounts.go
@@ -107,8 +107,10 @@ func mapToDockerMounts(containerMounts ContainerMounts) []mount.Mount {
 			containerMount.VolumeOptions = typedMounter.GetVolumeOptions()
 		case TmpfsMounter:
 			containerMount.TmpfsOptions = typedMounter.GetTmpfsOptions()
-		default:
+		case BindMounter:
 			Logger.Printf("Mount type %s is not supported by Testcontainers for Go", m.Source.Type())
+		default:
+			// The provided source type has no custom options
 		}
 
 		mounts = append(mounts, containerMount)


mdelapenya:
I have to admit that the current implementation for volumes/mounts is so complex that it should be fine to rewrite it in a simpler manner. Because we are working on an eventual v1 branch, and for the sake of simplicity, it would eventually contain breaking changes.

docker_mounts.go
```diff
@@ -108,7 +108,7 @@ func mapToDockerMounts(containerMounts ContainerMounts) []mount.Mount {
 		case TmpfsMounter:
 			containerMount.TmpfsOptions = typedMounter.GetTmpfsOptions()
 		default:
-			Logger.Printf("Mount type %s is not supported by Testcontainers for Go", m.Source.Type())
+			// The provided source type has no custom options
```
mdelapenya, line 111: Do you think we can add a case for the `BindMounter` with a deprecation logging, and then include your default case?

srabraham: That's a good idea! Done

mdelapenya (approved):
LGTM, thanks!

srabraham:
> LGTM, thanks!

Go ahead and merge when you'd like to do so. I don't have the relevant permissions on the repo

Issue #2163, opened 2024-01-25
[Documentation]: document volumes and mounts


mdelapenya:
### Proposal

They are not properly documented in our docs

srabraham:
My bind mounts quietly stopped working when I upgraded from 0.26.0 to 0.27.0, for some reason I couldn't figure out (the deprecation of bind mounts in testcontainers wasn't supposed to break them). So, I switched over to volumes instead. That created a lot more pain, since our CI system got overrun with volumes. As it turned out, named volumes don't get cleaned up by ryuk, so they stuck around indefinitely. The testcontainers-go API makes it really tempting to set a non-empty name for volumes, but the consequences of that are quite bad.

Anyway, yeah, more documentation on this stuff would be great :).

Also, there was some quite misleading logging in place as of when bind mounts were deprecated, and I made a PR to improve that. https://github.com/testcontainers/testcontainers-go/pull/2178/files

mdelapenya:
@srabraham thanks for raising your voice on this issue. I do appreciate your feedback.

> My bind mounts quietly stopped working when I upgraded from 0.26.0 to 0.27.0, for some reason I couldn't figure out (the deprecation of bind mounts in testcontainers wasn't supposed to break them)

I'm interested in this. Could you share a repo where it happened? As you pointed out, they should not break any client code as per the deprecation, although I could have committed a mistake during the deprecation path.

> named volumes don't get cleaned up by ryuk

Did you manually created the volumes? If not I'd expect testcontainers-go to label the volumes accordingly, so that Ryuk prunes them. Could you run Ryuk in verbose mode, so you're be able to check what volumes are pruned?: https://github.com/testcontainers/moby-ryuk/blob/430d3338e3af80816a0a7445a90b4a1e58b1c710/main.go#L330-L337.

In order to do so: https://golang.testcontainers.org/features/configuration/#customizing-ryuk-the-resource-reaper. Since `v0.27.0` it's possible to run Ryuk in verbose mode.

> The testcontainers-go API makes it really tempting to set a non-empty name

Mmmm could you share how did you create the volumes? I think this answers my previous question so you probably used the API. I'm interested in how so I can double check and identify where the experience is broken.

srabraham:
Regarding bind mount breakage at 0.27.0, I made a separate bug report here with a very simple reproduction: https://github.com/testcontainers/testcontainers-go/issues/2179

srabraham:
[...]

PR #2191, opened 2024-02-01
fix: all mounts should contain the testcontainers labels

mdelapenya:
## What does this PR do?
Adds the tescontainers-go labels to all the mounts.

## Why is it important?
Otherwise, Ryuk won't remove them.

## Related issues

- Relates #2163
- Relates https://github.com/testcontainers/moby-ryuk/issues/103

## How to test this PR

Run this test with and without the fix. Then monitor the volumes with `watch -n 1 docker volume ls`

```shell
go test -timeout 600s -run ^TestContainerCreationWithVolumeAndFileWritingToIt$ github.com/testcontainers/testcontainers-go -count=1 -v
```

After the fix, the `volumeName` volume should be pruned by Ryuk. Before it, not.


srabraham (approved):
Thank you!

PR #2195, opened 2024-02-04
fix: go doc reference broken image

mmorel-35:
## What does this PR do?

fix go doc reference broken image

## Why is it important?

Make link more accessible


mdelapenya (approved):
Good catch! Thanks!

PR #2199, opened 2024-02-05
feat (postgres): support for creating and restoring Snapshots

Minivera:
## What does this PR do?

I recently went through building a complex testing setup for an entire webservice using testcontainers-go, which taught me some important tips and tricks about the library. I have added these in the documentation wherever I thought they could be useful.

I also added a new example for using the postgres module and making each test use a clean database. It's a pretty useful tool, and it's available as an example in the Java version, so I've added it here as well.

## Why is it important?

During my own work, I ran into some roadblocks where the docs did not offer all the information I needed to get things working. I worked things with the team and by reading the official Java example. I thought it'd be a good idea to upstream those learnings to the library.

## Related issues

- 

## How to test this PR

Run the postgres example, it should pass and should show how we can use a clean database in multiple tests.


docs/examples/postgres.md
Minivera, line 1: Self-review: I'm not entirely sure how to get this example in the sidebar 🤔

mdelapenya (commented):
Thanks for this PR! I do appreciate any improvement to the docs, reverb not when they come after suffering them first hand.

This LGTM, although I'd recommend moving the coffee in the examples/postgres to the module. If like to have the code you added into the testable examples (e.g. func ExampleFoo) of the postgres module, so they are automatically run on the CI.

Does it make sense to you?

docs/examples/postgres.md
mdelapenya: Please see https://github.com/testcontainers/testcontainers-go/pull/2199#pullrequestreview-1863997990

The docs are using mkdocs system, so the menu is driven by the `mkdocs.yml` file in the root directory. But I'd try to merge your proposal with the existing testable examples in the postgres module itself.

Minivera:
@mdelapenya do you mean to add the snapshot and function officially to the module so anyone can use them, then move the test there once those functions are added? I'd assume I would also change the example accordingly.

mdelapenya:
Yeah, I'd remove anything related to the postgres example, having everything under the postgres module.

mkdocs.yml
```diff
@@ -95,6 +95,7 @@ nav:
         - examples/consul.md
         - examples/nginx.md
         - examples/toxiproxy.md
+        - examples/postgres.md
```
mdelapenya, line 98: Let's remove this from the docs site, as we want to have everything in module page
```suggestion
```

docs/examples/postgres.md
```diff
@@ -0,0 +1,13 @@
+# Reusable Postgres Database
+
+## Introduction
+
+This example shows the usage of the postgres module's Snapshot feature to give each test a clean database without having
```
mdelapenya, line 5: I have this idea: we can add an `Examples` section for each module, as the last element in the menu for the postgres module. Then I'd move the entire content of this file there, looking like this:

```md
## Examples

### Using Snapshots
This example shows the usage of the postgres module's Snapshot feature to give each test a clean database without having
to recreate the database container on every test or run heavy scripts to clean your database. This makes the individual
tests very modular, since they always run on a brand-new database.

<!--codeinclude-->
[Test with a reusable Postgres container](../../modules/postgres/postgres_test.go) inside_block:snapshotAndReset
<!--/codeinclude-->
```

Wdyt?

mdelapenya (commented):
I added a few comments regarding the example. Once resolved I think we are done and we can merge this indeed needed improvement.

docs/examples/postgres.md
mdelapenya: The thing in the Go project is that we have pkg.go.dev, so all the testable examples land there. That's why I like testable examples, although a balance between real tests and examples is needed.

docs/features/files_and_mounts.md
```diff
@@ -67,7 +71,27 @@ It's important to notice that, when copying the directory to the container, the
 
 You can leverage the very same mechanism used for copying files to a container, but for directories.:
 
-1. The first way is using the `Files` field in the `ContainerRequest` struct, as shown in the previous section, but using the path of a directory as `HostFilePath`.
+1. The first way is using the `Files` field in the `ContainerRequest` struct, as shown in the previous section, but using the path of a directory as `HostFilePath`. Like so:
+
+```go
+ctx := context.Background()
```
mdelapenya, line 77: If possible, this code snippet should come from an embedded test, so that the snippet does not get outdated. I think it relates to #2202

modules/postgres/postgres.go
```diff
@@ -141,3 +143,85 @@ func RunContainer(ctx context.Context, opts ...testcontainers.ContainerCustomize
...
+
+// SnapshotOption is the type for passing options to the snapshot function of the database
+type SnapshotOption func(container *snapshotConfig) *snapshotConfig
+
+// WithSnapshotName adds a specific name to the snapshot database created from the main database defined on the
+// container. The snapshot must not have the same name as your main database, otherwise it will be overwritten
+func WithSnapshotName(name string) SnapshotOption {
+	return func(container *snapshotConfig) *snapshotConfig {
+		container.snapshotName = name
+		return container
```
mdelapenya, line 159:
```suggestion
	return func(cfg *snapshotConfig) *snapshotConfig {
		cfg.snapshotName = name
		return cfg
```

mdelapenya (approved):
LGTM, thanks!

PR #2210, opened 2024-02-14
chore(ci): do not fail fast main tests in different Go versions

mdelapenya:
## What does this PR do?
This PR removes the fail-fast situation when a CI build fails, making all the build in the following matrices to continue:

- main job
- reaper-off
- rootless docker

## Why is it important?
Many times the failure could be caused for some flakiness in that particular build, and its sibling could not be affected.

We run all the tests in the two current releases of Go, so if it fails in Go 1.20.x we still want to continue the execution for Go 1.21.x

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,7 @@ jobs:
     uses: ./.github/workflows/ci-test-go.yml
     with:
       go-version: ${{ matrix.go-version }}
-      fail-fast: false
+      fail-fast: true
       platform: ${{ matrix.platform }}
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -33,7 +33,7 @@ jobs:
     uses: ./.github/workflows/ci-test-go.yml
     with:
       go-version: ${{ matrix.go-version }}
-      fail-fast: true
+      fail-fast: false
```
mdelapenya, line 36:
```suggestion
      fail-fast: true
```

PR #2240, opened 2024-02-16
feat(modules.clickhouse): add `WithContainerName` option

dozer111:
Feature

## What does this PR do?

Add `WithContainerName` option to clickhouse module

## Why is it important?

Because in my pratice sometimes I need to interact with docker containers. 

I have macros with concrete names, so default docker naming behavior break down this possibility

## How to test this PR

I wrote test to it

### DIFF
--- a/modules/clickhouse/clickhouse.go
+++ b/modules/clickhouse/clickhouse.go
@@ -170,6 +170,15 @@ func WithYamlConfigFile(configFile string) testcontainers.CustomizeRequestOption
 	}
 }
 
+// WithContainerName sets the custom container name
+// It can be used to define a concrete container name.
+// If it is not specified, then the default docker value will be used.
+func WithContainerName(containerName string) testcontainers.CustomizeRequestOption {
+	return func(req *testcontainers.GenericContainerRequest) {
+		req.Name = containerName
+	}
+}
+
 // WithDatabase sets the initial database to be created when the container starts
 // It can be used to define a different name for the default database that is created when the image is first started.
 // If it is not specified, then the default value("clickhouse") will be used.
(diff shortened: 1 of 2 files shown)


mdelapenya:
Hi @dozer111 thanks for opening this PR. Unfortunately, I don't think we want to add this behaviour to the public API of any container, as setting custom names is, in general, disregarded for testing: imagine running tests in parallel.

On the other hand, you could achieve this by leveraging the `CustomizeRequest` option, which allows overriding the entire container request: https://golang.testcontainers.org/modules/clickhouse/#customising-the-containerrequest

```golang
container, err := RunContainer(ctx,
    /* Other module options */
    testcontainers.CustomizeRequest(testcontainers.GenericContainerRequest{
            Name: "my-custom-name",
        },
    }),
)
```

Wdyt?

dozer111:
np, okay, thank u

mdelapenya:
Closing as pero the discussion here, thanks!

dozer111:
what about Hostname?(`WithHostname` for example)

now i`m trying to build clickouse cluster of 3 nodes with zookeeper

---

it more valuable, or also trash?

mdelapenya:
You can still configure the container request:

```go
container, err := RunContainer(ctx,
    /* Other module options */
    testcontainers.CustomizeRequest(testcontainers.GenericContainerRequest{
            Hostname: "my-custom-hostname",
        },
    }),
)
```

I'm wondering if we should deprecate the Hostname (and probably some others) in favor of using the ConfigModifier function (for advanced use cases like this one).

eddumelendez:
@dozer111 I think network aliases should make it work

PR #2243, opened 2024-02-17
Fix modulege template to succeed on make lint command

jespino:
## What does this PR do?

Added the `// nolint:gocritic` comment in the module `example_test.go` template
to make the `make lint` command succed and in consequence allow to create
modules without errors.

## Why is it important?

It is not critical, but it really confuses the module creator, also it requires
manual fix of the problem to be able to pass the linter check.

## How to test this PR

You can try to create a new module before and after the PR, you will see that
the before state is going to lead you to an error executing make lint, and the
new one passes properly.


mdelapenya (approved):
Good catch! LGTM

PR #2244, opened 2024-02-17
Smocker module

jespino:
## What does this PR do?

Add smocker module that is useful to generate http mocks

## Why is it important?

Similar to mock-server, smocker can provide a good option for mocking third party http services for testing purposes.

## How to test this PR

This PR include some unit tests in it


mdelapenya:
@jespino We have added a new section in the module docs regarding hosting modules. I'd recommend you taking a look first: https://golang.testcontainers.org/modules/#before-creating-a-new-module

Please let us know what you think after that 🙏

jespino:
@mdelapenya I'm not 100% sure where is the "cut-line" here. I think the document doesn't explicitly say "What goes in community modules and what goes in the official repo", not having that clear puts contributors in a difficult situation.

mdelapenya:
@jespino indeed, I don't think there is a clear cut-line. We cannot force anybody to host the repo if they don't want or don't know what to do. In that doc we explain both alternatives (pros and cons), but I'm not sure we can draw that clear red line. We can discuss it case by case.

jespino:
@mdelapenya I'm moving this to my own personal repo here: https://github.com/jespino/testcontainers-go-smocker

PR #2247, opened 2024-02-20
bug: allow start container with reuse in different test package

### LINKED ISSUE #1277: [Bug]: Reuse ignores WaitingFor in container request

Alviner:
## What does this PR do?

When creating reusable container from different test binaries in parallel, you can get "name conflict" from docker daemon, because different binaries don't have a shared mutex

Instead of exiting with error when receiving "name conflict", we will try to wait for the container creation to complete using the ContainerList method and call ready/started hooks after.

## Why is it important?
This PR allows a reusable container to be used in parallel test binaries run from the "go test ./..." command.

## Related issues

- Closes #1277
- Closes #1572 
- Relates #809


Alviner:
@mdelapenya Hi! Please take a look, I encountered the same issue, refreshed the code from #1572

docker.go
```diff
@@ -1187,7 +1224,24 @@ func (p *DockerProvider) ReuseOrCreateContainer(ctx context.Context, req Contain
 		terminationSignal:   termSignal,
 		stopLogProductionCh: nil,
 		logger:              p.Logger,
-		isRunning:           c.State == "running",
+		lifecycleHooks: []ContainerLifecycleHooks{
+			{
+				PostStarts:  allHooks.PostStarts,
+				PostReadies: allHooks.PostReadies,
```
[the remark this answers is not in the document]
Alviner: let's add all of them, but not all of them will be executed, because this flow doesn't have all steps like with creation one. Does it acceptable?

docker.go
```diff
@@ -1178,6 +1208,13 @@ func (p *DockerProvider) ReuseOrCreateContainer(ctx context.Context, req Contain
 		}
 	}
 
+	// default hooks include logger hook and pre-create hook
```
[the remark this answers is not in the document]
mdelapenya: Yeah, let's not add the copy file hook. I think it's not need when reused

Alviner:
@mdelapenya Are there some flaky tests? I just successfully flew the checks in my fork)

https://github.com/Alviner/testcontainers-go/actions/runs/7993042944/

mdelapenya (approved):
LGTM, thanks!

Just to let you know, we are thinking about revamping how the reusable containers work, in a more consistent and deterministic manner. So this code could change or be removed. But until that, this is perfectly fine, thanks for your time contributing it!

mdelapenya:
BTW I'm renaming the PR to bug, as this is fixing the existing behaviour.

docker.go
```diff
@@ -1178,6 +1208,13 @@ func (p *DockerProvider) ReuseOrCreateContainer(ctx context.Context, req Contain
...
+	defaultHooks := []ContainerLifecycleHooks{
+		DefaultLoggingHook(p.Logger),
+		defaultReadinessHook(),
```
mdonkers, line 1214: Question @Alviner / @mdelapenya : is there any good reason why the `defaultReadinessHook()` should be added _also_ for a container being reused?

Reason I'm asking is that I noticed quite some overhead when changing testcontainers from `v0.28.0` to `v0.29.1` and was able to 'bisect' the issue to this PR.

As we're heavily relying on reused containers, this overhead is quite noticeable:
```
$ time make integration-test
...
real	0m4.555s
user	0m10.373s
sys	0m1.093s
```
before. Opposed to now:
```
real	0m9.177s
user	0m8.254s
sys	0m1.094s
```

When the container is initially created (and due to the mutex) it should have been fully started and the 'readiness' state was already checked. So from my perspective I don't see a reason to keep checking this.

Alviner: > Question @Alviner / @mdelapenya : is there any good reason why the `defaultReadinessHook()` should be added _also_ for a container being reused?
> 
> Reason I'm asking is that I noticed quite some overhead when changing testcontainers from `v0.28.0` to `v0.29.1` and was able to 'bisect' the issue to this PR.
> 
> As we're heavily relying on reused containers, this overhead is quite noticeable:
> 
> ```
> $ time make integration-test
> ...
> real	0m4.555s
> user	0m10.373s
> sys	0m1.093s
> ```
> 
> before. Opposed to now:
> 
> ```
> real	0m9.177s
> user	0m8.254s
> sys	0m1.094s
> ```
> 
> When the container is initially created (and due to the mutex) it should have been fully started and the 'readiness' state was already checked. So from my perspective I don't see a reason to keep checking this.

When you run tests from different packages, they run in separate processes with a single common parent. Therefore, if you reuse containers in these packages, you need to wait for the readiness hook, which effectively triggers waitFor in most situations.

mdonkers: Right, ok. Then the [mutex](https://github.com/testcontainers/testcontainers-go/blob/09648d156cd13a650925f423768c5ee7b1f72734/generic.go#L68) would of course no longer guarantee readiness of the Container.

But then I do think a solution for another bug I found (https://github.com/testcontainers/testcontainers-go/issues/2445) might resolve it to some extend. Because if there's a way to (explicitly) 'scope' a reusable container to the SessionID, then it's guaranteed its also always the same mutex instance. But I'm waiting for some feedback there to do a PR :-)

mdelapenya: Thank you all for bringing this topic, and for being so proactive in the discussion 🙇 

We are working in a better reusable experience for testcontainers-go. At the moment, it was initially built as an experimental feature that was immediately used when landed, so now it's part of the public API 😅 so difficult to remove without causing trouble to users.

The new experience will be improved and more robust and provided through Testcontainers Desktop.

mdonkers: Thx for the additional insight.
But does that mean no changes are expected or even approved just for the library? I don't think it needs to be removed, but there must be a way to both fix the existing behaviour and not break others...
I surely hope it would not require using the TestContainers desktop app since that would not work in our CI pipeline. And currently the "reuse" flag is causing actual issues: https://github.com/testcontainers/testcontainers-go/issues/2445#issuecomment-2016773569

PR #2257, opened 2024-02-20
chore: use logger.PrintXX instead of fmt.PrintXX

stevenh:
## What does this PR do?

Use logger for printing output instead of raw output to Stdout which causes issues with running tests.

## Why is it important?

Using raw Stdout via fmt.PrintXX methods corrupts the output when running tests and benchmarks.


mdelapenya (approved):
Thanks for this quick fix! Much appreciated! 🙇

PR #2259, opened 2024-02-21
feat: WithLogger ContainerCustomizer support

stevenh:
## What does this PR do?

Add support to use `WithLogger` as a `ContainerCustomizer` so that callers have an easy way to configure the logger of a container.

Add tests for `WithLogger`.

Validate `Logger` and `LoggerOption` implement the required interfaces.

Also fix a comment typo and remove an unneeded return.

## Why is it important?

Allows users to easily configure container loggers which is useful when using modules e.g. postgres or redis.

## How to test this PR

```shell
go test -run=^TestWithLogger$
```

### DIFF
--- a/docs/features/common_functional_options.md
+++ b/docs/features/common_functional_options.md
@@ -31,6 +31,26 @@ func (g *TestLogConsumer) Accept(l Log) {
 }
 ```
 
+#### WithLogger
+
+- Not available until the next release of testcontainers-go <a href="https://github.com/testcontainers/testcontainers-go"><span class="tc-version">:material-tag: main</span></a>
+
+If you need to either pass logger to a container, you can use `testcontainers.WithLogger`.
+
+In this example we also use `TestLogger` which writes to the passed in `testing.TB` using `Logf`.
+The result is that we capture all logging from the container into the test context meaning its
+hidden behind `go test -v` and is associated with the relevant test, providing the user with
+useful context instead of appearing out of band.
+
+```golang
+func TestHandler(t *testing.T) {
+    logger := TestLogger(t)
+    _, err := postgresModule.RunContainer(ctx, testcontainers.WithLogger(logger))
+    require.NoError(t, err)
+    // Do something with container.
+}
+```
+
 Please read the [Following Container Logs](/features/follow_logs) documentation for more information about creating log consumers.
 
 #### Wait Strategies
--- a/docs/features/common_functional_options.md
+++ b/docs/features/common_functional_options.md
@@ -37,6 +37,9 @@ func (g *TestLogConsumer) Accept(l Log) {
 
 If you need to either pass logger to a container, you can use `testcontainers.WithLogger`.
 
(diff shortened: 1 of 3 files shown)


logger.go
```diff
@@ -24,37 +32,51 @@ func LogDockerServerInfo(ctx context.Context, client client.APIClient, logger Lo
...
 func (o LoggerOption) ApplyGenericTo(opts *GenericProviderOptions) {
 	opts.Logger = o.logger
 }
 
+// ApplyDockerTo implements DockerProviderOption.
 func (o LoggerOption) ApplyDockerTo(opts *DockerProviderOptions) {
 	opts.Logger = o.logger
 }
 
+// Customize implements ContainerCustomizer.
```
mdelapenya, line 70: Because this option is now available to the modules, we need to document it. You already know how to do it. See https://github.com/testcontainers/testcontainers-go/pull/2260/files#r1498933117

Thanks! 🙏

options.go
```diff
@@ -67,7 +66,6 @@ func WithImage(image string) CustomizeRequestOption {
 	}
 }
 
-// imageSubstitutor {
```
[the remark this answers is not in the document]
stevenh: Have restored but noted these are [causing corrupt in the docs](https://pkg.go.dev/github.com/testcontainers/testcontainers-go#ImageSubstitutor).
image


Seems like the fix is to just add a blank line between the tag and the method, if there's no objections I'll do that a separate PR?

docs/features/common_functional_options.md
```diff
@@ -31,6 +31,21 @@ func (g *TestLogConsumer) Accept(l Log) {
...
 
+#### WithLogger
+
+- Not available until the next release of testcontainers-go <a href="https://github.com/testcontainers/testcontainers-go"><span class="tc-version">:material-tag: main</span></a>
+
+If you need to either pass logger to a container, you can use `testcontainers.WithLogger` for example:
+
+```golang
+func TestHandler(t *testing.T) {
+    logger := TestLogger(t)
```
mdelapenya, line 42: Let's add a short paragraph explaining what `TestLogger` is and does. It can go right before the code snippet.

options.go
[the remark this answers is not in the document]
mdelapenya: Yes, please and thanks!

docs/features/common_functional_options.md
```diff
@@ -31,6 +31,26 @@ func (g *TestLogConsumer) Accept(l Log) {
 }
 ```
 
+#### WithLogger
+
+
+If you need to either pass logger to a container, you can use `testcontainers.WithLogger`.
+
```
mdelapenya, line 39: This is coming from the function docs, and I think it should live here too
```suggestion

!!!info
	Consider calling this before other "With" functions as these may generate logs.

```

mdelapenya (approved):
LGTM, thanks for adding this customisation of the container request!

PR #2261, opened 2024-02-22
Move the client tests into a test package

Minivera:
## What does this PR do?

## Why is it important?

## Related issues

- Relates to #2180


docker_mounts.go
```diff
@@ -81,6 +81,12 @@ func (s DockerTmpfsMountSource) GetTmpfsOptions() *mount.TmpfsOptions {
 	return s.TmpfsOptions
 }
 
+// PrepareMounts maps the given []ContainerMount to the corresponding
```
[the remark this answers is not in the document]
Minivera: Only in future tests at the moment. This method existed in a prior version of the repository and got removed (we still have a `docker_mount_test.go` test that is called `TestPrepareMount` or something like that. It'll be the main place where this public method will be used, https://github.com/testcontainers/testcontainers-go/pull/2202/files#diff-49f2ad5a83d38c239147ec1f8520f6dae1d9c82b1296dfa6ff15c461ed0594fbR45. This was likely a bad cherry-pick. I will remove it and leave it for when we review the mount tests 😄

utils_test.go
```diff
@@ -0,0 +1,41 @@
+package testcontainers_test
```
mdelapenya, line 1: I think this helper code and functions in it is outdated now that #2270 was merged

Minivera: Yep, I'll update now that this has merged 🙂 It all depended on which one merge first

PR #2262, opened 2024-02-22
docs: fix comment corruption

stevenh:
## What does this PR do?

Fix doc generation comments being included package documentation by ensuring they are separated by a blank line.

## Why is it important?

So that package documentation doesn't have processing information in them which is confusing to users.


mdelapenya (approved):
LGTM, thanks for taking care of the docs 🙏

PR #2265, opened 2024-02-23
feat: add ollama module

mdelapenya:
## What does this PR do?
This PR adds a Go module for Ollama, using the module generator to bootstrap the module.

We present examples on how to use it with an HTTP client and with Langchaingo.

To improve the performance, the module will automatically request GPUs if the docker host supports it, and will allow to Commit the container after the module in order to speed the following containers using that image as base.

## Why is it important?
A module for Ollama!


modules/ollama/options.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+// WithModel will run the given model, without any prompt.
+// If Ollama is not able to run the given model, it will fail to initialise.
+func WithModel(model string) testcontainers.CustomizeRequestOption {
+	return withModel(model, "")
+}
+
+func withModel(model string, prompt string) testcontainers.CustomizeRequestOption {
+	runCmds := []string{"ollama", "run", model}
```
eddumelendez, line 58: There is a difference between `run` and `pull` command. I think the command should be `pull` here in order to let the container ready with the model in it. WDYT?

Also the `run` command works with models that support chat. For example, it works with `orca-mini` but it doesn't with `all-minilm`

mdelapenya: @eddumelendez I added the pull option, so that the models are always pull & run.

Besides, I've added the Commit image support (see https://github.com/testcontainers/testcontainers-java/pull/8369), but not forcing users to pass an image but instead generate a random suffix at the end, which is returned in the Commit method

modules/ollama/ollama_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+	// commitOllamaContainer {
+	newImage, err := ollamaContainer.Commit(context.Background())
+	// }
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if newImage == "" {
+		t.Fatal("commit should not be empty")
```
mdelapenya, line 128:
```suggestion
		t.Fatal("new image should not be empty")
```

modules/ollama/options.go
mdelapenya: Solved in 343194004aee0b24ec331b3ef73d386880755159

eddumelendez (approved):

PR #2269, opened 2024-02-28
Move the docker tests into a test package

Minivera:
## What does this PR do?

## Why is it important?

## Related issues

- Relates to#2180


docker_test.go
```diff
@@ -1955,7 +1955,7 @@ func assertExtractedFiles(t *testing.T, ctx context.Context, container Container
 	}
 }
 
-func terminateContainerOnEnd(tb testing.TB, ctx context.Context, ctr Container) {
+func terminateContainerOnEnd(tb testing.TB, ctx context.Context, ctr testcontainers.Container) {
```
mdelapenya, line 1958: We need to remove it from here, as it's duplicated. But then, there will be other places where this function is still used. We probably need to do this move last.

Minivera: Yeah, I'm moving everything temporarily to the testhelpers file, then we should be able to move this back here once the whole project is done. Still a few PRs to go 😅

lifecycle.go
```diff
@@ -180,8 +180,8 @@ var defaultLogConsumersHook = func(cfg *LogConsumerConfig) ContainerLifecycleHoo
 	}
 }
 
-// defaultReadinessHook is a hook that will wait for the container to be ready
-var defaultReadinessHook = func() ContainerLifecycleHooks {
+// DefaultReadinessHook is a hook that will wait for the container to be ready
+func DefaultReadinessHook() ContainerLifecycleHooks {
```
mdelapenya, line 184: I don't think we need to expose these variables to the public API in order to run the tests. Do you see any other way to achieve that?

lifecycle.go
```diff
@@ -475,13 +475,13 @@ func (p *DockerProvider) preCreateContainerHook(ctx context.Context, req Contain
 	return nil
 }
 
-// combineContainerHooks it returns just one ContainerLifecycle hook, as the result of combining
+// CombineContainerHooks it returns just one ContainerLifecycle hook, as the result of combining
```
mdelapenya, line 478: Same here: we don't want users to combine the hooks, as this is internal to the project. Which makes me think we could move this to the internal package 🤔

mdelapenya (commented):
Left a few comments regarding exposing private functions and vars to the public API. Let's discuss about them before moving on

Thank you!

lifecycle.go
Minivera: The `DefaultSomething` ones I think I might have to keep as exposed because they're used in tests that end up imported as examples, like this one: https://github.com/testcontainers/testcontainers-go/pull/2269/files/2f20304bbcd6ef52418ae330efb70be4af90cef4#diff-8c1de2f5177b67b15485d779c1510b2fa4abeb3cd1e82c18d43d0c2e7871ed19R74. If we keep that test in the internal package, that would mean losing the `testcontainer` prefix 🤔

lifecycle.go
Minivera: This one could probably be moved yeah, I did a thorough search and none of the tests using it seem to be loaded as examples for the docs 🤔

lifecycle.go
Minivera: After some shuffling around, I managed to rename them 😄

PR #2294, opened 2024-03-01
feature: allowing building of a target in multi-stage build

ClaytonNorthey92:
feature: if a Dockerfile specifies build stages, allow the user to set the target build stage in ContainerRequest

## What does this PR do?

Adds a `Target` option to `ContainerRequest` to allow a user to specify a multi-stage build stage.

## Why is it important?

Many projects use multi-stage builds, we should enable those to be used via testcontainers-go

## Related issues

- None

## How to test this PR

* automated tests handle this


ClaytonNorthey92:
realizing this is already solved with `BuildOptionsModifier` 🤦 closing

mdelapenya:
I was about to suggest it! Thanks for taking care of it 🙇

PR #2336, opened 2024-03-08
fix: consul race on HTTP port

### LINKED ISSUE #2335: consul: Update wait strategy
note this introduced a flake. The former example was waiting for the listen port (as are most ad-hoc usages on github). This passes before the api is ready failing tests. I think it should be reverted to what the example used:

```
		testcontainers.WithWaitStrategy(wait.ForListeningPort("8500/tcp")),
```

_Originally posted by @codefromthecrypt in https://github.com/testcontainers/testcontainers-go/pull/2194#discussion_r1516991862_

codefromthecrypt:
## What does this PR do?

Fixes a race where tests can fail due to trying to use the HTTP port before it is listening

## Why is it important?

I swapped out host-based consul in a private project with testcontainers, and all the tests failed until I swapped the health check to block until the port is ready

## Related issues

- Closes #2335

## How to test this PR

run `make test` in the modules/consul directory

### DIFF
--- a/modules/consul/consul.go
+++ b/modules/consul/consul.go
@@ -14,7 +14,7 @@ const (
 )
 
 const (
-	DefaultBaseImage = "docker.io/hashicorp/consul:1.18"
+	DefaultBaseImage = "docker.io/hashicorp/consul:1.15"
 )
 
 // ConsulContainer represents the Consul container type used in the module.
--- a/modules/consul/consul.go
+++ b/modules/consul/consul.go
@@ -14,7 +14,7 @@ const (
 )
 
 const (
-	DefaultBaseImage = "docker.io/hashicorp/consul:1.15"
+	DefaultBaseImage = "docker.io/hashicorp/consul:1.18"
 )
 
 // ConsulContainer represents the Consul container type used in the module.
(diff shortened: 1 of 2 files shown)


modules/consul/consul.go
```diff
@@ -14,7 +14,7 @@ const (
 )
 
 const (
-	DefaultBaseImage = "docker.io/hashicorp/consul:1.15"
+	DefaultBaseImage = "docker.io/hashicorp/consul:1.18"
```
codefromthecrypt, line 17: I also updated the version. when this change was made it ported the same version of consul used in the example, but it isn't the current one.

mdelapenya: I'll give you some historical context about this: we usually define a base image as a reference, but we encourage always passing the `WithImage` option to make it clear what image to use: our examples in the website always include the image. So updating the default base image is just a matter of checking that the code of the module behaves as expected for that image. Like a minimal supported version. For this particular case, it will work because there is no specific logic, but if you check Elasticsearch module, it has custom code to detect versions and perform tasks according to that version. Therefore a simple bump in the default image could lead to changes in the implementation of the module.

But because we are not doing/building a support matrix for any module, it's fine to keep the default image as is, and probably change the code examples, only.

We are having conversations internally on how to address this: testcontainers-java forces passing the image as part of the "RunContainer" method, which is something we could do. But it's not decided yet.

codefromthecrypt: ok maybe it is best to just keep it coherent as it was vs different versions here and there. I'll revert this part. thanks for the context!

mdelapenya (approved):
LGTM, thanks!!

I'm keeping this is a bug instead of an enhancement as per 2 reasons: your project failing because of this, and our mistake converting the former example into a module. Good catch!

PR #2341, opened 2024-03-11
feat: add distribution-registry module

mdelapenya:
Render URL: https://deploy-preview-2341--testcontainers-go.netlify.app/modules/registry/

## What does this PR do?
This PR adds a new module for the distribution-registry (https://hub.docker.com/_/registry).

The module piggybacks from existing tests for the Docker auth feature, so it was really easy to convert them into a module.

The module comes with the following options:

- allows to add data including custom images
- allows to set up authentication, in two manners:
  - passing the string representation of an htpasswd file
  - an htpasswd file that will be copied into the container

The module also exposes the following methods:

- retrieve the HTTP address of the registry, so that it's possible to connect to it with an HTTP client.
- check if an image ref exists in the registry
- push an image ref to the registry
- delete an image red from the registry

Tests and testable examples have been added, including some for building images from the container registry the module creates. Please check the `testdata` directory including the blobs and layers for a custom redis image, that will be used in those tests.

## Why is it important?
Support for writing tests when accessing a private registry.

## Related issues

- Supersedes and Closes #1165


mdelapenya:
Q: should we remove the docker auth tests for the private registry, given they now live in the module? 🤔  Thoughts?

mdelapenya:
Merge after #2349, as it contained a commit for the HTTP headers

davidesalerno (approved):

davidesalerno:
I think that the changes can be merged

Issue #2134, opened 2024-01-22
Add Must methods for other database modules


mdelapenya:
Replicate what the CockroachDB module is doing with `MustConnectionString`, adding them to any database module.

              Thanks! Also, asking because I don't see this for other database modules

_Originally posted by @eddumelendez in https://github.com/testcontainers/testcontainers-go/pull/2131#discussion_r1462000993_

bearrito:
@mdelapenya I haven't contributed to this project before, but I can take this. It looks fairly straightforward.

mdelapenya:
Fantastic! Looking forward to it 😎

bearrito:
This could be closed.

PR #2343, opened 2024-03-12
feat(MustConn): Add MustConnectionString on (some) dbs

bearrito:
## What does this PR do?

Adds `MustConnectionString` on a variety of DB containers. I skipped MSSQL because that was failing to run tests for me locally.

There are probably more I can add them. Should I just add them to all containers with a `ConnectionString` method?

## Why is it important?

Not sure again, just looking at `good first issue` tags. 
https://github.com/testcontainers/testcontainers-go/issues/2134

## How to test this PR
Run the unit tests.

One point about the unit tests, I could not figure out a way to get a `panic` out of the `ConnectionString` methods. I looked for ways to corrupt/mutate the exposed_ports and host, but could not figure out an entrypoint. So there is no test that it actually panics if unable to provide the connstring.

### DIFF
--- a/modules/mysql/mysql_test.go
+++ b/modules/mysql/mysql_test.go
@@ -8,6 +8,7 @@ import (
 
 	// Import mysql into the scope of this package (required)
 	_ "github.com/go-sql-driver/mysql"
+	"github.com/stretchr/testify/require"
 
 	"github.com/testcontainers/testcontainers-go/modules/mysql"
 )
@@ -34,6 +35,8 @@ func TestMySQL(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+	mustConnectionString := container.MustConnectionString(ctx,"tls=skip-verify")
+	require.Equal(t,connectionString,mustConnectionString)
 
 	db, err := sql.Open("mysql", connectionString)
 	if err != nil {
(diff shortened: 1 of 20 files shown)


modules/mysql/mysql_test.go
```diff
@@ -34,6 +35,8 @@ func TestMySQL(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+	mustConnectionString := container.MustConnectionString(ctx,"tls=skip-verify")
+	require.Equal(t,connectionString,mustConnectionString)
```
mdelapenya, line 39: Super NIT: all the assertions in this module are done without testify.

I have to think about if using testify everywhere, or remove it everywhere, but not there yet. I'd keep this file at it was, without testify 🙏

modules/nats/nats_test.go
```diff
@@ -32,6 +33,8 @@ func TestNATS(t *testing.T) {
 	if err != nil {
 		t.Fatalf("failed to get connection string: %s", err)
 	}
+	mustUri := container.MustConnectionString(ctx)
+	require.Equal(t,uri,mustUri)
```
mdelapenya, line 37: Same here about testify.

mdelapenya (commented):
> One point about the unit tests, I could not figure out a way to get a panic out of the ConnectionString methods.

It's fine, I would not worry to get 100% coverage meanwhile it's clear, which this PR is.

Other than that, LGTM, thank you!

modules/mysql/mysql_test.go
bearrito: Will do.

mdelapenya (approved):
LGTM, thanks for implementing this :clap:

PR #2345, opened 2024-03-12
fix: data race on container run, caused by the otelhttp dependency

stevenh:
## What does this PR do?
Fix a data race on container run by updating otelhttp:
https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4895

## Why is it important?

Usage randomly fails due to a data race when testing.


mdelapenya:
Hey @stevenh thanks for bumping the dependency. Could you check the compose module too? That dependency makes the module to not compile.

Thanks!

stevenh:
Thanks @mdelapenya should be all fixed.

mdelapenya (approved):
LGTM, thanks!

PR #2346, opened 2024-03-12
fix: logging deadlock

### LINKED ISSUE #2328: [Bug]: Container logging exits prematurely with error
### Testcontainers version

0.29.1

Yes

### Host OS

Linux

### Host arch

x86

### Go version

1.22

### Docker version

```shell
(NOTE: using podman, not docker)

Version:      3.4.4
API Version:  3.4.4
Go Version:   go1.18.1
Built:        Wed Dec 31 19:00:00 1969
OS/Arch:      linux/amd64
```

### Docker info

```shell
host:
  arch: amd64
  buildahVersion: 1.23.1
  cgroupControllers:
  - memory
  - pids
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: 'conmon: /usr/bin/conmon'
    path: /usr/bin/conmon
    version: 'conmon version 2.0.25, commit: unknown'
  cpus: 32
  distribution:
    codename: jammy
    distribution: ubuntu
    version: "22.04"
  eventLogger: journald
  hostname: noah-workstation
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 2109978069
      size: 1
    - container_id: 1
      host_id: 10000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 2109978069
      size: 1
    - container_id: 1
      host_id: 10000
      size: 65536
  kernel: 6.1.0-1033-oem
  linkmode: dynamic
  logDriver: journald
  memFree: 101372432384
  memTotal: 134160068608
  ociRuntime:
    name: crun
    package: 'crun: /usr/bin/crun'
    path: /usr/bin/crun
    version: |-
      crun version 0.17
      commit: 0e9229ae34caaebcb86f1fde18de3acaf18c6d9a
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +YAJL
  os: linux
  remoteSocket:
    exists: true
    path: /run/user/2109978069/podman/podman.sock
  security:
    apparmorEnabled: false
    capabilities: CAP_CHOWN,CAP_DAC_OVERRIDE,CAP_FOWNER,CAP_FSETID,CAP_KILL,CAP_NET_BIND_SERVICE,CAP_SETFCAP,CAP_SETGID,CAP_SETPCAP,CAP_SETUID,CAP_SYS_CHROOT
    rootless: true
    seccompEnabled: true
    seccompProfilePath: /usr/share/containers/seccomp.json
    selinuxEnabled: false
  serviceIsRemote: false
  slirp4netns:
    executable: /usr/bin/slirp4netns
    package: 'slirp4netns: /usr/bin/slirp4netns'
    version: |-
      slirp4netns version 1.0.1
      commit: 6a7b16babc95b6a3056b33fb45b74a6f62262dd4
      libslirp: 4.6.1
  swapFree: 8589930496
  swapTotal: 8589930496
  uptime: 363h 0m 53.36s (Approximately 15.12 days)
plugins:
  log:
  - k8s-file
  - none
  - journald
  network:
  - bridge
  - macvlan
  volume:
  - local
registries: {}
store:
  configFile: /home/noah/.config/containers/storage.conf
  containerStore:
    number: 4
    paused: 0
    running: 0
    stopped: 4
  graphDriverName: overlay
  graphOptions: {}
  graphRoot: /home/noah/.local/share/containers/storage
  graphStatus:
    Backing Filesystem: extfs
    Native Overlay Diff: "true"
    Supports d_type: "true"
    Using metacopy: "false"
  imageStore:
    number: 25
  runRoot: /run/user/2109978069/containers
  volumePath: /home/noah/.local/share/containers/storage/volumes
version:
  APIVersion: 3.4.4
  Built: 0
  BuiltTime: Wed Dec 31 19:00:00 1969
  GitCommit: ""
  GoVersion: go1.18.1
  OsArch: linux/amd64
  Version: 3.4.4
```

### What happened?

I've found that `LogConsumer`s sometimes stop receiving logs prematurely and dump an error. 

I've built a minimal reproducible example of the specific issue. It consists of a very small container image based on [`distroless/base`](https://github.com/GoogleContainerTools/distroless). The container executes the following program:

```go
package main

import (
	"log"
	"time"
)

func main() {
	i := 0
	for {
		log.Printf("tick %d", i)
		i += 1
		time.Sleep(1 * time.Second)
	}
}
```

I've attached a prebuilt container to the issue as a tarball.

In order to reproduce the issue, first load the container tarball like so:

```
gunzip container.tar.gz
podman load -i container.tar
```

And then compile and run the following test:

```go
package container_test

import (
	"context"
	"fmt"
	"testing"
	"time"

	tc "github.com/testcontainers/testcontainers-go"
)

type StdoutLogConsumer struct{}

func (lc *StdoutLogConsumer) Accept(l tc.Log) {
	fmt.Printf("%s", l.Content)
}

func TestLogIssue(t *testing.T) {
	req := tc.ContainerRequest{
		Image: "localhost/log_repro:latest",
		LogConsumerCfg: &tc.LogConsumerConfig{
			Consumers: []tc.LogConsumer{&StdoutLogConsumer{}},
		},
	}

	ctx := context.Background()
	container, err := tc.GenericContainer(ctx, tc.GenericContainerRequest{
		ContainerRequest: req,
		Started:          true,
	})
	if err != nil {
		t.Fatal(err)
	}
	t.Cleanup(func() {
		if err := container.Terminate(ctx); err != nil {
			t.Fatal(err)
		}
	})

	time.Sleep(10 * time.Second)

	fmt.Println("test over")
}
```

I'd expect this test to print container logs showing 9-10 ticks and exit cleanly. Instead, it consistently stops displaying output after 6 ticks, and always outputs an error message:

```
2024/03/06 19:21:03 github.com/testcontainers/testcontainers-go - Connected to docker:
  Server Version: 3.4.4
  API Version: 1.40
  Operating System: ubuntu
  Total Memory: 127945 MB
  Resolved Docker Host: unix:///run/user/2109978069/podman/podman.sock
  Resolved Docker Socket Path: /run/user/2109978069/podman/podman.sock
  Test SessionID: c7ce1db5c519f0188cc6943c86f8faccf62034e6b4d4c1ab5bf6f8ff896ecf3b
  Test ProcessID: 36d03253-4e13-441e-8149-88e46fb4fa15
2024/03/06 19:21:03 🐳 Creating container for image testcontainers/ryuk:0.6.0
2024/03/06 19:21:03 ✅ Container created: 15f55739196d
2024/03/06 19:21:03 🐳 Starting container: 15f55739196d
2024/03/06 19:21:03 ✅ Container started: 15f55739196d
2024/03/06 19:21:03 🚧 Waiting for container id 15f55739196d image: testcontainers/ryuk:0.6.0. Waiting for: &{Port:8080/tcp timeout:<nil> PollInterval:100ms}
2024/03/06 19:21:03 🔔 Container is ready: 15f55739196d
2024/03/06 19:21:03 🐳 Creating container for image localhost/log_repro:latest
2024/03/06 19:21:03 ✅ Container created: f8583fd38a61
2024/03/06 19:21:03 🐳 Starting container: f8583fd38a61
2024/03/06 19:21:03 ✅ Container started: f8583fd38a61
2024/03/06 19:21:03 🔔 Container is ready: f8583fd38a61
2024/03/06 19:21:03 tick 0
2024/03/06 19:21:04 tick 1
2024/03/06 19:21:05 tick 2
2024/03/06 19:21:06 tick 3
2024/03/06 19:21:07 tick 4
2024/03/06 19:21:08 tick 5
container log error: EOF. Stopping log consumer: Headers out of sync2024/03/06 19:21:13 🐳 Terminating container: f8583fd38a61
test over
2024/03/06 19:21:14 🚫 Container terminated: f8583fd38a61
PASS
```

I did notice some interesting things about this while trying to debug, but I'm not sure what to make of them. In the main loop of `docker.startLogProduction()`, it seems like the `"use of closed network connection"` error block consistently triggers before the EOF error is triggered. In addition, it seems like attempting to restart the connection when EOF is received allows the log consumer to get a few more logs, e.g.:

```go
//_, _ = fmt.Fprintf(os.Stderr, "container log error: %+v. %s", err, logStoppedForOutOfSyncMessage)
//// if we would continue here, the next header-read will result into random data...
//return

now := time.Now()
since = fmt.Sprintf("%d.%09d", now.Unix(), int64(now.Nanosecond()))
goto BEGIN
```

### Relevant log output

_No response_

### Additional information

[container.tar.gz](https://github.com/testcontainers/testcontainers-go/files/14514644/container.tar.gz)

stevenh:
## What does this PR do?

Fix logging deadlock, causing lots of test timeouts.

This refactors how logging shutdown is handled, eliminating unnecessary captures, use idiomatic wait group to signal processor completion and remove unnecessary `nil` initialisation.

Fix race condition in log testing which was reading `Msg` while the processor was still running.

Switch to checking `GITHUB_RUN_ID` environment variable to detect GitHub as `XDG_RUNTIME_DIR` can be present in other situations.

@mdelapenya edit:
> Capture all errors in the lifecycle hooks, refactoring the code in a single `apply` function that will be used in all the lifecycle hook methods.

## Why is it important?

Tests and runtime is currently flakey due to race condition during start and stop of logging.

## How to test this PR

Testing is hard as it's fixing a random race condition, but tests should now be more reliable.

### DIFF
--- a/lifecycle.go
+++ b/lifecycle.go
@@ -2,6 +2,7 @@ package testcontainers
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"strings"
@@ -225,65 +226,40 @@ var defaultReadinessHook = func() ContainerLifecycleHooks {
 
 // creatingHook is a hook that will be called before a container is created.
 func (req ContainerRequest) creatingHook(ctx context.Context) error {
-	for _, lifecycleHooks := range req.LifecycleHooks {
-		err := lifecycleHooks.Creating(ctx)(req)
-		if err != nil {
-			return err
-		}
+	errs := make([]error, len(req.LifecycleHooks))
+	for i, lifecycleHooks := range req.LifecycleHooks {
+		errs[i] = lifecycleHooks.Creating(ctx)(req)
 	}
 
-	return nil
+	return errors.Join(errs...)
 }
 
-// createdHook is a hook that will be called after a container is created
+// createdHook is a hook that will be called after a container is created.
 func (c *DockerContainer) createdHook(ctx context.Context) error {
-	for _, lifecycleHooks := range c.lifecycleHooks {
-		err := containerHookFn(ctx, lifecycleHooks.PostCreates)(c)
-		if err != nil {
-			return err
-		}
-	}
-
-	return nil
+	return c.applyLifecycleHooks(ctx, false, func(lifecycleHooks ContainerLifecycleHooks) []ContainerHook {
+		return lifecycleHooks.PostCreates
+	})
 }
 
-// startingHook is a hook that will be called before a container is started
+// startingHook is a hook that will be called before a container is started.
 func (c *DockerContainer) startingHook(ctx context.Context) error {
-	for _, lifecycleHooks := range c.lifecycleHooks {
-		err := containerHookFn(ctx, lifecycleHooks.PreStarts)(c)
-		if err != nil {
-			c.printLogs(ctx, err)
-			return err
-		}
-	}
-
-	return nil
+	return c.applyLifecycleHooks(ctx, true, func(lifecycleHooks ContainerLifecycleHooks) []ContainerHook {
+		return lifecycleHooks.PreStarts
+	})
 }
 
-// startedHook is a hook that will be called after a container is started
+// startedHook is a hook that will be called after a container is started.
 func (c *DockerContainer) startedHook(ctx context.Context) error {
-	for _, lifecycleHooks := range c.lifecycleHooks {
-		err := containerHookFn(ctx, lifecycleHooks.PostStarts)(c)
-		if err != nil {
-			c.printLogs(ctx, err)
-			return err
-		}
-	}
-
-	return nil
+	return c.applyLifecycleHooks(ctx, true, func(lifecycleHooks ContainerLifecycleHooks) []ContainerHook {
+		return lifecycleHooks.PostStarts
+	})
 }
 
-// readiedHook is a hook that will be called after a container is ready
+// readiedHook is a hook that will be called after a container is ready.
 func (c *DockerContainer) readiedHook(ctx context.Context) error {
-	for _, lifecycleHooks := range c.lifecycleHooks {
-		err := containerHookFn(ctx, lifecycleHooks.PostReadies)(c)
-		if err != nil {
-			c.printLogs(ctx, err)
-			return err
-		}
-	}
-
-	return nil
+	return c.applyLifecycleHooks(ctx, true, func(lifecycleHooks ContainerLifecycleHooks) []ContainerHook {
+		return lifecycleHooks.PostReadies
+	})
 }
 
 // printLogs is a helper function that will print the logs of a Docker container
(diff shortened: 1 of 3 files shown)


stevenh:
Looks like 95% of tests are in the green, `Test with reaper off` seems to be a common failure is that a know issue?

stevenh:
Could you test with the reaper built with [this fix](https://github.com/testcontainers/moby-ryuk/pull/121) @bearrito and see if it corrects your test case?

You will need to build locally and tag a version then update testcontainers to use the new tagged image

stevenh:
> @mdelapenya @stevenh Not sure if this is helpful, but I ran into this issue (linked above) and debugged it.
> 
> The multi-consumer aspect didn't have anything to do with it. I could replicate this with a single consumer. I can force the bug about 1/15 times.
> 
> Here is the flow
> 
> 1. https://github.com/testcontainers/testcontainers-go/blob/main/docker.go#L726 This condition is met. We hit the GOTO
> 2. https://github.com/testcontainers/testcontainers-go/blob/main/docker.go#L710 This condition matches, we return. I don't have the error off hand by I believe it was `context cancelled`
> 3. https://github.com/testcontainers/testcontainers-go/blob/main/docker.go#L686 Defer runs
> 4. https://github.com/testcontainers/testcontainers-go/blob/main/docker.go#L791 There is no longer a consumer for stop. This deadlocks here, as that channel is unbuffered.

That seems to be an analysis on the old code, that shouldn't be possible with the new version?

bearrito:
@stevenh Probably not, I was more or less replying to https://github.com/testcontainers/testcontainers-go/pull/2346#issuecomment-2018211902. 

I was also hitting this in CI and wondered what was happening. I think your waitGroup approach should work.

If you like, I can commit my reduced test case where I was able to replicate this ( at least with high-probability), then test against that?

stevenh:
The `Test with reaper off` looks like a secondary issue. What's happening is the test hits cancels the main context causing the logger to stop. This returns  context cancelled error which subsequently returned by `stopLogProduction` in `terminatingHook` during the tests deferred called to `c.Terminate` and `c2.Terminate` which prevents `Terminate` from cleaning up properly as it returns early.

There's a few ways to fix:
1. Don't return context cancelled error from `stopLogProduction`.
2. Don't return context cancelled error from `defaultLogConsumersHook.PreTerminates`.
3. Ignore context cancelled errors from hooks in `Terminate`.
4. Perform best effort clean up in `Terminate` and return a multi error if needed.

It feels to my like 4. would be the best solution as it puts the caller in control of the behaviour.

A tertiary issue then becomes should all lifecycleHooks iterators e.g. `terminatedHook` and hence `containerHookFn` also be best effort?

Thoughts?

stevenh:
> @stevenh Probably not, I was more or less replying to [#2346 (comment)](https://github.com/testcontainers/testcontainers-go/pull/2346#issuecomment-2018211902).
> 
> I was also hitting this in CI and wondered what was happening. I think your waitGroup approach should work.
> 
[best-effort.patch](https://github.com/testcontainers/testcontainers-go/files/14844536/best-effort.patch)

> If you like, I can commit my reduced test case where I was able to replicate this ( at least with high-probability), then test against that?

If you have a reproduction and want to try with this branch + the attached patch which makes lifecycles and Terminate best effort that would be helpful to confirm it fixes your case too.

mdelapenya:
I like the best-effort patch for the lifecycle, which will make the potential errors more evident. For the terminate method, you want to continue with the terminated hook even if we failed the removal of the container, right?

stevenh:
> I like the best-effort patch for the lifecycle, which will make the potential errors more evident. For the terminate method, you want to continue with the terminated hook even if we failed the removal of the container, right?

Yes that makes sense to me, I have a version of this from last night I'll tweak and put up see what folks think.

lifecycle.go
```diff
@@ -304,49 +280,47 @@ func (c *DockerContainer) printLogs(ctx context.Context, cause error) {
...
-		err := containerHookFn(ctx, lifecycleHooks.PostTerminates)(c)
-		if err != nil {
-			return err
+	return c.applyLifecycleHooks(ctx, false, func(lifecycleHooks ContainerLifecycleHooks) []ContainerHook {
+		return lifecycleHooks.PostTerminates
+	})
+}
+
+// applyLifecycleHooks applies all lifecycle hooks reporting the container logs on error if logError is true.
+func (c *DockerContainer) applyLifecycleHooks(ctx context.Context, logError bool, hooks func(lifecycleHooks ContainerLifecycleHooks) []ContainerHook) error {
```
mdelapenya, line 312: 💯

mdelapenya (approved):
LGTM! Thank you for dedicating time to resolve concurrency issues, that are always hard to detect and fix, even more with legacy code, so I appreciate your efforts here 🙇

Issue #2264, opened 2024-02-23
[Bug]: Ryuk does not work with podman


dfioravanti:
### Testcontainers version

0.28

### Using the latest Testcontainers version?

Yes

### Host OS

MacOs

### Host arch

ARM

### Go version

1.22

### Docker version

```shell
Client: Docker Engine - Community
 Version:           25.0.3
 API version:       1.41 (downgraded from 1.44)
 Go version:        go1.21.6
 Git commit:        4debf411d1
 Built:             Tue Feb  6 20:42:40 2024
 OS/Arch:           darwin/arm64
 Context:           default

Server: linux/arm64/fedora-39
 Podman Engine:
  Version:          4.9.0
  APIVersion:       4.9.0
  Arch:             arm64
  BuildTime:        2024-01-24T11:07:09+01:00
  Experimental:     false
  GitCommit:
  GoVersion:        go1.21.6
  KernelVersion:    6.7.4-200.fc39.aarch64
  MinAPIVersion:    4.0.0
  Os:               linux
 Conmon:
  Version:          conmon version 2.1.8, commit:
  Package:          conmon-2.1.8-2.fc39.aarch64
 OCI Runtime (crun):
  Version:          crun version 1.14
commit: 667e6ebd4e2442d39512e63215e79d693d0780aa
rundir: /run/crun
spec: 1.0.0
+SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +CRIU +LIBKRUN +WASM:wasmedge +YAJL
  Package:          crun-1.14-1.fc39.aarch64
 Engine:
  Version:          4.9.0
  API version:      1.41 (minimum version 1.24)
  Go version:       go1.21.6
  Git commit:
  Built:            Wed Jan 24 11:07:09 2024
  OS/Arch:          linux/arm64
  Experimental:     false
```

### Docker info

```shell
Client: Docker Engine - Community
 Version:    25.0.3
 Context:    default
 Debug Mode: false
 Plugins:

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 1
 Server Version: 4.9.0
 Storage Driver: overlay
  Backing Filesystem: xfs
  Supports d_type: true
  Native Overlay Diff: false
  Using metacopy: true
  Supports shifting: true
  Supports volatile: true
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
[...]

Issue #538, opened 2022-09-29
[Bug]: Podman - Reaper creation fails


fpozzobon:
### Testcontainers version

0.14.0

### Using the latest Testcontainers version?

Yes

### Host OS

MacOs

### Host Arch

amd64

### Go Version

1.18

### Docker version

```shell
podman 4.0.1
```

### Docker info

```shell
host:
  arch: amd64
  buildahVersion: 1.24.1
  cgroupControllers:
  - memory
  - pids
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: conmon-2.1.0-2.fc35.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.1.0, commit: '
  cpus: 1
  distribution:
    distribution: fedora
    variant: coreos
    version: "35"
  eventLogger: journald
  hostname: localhost.localdomain
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 5.15.18-200.fc35.x86_64
  linkmode: dynamic
  logDriver: journald
  memFree: 1611247616
  memTotal: 2061381632
  networkBackend: netavark
  ociRuntime:
    name: crun
    package: crun-1.4.2-1.fc35.x86_64
    path: /usr/bin/crun
    version: |-
      crun version 1.4.2
      commit: f6fbc8f840df1a414f31a60953ae514fa497c748
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +CRIU +YAJL
  os: linux
  remoteSocket:
    exists: true
    path: /run/user/1000/podman/podman.sock
  security:
    apparmorEnabled: false
    capabilities: CAP_CHOWN,CAP_DAC_OVERRIDE,CAP_FOWNER,CAP_FSETID,CAP_KILL,CAP_NET_BIND_SERVICE,CAP_SETFCAP,CAP_SETGID,CAP_SETPCAP,CAP_SETUID,CAP_SYS_CHROOT
    rootless: true
    seccompEnabled: true
    seccompProfilePath: /usr/share/containers/seccomp.json
    selinuxEnabled: true
  serviceIsRemote: true
  slirp4netns:
    executable: /usr/bin/slirp4netns
[...]

PR #2347, opened 2024-03-13
redpanda: set entrypoint to the custom entrypoint file

bojand:
## What does this PR do?

Redpanda module has a custom entrypoint file `/entrypoint-tc.sh` that waits until the actual Redpanda broker node config file is mounted.
Then proceeds to call the normal Redpanda entryfile `/entrypoint.sh`.

The current implementation of the module sets the entrypoint file into the `Cmd` of the `ContainerRequest`. 
This seems to work with `Docker`, but does not work with `Podman`.

This results in an error when starting the container:

```sh
+ '[' '' = true ']'
+ exec /usr/bin/rpk /entrypoint-tc.sh redpanda start --mode=dev-container --smp=1 --memory=1G
Error: unknown command "/entrypoint-tc.sh" for "rpk"
Run 'rpk --help' for usage.
```

Since the intent is to override the entrypoint, we should be setting `Entrypoint` property as this is more appropriate.

In more detail, if we inspect the Docker container:

```json
{
    "Id": "52d5be52543a60e793fa4920fd81794fe646fff06d917c96d56cbdf164475925",
    "Created": "2024-03-12T18:42:46.003923967Z",
    "Path": "/entrypoint-tc.sh",
    "Args": [
        "redpanda",
        "start",
        "--mode=dev-container",
        "--smp=1",
        "--memory=1G"
    ],
    ...
    "Config": {
        "Hostname": "52d5be52543a",
        "Domainname": "",
        "User": "root:root",
        "AttachStdin": false,
        "AttachStdout": false,
        "AttachStderr": false,
        "ExposedPorts": {
            "8081/tcp": {},
            "8082/tcp": {},
            "9092/tcp": {},
            "9644/tcp": {}
        },
        "Tty": false,
        "OpenStdin": false,
        "StdinOnce": false,
        "Env": [
            "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
            "EDITOR=vi"
        ],
        "Cmd": [
            "/entrypoint-tc.sh",
            "redpanda",
            "start",
            "--mode=dev-container",
            "--smp=1",
            "--memory=1G"
        ],
        "Image": "redpandadata/redpanda:v23.3.2",
        "Volumes": {
            "/var/lib/redpanda/data": {}
        },
        "WorkingDir": "",
        "Entrypoint": [],
        "OnBuild": null,
        "Labels": {
            "org.opencontainers.image.authors": "Redpanda Data <hi@redpanda.com>",
            "org.testcontainers": "true",
            "org.testcontainers.lang": "go",
            "org.testcontainers.sessionId": "353c39fb1e5c8ee2313ec527210b4ab8fcda2274b067bd4d2360bff88574f774",
            "org.testcontainers.version": "0.28.0"
        }
    },
    ...
}
```

But if we inspect the Podman container:

```json
{
  "Id": "3f9bb85cc4761e06082d7add0c02937439eba3d05ec1febf48ad09667b7abae5",
  "Created": "2024-03-12T19:05:31.255025847Z",
  "Path": "/entrypoint.sh",
  "Args": [
    "/entrypoint-tc.sh",
    "redpanda",
    "start",
    "--mode=dev-container",
    "--smp=1",
    "--memory=1G"
  ],
...
"Config": {
    "Hostname": "3f9bb85cc476",
    "Domainname": "",
    "User": "root:root",
    "AttachStdin": false,
    "AttachStdout": false,
    "AttachStderr": false,
    "ExposedPorts": {
      "8081/tcp": {},
      "8082/tcp": {},
      "9092/tcp": {},
      "9644/tcp": {}
    },
    "Tty": false,
    "OpenStdin": false,
    "StdinOnce": false,
    "Env": [
      "container=podman",
      "EDITOR=vi",
      "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
      "HOME=/root",
      "HOSTNAME=3f9bb85cc476"
    ],
    "Cmd": [
      "/entrypoint-tc.sh",
      "redpanda",
      "start",
      "--mode=dev-container",
      "--smp=1",
      "--memory=1G"
    ],
    "Image": "docker.io/redpandadata/redpanda:v23.3.2",
    "Volumes": null,
    "WorkingDir": "/",
    "Entrypoint": [
      "/entrypoint.sh"
    ],
    "OnBuild": null,
    "Labels": {
      "org.opencontainers.image.authors": "Redpanda Data <hi@redpanda.com>",
      "org.testcontainers": "true",
      "org.testcontainers.lang": "go",
      "org.testcontainers.sessionId": "acef449fb3c94d212abe04df036b9e045292ce128eaf4a666f3d7340f9e421ad",
      "org.testcontainers.version": "0.28.0"
    },
    "StopSignal": "15",
    "StopTimeout": 10
  },
  ...
}
```

We can see the Entrypoint for Podman container includes the stock Redpanda `entrypoint.sh`. 

## Why is it important?

It is more correct to override entrypoint this way.

## How to test this PR

I have tested this change using Docker, and also using Podman, without changing code under test.

`TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE`, `TESTCONTAINERS_RYUK_CONTAINER_PRIVILEGED`, and `DOCKER_HOST` env vars were set to accommodate proper running of Ryuk, and test environment. I ran into issues https://github.com/testcontainers/testcontainers-go/issues/2264 and https://github.com/testcontainers/testcontainers-go/issues/538.

### DIFF
--- a/modules/redpanda/redpanda.go
+++ b/modules/redpanda/redpanda.go
@@ -72,9 +72,8 @@ func RunContainer(ctx context.Context, opts ...testcontainers.ContainerCustomize
 				defaultAdminAPIPort,
 				defaultSchemaRegistryPort,
 			},
-			Entrypoint: []string{},
+			Entrypoint: []string{entrypointFile},
 			Cmd: []string{
-				entrypointFile,
 				"redpanda",
 				"start",
 				"--mode=dev-container",


kiview:
Hi @bojand, thanks for the PR. 
You are right, the intent is to change the entrypoint (although I would like to loop in @eddumelendez, who originally implemented this pattern in tc-java). If we decide to merge this change, we should align the implementations in other languages as wel.

However, it is strange to me, that Podman and Docker behave different in this specific instance. So there are cases, where a `withCommand()` will effectively set an entrypoint in the case of Docker, while it won't in case of Podman?

kiview:
I checked that we have the same situation in tc-java. For me it looks, like Podman does not handle setting an empty Entrypoint correctly (i.e. in the same way as Docker):
https://github.com/testcontainers/testcontainers-go/blob/main/modules/redpanda/redpanda.go#L75

If the Entrypoint is empty, the Cmd is supposed to become the Entrypoint (see Docker [docs](https://docs.docker.com/reference/dockerfile/#understand-how-cmd-and-entrypoint-interact)). While this PR would fix the issue for Podman users, we generally try to not perform Podman-specific changes, and rather prefer raising this as an issue with Podman (which is generally appreciated in bringing parity to Docker behavior).

kiview (approved):

mdelapenya (approved):
Thanks @kiview for your review and coming with your insights.

LGTM

PR #2402, opened 2024-03-19
chore(deps): Bumping ChromaGo client version

tazarov:
## What does this PR do?

- I updated the chroma-go dependency and some tests with minor API changes breaking.
- I have also added a `GetClient()` chroma container option to make it easier for users to create a client for the container with options (part of the new chroma-go release)
- Docs have also been updated

## Why is it important?

- `chroma-go` has undergone a complete rewrite with some minor breaking changes. The new rewrite makes it easier for users to interface with Chroma via the Go client.
- Adding a utility method for getting the client makes for a good DX.
- Added examples of adding and querying collection data

## How to test this PR

Existing tests are passing. A new test was added for the `GetClient` utility method.

## Keep In Mind

Once #2391 is merged, I will rebase this to fix conflicts (sorry, Graphite stacks don't work well with upstream repos)


modules/chroma/chroma.go
```diff
@@ -58,3 +60,11 @@ func (c *ChromaContainer) RESTEndpoint(ctx context.Context) (string, error) {
 
 	return fmt.Sprintf("http://%s:%s", host, containerPort.Port()), nil
 }
+
+func (c *ChromaContainer) GetClient(opt ...chromago.ClientOption) (*chromago.Client, error) {
```
mdelapenya, line 64: In general we prefer not coupling a module with its client/SDK technology, to avoid users to get that dependency as part of the upstream dependencies from the module.

It's true that, in general, users would already use that go-sdk in their projects, but it could be the case they prefer using raw calls (http, grpc)... You know, in Go we many times prefer cooking everything ourselves...

Imagine the postgres module: there are a few popular clients, so we do not include a given one in the module. But we do add an example to the testable examples, so the dependency is not added because is a testing one.

For those reasons, I'd remove this GetClient method 🙏

tazarov: @mdelapenya, thank you. I've added it as I followed a pattern I saw in the Python project - https://github.com/testcontainers/testcontainers-python/blob/d61af383def6eadcd7f2b5ba667eb587c6cc84f1/modules/minio/testcontainers/minio/__init__.py#L69

I am completely on board with the looser coupling of things. And yes, there is definitely more than one way to integrate with Chroma. That said, the Chroma API needs some work to make it simpler for third-party clients. For example, right now, it doesn't work well with generated clients for strongly typed languages like Go. Another point in case is that soon enough we'll move the chroma-go from a third-party package to a first-party under the Chroma org.

tazarov: @mdelapenya, `GetClient()` removed.

mdelapenya (approved):
LGTM, thank you for your work here 🙏

PR #2403, opened 2024-03-19
chore: resolve IPv4-IPv6 issues

mdelapenya:
- **feat: detect IPv4 and IPv6**
- **chore: get container host from the hostIPs**
- **chore: simplify**
- **chore: resolve host port bindings**
- **chore: get bound ports from port bindings**
- **chore: update wait strategies to use bound ports**
- **chore: lint**
- **fix: use a fixed port**

## What does this PR do?
It looks up the bounded ports every time a container needs to get a mapped port. It also improves the wait strategies (http and host-port) to get the bound port.

To get the bound port, we'll get the host IP addresses from the current Docker Host, and with that list of values, and the ports from a container, we will get the bound ports.

Finally, we are deprecating the `ForceIPv4LocalHost` field from the HTTPStrategy, as it's not needed anymore. The example test is updated to fix the port  in the wait strategy.

## Why is it important?
Resolves flakiness in the CI, as the testable example is failing in a non consistent manner: because the test exposes two ports, 8080 and 80, the way Docker is retrieving them from the JSON representation is not deterministic (it's a port map), therefore, the test is sometimes returning the 80 port for getting the HTTP endpoint and others, the 8080 port.

At the same time, it adds a more consistent manner of getting the host and the port, relying on the Docker host, taking into consideration the IPv4 and IPv6 families.

## Related issues

- Relates #1775


mdelapenya:
@cristianrgreco I ping you as you already implemented this for NodeJS, and I used it as reference. Could you please take a look? 🙏

docker_client.go
```diff
@@ -59,17 +59,22 @@ func (c *DockerClient) Info(ctx context.Context) (system.Info, error) {
...
   Test ProcessID: %s
 `
 
+	dockerHost := core.ExtractDockerHost(ctx)
+
 	Logger.Printf(infoMessage, packagePath,
 		dockerInfo.ServerVersion, c.Client.ClientVersion(),
 		dockerInfo.OperatingSystem, dockerInfo.MemTotal/1024/1024,
+		dockerHost,
+		core.GetDockerHostIPs(),
```
mdelapenya, line 74: Wrong conflict resolution
```suggestion
```

kiview:
Closing here, because resolved in the Moby upstream at https://github.com/moby/moby/pull/47871.
If in the future we see the need to revisit this on the library side, we can re-open, or reuse parts of this PR.

PR #1685, opened 2023-09-25
Wait for k3s node to sync

pablochacin:
## What does this PR do?

Changes the waiting condition for k3s container to wait for the k3s node to finish the synchronization.

## Why is it important?

This change ensures the API server is able to respond to requests preventing issues such as described in [this thread.](https://testcontainers.slack.com/archives/CBWBNR76G/p1692981129009009) 

## Related issues

- Closes #1547

PR #2446, opened 2024-03-23
feat: authenticate docker on PullImage

codefromthecrypt:
## What does this PR do?

This authenticates on PullImage and improves the k3s LoadImages test so that a failure to start the pod will fail the test.

## Why is it important?

I'm using k3s with private images, so I want to use the `LoadImages` feature. The instructions to do this included running `PullImage` first, which works on anonymous images, but not authenticated ones. By re-using the same auth code used to run a container, I can now use the `LoadImages` feature in k3s with private images.

While trying to figure out what was wrong I used `Test_LoadImages` which is quite nice, but accidentally had a problem where a container can fail to run, yet still pass the test. This lost me time, so I figured I would improve it for the next person.

## Related issues

maybe #1685?

## How to test this PR

Change the image in Test_LoadImages from "nginx" to a privately hosted image (like a private github repository). Before, it would fail to pull, and now it won't.

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -1027,20 +1027,6 @@ func (p *DockerProvider) CreateContainer(ctx context.Context, req ContainerReque
 			pullOpt := types.ImagePullOptions{
 				Platform: req.ImagePlatform, // may be empty
 			}
-
-			registry, imageAuth, err := DockerImageAuth(ctx, imageName)
-			if err != nil {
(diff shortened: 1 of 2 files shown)


mdelapenya (approved):
This change makes total sense, thanks for adding it 🙇 

LGTM!

PR #2447, opened 2024-03-24
refactor: Add Weaviate modules tests

antas-marcin:
## What does this PR do?

In order to be in parity with `java` `node` and `python` modules, Weaviate container should be created with only minimal environment variables. Enabled modules settings (and other) should be optional to set.

## Why is it important?

This PR brings the same consistent behaviour across all `Weaviate testcontainers modules`.

## Related issues

none

## Follow-ups

none


modules/weaviate/weaviate_test.go
```diff
@@ -74,3 +79,109 @@ func TestWeaviate(t *testing.T) {
...
+			tt.Fatal(err)
+		}
+
+		if meta == nil || meta.Version == "" {
+			tt.Fatal("failed to get /v1/meta response")
+		}
+	})
+}
+
+func TestWeaviateWithClientWithModules(t *testing.T) {
```
mdelapenya, line 123: Hey @antas-marcin isn't this test the same one as the testable example?

I always find difficult to choose if writing one or the other, but in the end I try (not always with success) to have only one of them. If the main purpose of this code is to demonstrate how to bootstrap the client, I'd choose a testable example, else a test is fine.

Because both of them takes ~80 seconds, I think it's not meaningful for the total build time, so this comment is not a blocker for the merge. I'd appreciate your thoughts regarding choosing one or the other, or even both.

Thanks!

antas-marcin: @mdelapenya that's a good point! I was also thinking about this, but I went with leaving both of those tests, so let me remove this test and just leave 1 in examples.

modules/weaviate/weaviate.go
```diff
@@ -5,10 +5,16 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/docker/go-connections/nat"
 	"github.com/testcontainers/testcontainers-go"
 	"github.com/testcontainers/testcontainers-go/wait"
 )
 
+const (
+	httpPort = nat.Port("8080/tcp")
```
mdelapenya, line 14: I have just only one final concern before merging this PR: do we need the dependency with nat here, or we can simply go with strings for the ports?

Not a blocker, as we already get that dependency from the core library, but wanted to double check with you first.

antas-marcin: I can remove this dependency if you want, no problem with that. I will change that to strings then

mdelapenya (approved):
LGTM thank you!

PR #2449, opened 2024-03-26
chore: run tests on windows workers

mdelapenya:
- **chore: run core on windows**

## What does this PR do?
It adds Windows workers to the matrix to run the core tests.

## Why is it important?
Run tests on more platforms

## Related issues

- Supersedes and Closes #2398


internal/core/docker_socket_test.go
```diff
@@ -0,0 +1,139 @@
...
+		defer func() {
+			if r := recover(); r == nil {
+				tt.Errorf("expected panic")
+			}
+		}()
+
+		checkDefaultDockerSocket(context.Background(), mockCli{infoErr: errors.New("info should panic")}, "")
+	})
+
+	t.Run("Local Docker on Unix", func(tt *testing.T) {
```
mdelapenya, line 20: @kiview could you double check that these expectations are correct?

PR #2450, opened 2024-03-27
feat: optimizes file copies to and from containers

codefromthecrypt:
## What does this PR do?

This changes code interacting with file copies to and from the container to use optimized functions. Doing so reduces buffering and gives a chance for to use [Go 1.22's optimized paths for linux](https://github.com/golang/go/issues/58808).

## Why is it important?

I'm using rather large images in k3s. For example, nodejs images get easily over a GB each. I found this copy logic accounts for the majority of test fixture setup, in our case sometimes over a minute is spent here even when images are available locally.

## Related issues

Originally added in #347

## How to test this PR

you can use k3s and its `LoadImages` function which will hit all the paths here.

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -602,19 +602,34 @@ func (c *DockerContainer) CopyFileToContainer(ctx context.Context, hostFilePath 
 		return c.CopyDirToContainer(ctx, hostFilePath, containerFilePath, fileMode)
 	}
 
-	fileContent, err := os.ReadFile(hostFilePath)
+	file, err := os.Open(hostFilePath)
 	if err != nil {
 		return err
 	}
-	return c.CopyToContainer(ctx, fileContent, containerFilePath, fileMode)
+	defer file.Close()
+
+	info, err := file.Stat()
+	if err != nil {
+		return err
+	}
+
+	return c.copyToContainer(ctx, func(tw io.Writer) error {
+		// Attempt optimized writeTo, implemented in linux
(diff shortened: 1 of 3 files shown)


codefromthecrypt:
I think the failure is a flake?
```
           	            	-  RyukDisabled: (bool) true,
          	            	+  RyukDisabled: (bool) false,
```

docker.go
```diff
@@ -602,19 +603,41 @@ func (c *DockerContainer) CopyFileToContainer(ctx context.Context, hostFilePath
...
-	return c.CopyToContainer(ctx, fileContent, containerFilePath, fileMode)
+	defer f.Close()
+
+	info, err := f.Stat()
+	if err != nil {
+		return err
+	}
+
+	// In Go 1.22 os.File is always an io.WriterTo. However, testcontainers
+	// currently allows Go 1.21, so we need to trick the compiler a little.
```
mdelapenya, line 618: We test against both versions of the language, although we always develop in the lowest one. It could be probably interesting working the other way around: always develop in the latest release, and run the tests for both (current and current -1). Thoughts?

mdelapenya (approved):
LGTM, thanks!

docker.go
codefromthecrypt: To me, I think we should test at least 2 versions of Go anyway, and usually things like this don't come up too often.

In some projects I tend to do the develop in latest and test the floor model, just because a lot of devs always use latest first.

PR #2451, opened 2024-03-27
docs: add documentation for Exec method

gustavosbarreto:
## What does this PR do?

Adds documentation for the `Exec` method in the `DockerContainer` struct.

## Why is it important?

The documentation emphasizes that reading directly from the `io.Reader` may result in unexpected bytes due to custom stream multiplexing headers. It recommends using `stdcopy.StdCopy()` from the Docker API to properly separate the stdout and stderr.

See: https://gitlab.com/gitlab-org/gitlab-runner/-/merge_requests/503

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -478,6 +478,13 @@ func (c *DockerContainer) NetworkAliases(ctx context.Context) (map[string][]stri
 	return a, nil
 }
 
+// Exec executes a command in the current container.
+// It returns the exit status of the executed command, an [io.Reader] containing the combined
+// stdout and stderr, and any encountered error. Note that reading directly from the [io.Reader]
+// may result in unexpected bytes due to custom stream multiplexing headers.
+// Use [tcexec.Multiplexed] option to read the combined output without the multiplexing headers.
+// Alternatively, to separate the stdout and stderr from [io.Reader] and interpret these headers properly,
+// [github.com/docker/docker/pkg/stdcopy.StdCopy] from the Docker API should be used.
 func (c *DockerContainer) Exec(ctx context.Context, cmd []string, options ...tcexec.ProcessOption) (int, io.Reader, error) {
 	cli := c.provider.client
 
--- a/exec/processor.go
+++ b/exec/processor.go
@@ -60,6 +60,8 @@ func WithEnv(env []string) ProcessOption {
 	})
 }
 
+// Multiplexed returns a [ProcessOption] that configures the command execution
+// to combine stdout and stderr into a single stream without Docker's multiplexing headers.
 func Multiplexed() ProcessOption {
 	return ProcessOptionFunc(func(opts *ProcessOptions) {
 		// returning fast to bypass those options with a nil reader,


docker.go
```diff
@@ -478,6 +478,13 @@ func (c *DockerContainer) NetworkAliases(ctx context.Context) (map[string][]stri
 	return a, nil
 }
 
+// Exec executes a command in the current container.
+// It returns the exit status of the executed command, an `io.Reader` containing the combined
+// stdout and stderr, and any encountered error. Note that reading directly from the `io.Reader`
+// may result in unexpected bytes due to custom stream multiplexing headers.
+// To separate the stdout and stderr from `io.Reader` and interpret these headers properly,
+// `stdcopy.StdCopy()` from the Docker API should be used.
+// More information about this can be found at: https://godoc.org/github.com/docker/docker/pkg/stdcopy
```
mdelapenya, line 487: Thanks for contributing this. It probably caused some pain while using the method and I'm very thankful you decide to add it for others not to suffer the same 🙇 

We have the `Multiplexed` process option that separates the stdout and stderr using the stdcopy API, so I'd change this comment to use it.

I love having docs for this method, but I'd enforce the usage of the right process option instead. Wdyt?

gustavosbarreto: I've tested the `Multiplexed` option and found that it doesn't combine the output as expected, similar to how it's done in [`cmd.CombinedOutput()`](https://pkg.go.dev/os/exec#Cmd.CombinedOutput). Specifically, in the scenario where stderr contains data, it's preferred over stdout:
 
https://github.com/testcontainers/testcontainers-go/blob/32f372c44be48f9a4b271555f5e6a1e20238cbaf/exec/processor.go#L85-L89

It seems there might be a need to change the behavior of the `Multiplexed` option or introduce a new option like `CombinedOutput`.

What do you think?

mdelapenya: I like the idea: having different options or fixing the existing one makes sense to me.

Would you like to work on that?

gustavosbarreto: I can work on a new option called `CombinedOutput`, or I can investigate fixing the behavior of the existing `Multiplexed`. However, my concern is that changing the behavior of something existing might break compatibility with code that relies on the current behavior (preferring stderr if it contains data).

Let me know your thoughts on how you'd like to proceed.

mdelapenya: We can parametrize the Multiplexed option. It would mean a breaking change, but for the sake of consistency, it could be seen as needed.

gustavosbarreto:
If PR #2452 gets merged, I'll make sure to update the documentation to reflect the changes, specifically mentioning the utilization of the `Multiplexed` option.

mdelapenya:
> If PR #2452 gets merged, I'll make sure to update the documentation to reflect the changes, specifically mentioning the utilization of the `Multiplexed` option.

PR was merged, so I think it's time to update the docs in this one. Thank you!!

gustavosbarreto:
> > If PR #2452 gets merged, I'll make sure to update the documentation to reflect the changes, specifically mentioning the utilization of the `Multiplexed` option.
> 
> PR was merged, so I think it's time to update the docs in this one. Thank you!!

Docs updated!

PR #2452, opened 2024-03-28
fix(exec): updates the `Multiplexed` opt to combine stdout and stderr

gustavosbarreto:
## What does this PR do?

Updates the `Multiplexed` option to combine the output of stdout and stderr, changing the previous behavior where stderr was preferred over stdout. Now, both streams are properly combined into a single stream.

## Why is it important?

Previously, the `Multiplexed` option prioritized stderr over stdout, potentially leading to unexpected results. By updating it to combine both stdout and stderr into a single stream, we ensure consistency and accuracy in handling the output of executed commands. This change prevents potential confusion or errors resulting from the previous preference for stderr over stdout.

## Related issues

- Relates #2451

### DIFF
--- a/docker_exec_test.go
+++ b/docker_exec_test.go
@@ -51,7 +51,6 @@ func TestExecWithOptions(t *testing.T) {
 			}
 
 			container, err := GenericContainer(ctx, GenericContainerRequest{
-				ProviderType:     providerType,
 				ContainerRequest: req,
 				Started:          true,
 			})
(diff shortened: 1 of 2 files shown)


mdelapenya (approved):
LGTM! Thank you for contributing this nice improvement.

While reviewing the PR I added a commit on top removing the provider type from the containers in the tests, to avoid causing you more noise for this PR. Please take a look, you'll see it very simple, although not related, but they make the code more readable.

Once the CI passes, I'll merge it to include it in the upcoming releae

Cheers!

PR #2462, opened 2024-04-02
bug:Fix AMQPS url

### LINKED ISSUE #2457: [Bug]: AmqpsURL() function returns AMQP url
### Testcontainers version

0.29.1

Yes

### Host OS

MacOS

### Host arch

amd64

### Go version

1.21

### Docker version

```shell
N/A
```

### Docker info

```shell
N/A
```

### What happened?

I am using the testcontainers-go framework for our integration tests and I noticed that the AmqpsURL() function returns an AMQP URL:

`// AmqpURL returns the URL for AMQPS clients.
func (c *RabbitMQContainer) AmqpsURL(ctx context.Context) (string, error) {
	endpoint, err := c.PortEndpoint(ctx, nat.Port(DefaultAMQPPort), "")
	if err != nil {
		return "", err
	}

	return fmt.Sprintf("amqps://%s:%s@%s", c.AdminUsername, c.AdminPassword, endpoint), nil
}`

### Relevant log output

_No response_

### Additional information

_No response_

bearrito:
Fixes AQMPS url

## What does this PR do?

Fixes the AMQPS URL. Adds test.

## Why is it important?

User reported an issue - https://github.com/testcontainers/testcontainers-go/issues/2457

## Related issues

- Closes #2457

## How to test this PR

Will include unit-tests covering aqmps/tls

This currently won't work, while the AMQPS url is fixed it looks like container doesn't support TLS. I'll see if I can resolve that without too much difficulty.

### DIFF
--- a/modules/rabbitmq/rabbitmq.go
+++ b/modules/rabbitmq/rabbitmq.go
@@ -48,7 +48,7 @@ func (c *RabbitMQContainer) AmqpURL(ctx context.Context) (string, error) {
 
 // AmqpURL returns the URL for AMQPS clients.
 func (c *RabbitMQContainer) AmqpsURL(ctx context.Context) (string, error) {
-	endpoint, err := c.PortEndpoint(ctx, nat.Port(DefaultAMQPPort), "")
+	endpoint, err := c.PortEndpoint(ctx, nat.Port(DefaultAMQPSPort), "")
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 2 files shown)


mdelapenya (approved):
LGTM, thanks!

bearrito:
@mdelapenya  Just wondering, what sort of tests belong in examples vs in the modules test? 
I think the tests I added belong in the modules test because they cover a regression , but I often see the examples covering more code surface.

mdelapenya:
@bearrito I always suffer from the same: if we want to create a code snippet that lands into the pkg.go.dev docs, something that is completely copy&pasteable, then I'd say it goes as a testable example. Otherwise, covering regressions, then to real tests. But I see you, I many times see duplication in how we write tests Vs testable examples

PR #2464, opened 2024-04-03
chore: use "docker compose" (v2) instead of "docker-compose" (v1)

mdelapenya:
- **chore: migrate from docker-compose (v1) to compose (v2)**
- **chore: lowercase error message**
- **chore: lint**

## What does this PR do?
This PR changes the implementation of the local docker compose support to shell-out to the `docker` binary, with `compose` as first argument, instead of to the `docker-compose` binary.

At the same time, we are deprecating all the methods and types from the LocalDockerCompose implementation.

## Why is it important?
compose v1 was deprecated back in July '23, and the Github workers are not installing `docker-compose` in the workers by default.

## Related issues

- Github notice: https://github.com/actions/runner-images/issues/9557

### DIFF
--- a/docs/features/docker_compose.md
+++ b/docs/features/docker_compose.md
@@ -7,7 +7,7 @@ This is intended to be useful on projects where Docker Compose is already used
 in dev or other environments to define services that an application may be
 dependent upon.
 
-## Using `docker-compose` directly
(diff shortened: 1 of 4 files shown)


glours (approved):
LGTM 😉

PR #2466, opened 2024-04-04
Fix: HTTP wait strategy does not take query params into account

benja-M-1:
## What does this PR do?

When specifying a wait strategy using HTTP with query params, they are not sent to the container.
For example, to ensure the elasticmq container is ready, as advised in [this issue](https://github.com/softwaremill/elasticmq/issues/776), one must
poll the `/?Action=ListQueues` endpoint.
Currently this returns a 400 Bad request because the query parameters are not sent correctly to the endpoint.

The way the fix work is to parse the given path to create a `url.URL` instance and update the fields with the other values. This way the query parameters and the path are correclty handled.

## Related issues

Here is a reproduction with the elasticmq container:

```go
	Image:        "softwaremill/elasticmq-native:1.5.7",
	ExposedPorts: []string{"9324/tcp"},
	WaitingFor:   wait.ForHTTP("/?Action=ListQueues").WithPort("9324").WithStartupTimeout(3 * time.Second),
}

	ContainerRequest: req,
	Started:          true,
}

c, err := testcontainers.GenericContainer(ctx, genericContainerReq)
if err != nil {
	return fmt.Errorf("failed to start repos-new elasticmq container: %w", err)
}
```

### DIFF
--- a/wait/http.go
+++ b/wait/http.go
@@ -264,11 +264,12 @@ func (ws *HTTPStrategy) WaitUntilReady(ctx context.Context, target StrategyTarge
 	client := http.Client{Transport: tripper, Timeout: time.Second}
 	address := net.JoinHostPort(ipAddress, strconv.Itoa(mappedPort.Int()))
 
-	endpoint := url.URL{
-		Scheme: proto,
-		Host:   address,
-		Path:   ws.Path,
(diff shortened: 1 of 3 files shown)


mdelapenya (approved):
LGTM, thanks!

PR #2469, opened 2024-04-05
feat(modules.mongodb): add replica set support via opts

### LINKED ISSUE #2138: [Feature]: MongoDB ReplicaSet

heiytor:
## What does this PR do?

Implementing a `WithReplicaSet` function to configure the container to run with a default single-node replica set named "rs". This function appends "--replSet rs" to `req.Cmd` to initialize MongoDB in preparation to receive a replica set. Additionally, after execution, it initializes the replica set using "rs.initiate()", configuring the current host as the PRIMARY.

## Why is it important?

I've been utilizing testcontainers to test my service environment, which necessitates a replica set to execute transactions. As I've developed this functionality for internal use, I believe it would be beneficial to others as well.

## Related issues

This PR closes #2138. Additionally, I noticed that there is another PR, #2139, which also aims to implement this feature. However, it has not been updated for a while, and this branch has certain differences, such as utilizing both `mongosh` and `mongo` to ensure version compatibility.

## How to test this PR

I've tested this in a production environment, so it should work fine. Additionally, I've added new test cases to `TestMongoDB`.

## Follow-ups

Maybe adding a sharding option would be great.

### DIFF
--- a/modules/mongodb/mongodb.go
+++ b/modules/mongodb/mongodb.go
@@ -79,6 +79,29 @@ func WithPassword(password string) testcontainers.CustomizeRequestOption {
 	}
 }
 
+// WithReplicaSet configures the container to run a single-node MongoDB replica set named "rs".
+// It will wait until the replica set is ready.
+func WithReplicaSet() testcontainers.CustomizeRequestOption {
+	return func(req *testcontainers.GenericContainerRequest) error {
+		req.Cmd = append(req.Cmd, "--replSet", "rs")
+		req.LifecycleHooks = append(req.LifecycleHooks, testcontainers.ContainerLifecycleHooks{
+			PostStarts: []testcontainers.ContainerHook{
+				func(ctx context.Context, c testcontainers.Container) error {
+					ip, err := c.ContainerIP(ctx)
+					if err != nil {
+						return err
+					}
+
+					cmd := eval("rs.initiate({ _id: 'rs', members: [ { _id: 0, host: '%s:27017' } ] })", ip)
+					return wait.ForExec(cmd).WaitUntilReady(ctx, c)
+				},
+			},
+		})
+
+		return nil
+	}
+}
+
 // ConnectionString returns the connection string for the MongoDB container.
 // If you provide a username and a password, the connection string will also include them.
 func (c *MongoDBContainer) ConnectionString(ctx context.Context) (string, error) {
@@ -94,4 +117,16 @@ func (c *MongoDBContainer) ConnectionString(ctx context.Context) (string, error)
 		return fmt.Sprintf("mongodb://%s:%s@%s:%s", c.username, c.password, host, port.Port()), nil
 	}
 	return c.Endpoint(ctx, "mongodb")
+}
+
+// eval builds an mongosh|mongo eval command.
+func eval(command string, args ...any) []string {
+	command = "\"" + fmt.Sprintf(command, args...) + "\""
+
+	return []string{
+		"sh",
+		"-c",
+		// In previous versions, the binary "mongosh" was named "mongo".
+		"mongosh --quiet --eval " + command + " || mongo --quiet --eval " + command,
+	}
 }
--- a/modules/mongodb/mongodb_test.go
+++ b/modules/mongodb/mongodb_test.go
@@ -14,50 +14,70 @@ import (
 
 func TestMongoDB(t *testing.T) {
 	type tests struct {
-		name  string
-		image string
+		name string
+		opts []testcontainers.ContainerCustomizer
 	}
 	testCases := []tests{
 		{
-			name:  "From Docker Hub",
-			image: "mongo:6",
+			name: "From Docker Hub",
+			opts: []testcontainers.ContainerCustomizer{
+				testcontainers.WithImage("mongo:6"),
+			},
 		},
 		{
-			name:  "Community Server",
-			image: "mongodb/mongodb-community-server:7.0.2-ubi8",
+			name: "Community Server",
+			opts: []testcontainers.ContainerCustomizer{
+				testcontainers.WithImage("mongodb/mongodb-community-server:7.0.2-ubi8"),
+			},
 		},
 		{
-			name:  "Enterprise Server",
-			image: "mongodb/mongodb-enterprise-server:7.0.0-ubi8",
+			name: "Enterprise Server",
+			opts: []testcontainers.ContainerCustomizer{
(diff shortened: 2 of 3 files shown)


heiytor:
Some differences between this PR and #2139 include:

1. the utilization of `PostReadies` instead of `PostStart`. I observed that occasionally, in `PostStart`, the availability of `mongosh` or `mongo` was unreliable.
2.  does not implement a wait function because the call to `rs.initiate` automatically designates the current container as the PRIMARY node (resulting in `db.runCommand({ isMaster: 1 }).ismaster` being automatically set to true).
3. Does not modify any existing functions (such as `ConnectionString`) and does not require saving the replica set name.

modules/mongodb/mongodb.go
```diff
@@ -73,6 +73,31 @@ func WithPassword(password string) testcontainers.CustomizeRequestOption {
...
+		req.Cmd = append(req.Cmd, "--replSet", "rs")
+		req.LifecycleHooks = append(req.LifecycleHooks, testcontainers.ContainerLifecycleHooks{
+			PostReadies: []testcontainers.ContainerHook{
+				func(ctx context.Context, c testcontainers.Container) error {
+					cIP, err := c.ContainerIP(ctx)
+					if err != nil {
+						return err
+					}
+
+					cmd := eval("rs.initiate({ _id: 'rs', members: [ { _id: 0, host: '%s:27017' } ] })", cIP)
```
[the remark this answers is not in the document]
heiytor: In my experience, the lack of the 'members' attribute occasionally makes the replica set [unreachable](https://www.mongodb.com/docs/manual/reference/replica-configuration/#mongodb-rsconf-rsconf.members-n-.host). Consequently, I consistently use this attribute whenever configuring replica sets. Given its usage in the provided documentation example, I consider it appropriate here as well.

Additionally, the documentation suggests using DNS only when the IP address can change. Since I don't fully understand the testcontainers source code, I'm unsure if this is a concern.

I can change them if necessary.

modules/mongodb/mongodb.go
```diff
@@ -73,6 +73,31 @@ func WithPassword(password string) testcontainers.CustomizeRequestOption {
...
+					if exitCode, _, err := c.Exec(ctx, cmd); err != nil || exitCode != 0 {
+						return fmt.Errorf("failed to initiate the replica set with status %d: %s", exitCode, err)
+					}
+
+					return nil
+				},
+			},
+		})
+	}
+}
```
mdelapenya, line 99: Could you convert this postReadies hook into a wait strategy that is added to the existing one? The wait strategies run and after that the container is marked as ready. So running the replica set as a wait strategy will define a readiness check in the case the functional opt is passed. Wdyt?

I can help you out if needed.

mdelapenya (commented):
Hey @heiytor I think we are going to build on top of this PR instead of #2139. Could you address the comments in there? I can do it myself adding commits on top of yours, in the case you have no bandwidth. Wdyt?

modules/mongodb/mongodb.go
heiytor: Hi @mdelapenya, I've been working on converting to a wait strategy. However, the members need to use either `localhost:[container-ip]` or `[container-name]:27017`, and I'm unsure how to obtain the container's name or IP using a `GenericContainerRequest`. Do you have any suggestions?  https://github.com/testcontainers/testcontainers-go/pull/2469/commits/504cd35666937da3063f42da5d9b506ba7a955ee#diff-ad47d995563c853ad9c6745c55e4ffda6237cb13df6917c8f76bc2ad3c78609aR82-R92

mdelapenya: Oh you're totally right: because the wait strategy must be built before the container is created, we do not have access to the container IP. I think you can use this very same approach: https://github.com/testcontainers/testcontainers-go/blob/5fa65489ac33e7a6a15e4599739b2c717cf3a1a9/modules/kafka/kafka.go#L84-L87

- create an adhoc wait strategy in the post starts of the container, and there you have access to the container instance.

modules/mongodb/mongodb.go
```diff
@@ -79,6 +79,29 @@ func WithPassword(password string) testcontainers.CustomizeRequestOption {
 	}
 }
 
+// WithReplicaSet configures the container to run a single-node MongoDB replica set named "rs".
+// It will wait until the replica set is ready.
+func WithReplicaSet() testcontainers.CustomizeRequestOption {
+	return func(req *testcontainers.GenericContainerRequest) error {
+		req.Cmd = append(req.Cmd, "--replSet", "rs")
```
mdelapenya, line 86: One last question: do we want users to provide their replicasets? Or is it just fine to keep it constant, as `rs`?

modules/mongodb/mongodb_test.go
```diff
@@ -14,32 +14,52 @@ import (
...
+				testcontainers.WithImage("mongo:6"),
+				mongodb.WithReplicaSet(),
+			},
 		},
 	}
 
-	for _, tc := range testCases {
-		image := tc.image
-		t.Run(image, func(t *testing.T) {
+	for _, tt := range testCases {
```
mdelapenya, line 55: BTW let's keep this as `tc`. Can you rename the inner `t` instance as `tt`? Then please call `tt.Parallel` in L58 🙏 that will avoid possible confusions around what instance calls the parallel

modules/mongodb/mongodb.go
heiytor: According to the [official documentation](https://www.mongodb.com/docs/manual/reference/configuration-options/#mongodb-setting-replication.replSetName), the name serves to define the replica set that the mongo belongs. Since this will be a single-node replica set, I believe it is simpler to use a constant name.

modules/mongodb/mongodb_test.go
heiytor: Sure, I also added `opts := tc.opts` outside the `t.Run` function to prevent the 'loop variable tc captured by func literal' on L61.

heiytor: I ended up not reading the last line, sorry. I converted `opts := tc.opts` to `tc := tc`.

mdelapenya (approved):
LGTM, thanks!

mdelapenya (approved):
LGTM thanks!

PR #2467, opened 2024-04-04
[draft] feat(portforward): Add SSHD module + Ability to port forward

JulienBreux:
## What does this PR do?

Rel: https://github.com/testcontainers/testcontainers-go/issues/2212

## Why is it important?

To call the host from a container inside a testcontainers container.
Related to Microcks stuff.
(because a Microcks async minion had to call a host development webserver for example)

## Related issues

- Relates #2212

## How to test this PR

Use `.WithAccessToHost(true)`

PR #2471, opened 2024-04-08
feat: forward host ports to a container using an SSH tunnel

### LINKED ISSUE #2212: [Feature] Add withAccessToHost option to access to host from containers
### Problem

In my case, I've a container that needs to access to host in order to function.

### Solution

Add the `withAccessToHost(bool: access)` from Java implem to Go to do that.

### Benefit

The main advantage is to allow containers to communicate with the host.
(e.g. Redis hosted on a host or a service under development, etc.)

### Alternatives

IDK

Yes

mdelapenya:
- **chore: start a foundational package for interacting with Docker networks**
- **feat: add an SSH tunnel forwarding a host port to a container**

## What does this PR do?
This PR defines a new field in the container request struct, so that users can define which host ports are needed to be exposed to a given container.

If set, an internal SSHD server container will be spun before the container, forwarding each exposed host port to the container using an SSH tunnel.

All this logic will happen thanks to the lifecycle container hooks:

1. if the container declares host ports, we will define a new container lifecycle hook:
  - after the container is ready, we will expose the host ports using a SSH tunnel.
  - before the container is terminates, we will stop the sshd container.
2. we'll take the first network of the container to attach the sshd container to it
3. the IP fo the SSHD container in that network will be passed as extra-hosts of the container, so that the `host.testcontainers.internal` key will land in the container's `/etc/hosts` file (Linux).

As part of this PR we are adding an `internal/core/network` package to start adding core, network-related logic in it. This is needed because we need to get the Docker network of the container to pass it to the SSHD container.

## Why is it important?
At the moment it's not possible for a container to access a port in the host, which is handy for multiple use cases (please see #2212)

## Related issues

- Closes #2212
- Superseded and Closes #2467 
- Relates https://github.com/testcontainers/sshd-docker/pull/7
- Relates (java) https://github.com/testcontainers/testcontainers-java/pull/8574
- Relates (.NET) https://github.com/testcontainers/testcontainers-dotnet/pull/1160
- Relates (node) https://github.com/testcontainers/testcontainers-node/pull/758

## Follow-ups
This PR is in draft for one single reason: I'm pushing this branch to receive feedback from the community. I do not want to add any new library for the SSH tunnel, just using the plain stdlib. I see it not very friendly to use it and would like to have more eyes on this feature.

Both Java and .Net use a library to forward the port, and in Go I'd like to avoid that.

At the same time, there is a TODO in the code, as we are adding the network to the sshd container with a functional option that is an exact clone of the `network.WithNetwork`. If we use it, then a circular dependency will appear, which is not desired. This demonstrates that our efforts for refactoring the API with a more decoupled API is in the right direction.

One final follow-up is to consider starting just one SSHD container per test session, ala Ryuk: at the moment, there is one sshd container per container request. I think we can start this way, make progress, and move to that approach once needed. Thoughts?


port_forwarding.go
```diff
@@ -0,0 +1,386 @@
...
+
+	if len(p) == 0 {
+		return sshdConnectHook, fmt.Errorf("no ports to expose")
+	}
+
+	// Use the first network of the container to connect to the SSHD container.
+	var sshdFirstNetwork string
+	if len(req.Networks) > 0 {
+		sshdFirstNetwork = req.Networks[0]
+	} else if sshdFirstNetwork == "bridge" && len(req.Networks) > 1 {
```
codefromthecrypt, line 50: It seems like this branch will never be reached because  `sshdFirstNetwork` is set to the zero value and `len(req.Networks)` overlaps with the first branch. Did you mean to put this into a separate if *after* the first?

port_forwarding.go
```diff
@@ -0,0 +1,386 @@
...
+	}
+
+	// do not override the original HostConfigModifier
+	originalHCM := req.HostConfigModifier
+	req.HostConfigModifier = func(hostConfig *container.HostConfig) {
+		// adding the host internal alias to the container as an extra host
+		// to allow the container to reach the SSHD container.
+		hostConfig.ExtraHosts = append(hostConfig.ExtraHosts, fmt.Sprintf("%s:%s", hostInternal, sshdIP))
+
+		modes := []container.NetworkMode{container.NetworkMode(sshdFirstNetwork), "none", "host"}
```
codefromthecrypt, line 110: the edge case where `sshdFirstNetwork` == "" .. is this ok here? If not, I would early exit if `len(req.Networks) == 0` a lot higher in this function, as it also reduces indentation and breaks apart some other logic

port_forwarding.go
```diff
@@ -0,0 +1,386 @@
...
+		return sshdConnectHook, fmt.Errorf("failed to get IP for the SSHD container: %w", err)
+	}
+
+	if req.HostConfigModifier == nil {
+		req.HostConfigModifier = func(hostConfig *container.HostConfig) {}
+	}
+
+	// do not override the original HostConfigModifier
+	originalHCM := req.HostConfigModifier
+	req.HostConfigModifier = func(hostConfig *container.HostConfig) {
```
codefromthecrypt, line 105: nit: not sure about style in this project, but functions like this are easier to unit test if they are not anonymous.

port_forwarding.go
```diff
@@ -0,0 +1,386 @@
...
+		endpoint.Port, _ = strconv.Atoi(parts[1])
+	}
+	return endpoint
+}
+
+func (endpoint *sshEndpoint) String() string {
+	return fmt.Sprintf("%s:%d", endpoint.Host, endpoint.Port)
+}
+
+type ssHTunnel struct {
```
codefromthecrypt, line 304:
```suggestion
type sshTunnel struct {
```

port_forwarding.go
```diff
@@ -0,0 +1,386 @@
...
+		if err != nil {
+			return err
+		}
+		tunnel.logf("accepted connection")
+		go tunnel.forward(conn)
+	}
+}
+
+func (tunnel *ssHTunnel) forward(localConn net.Conn) {
+	serverConn, err := ssh.Dial("tcp", tunnel.Server.String(), tunnel.Config)
```
codefromthecrypt, line 363: maybe DialContext so that an incoming context with a deadline is paid attention to? Same below

```suggestion
	serverConn, err := ssh.DialContext("tcp", tunnel.Server.String(), tunnel.Config)
```

port_forwarding.go
```diff
@@ -0,0 +1,386 @@
...
+func (tunnel *ssHTunnel) Close() error {
+	if tunnel.Log != nil {
+		tunnel.logf("closing tunnel")
+	}
+
+	return nil
+}
+
+func (tunnel *ssHTunnel) Start() error {
+	listener, err := net.Listen("tcp", tunnel.Local.String())
```
codefromthecrypt, line 345: maybe `lc := net.ListenConfig{}` then `lc.Listen(ctx, ...` so an incoming context with deadline can be paid attention to?

codefromthecrypt (commented):
Thanks for the start!

some comments about where context can be propagated to facilitate an incoming context with deadline (I use one all the time with tc). Also, some notes about a logic problem I think.

Finally, if anything is replicating logic from another project, maybe worthwhile citing it even if in the same org, as not all maintainers will know to look in the java side for congruence or practice updates.

port_forwarding.go
mdelapenya: Indeed. I just realised the second branch came from a previous state. Changing it locally, thanks!

port_forwarding.go
mdelapenya: I do not see Dial context in the latest release of the crypto package: https://cs.opensource.google/go/x/crypto/+/master:ssh/client.go;l=176 Is it not merged yet?

codefromthecrypt: my mistake.. I must have looked at the wrong type at the time

mdelapenya:
> A big thanks from Microcks team 🙏🏻

🙇 I need to focus on this task, as I found a blocker here: I cannot see the communication being produced between the container and the exposed port in the host. The code is trying to mimic what the Java team is doing:

- add an extra-hosts to the container with the IP of the sshd server
- create a ssh client to the sshd server, using the mapped port of the sshd server for the well-known 22.

Could anybody give this PR a try and help if possible?

codefromthecrypt:
@joonas-fi I noticed you are doing something similar in your holepunch-client. Though maybe you aren't using testcontainers, yet. If you have some time to spare to help folks get container port forwarding working, much oblidged! https://github.com/function61/holepunch-client/blob/master/cmd/holepunch/client.go

JulienBreux:
> @codefromthecrypt @JulienBreux could you give this one a try? 🙏 I think it's ready for review

Let's go to testing at 2 PM today.

port_forwarding.go
```diff
@@ -0,0 +1,335 @@
...
+	"golang.org/x/crypto/ssh"
+
+	"github.com/testcontainers/testcontainers-go/internal/core/network"
+)
+
+const (
+	image string = "testcontainers/sshd:1.2.0"
+	// hostInternal is the internal hostname used to reach the host from the container,
+	// using the SSHD container as a bridge.
+	hostInternal string = "host.testcontainers.internal"
```
JulienBreux, line 20:
```suggestion
	// HostInternal is the internal hostname used to reach the host from the container,
	// using the SSHD container as a bridge.
	HostInternal string = "host.testcontainers.internal"
```

Maybe make it public so that you can use it easily from the modules?

```
testcontainers.HostInternal
```

mdelapenya: I like it, let's do it, thanks!

codefromthecrypt:
probably just me, but I struggled on this in k3s (wanting a pod to call a service on my host). 1000x thanks if someone has a gist or contributes an integration test to the k3s module for this.

JulienBreux:
Hi @codefromthecrypt,

Do you've a reproducible architecture? I need to test more.
Thanks!

codefromthecrypt:
@JulienBreux (cc @salaboy as he probably is interested in this)

So, I was testing coredns dnstap plugin (wanting to have traffic flow out to a listener I setup in the go test). However, since the PR here is about http and should work similar, I've a rough test below (pasted into k3s_test.go). I'm not sure I need an external name, but with or without it, I get "bad address" from alpine.

```go
const (
	expectedResponse = "Hello, World!"
)

func Test_ExternalNameService(t *testing.T) {
	ctx := context.Background()

	port, err := getFreePort()
	if err != nil {
		t.Fatal(err)
	}
	portString := fmt.Sprintf("%d", port)

	// Create a local HTTP server
	server := http.Server{Addr: ":" + portString, Handler: http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		fmt.Fprintln(w, expectedResponse)
	})}
	go server.ListenAndServe()
	defer server.Close()

	k3sContainer, err := k3s.RunContainer(ctx,
		testcontainers.WithImage("docker.io/rancher/k3s:v1.27.1-k3s1"),
		testcontainers.WithHostPortAccess(port),
	)
	if err != nil {
		t.Fatal(err)
	}

	// Clean up the container
	defer func() {
		if err := k3sContainer.Terminate(ctx); err != nil {
			t.Fatal(err)
		}
	}()

	kubeConfigYaml, err := k3sContainer.GetKubeConfig(ctx)
	if err != nil {
		t.Fatal(err)
	}

	restcfg, err := clientcmd.RESTConfigFromKubeConfig(kubeConfigYaml)
	if err != nil {
		t.Fatal(err)
	}

	k8s, err := kubernetes.NewForConfig(restcfg)
	if err != nil {
		t.Fatal(err)
	}

	// Create a CNAME httptest pointing to host.testcontainers.internal
	service := &corev1.Service{
		ObjectMeta: metav1.ObjectMeta{
			Name: "httptest",
		},
		Spec: corev1.ServiceSpec{
			Type:         corev1.ServiceTypeExternalName,
			ExternalName: "host.testcontainers.internal",
		},
	}
	services := k8s.CoreV1().Services("default")
	if _, err = services.Create(ctx, service, metav1.CreateOptions{}); err != nil {
		t.Fatal(err)
	}

	// Block until the service is ready to avoid race conditions
	if err = kwait.PollUntilContextCancel(ctx, time.Second, true, func(ctx context.Context) (bool, error) {
		if s, err := services.Get(ctx, service.Name, metav1.GetOptions{}); err != nil {
			return false, err
		} else if s.Spec.Type != service.Spec.Type || s.Spec.ExternalName != service.Spec.ExternalName {
			return false, fmt.Errorf("unexpected service spec: %v", s.Spec)
		}
		return true, nil
	}); err != nil {
		t.Fatal(err)
	}

	// Create a pod that attempts to access the host's http service
	pod := &corev1.Pod{
		ObjectMeta: metav1.ObjectMeta{
			Name: "test-pod",
		},
		Spec: corev1.PodSpec{
			Containers: []corev1.Container{
				{
					Name:    "alpine",
					Image:   "alpine:3.17",
					Command: []string{"wget", "-qO-", fmt.Sprintf("http://%s:%d", service.Name, port)},
				},
			},
			RestartPolicy: corev1.RestartPolicyNever,
		},
	}

	pods := k8s.CoreV1().Pods("default")
	if _, err = pods.Create(ctx, pod, metav1.CreateOptions{}); err != nil {
		t.Fatal(err)
	}

	// Wait for the Pod to be ready
	if err = kwait.PollUntilContextCancel(ctx, time.Second, true, func(ctx context.Context) (bool, error) {
		pod, err := pods.Get(ctx, "test-pod", metav1.GetOptions{})
		if err != nil {
			return false, err
		}

		for _, status := range pod.Status.ContainerStatuses {
			if status.State.Terminated != nil {
				return true, nil
			}
		}

		return false, nil
	}); err != nil {
		t.Fatal(err)
	}

	// Check the logs of the Pod to ensure that we read the expected response
	logs, err := pods.GetLogs("test-pod", &corev1.PodLogOptions{}).DoRaw(ctx)
	if err != nil {
		t.Fatal(err)
	}
	if !strings.Contains(string(logs), expectedResponse) {
		t.Fatalf("wget did not return the expected response, got: %s", logs)
	}
}

// getFreePort asks the kernel for a free open port that is ready to use.
func getFreePort() (int, error) {
	addr, err := net.ResolveTCPAddr("tcp", "localhost:0")
	if err != nil {
		return 0, err
	}

	l, err := net.ListenTCP("tcp", addr)
	if err != nil {
		return 0, err
	}
	defer l.Close()
	return l.Addr().(*net.TCPAddr).Port, nil
}
```

FYI I'm moving to singapore this weekend, so won't be able to revise this, but anyone has license to this if they want to make a PR out of it.

PR #2473, opened 2024-04-08
feat(postgres): ssl for postgres

### LINKED ISSUE #2404: [Bug]: File permissions with custom PostgreSQL image
### Testcontainers version

v0.29.1

Yes

### Host OS

MacOS

### Host arch

Arm

### Go version

1.21.8

### Docker version

```shell
Client:
 Cloud integration: v1.0.35+desktop.11
 Version:           25.0.3
 API version:       1.44
 Go version:        go1.21.6
 Git commit:        4debf41
 Built:             Tue Feb  6 21:13:26 2024
 Context:           desktop-linux

Server: Docker Desktop 4.28.0 (139021)
 Engine:
  Version:          25.0.3
  API version:      1.44 (minimum version 1.24)
  Go version:       go1.21.6
  Git commit:       f417435
  Built:            Tue Feb  6 21:14:22 2024
  OS/Arch:          linux/arm64
  Experimental:     false
 containerd:
  Version:          1.6.28
  GitCommit:        ae07eda36dd25f8a1b98dfbf587313b99c0190bb
 runc:
  Version:          1.1.12
  GitCommit:        v1.1.12-0-g51d5e94
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### Docker info

```shell
Client:
 Version:    25.0.3
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
    Version:  v0.12.1-desktop.4
    Path:     /Users/a/.docker/cli-plugins/docker-buildx
    Version:  v2.24.6-desktop.1
    Path:     /Users/a/.docker/cli-plugins/docker-compose
  debug: Get a shell into any image or container. (Docker Inc.)
    Version:  0.0.24
    Path:     /Users/a/.docker/cli-plugins/docker-debug
  dev: Docker Dev Environments (Docker Inc.)
    Version:  v0.1.0
    Path:     /Users/a/.docker/cli-plugins/docker-dev
  extension: Manages Docker extensions (Docker Inc.)
    Version:  v0.2.22
    Path:     /Users/a/.docker/cli-plugins/docker-extension
  feedback: Provide feedback, right in your terminal! (Docker Inc.)
    Version:  v1.0.4
    Path:     /Users/a/.docker/cli-plugins/docker-feedback
  init: Creates Docker-related starter files for your project (Docker Inc.)
    Version:  v1.0.1
    Path:     /Users/a/.docker/cli-plugins/docker-init
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc.)
    Version:  0.6.0
    Path:     /Users/a/.docker/cli-plugins/docker-sbom
  scout: Docker Scout (Docker Inc.)
    Version:  v1.5.0
    Path:     /Users/a/.docker/cli-plugins/docker-scout

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 250
 Server Version: 25.0.3
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: ae07eda36dd25f8a1b98dfbf587313b99c0190bb
 runc version: v1.1.12-0-g51d5e94
 init version: de40ad0
 Security Options:
  seccomp
   Profile: unconfined
  cgroupns
 Kernel Version: 6.6.16-linuxkit
 OSType: linux
 Architecture: aarch64
 CPUs: 12
 Total Memory: 19.51GiB
 Name: docker-desktop
 ID: c8e87756-6c7a-42fc-b2c7-7740dc3ca665
 Debug Mode: false
 Experimental: false
 Insecure Registries:
  hubproxy.docker.internal:5555
  127.0.0.0/8
 Live Restore Enabled: false

WARNING: daemon is not using the default seccomp profile
```

### What happened?

There is a miss match between running `testcontainer` and `docker run` when I mount files and cannot seem to figure out what is the reason.

Example testcontainer code:

```golang
postgresContainer, err := testcontainers.GenericContainer(ctx, testcontainers.GenericContainerRequest{
			Image: "postgresql-mock",
			Name:  "postgresql_mock",
			Env: map[string]string{
				"POSTGRES_USER":     "random",
				"POSTGRES_PASSWORD": "random",
				"POSTGRES_DB":       "random",
				"POSTGRES_HOST":     "postgresql",
			},
			Privileged:   true,  // tested with false also
			ExposedPorts: []string{"5432/tcp", "8080/tcp"},
			WaitingFor: wait.ForExec([]string{"pg_isready && curl --fail localhost:8080/api/tuner/health || exit 1"}),
			Files: []testcontainers.ContainerFile{
				{
					HostFilePath:      "../cert/server.crt",
					ContainerFilePath: "/var/lib/postgresql/server.crt",
					FileMode:          0o640,
				},
				{
					HostFilePath:      "../cert/server.key",
					ContainerFilePath: "/var/lib/postgresql/server.key",
					FileMode:          0o640,
				},
				{
					HostFilePath:      "../cert/root.crt",
					ContainerFilePath: "/var/lib/postgresql/root.crt",
					FileMode:          0o640,
				},
			}
		},
	})
```

I tested with various file modes also like `0o600`, `0o444` `0o440` but no luck.

The errors I get are either this:
```
FATAL:  could not load server certificate file "/var/lib/postgresql/server.crt": SSL error code 2147483661
```

or that:
```
File must have permissions u=rw (0600) or less if owned by the database user, or permissions u=rw,g=r (0640) or less if owned by root
```

The docker run command seems to work fine:
```
docker run -d -p 5432:5432 -p 8080:8080 --name postgresql-mock --memory="1250m" \
        -e POSTGRES_PASSWORD=random \
        -e POSTGRES_USER=random \
        -e POSTGRES_DB=random \
        -v "$(pwd)/../cert/server.crt:/var/lib/postgresql/server.crt:ro" \
        -v "$(pwd)/../cert/server.key:/var/lib/postgresql/server.key:ro" \
        -v "$(pwd)/../cert/root.crt:/var/lib/postgresql/root.crt:ro" \
        -v /var/run/docker.sock:/var/run/docker.sock \
        postgresql-mock
```

### Relevant log output

```shell
FATAL:  could not load server certificate file "/var/lib/postgresql/server.crt": SSL error code 2147483661
```

### Additional information

_No response_

bearrito:
## What does this PR do?
Enables SSL for postgres

The main thing to recognize when reviewing this is that the secret material must be owned by the postgres user.
The docker file copy api doesn't allowfor setting a user when copying. The only way I could think was to take wrap the entrypoint script with one that does what we want.

## Why is it important?

User ran into issue trying to use SSL. It's not obviously supported (able) due to file permissions.

## Related issues

Link related issues below. Insert the issue link or reference after the word "Closes" if merging this should automatically close it.

- Closes #2404 

## How to test this PR

Unit tests will work when complete.


modules/postgres/testdata/docker-entrypoint-ssl.bash
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+set -Eeo pipefail
+
+# Hardcoded user id from pg11-alpine image
+chown 70:70 /tmp/data/ca_cert.pem
+chown 70:70 /tmp/data/server.cert
+chown 70:70 /tmp/data/server.key
+
+ls -lah /tmp/data
```
bearrito, line 9: I need to remove this but it is helpful during dev.

modules/postgres/options.go
```diff
@@ -0,0 +1,14 @@
...
+
+type SSLSettings struct {
+	// Path to the CA certificate file
+	CACertFile string
+	// Path to the client certificate file
+	CertFile string
+	// Path to the key file
+	KeyFile string
+	// Entrypoint used to override and set SSL ownership
+	Entrypoint string
```
mdelapenya, line 13: I'd not mix the SSL Settings with an entrypoint to change ownership. I instead would pass another functional option to add a file, separating concerns. Wdyt? Doing that will allow to, at some point, extract the SSL capabilities and add them to the core, so any module can benefit from it (We have some modules already doing it).

mdelapenya (changes requested):
We are in the right track with this one, although I left some comments that need to be addressed.

Thanks!

modules/postgres/options.go
bearrito: No problem. I was thinking the same thing this morning.

In some ways I don't like exposing the `Entrypoint` because it's such a large escape hatch.

modules/postgres/testdata/docker-entrypoint-ssl.bash
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+set -Eeo pipefail
+
+# Hardcoded user id from pg11-alpine image
```
bearrito, line 4: I'll see if I can extract the `uid` from `/etc/passwd` instead of hardcoding it here. 
At somepoint we have to make an assumption about the image . i think its safer to assume postgres runs under the `postgres` user rather than always `uid` 70

mdelapenya:
@bearrito could you please take a look at https://github.com/testcontainers/testcontainers-go/pull/2478? I'm adding helper code to deal with TLS certificates in order to simplify the story for the client code. Wdyt?

bearrito:
@mdelapenya  That makes a lot of sense. Much cleaner. 

Do you want me to wait until that's done then base this on top of that?
I also recently added TLS to rabbitmq. Should I do that one or are you going to take care of all of them in one go?

mdelapenya:
@bearrito I went ahead and extracted the TLS cert generation to a separate go package, which makes more sense: https://github.com/mdelapenya/tlscert

Please take a look and use it as you need here!

bearrito:
Trying from vscode 
```
could not import github.com/mdelapenya/tlscert (no required module provides package "github.com/mdelapenya/tlscert")
```

bearrito:
@mdelapenya  Updated to use your package.

modules/postgres/go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	github.com/docker/go-connections v0.5.0
 	github.com/jackc/pgx/v5 v5.5.4
 	github.com/lib/pq v1.10.9
+	github.com/mdelapenya/tlscert v0.0.0-20240411155228-ea0834bcfa84
```
mdelapenya, line 9: I released the v0.1.0 this morning.
```suggestion
	github.com/mdelapenya/tlscert v0.1.0
```

modules/postgres/testdata/docker-entrypoint-ssl.bash
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
```
mdelapenya, line 1: IIUC this shell script is to force the ownership in the cert files, right? What will happen if users don't do it to their projects?

mdelapenya (commented):
Just one final concern regarding the entrypoint: do you think we need to pass it? Or is it something we can internally handle in the module, being transparent to the end users?

modules/postgres/testdata/docker-entrypoint-ssl.bash
bearrito: For postgres images, the container will fail to launch. 

I tried various file modes to get postgres to launch when the files are owned by root, but none of them worked. This is contra the docs where it appears root can own the docs.

The original reporter of this issue tried various approaches as well - https://github.com/testcontainers/testcontainers-go/issues/2404

bearrito: We could do that. If the user is enabling ssl then inject the entrypoint for them.

bearrito:
@mdelapenya Update so that the entrypoint is coupled with the ssl settings. What about doing something similar for the conf file?

modules/postgres/postgres.go
```diff
@@ -170,6 +170,98 @@ func WithSnapshotName(name string) SnapshotOption {
...
+		Reader:            reader,
+		ContainerFilePath: entrypointPath,
+		FileMode:          0o666,
+	})
+
+	req.Entrypoint = []string{"sh", entrypointPath}
+
+}
+
+func WithEntrypoint(hostEntrypointPath string) testcontainers.CustomizeRequestOption {
```
bearrito, line 248: I can remove this entirely if we don't want this exposed to the user.

modules/postgres/postgres_test.go
```diff
@@ -188,6 +244,43 @@ func TestWithConfigFile(t *testing.T) {
...
+	sslSettings := createSSLSettings(t)
+
+	container, err := postgres.RunContainer(ctx,
+		postgres.WithConfigFile(filepath.Join("testdata", "my-postgres-ssl.conf")),
+		postgres.WithInitScripts(filepath.Join("testdata", "init-user-db.sh")),
+		postgres.WithDatabase(dbname),
+		postgres.WithUsername(user),
+		postgres.WithPassword(password),
+		testcontainers.WithWaitStrategy(wait.ForLog("database system is ready to accept connections").WithOccurrence(2).WithStartupTimeout(5*time.Second)),
+		postgres.WithSSLSettings(sslSettings),
```
bearrito, line 259: No more entrypoint setting here.

modules/postgres/postgres_test.go
```diff
@@ -188,6 +244,43 @@ func TestWithConfigFile(t *testing.T) {
 	defer db.Close()
 }
 
+func TestWithSSLEnabledConfigFile(t *testing.T) {
+	ctx := context.Background()
+
+	sslSettings := createSSLSettings(t)
+
+	container, err := postgres.RunContainer(ctx,
+		postgres.WithConfigFile(filepath.Join("testdata", "my-postgres-ssl.conf")),
```
bearrito, line 253: User still has to remember to set this though.

modules/postgres/postgres.go
```diff
@@ -170,6 +170,98 @@ func WithSnapshotName(name string) SnapshotOption {
...
+	#!/usr/bin/env bash
+	set -Eeo pipefail
+
+
+	pUID=$(id -u postgres)
+	pGID=$(id -g postgres)
+
+	if [ -z "$pUID" ]
+	then
+      exit 1
```
bearrito, line 220: Not sure why github is rendering it like this. In code the `exit 1` is correctly indented.

modules/postgres/testdata/docker-entrypoint-ssl.bash
mdelapenya: I'm not sure if we want to expose a `WithEntrypoint`, at least for the modules, as it would allow changing the behaviour of a given module. I'd like to know when/if this is needed 🤔 Else, the module could behave differently from the expected behaviour, and can cause issues to users. For advanced uses cases, they can override the entire container request (see https://golang.testcontainers.org/modules/postgres/#customising-the-containerrequest)

modules/postgres/postgres.go
mdelapenya: Have you checked spaces Vs tabs?? 🤔

bearrito:
@mdelapenya added docs.

modules/postgres/testdata/docker-entrypoint-ssl.bash
bearrito: Okay, I wont export it.

modules/postgres/postgres_test.go
bearrito: @mdelapenya I'm thinking I get rid of the `WithConfigFile`

According to postgres there is another config  option available. I think I can just embed the relevant file, set that as the postgres.auto.conf. A user won't have to have a special conf file just to enable ssl, but also if they want to have a custom conf file, they can without worrying about the ssl settings.

wdyt?

```
In addition to postgresql.conf, a PostgreSQL data directory contains a file postgresql.auto.conf, which has the same format as postgresql.conf but is intended to be edited automatically, not manually. This file holds settings provided through the [ALTER SYSTEM](https://www.postgresql.org/docs/current/sql-altersystem.html) command. This file is read whenever postgresql.conf is, and its settings take effect in the same way. Settings in postgresql.auto.conf override those in postgresql.conf.
```

modules/postgres/postgres.go
bearrito: gone

docs/modules/postgres.md
```diff
@@ -69,6 +69,27 @@ In the case you have a custom config file for Postgres, it's possible to copy th
 !!!tip
     For information on what is available to configure, see the [PostgreSQL docs](https://www.postgresql.org/docs/14/runtime-config.html) for the specific version of PostgreSQL that you are running.
 
+#### SSL Configuration
+
```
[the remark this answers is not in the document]
mdelapenya: I need to improve the contribution guidelines 😞 I'm sorry 🙏

modules/postgres/postgres.go
```diff
@@ -170,6 +170,81 @@ func WithSnapshotName(name string) SnapshotOption {
 	}
 }
 
+// WithSSLSettings configures the Postgres server to run with the provided CA Chain
+// This will not function if the corresponding postgres conf is not correctly configured.
+// Namely the paths below must match what is set in the conf file
+func WithSSLSettings(sslSettings SSLSettings) testcontainers.CustomizeRequestOption {
```
mdelapenya, line 176: Do we want users to customise the certs, or just use the tlscert library to happen under the hood?

docs/modules/postgres.md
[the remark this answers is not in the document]
bearrito: No I'm the worst at docs. mea culpa.

modules/postgres/postgres.go
bearrito: I like things to work out of the box, so I could see removing that. But inevitably there is going to be a issue opened titled "Allow users to use their own certificate chain"

docs/modules/postgres.md
```diff
@@ -66,9 +66,35 @@ An example of a `*.sh` script that creates a user and database is shown below:
...
+- Not available until the next release of testcontainers-go <a href="https://github.com/testcontainers/testcontainers-go"><span class="tc-version">:material-tag: main</span></a>
+
+If you would like to use SSL with the container you can use the `WithSSLSettings`. This function accepts a `SSLSettings` which has the required secret material, namely the ca-certificate, server certificate and key. The container will copy this material to `/tmp/data/ca_cert.pem`, `tmp/data/server.cert` and `/tmp/data/server.key`
+
+This function will inject a custom postgres configuration file that enables SSL and correctly sets the paths on the key material.
+
```
mdelapenya, line 81:
```suggestion
```

docs/modules/postgres.md
```diff
@@ -66,9 +66,35 @@ An example of a `*.sh` script that creates a user and database is shown below:
...
+
+If you use this function in conjuction with `WithConfigFile` your custom conf must set the require ssl fields. The configuration must correctly align the key material provided via `SSLSettings` with the server configuration, namely the paths. Your configuration will need to contain the following
+ ```
+ ssl = on
+ssl_ca_file = '/tmp/data/ca_cert.pem'
+ssl_cert_file = '/tmp/data/server.cert'
+ssl_key_file = '/tmp/data/server.key'
+ ```
+
```
mdelapenya, line 90:
```suggestion
```

docs/modules/postgres.md
```diff
@@ -66,9 +66,35 @@ An example of a `*.sh` script that creates a user and database is shown below:
...
+#### SSL Configuration
+
+- Not available until the next release of testcontainers-go <a href="https://github.com/testcontainers/testcontainers-go"><span class="tc-version">:material-tag: main</span></a>
+
+If you would like to use SSL with the container you can use the `WithSSLSettings`. This function accepts a `SSLSettings` which has the required secret material, namely the ca-certificate, server certificate and key. The container will copy this material to `/tmp/data/ca_cert.pem`, `tmp/data/server.cert` and `/tmp/data/server.key`
+
+This function will inject a custom postgres configuration file that enables SSL and correctly sets the paths on the key material.
+
+
+If you use this function in conjuction with `WithConfigFile` your custom conf must set the require ssl fields. The configuration must correctly align the key material provided via `SSLSettings` with the server configuration, namely the paths. Your configuration will need to contain the following
```
mdelapenya, line 83:
```suggestion
If you use this function in conjuction with `WithConfigFile` your custom conf must set the require ssl fields. The configuration must correctly align the key material provided via `SSLSettings` with the server configuration, namely the paths. Your configuration will need to contain the following:

```

modules/postgres/postgres.go
```diff
@@ -170,6 +179,67 @@ func WithSnapshotName(name string) SnapshotOption {
...
+	const postgresCaCertPath = "/tmp/data/ca_cert.pem"
+	const postgresCertPath = "/tmp/data/server.cert"
+	const postgresKeyPath = "/tmp/data/server.key"
+
+	const defaultPermission = 0o600
+
+	return func(req *testcontainers.GenericContainerRequest) {
```
mdelapenya, line 192: After #2267 (merged today) all customisers need to return error. Could you update this one? 🙏

modules/postgres/postgres_test.go
```diff
@@ -26,6 +29,59 @@ const (
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	return postgres.SSLSettings{
+		CACertFile: caCert.CertPath,
+		CertFile:   serverCerts.CertPath,
+		KeyFile:    serverCerts.KeyPath,
+	}
+
```
mdelapenya, line 82:
```suggestion
```

modules/postgres/postgres_test.go
```diff
@@ -26,6 +29,59 @@ const (
...
+	if cert == nil {
+		return caCert, cert, errors.New("Unable to create Server Certificates")
+	}
+
+	return caCert, cert, nil
+
+}
+
+func createSSLSettings(t *testing.T) postgres.SSLSettings {
+
```
mdelapenya, line 71:
```suggestion
```

modules/postgres/postgres_test.go
```diff
@@ -26,6 +29,59 @@ const (
 	password = "password"
 )
 
+func createSSLCerts(t *testing.T) (*tlscert.Certificate, *tlscert.Certificate, error) {
+
```
mdelapenya, line 33:
```suggestion
```

modules/postgres/postgres_test.go
```diff
@@ -26,6 +29,59 @@ const (
...
+		Name:      "client-cert",
+		Parent:    caCert,
+		ParentDir: certsDir,
+	})
+	if cert == nil {
+		return caCert, cert, errors.New("Unable to create Server Certificates")
+	}
+
+	return caCert, cert, nil
+
```
mdelapenya, line 67:
```suggestion
```

modules/postgres/resources/postgres-ssl.conf
```diff
@@ -0,0 +1,80 @@
+# -----------------------------
```
mdelapenya, line 1: I have one question regarding the default ssl conf: will it work with different versions of the Postgres image? If a user sets the postgres version to 13, will it work the same? What will happen with new versions of Postgres? Must we maintain this file once new versions appear? If so, I'd advocate for removing that and letting users to provide their configs.

For our tests, I do see it valuable, as we control the versions, but not sure if we should put a fixed config file into the postgres config for all users of the module. Wdyt?

mdelapenya (commented):
I left a few more comments, sorry if that's taking too long 😞  My only concern is about the default SSL conf file, please see my comments about it.

modules/postgres/resources/postgres-ssl.conf
bearrito: Reverted to use the previous approach of passing in.

mdelapenya: Thank you! I think that makes more sense 🙏

bearrito:
@mdelapenya Anything else on this? I'm going on holiday and am closing out my open pr's

modules/postgres/postgres_test.go
```diff
@@ -26,6 +29,54 @@ const (
...
+	if err := os.MkdirAll(certsDir, 0o755); err != nil {
+		t.Fatal(err)
+	}
+
+	t.Cleanup(func() {
+		os.RemoveAll(tmpDir)
+	})
+
+	caCert := tlscert.SelfSignedFromRequest(tlscert.Request{
```
mdelapenya, line 44: Have you considered using tlscert.SelfSignedCA here instead? See https://github.com/mdelapenya/tlscert/blob/a038576e80ef259d2a60317ab51e37cdda8da3c8/tlscert.go#L135

modules/postgres/resources/customEntrypoint.sh
```diff
@@ -0,0 +1,22 @@
...
+fi
+
+if [ -z "$pGID" ]
+then
+    exit 1
+fi
+
+chown "$pUID":"$pGID" /tmp/data/ca_cert.pem
+chown "$pUID":"$pGID" /tmp/data/server.cert
+chown "$pUID":"$pGID" /tmp/data/server.key
```
mdelapenya, line 20: Would this be enough?
```suggestion
chown -R "${pUID}:${pGID}" /tmp/data
```

mdelapenya (commented):
Sorry for the late response :(

I think we are close to make it into the project, but let's clarify the custom entry point, as I'm not sure if I understand its use case properly

Thanks!

mdelapenya:
@stevenh with the latest merges, I think we can prioritise this one using the new approach for TLS, wdyt?

bearrito:
> @stevenh with the latest merges, I think we can prioritise this one using the new approach for TLS, wdyt?

I dropped off from this for a while. Two questions
1) Is the PR still valid?
2) Are there merged changes that invalidate any of the approaches?

modules/postgres/postgres.go
```diff
@@ -182,6 +189,62 @@ func WithSnapshotName(name string) SnapshotOption {
...
+	reader := strings.NewReader(embeddedCustomEntrypoint)
+
+	req.Files = append(req.Files, testcontainers.ContainerFile{
+		Reader:            reader,
+		ContainerFilePath: entrypointPath,
+		FileMode:          0o666,
+	})
+
+	req.Entrypoint = []string{"sh", entrypointPath}
+}
```
stevenh, line 246: suggestion: roll up to simplify, also don't think you need the additional way do you?
```suggestion
		const entrypointPath = "/usr/local/bin/docker-entrypoint-ssl.bash"

		req.Files = append(req.Files,
			testcontainers.ContainerFile{
				HostFilePath:      caCertFile,
				ContainerFilePath: "/tmp/certs/ca_cert.pem",
				FileMode:          defaultPermission,
			},
			testcontainers.ContainerFile{
				HostFilePath:      certFile,
				ContainerFilePath: "/tmp/certs/server.cert",
				FileMode:          defaultPermission,
			},
			testcontainers.ContainerFile{
				HostFilePath:      keyFile,
				ContainerFilePath: "/tmp/data/server.key",
				FileMode:          defaultPermission,
			},
			testcontainers.ContainerFile{
				Reader:            strings.NewReader(embeddedCustomEntrypoint),
				ContainerFilePath: ",
				FileMode:          0o666,
			},
		)
		req.Entrypoint = []string{"sh", entrypointPath}
		
		return nil
	}
}
```

modules/postgres/postgres.go
```diff
@@ -182,6 +189,62 @@ func WithSnapshotName(name string) SnapshotOption {
 	}
 }
 
+// WithSSLSettings configures the Postgres server to run with the provided CA Chain
+// This will not function if the corresponding postgres conf is not correctly configured.
+// Namely the paths below must match what is set in the conf file
+func WithSSLSettings(sslSettings SSLSettings) testcontainers.CustomizeRequestOption {
```
stevenh, line 195: suggestion: pass in values not struct to avoid ambiguity.
question: should this be renamed as its not ssl settings to me?
```suggestion
func WithSSLCert(caCertFile, certFile, keyFile) testcontainers.CustomizeRequestOption {
```

stevenh (changes requested):
I don't believe we can use the new SSL wait as this container doesn't have the option to generate its own certs?

Assuming that's the case I've done review pass, so we can get this moving.

modules/postgres/postgres_test.go
```diff
@@ -26,6 +29,54 @@ const (
...
+		Host:      "localhost",
+		Name:      "ca-cert",
+		ParentDir: certsDir,
+	})
+
+	if caCert == nil {
+		return caCert, nil, errors.New("Unable to create CA Authority")
```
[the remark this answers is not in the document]
bearrito: Unfortunately don't think so, only returns the pointer. This uses the library from @mdelapenya and was abstracted out into a lib from this project in previous commits.

modules/postgres/options.go
```diff
@@ -0,0 +1,12 @@
+package postgres
+
+type SSLVerificationMode string
+
+type SSLSettings struct {
```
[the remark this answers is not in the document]
bearrito: Done, but I'll note I followed what's being done in the rabbit module: https://github.com/testcontainers/testcontainers-go/blob/main/modules/rabbitmq/options.go#L15

modules/postgres/postgres_test.go
```diff
@@ -25,6 +30,51 @@ const (
...
+		ParentDir: certsDir,
+	})
+	if cert == nil {
+		return caCert, cert, errors.New("unable to create Server Certificates")
+	}
+
+	return caCert, cert, nil
+}
+
+func createSSLSettings(t *testing.T) postgres.SSLSettings {
```
stevenh, line 67: bug: return type not needed any more due to change of args.

modules/postgres/postgres_test.go
```diff
@@ -171,6 +221,55 @@ func TestWithConfigFile(t *testing.T) {
...
+
+	result, err := db.Exec("SELECT * FROM testdb;")
+	require.NoError(t, err)
+	assert.NotNil(t, result)
+}
+
+func TestSSLValidatesKeyMaterialPath(t *testing.T) {
+	ctx := context.Background()
+
+	sslSettings := postgres.SSLSettings{}
```
stevenh, line 258: bug: type not needed any more.

modules/postgres/resources/customEntrypoint.sh
```diff
@@ -0,0 +1,24 @@
...
+
+if [ -z "$pGID" ]
+then
+    echo "Unable to find postgres group id, required in order to chown key material"
+    exit 1
+fi
+
+chown "$pUID":"$pGID" /tmp/testcontainers-go/postgres/ca_cert.pem
+chown "$pUID":"$pGID" /tmp/testcontainers-go/postgres/server.cert
+chown "$pUID":"$pGID" /tmp/testcontainers-go/postgres/server.key
```
stevenh, line 22: suggestion: this can be done with a single call
```suggestion
chown "$pUID":"$pGID" \
    /tmp/testcontainers-go/postgres/ca_cert.pem \
    /tmp/testcontainers-go/postgres/server.cert \
    /tmp/testcontainers-go/postgres/server.key
```

modules/postgres/options.go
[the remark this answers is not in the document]
stevenh: Yep we're trying to make some improvements as we continue to develop options.

stevenh (changes requested):
Thanks for the changes, there's some outstanding bits needed to complete the switch to parameters, think I've caught them all in this pass.

bearrito:
Addressed latest.

stevenh (dismissed):
Thanks for the fast turnaround on the changes, looks good to me, happy holidays!

bearrito:
@mdelapenya Is this good to go from your end?

bearrito:
@mdelapenya Is this good?

docs/modules/postgres.md
```diff
@@ -74,9 +74,34 @@ An example of a `*.sh` script that creates a user and database is shown below:
...
+- Not available until the next release of testcontainers-go <a href="https://github.com/testcontainers/testcontainers-go"><span class="tc-version">:material-tag: main</span></a>
+
+If you would like to use SSL with the container you can use the `WithSSLSettings`. This function accepts a `SSLSettings` which has the required secret material, namely the ca-certificate, server certificate and key. The container will copy this material to `/tmp/testcontainers-go/postgres/ca_cert.pem`, `/tmp/testcontainers-go/postgres/server.cert` and `/tmp/testcontainers-go/postgres/server.key`
+
+This function requires a custom postgres configuration file that enables SSL and correctly sets the paths on the key material.
+
+If you use this function by itself or in conjuction with `WithConfigFile` your custom conf must set the require ssl fields. The configuration must correctly align the key material provided via `SSLSettings` with the server configuration, namely the paths. Your configuration will need to contain the following:
+
+ ```
+ ssl = on
```
mdelapenya, line 93: Nit: not sure if pg config is sensible to white spaces 🤞 
```suggestion
ssl = on
```

docs/modules/postgres.md
```diff
@@ -74,9 +74,34 @@ An example of a `*.sh` script that creates a user and database is shown below:
...
+ssl_ca_file = '/tmp/testcontainers-go/postgres/ca_cert.pem'
+ssl_cert_file = '/tmp/testcontainers-go/postgres/server.cert'
+ssl_key_file = '/tmp/testcontainers-go/postgres/server.key'
+ ```
+
+This function assumes the postgres user in the container is `postgres`
+
+There is no current support for mutual authentication.
+
+The `SSLSettings` function will modify the container `entrypoint`. This is done so that key material copied over to the container is chowned by `postgres`. All other container arguments will be passed through to the original container entrypoint.
```
mdelapenya, line 103: Suggestion: can we wrap this block in a `warning`?

modules/postgres/postgres_test.go
```diff
@@ -171,6 +209,54 @@ func TestWithConfigFile(t *testing.T) {
 	defer db.Close()
 }
 
+func TestWithSSL(t *testing.T) {
+	ctx := context.Background()
+
+	caCert, serverCerts, err := createSSLCerts(t)
+	require.NoError(t, err)
+
+	container, err := postgres.RunContainer(ctx,
```
mdelapenya, line 218: bug: let's use non-deprecated API
```suggestion
	container, err := postgres.Run(ctx,
```

mdelapenya, line 218: suggestion: let's name the variable `ctr` to avoid potential conflicts with a `container` package (docker, moby exposes it)
```suggestion
	ctr, err := postgres.RunContainer(ctx,
```

modules/postgres/postgres_test.go
```diff
@@ -171,6 +209,54 @@ func TestWithConfigFile(t *testing.T) {
...
+		postgres.WithConfigFile(filepath.Join("testdata", "postgres-ssl.conf")),
+		postgres.WithInitScripts(filepath.Join("testdata", "init-user-db.sh")),
+		postgres.WithDatabase(dbname),
+		postgres.WithUsername(user),
+		postgres.WithPassword(password),
+		testcontainers.WithWaitStrategy(wait.ForLog("database system is ready to accept connections").WithOccurrence(2).WithStartupTimeout(5*time.Second)),
+		postgres.WithSSLCert(caCert.CertPath, serverCerts.CertPath, serverCerts.KeyPath),
+	)
+
+	testcontainers.CleanupContainer(t, container)
```
mdelapenya, line 228: suggestion: as above, more below
```suggestion
	testcontainers.CleanupContainer(t, ctr)
```

modules/postgres/postgres_test.go
```diff
@@ -171,6 +209,54 @@ func TestWithConfigFile(t *testing.T) {
...
+
+	result, err := db.Exec("SELECT * FROM testdb;")
+	require.NoError(t, err)
+	assert.NotNil(t, result)
+}
+
+func TestSSLValidatesKeyMaterialPath(t *testing.T) {
+	ctx := context.Background()
+
+	_, err := postgres.RunContainer(ctx,
```
mdelapenya, line 247: bug: let's use non-deprecated API
```suggestion
	_, err := postgres.Run(ctx,
```

mdelapenya (commented):
> @mdelapenya Is this good?

@bearrito sorry, I was on XMas PTO, and finish it tomorrow 7th. I left some minor comments that need to be addressed before merging.

As a follow-up, not for this PR to make progress and see it into main asap, I'd allow different users than postgres when enabling SSL.

Other than that, once the comments are addressed, LGTM

mdelapenya:
@bearrito I went ahead and added a few commits with my suggestions to avoid causing you more trouble 🙏 

Hopefully this build will pass and we can proceed with the merge.

Sorry about it taking too long 🙇

mdelapenya (approved):
LGTM, thanks for your patience and the long time it took for the review. My apologies

PR #2478, opened 2024-04-10
chore: create TLS certs in a consistent manner

mdelapenya:
## What does this PR do?
This PR uses the https://github.com/mdelapenya/tlscert package to handle the creation of TLS certificates for testing purposes, and doing it again and again over all places.

Render URL for the docs: https://deploy-preview-2478--testcontainers-go.netlify.app/features/tls/

We have updated the tests for cockroachdb, rabbitmq and redpanda, so that they use the new package to get the TLS cert.

## Why is it important?
Modules are starting to add SSL support, in order to connect to the technologies using a TLS certificate. This is fine, but we are forcing them to generate a valid TLS cert to be added to the tests. Using this library, it's possible to retrieve the TLS cert, its private key, and the bytes for each other. The cert struct can be used then in the tests to create the SSL configuration.

NOTE: we started the Go library as a Go package in this project, but in the end, we thought it makes more sense to maintain it as a separate project, with benefits for the Go community.


docker.go
```diff
@@ -637,6 +637,9 @@ func (c *DockerContainer) CopyToContainer(ctx context.Context, fileContent []byt
 
 func (c *DockerContainer) copyToContainer(ctx context.Context, fileContent func(tw io.Writer) error, fileContentSize int64, containerFilePath string, fileMode int64) error {
 	buffer, err := tarFile(containerFilePath, fileContent, fileContentSize, fileMode)
+	if err != nil {
```
mdelapenya, line 640: Introduced in #2450

mdelapenya:
There is a chance that we move this package to a separate lib. Thoughts?

bearrito:
Being used here - https://github.com/testcontainers/testcontainers-go/pull/2473

PR #2485, opened 2024-04-16
feat: support Ryuk for the compose module

### LINKED ISSUE #372: Containers created by Docker Compose are not cleaned by Ryuk

mdelapenya:
- **feat: add testcontainers labels to compose containers**
- **feat: support reaper for compose**

## What does this PR do?
This PR enables the usage of Ryuk in the compose module, following the same approach than for a regular container:

- check if Ryuk is enabled at the config
- start the reaper, and store it in the compose struct
- connect to the reaper getting a termination signal for each container service, and delegate to container lifecycle the termination of that signal
- pass all the required testcontainers labels to each service in a compose file, so that they receive the same behaviors than the generic containers: be killed by Ryuk and aggregate test executions with the sessionID.
- detect the networks of the compose project, adding the testcontainers labels to each of them. Each network will receive a termination signal from ryuk.

We had to rewrite a bit all the tests for compose, as they were using lots os fixed ports that could lead to errors while testing the library. Instead, we are converting them into YAML templates and getting a random free port from the OS to render the template into a temporary compose file (created by the testing.T instance). The service names has been modified in the template too, to distinguish the usage from the local compose to the api compose.

Finally, we noticed that is important to fine tune the Ryuk reconnection timeout on CI, default is 10s: with compose, for the tests in the module, I had to update it to 30s; otherwise Ryuk could die before getting the new compose container. I have not see the same behaviour in the generic container approach.

## Why is it important?
Will allow not forcing the call to compose.Down, as Ryuk will take care of the compose containers.

## Related issues

- Closes #372
- Relates #465

## How to test this PR

Added tests demonstrating that:
- the labels are added to all services in a compose file
- with the reaper enabled, there is no need to call compose.Down
- with the reaper disabled, we need to call compose.Down in the case we want to clean the state

## Follow-ups
A proper abstraction on how to bootstrap the reaper would be needed, so that the compose module can call it without using a deprecated API. BTW I noticed the `NewReaper` func is not deprecated because the `// Deprecated` comment does not go first in the function comments 🤯


.github/workflows/ci-test-go.yml
```diff
@@ -50,6 +50,7 @@ jobs:
     continue-on-error: ${{ !inputs.fail-fast }}
     env:
       TESTCONTAINERS_RYUK_DISABLED: "${{ inputs.ryuk-disabled }}"
+      RYUK_RECONNECTION_TIMEOUT: "${{ inputs.project-directory == 'modules/compose' && '30s' || '10s' }}"
```
mdelapenya, line 53: If the CI passes I'm going to document this in the compose module docs

modules/compose/compose_api.go
```diff
@@ -134,6 +135,10 @@ type dockerCompose struct {
 	// used in ServiceContainer(...) function to avoid calls to the Docker API
 	containers map[string]*testcontainers.DockerContainer
 
+	// cache for containers that are part of the stack
+	// used in ServiceContainer(...) function to avoid calls to the Docker API
```
mdelapenya, line 139:
```suggestion
	// cache for networks in the compose stack
```

PR #2504, opened 2024-04-20
feat: Add DynamoDB local module

fhke:
## What does this PR do?

This PR adds a module to allow testing with a [DynamoDB local](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBLocal.html) container.

## Why is it important?

This allows users to test code that interacts with DynamoDB via a pre-canned module.


mdelapenya:
Hi @fhke, thanks for creating this PR. We already have a community module, maintained by an AWS advocate: https://github.com/abhirockzz/dynamodb-local-testcontainers-go. See the module here in our module catalog: https://testcontainers.com/modules/dynamodb/

In our docs site (golang.testcontainers.org) we only include the modules in the repository, although I think it could be a good idea to include all from the modules catalog 🤔

PR #2506, opened 2024-04-21
fix: don't retry on permanent APIClient errors

### LINKED ISSUE #2502: [Enhancement]: define more errors as non-retryable during image pull
### Proposal

If the docker client returns an error during image pull then the command is retried, except for [ErrNotFound](https://github.com/testcontainers/testcontainers-go/blob/main/docker.go#L1269).

There are more error types defined in https://github.com/moby/moby/blob/master/errdefs/defs.go that are also permanent, as far as I can tell.

I ran into the issue that, with invalid credentials, the pull was retried until I got temp blocked by the container registry. Pulling with invalid credentials should not be tried again.

Suggested errors to be seen as permanent/non-retryable:
* `ErrInvalidParameter`
* `ErrUnauthorized`
* `ErrForbidden`
* `ErrNotImplemented`

I'm unsure if all types might be returned by `ImagePull` (e.g. `ErrNotImplemented` seems unlikely). I also didn't check if there are similar places where the docker client is called within a retry mechanism. If so, it might be worth using a central definition that just contains all permanent types.

Happy to provide a PR if you like the suggestion.

p-jahn:
## What does this PR do?

This PR introduces the concept of non-retryable errors that might be returned by the APIClient. If such an error is encountered within a retry loop then the loop is broken since retrying will not change the result.

One exemption is `IsNotFound` during `waitContainerCreation` since the flow relies on that condition.

## Why is it important?

The docker API should not be called repeatedly since this might result in negative side effects (e.g. being blocked by a registry due to repeated failed authorization attempts).

## Related issues

- Closes #2502

## How to test this PR

Some unit tests around the retry behaviour where added.

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -886,8 +886,7 @@ func (p *DockerProvider) BuildImage(ctx context.Context, img ImageBuildInfo) (st
 		resp, err = p.client.ImageBuild(ctx, buildOptions.Context, buildOptions)
 		if err != nil {
 			buildError = errors.Join(buildError, err)
-			var enf errdefs.ErrNotFound
-			if errors.As(err, &enf) {
+			if isPermanentClientError(err) {
 				return backoff.Permanent(err)
 			}
 			Logger.Printf("Failed to build image: %s, will retry", err)
@@ -1166,6 +1165,9 @@ func (p *DockerProvider) waitContainerCreation(ctx context.Context, name string)
 	return container, backoff.Retry(func() error {
 		c, err := p.findContainerByName(ctx, name)
 		if err != nil {
+			if !errdefs.IsNotFound(err) && isPermanentClientError(err) {
+				return backoff.Permanent(err)
+			}
 			return err
 		}
 
@@ -1266,8 +1268,7 @@ func (p *DockerProvider) attemptToPullImage(ctx context.Context, tag string, pul
 	err = backoff.Retry(func() error {
 		pull, err = p.client.ImagePull(ctx, tag, pullOpt)
 		if err != nil {
-			var enf errdefs.ErrNotFound
-			if errors.As(err, &enf) {
+			if isPermanentClientError(err) {
 				return backoff.Permanent(err)
 			}
 			Logger.Printf("Failed to pull image: %s, will retry", err)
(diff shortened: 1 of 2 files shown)


docker.go
```diff
@@ -1603,3 +1604,20 @@ func (p *DockerProvider) SaveImages(ctx context.Context, output string, images .
 func (p *DockerProvider) PullImage(ctx context.Context, image string) error {
 	return p.attemptToPullImage(ctx, image, types.ImagePullOptions{})
 }
+
+var permanentClientErrors = []func(error) bool{
+	errdefs.IsNotFound,
+	errdefs.IsInvalidParameter,
+	errdefs.IsUnauthorized,
+	errdefs.IsForbidden,
+	errdefs.IsNotImplemented,
```
mdelapenya, line 1613: Are all of the values covered by the tests? I can see errors for the NotFound and the InvalidParameter err, but I'm not able to find the other ones yet.

p-jahn: No, I didn't check for all these conditions. I used one likely example condition for each retried client call but didn't want to repeat every condition in all calls since I think it's a bit noisy repeating all conditions in the tests, without much benefit.

I could check the `isPermanentClientError` function itself against all conditions but the function is also very simplistic and didn't seem to benefit much from checking the "completeness" of the list.

It's not an issue for me adding more testing. I have a few suggested approaches, what would you like to see:
* checking the `isPermanentClientError` itself for all expected conditions
* checking for other expectable conditions in the retry tests
* using another approach, being very specific in each client retry case on what errors can be expected and then test against that
* (another approach I did not think of?)

mdelapenya: As you shared, the helper method to check if the error is of the right type is very simplistic, so no need to add an unit test although it would be trivial.

For the rest of the errors not yet covered, I'd prefer having them as explicit tests. I know, we are creating a mock client which will produce those errors, but if at some point we reduce the list of the errors that are not permanent, I'd like to be warned by the tests.

Hope you understand

p-jahn: Hey, I added more scenarios to include the other error types as well.

mdelapenya (approved):
LGTM, thanks!

PR #2521, opened 2024-04-30
feat: Add support to seed data when using RunBigQueryContainer

mtellis2:
## What does this PR do?
Updates the [`RunBigQueryContainer`](https://github.com/testcontainers/testcontainers-go/blob/main/modules/gcloud/bigquery.go#L13) to optionally use the emulator's [`data-from-yaml` option](url), allowing users to seed the container with data.

## Why is it important?
This give the user of the `RunBigQueryContainer` the option to seed test data on the container creation/startup. 

## Related issues

-


modules/gcloud/gcloud.go
```diff
@@ -68,6 +70,20 @@ func WithProjectID(projectID string) Option {
 	}
 }
 
+// WithDataYamlFile seeds the Bigquery project for the GCloud container.
+func WithDataYamlFile(dataYamlFile string) testcontainers.CustomizeRequestOption {
+	return func(req *testcontainers.GenericContainerRequest) {
```
mdelapenya, line 75: Since #2267 all customiser options must return error, could you update that? [This line](https://github.com/testcontainers/testcontainers-go/pull/2521/files#diff-8faaed968761085a8ca6d01c0502d2c6159fabf32bec52d69cd76a86e56ecb87R28) will need to handle the errors 🙏

Issue #1373, opened 2023-07-31
[Bug]: Cannot connect to ryuk when running test inside docker container


n-tucker:
### Testcontainers version

0.21.0

### Using the latest Testcontainers version?

Yes

### Host OS

Mac

### Host arch

ARM

### Go version

1.20.3

### Docker version

```shell
Client:
 Cloud integration: v1.0.31
 Version:           23.0.5
 API version:       1.42
 Go version:        go1.19.8
 Git commit:        bc4487a
 Built:             Wed Apr 26 16:12:52 2023
 OS/Arch:           darwin/arm64
 Context:           default

Server: Docker Desktop 4.19.0 (106363)
 Engine:
  Version:          23.0.5
  API version:      1.42 (minimum version 1.12)
  Go version:       go1.19.8
  Git commit:       94d3ad6
  Built:            Wed Apr 26 16:17:14 2023
  OS/Arch:          linux/arm64
  Experimental:     false
 containerd:
  Version:          1.6.20
  GitCommit:        2806fc1057397dbaeefbea0e4e17bddfbd388f38
 runc:
  Version:          1.1.5
  GitCommit:        v1.1.5-0-gf19387a
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### Docker info

```shell
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.10.4
    Path:     /Users/nicholas.tucker/.docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.17.3
    Path:     /Users/nicholas.tucker/.docker/cli-plugins/docker-compose
  dev: Docker Dev Environments (Docker Inc.)
    Version:  v0.1.0
    Path:     /Users/nicholas.tucker/.docker/cli-plugins/docker-dev
  extension: Manages Docker extensions (Docker Inc.)
    Version:  v0.2.19
    Path:     /Users/nicholas.tucker/.docker/cli-plugins/docker-extension
  init: Creates Docker-related starter files for your project (Docker Inc.)
    Version:  v0.1.0-beta.4
    Path:     /Users/nicholas.tucker/.docker/cli-plugins/docker-init
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc.)
    Version:  0.6.0
[...]

PR #2522, opened 2024-04-30
fixed documentation on TESTCONTAINERS_HOST_OVERRIDE

### LINKED ISSUE #2209: TESTCONTAINERS_HOST_OVERRIDE is documented but not implemented
I'm running into similar issue. I'm following https://golang.testcontainers.org/system_requirements/ci/gitlab_ci/#example-using-docker-socket and so far it doesn't seem to work. Is the `TESTCONTAINERS_HOST_OVERRIDE` variable implemented?

_Originally posted by @arvenil in https://github.com/testcontainers/testcontainers-go/issues/1373#issuecomment-1938564264_

We should remove those references and update to the most updated ones.

hanapedia:
## What does this PR do?
Fixed the documentation around overriding the host of the docker daemon where the container port is exposed.
It was wrongfully documented to set `TESTCONTAINERS_HOST_OVERRIDE` environment variable. The correct environment variable is `TC_HOST`.

## Why is it important?
Current documentation is wrong as setting`TESTCONTAINERS_HOST_OVERRIDE` does not override the host of the docker daemon where the container port is exposed.

## Related issues
- Relates #1373 
- Closes #2209

-

### DIFF
--- a/docs/system_requirements/ci/dind_patterns.md
+++ b/docs/system_requirements/ci/dind_patterns.md
@@ -35,9 +35,9 @@ Where:
 
 
 !!! warning
-    If you are using Docker Desktop, you need to configure the `TESTCONTAINERS_HOST_OVERRIDE` environment variable to use the special DNS name
+    If you are using Docker Desktop, you need to configure the `TC_HOST` environment variable to use the special DNS name
     `host.docker.internal` for accessing the host from within a container, which is provided by Docker Desktop:
-    `-e TESTCONTAINERS_HOST_OVERRIDE=host.docker.internal`
+    `-e TC_HOST=host.docker.internal`
 
 ### Docker Compose example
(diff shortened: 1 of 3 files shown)


mdelapenya:
HI @hanapedia I think what we need here is to replace the usage of TC_HOST in the code with TESTCONTAINERS_HOST_OVERRIDE. That will be aligned with the other testcontainers implementations.

Wdyt?

hanapedia:
@mdelapenya 
Thanks for the feedback! I agree that it will be better to replace `TC_HOST` with `TESTCONTAINERS_HOST_OVERRIDE` if it aligns with other testcontainers projects. I just noticed that `TC_HOST` seemed to be deprecated in other projects.
I will make a new PR on this ASAP.

PR #2523, opened 2024-04-30
feat(gcloud)!: add support to seed data when using RunBigQueryContainer

mtellis2:
## What does this PR do?
Updates the [`RunBigQueryContainer`](https://github.com/testcontainers/testcontainers-go/blob/main/modules/gcloud/bigquery.go#L13) to optionally use the emulator's [`data-from-yaml` option](url), allowing users to seed the container with data.

## Why is it important?
This give the user of the `RunBigQueryContainer` the option to seed test data on the container creation/startup. 

## Related issues

-


modules/gcloud/gcloud.go
```diff
@@ -69,6 +71,22 @@ func WithProjectID(projectID string) Option {
 	}
 }
 
+// WithDataYamlFile seeds the Bigquery project for the GCloud container.
+func WithDataYamlFile(dataYamlFile string) testcontainers.CustomizeRequestOption {
+	return func(req *testcontainers.GenericContainerRequest) error {
```
mtellis2, line 76: Sorry @mdelapenya I recreated the PR to try to get the updates since I had last forked the project. I think I've made the error updates requested and updated the BQ docs to include information on this function.

modules/gcloud/bigquery.go
```diff
@@ -27,18 +27,24 @@ func RunBigQueryContainer(ctx context.Context, opts ...testcontainers.ContainerC
...
 	if err != nil {
 		return nil, err
 	}
 
 	// always prepend http:// to the URI
-	spannerContainer.URI = "http://" + spannerContainer.URI
+	bigqueryContainer.URI = "http://" + bigqueryContainer.URI
 
-	return spannerContainer, nil
+	return bigqueryContainer, nil
```
mdelapenya, line 49: Good catch! 🤦

modules/gcloud/bigquery_test.go
```diff
@@ -13,17 +15,19 @@ import (
...
 	"google.golang.org/grpc/credentials/insecure"
 
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"github.com/testcontainers/testcontainers-go"
 	"github.com/testcontainers/testcontainers-go/modules/gcloud"
 )
 
-func ExampleRunBigQueryContainer() {
+func TestBigQueryContainer(t *testing.T) {
```
mdelapenya, line 24: We would like to keep some testable examples, as they will land in the pkg.go.dev docs. If you have doubts on when creating a real test or a testable example, please use a testable example when you want to show case how to use the module. Else, a regular test is fine.

modules/gcloud/bigquery_test.go
```diff
@@ -78,8 +82,85 @@ func ExampleRunBigQueryContainer() {
...
+	// Output:
+	// [30]
+	expectedValue := int64(30)
+	actualValue := val[0]
+	fmt.Println(val[0])
+
+	require.NoError(t, err)
+	if assert.NotNil(t, val) {
+		assert.Equal(t, expectedValue, actualValue)
+	}
```
mdelapenya, line 94: Please use one way for assertions: testable examples (printing outputs) or regular tests (using the testing library)

modules/gcloud/bigquery_test.go
```diff
@@ -78,8 +82,85 @@ func ExampleRunBigQueryContainer() {
...
+}
+
+func TestBigQueryWithDataYamlFile(t *testing.T) {
+	// runBigQueryContainer {
```
mdelapenya, line 98: This markup is for embedding code snippets into the docs, and that key is probably already in use. Please remove it, alongside its closing mark.

modules/gcloud/bigquery.go
```diff
@@ -27,18 +27,24 @@ func RunBigQueryContainer(ctx context.Context, opts ...testcontainers.ContainerC
 
 	req.Cmd = []string{"--project", settings.ProjectID}
 
+	for _, opt := range opts {
+		if err := opt.Customize(&req); err != nil {
+			return nil, err
+		}
+	}
+
```
mdelapenya, line 35: Mmm I think the [applyOptions](https://github.com/testcontainers/testcontainers-go/pull/2523/files#diff-8faaed968761085a8ca6d01c0502d2c6159fabf32bec52d69cd76a86e56ecb87R23) call will already do this. Please see https://github.com/testcontainers/testcontainers-go/pull/2523/files#diff-658ed21fe179e3c4810748c731c20b75757c269dd8ce30f0bf98ce31f2addd24R91-R103

modules/gcloud/go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	cloud.google.com/go/pubsub v1.36.2
 	cloud.google.com/go/spanner v1.57.0
 	github.com/docker/go-connections v0.5.0
+	github.com/stretchr/testify v1.9.0
```
mdelapenya, line 13: I think we can live without testify. It's indeed used in other modules, but we can keep this one without it for now. Wdyt?

mtellis2: Yeah this shouldn't be needed

modules/gcloud/bigquery.go
mdelapenya: Let me test this locally, from this branch. Will come back to you later

mdelapenya: Indeed, with the addition of the YAML data file, now it's possible to have different settings contributing to the CMD. So it's fine if we append.

My only concern is now related to the usage of the gcloud options across all the emulators 🤔 Do you know if all of them would support passing this YAML file? Else, anybody could add this functional option to e.g. Firestore or Pubsub, and expect something else. I'm not an expert in those services, so you probably have more insights.

mtellis2: Good point, I myself have not used the other emulators but would assume this data yaml functionality would only apply for the bigquery emulator. I'm currently not sure of a great way to prevent this in the other gcloud emulators. I think the data.yaml option should be ignored though due to how that CMD is being set.

modules/gcloud/bigquery_test.go
```diff
@@ -78,8 +80,75 @@ func ExampleRunBigQueryContainer() {
...
+	// 30
+}
+
+func TestBigQueryWithDataYamlFile(t *testing.T) {
+	ctx := context.Background()
+
+	absPath, err := filepath.Abs(filepath.Join(".", "testdata", "data.yaml"))
+	if err != nil {
+		log.Fatalf("failed to run container: %v", err)
+	}
```
stevenh, line 94: suggestion: use require.NoError to simplify, more below
```suggestion
	require.NoError(t, err)
```

modules/gcloud/bigquery_test.go
```diff
@@ -78,8 +80,75 @@ func ExampleRunBigQueryContainer() {
...
+	)
+	if err != nil {
+		log.Fatalf("failed to run container: %v", err)
+	}
+
+	defer func() {
+		if err := bigQueryContainer.Terminate(ctx); err != nil {
+			log.Fatalf("failed to terminate container: %v", err)
+		}
+	}()
```
stevenh, line 110: suggestion: use `testcontainers.CleanupContainer` to simplify and ensure correct cleanup.

modules/gcloud/bigquery_test.go
```diff
@@ -78,8 +80,75 @@ func ExampleRunBigQueryContainer() {
...
+	var val []bigquery.Value
+	for {
+		err := it.Next(&val)
+		if errors.Is(err, iterator.Done) {
+			break
+		}
+		if err != nil {
+			log.Fatalf("failed to iterate: %v", err)
+		}
+	}
```
stevenh, line 145: bug: left over output comments below.

modules/gcloud/bigquery_test.go
```diff
@@ -78,8 +80,75 @@ func ExampleRunBigQueryContainer() {
...
 
 	// Output:
 	// [30]
+	expectedValue := int64(30)
+	actualValue := val[0]
+	if expectedValue != actualValue {
```
stevenh, line 151: suggestion: use require.Equal or EqualValue

modules/gcloud/gcloud.go
```diff
@@ -41,12 +41,14 @@ func newGCloudContainer(ctx context.Context, port int, c testcontainers.Containe
 }
 
 type options struct {
-	ProjectID string
+	ProjectID    string
+	DataYamlFile string
```
stevenh, line 45: bug: this doesn't seem to be ever used.

modules/gcloud/gcloud.go
```diff
@@ -69,6 +71,22 @@ func WithProjectID(projectID string) Option {
 	}
 }
 
+// WithDataYamlFile seeds the Bigquery project for the GCloud container.
+func WithDataYamlFile(dataYamlFile string) testcontainers.CustomizeRequestOption {
```
stevenh, line 75: suggestion: remove the duplication between function name and paramater

modules/gcloud/bigquery.go
```diff
@@ -25,20 +25,20 @@ func RunBigQueryContainer(ctx context.Context, opts ...testcontainers.ContainerC
...
-	req.Cmd = []string{"--project", settings.ProjectID}
+	req.Cmd = append(req.Cmd, "--project", settings.ProjectID)
 
 	container, err := testcontainers.GenericContainer(ctx, req)
 	if err != nil {
 		return nil, err
 	}
 
-	spannerContainer, err := newGCloudContainer(ctx, 9050, container, settings)
+	bigqueryContainer, err := newGCloudContainer(ctx, 9050, container, settings)
```
stevenh, line 35: Looks like this needs a rebase as there has been some work to ensure this correctly returns the container even on error.

stevenh (changes requested):
Thanks for the PR. I've done an initial pass but looks like this needs a rebase.

modules/gcloud/gcloud.go
mdelapenya: Hey @mtellis2 sorry for the late response. I'm taking this PR and resolving conflicts after the merge with main. I also addressed @stevenh's comments about using require and the new cleanup APIs for tests.

I have one concern regarding the functional option: if a user passes the option multiple times, it will be appending multiple files to be copied before the container is started, but overriding the file in the container (see https://github.com/testcontainers/testcontainers-go/pull/2523/files#diff-658ed21fe179e3c4810748c731c20b75757c269dd8ce30f0bf98ce31f2addd24R79), which means that the last added YAML file wins. Same for the CMD, as we are appending a new entry to the CMD args. I verified it's harmless regarding goccy's emulator to receive multiple values for the same flag, but then I verified that, if another data yaml file is added, the latest added yaml file wins.

To resolve it, I did a quick test and changed the option to support multiple files to be added, and the emulator just uses the last `--data-from-yaml` flag value. So I'm changing the implementation to support just one YAML file, adding a note in the docs for the functional option regarding this.

Hope you can take a look at the reviewed PR with my commits on top of yours.

modules/gcloud/bigquery.go
mdelapenya: I'm only processing the yaml data file in the bigQuery container. Until we have a better abstraction, this is fine I think

modules/gcloud/gcloud.go
stevenh: Yes, I would suggest just `filename`.

However based on recent discussions and changes in other modules should we pass a reader instead?

modules/gcloud/bigquery.go
```diff
@@ -31,7 +32,20 @@ func RunBigQuery(ctx context.Context, img string, opts ...testcontainers.Contain
...
+	// Process data yaml file only for the BigQuery container.
+	if settings.bigQueryDataYamlFile != "" {
+		containerPath := "/" + filepath.Base(settings.bigQueryDataYamlFile)
+
+		req.Cmd = append(req.Cmd, "--data-from-yaml", containerPath)
+
+		req.Files = append(req.Files, testcontainers.ContainerFile{
+			HostFilePath:      settings.bigQueryDataYamlFile,
+			ContainerFilePath: containerPath,
+			FileMode:          0o755,
```
stevenh, line 46: suggestion: no need for the file to be executable.

modules/gcloud/gcloud.go
```diff
@@ -72,6 +73,18 @@ func WithProjectID(projectID string) Option {
 	}
 }
 
+// WithDataYamlFile seeds the Bigquery project for the GCloud container
+// with the local path to the data yaml file, which is used to copy the file to the container,
+// and then process the file to seed the Bigquery project.
+//
+// Other GCloud containers will ignore this option.
+// If this option is passed multiple times, the last file will be used.
+func WithDataYamlFile(f string) Option {
```
stevenh, line 82: question: use a `ContainerCustomiser` and pass a `io.Reader` instead?

modules/gcloud/gcloud.go
mdelapenya: Yeah, makes sense. Let me change it right now. We can just simply generate a random name for the container file, can't we?

stevenh (changes requested):
Some comments taking into account recent discussions about other modules.

modules/gcloud/gcloud.go
mdelapenya: If we pass a customiser here, we'd need to append the file to the req.Files field. If users pass the option multiple times, as advised in the docs, the last one will win, but the code would unnecessarily copy multiple files to the container. That's why I went with the Option, to store the file (the reader) into the options type and then process it when needed.

modules/gcloud/gcloud.go
stevenh: Yep or use a fixed name, not sure what happens if there are multiple, that might be valid.

modules/gcloud/gcloud.go
stevenh: Yes but that too is confusing, I would suggest that two uses should error to make it clear to the user what they did wrong.

```go
func WithDataYAML(r io.Reader) testcontainers.CustomizeRequestOption {
		const dataFlag = "--data-from-yaml"
		for i, flag := range req.Cmd {
			if flag == dataFlag {
				if i+1 < len(req.Cmd) {
					return fmt.Errorf("%s already set to %q", dataFlag, req.Cmd[i+1])
				}
				return fmt.Errorf("%s already set without a value", dataFlag)
			}
		}

		containerPath := "/testcontainers-data.yaml"
		req.Cmd = append(req.Cmd, dataFlag, containerPath)

		req.Files = append(req.Files, testcontainers.ContainerFile{
			Reader:            r,
			ContainerFilePath: containerPath,
			FileMode:          0o644,
		})

		return nil
	}
}
```

Updated tests to match using embed to simplify:
```go

//go:embed testdata/data.yaml
var dataYaml []byte

func TestBigQueryWithDataYamlFile(t *testing.T) {
	ctx := context.Background()

	t.Run("valid", func(t *testing.T) {
		bigQueryContainer, err := gcloud.RunBigQuery(
			ctx,
			"ghcr.io/goccy/bigquery-emulator:0.6.1",
			gcloud.WithProjectID("test"),
			gcloud.WithDataYAML(bytes.NewReader(dataYaml)),
		)
		testcontainers.CleanupContainer(t, bigQueryContainer)
		require.NoError(t, err)

		projectID := bigQueryContainer.Settings.ProjectID

		opts := []option.ClientOption{
			option.WithEndpoint(bigQueryContainer.URI),
			option.WithGRPCDialOption(grpc.WithTransportCredentials(insecure.NewCredentials())),
			option.WithoutAuthentication(),
			internaloption.SkipDialSettingsValidation(),
		}

		client, err := bigquery.NewClient(ctx, projectID, opts...)
		require.NoError(t, err)
		defer client.Close()

		selectQuery := client.Query("SELECT * FROM dataset1.table_a where name = @name")
		selectQuery.QueryConfig.Parameters = []bigquery.QueryParameter{
			{Name: "name", Value: "bob"},
		}
		it, err := selectQuery.Read(ctx)
		require.NoError(t, err)

		var val []bigquery.Value
		for {
			err := it.Next(&val)
			if errors.Is(err, iterator.Done) {
				break
			}
			require.NoError(t, err)
		}

		require.Equal(t, int64(30), val[0])
	})

	t.Run("multi-value-set", func(t *testing.T) {
		bigQueryContainer, err := gcloud.RunBigQuery(
			ctx,
			"ghcr.io/goccy/bigquery-emulator:0.6.1",
			gcloud.WithProjectID("test"),
			gcloud.WithDataYAML(bytes.NewReader(dataYaml)),
			gcloud.WithDataYAML(bytes.NewReader(dataYaml)),
		)
		testcontainers.CleanupContainer(t, bigQueryContainer)
		require.EqualError(t, err, `--data-from-yaml already set to "/testcontainers-data.yaml"`)
	})

	t.Run("multi-value-not-set", func(t *testing.T) {
		noValueOption := func() testcontainers.CustomizeRequestOption {
				req.Cmd = append(req.Cmd, "--data-from-yaml")
				return nil
			}
		}

		bigQueryContainer, err := gcloud.RunBigQuery(
			ctx,
			"ghcr.io/goccy/bigquery-emulator:0.6.1",
			noValueOption(),
			gcloud.WithProjectID("test"),
			gcloud.WithDataYAML(bytes.NewReader(dataYaml)),
		)
		testcontainers.CleanupContainer(t, bigQueryContainer)
		require.EqualError(t, err, `--data-from-yaml already set without a value`)
	})
}
```

mdelapenya: Cool, I like the error if the option is passed more than once. Will implement the embed thing (always forget it's available!)

mdelapenya: One thing: at the moment we convert this option into a Customiser, then any other GCloud container will be able to use it. With the options approach, it's only processed in the right place, but I'm realising it's not following the "return error" API for the options. Should we add a BC here?

At some point we should tackle an specialisation of the current GCloud containers, or at least the options, so that they can be only used by their respective owners. Using packages for each container could be an option, so users will know the option belongs to a given package in an explicit manner.

stevenh: It's already documented that its only valid for bigquery. We could add a check based on an env var if you wanted to make it explicit, returning an error if its used in an unexpected way, however I would expect you get an error when used incorrectly anyway due to basic arg validation from the containers themselves.

Previously the arg would be silently ignored by other containers, which leads the user to believe its been applied making it confusing. I always prefer things to error if I use them wrong as it doesn't leave you scratching your head wondering why something didn't work.

mdelapenya: That's fair. In theory, any module is exporting options, so why not breaking my mysql container with a neo4j option? 🤷 

So failing is fine then, thanks!

modules/gcloud/bigquery_test.go
```diff
@@ -81,8 +87,74 @@ func ExampleRunBigQueryContainer() {
...
+
+	r, err := os.Open(filepath.Join(testDataPath, "data.yaml"))
+	require.NoError(t, err)
+
+	bigQueryContainer, err := gcloud.RunBigQuery(
+		ctx,
+		"ghcr.io/goccy/bigquery-emulator:0.6.1",
+		gcloud.WithProjectID("test"),
+		gcloud.WithDataYAML(r),
+	)
```
stevenh, line 109: suggestion: use dataYaml for this to simplify.
```suggestion
	bigQueryContainer, err := gcloud.RunBigQuery(
		ctx,
		"ghcr.io/goccy/bigquery-emulator:0.6.1",
		gcloud.WithProjectID("test"),
		gcloud.WithDataYAML(bytes.NewReader(dataYaml)),
	)
```

modules/gcloud/bigquery_test.go
```diff
@@ -81,8 +87,74 @@ func ExampleRunBigQueryContainer() {
...
+
+func TestBigQueryWithDataYamlFile_multiple(t *testing.T) {
+	ctx := context.Background()
+
+	bigQueryContainer, err := gcloud.RunBigQuery(
+		ctx,
+		"ghcr.io/goccy/bigquery-emulator:0.6.1",
+		gcloud.WithProjectID("test"),
+		gcloud.WithDataYAML(strings.NewReader(string(dataYaml))),
+		gcloud.WithDataYAML(strings.NewReader(string(dataYaml))), // last file will cause an error
```
stevenh, line 156: suggestion: use bytes.NewReader to avoid conversion
```suggestion
		gcloud.WithDataYAML(bytes.NewReader(dataYaml)),
		gcloud.WithDataYAML(bytes.NewReader(dataYaml)), // second use will cause an error
```

modules/gcloud/gcloud.go
```diff
@@ -67,8 +70,26 @@ func (o Option) Customize(*testcontainers.GenericContainerRequest) error {
...
+// WithDataYAML seeds the Bigquery project for the GCloud container with an io.Reader representing
+// the data yaml file, which is used to copy the file to the container, and then processed to seed
+// the Bigquery project.
+//
+// Other GCloud containers will ignore this option.
+// If this option is passed multiple times, the last added will be used.
+func WithDataYAML(r io.Reader) Option {
+	return func(o *options) error {
+		if o.bigQueryDataYaml != nil {
+			return errors.New("data yaml file already exists")
```
stevenh, line 88: suggestion: this isn't a file anymore so tweak the wording to match
```suggestion
			return errors.New("data yaml already set")
```

modules/gcloud/gcloud.go
```diff
@@ -67,8 +70,26 @@ func (o Option) Customize(*testcontainers.GenericContainerRequest) error {
...
 // WithProjectID sets the project ID for the GCloud container.
 func WithProjectID(projectID string) Option {
-	return func(o *options) {
+	return func(o *options) error {
 		o.ProjectID = projectID
+		return nil
+	}
+}
+
+// WithDataYAML seeds the Bigquery project for the GCloud container with an io.Reader representing
```
stevenh, line 79: suggestion: brace to type so it doc links and use the canonical name
```suggestion
// WithDataYAML seeds the BigQuery project for the GCloud container with an [io.Reader] representing
```

modules/gcloud/gcloud.go
```diff
@@ -67,8 +70,26 @@ func (o Option) Customize(*testcontainers.GenericContainerRequest) error {
...
+// the data yaml file, which is used to copy the file to the container, and then processed to seed
+// the Bigquery project.
+//
+// Other GCloud containers will ignore this option.
+// If this option is passed multiple times, the last added will be used.
```
stevenh, line 84: bug: this is no longer true.

mdelapenya: Oh my! 🤦🏻

mdelapenya:
Adding the breaking change label because the Options customiser type now returns an error

modules/gcloud/bigquery_test.go
```diff
@@ -81,8 +88,84 @@ func ExampleRunBigQueryContainer() {
 		}
 	}
 
-	fmt.Println(val)
-
+	fmt.Println(val[0])
```
stevenh, line 91: Question: could this change lead to a panic if no result was returned?

stevenh (approved):
Looks good just one question on the change to the example which might be better to leave as it was?

modules/gcloud/bigquery_test.go
mdelapenya: It's a testable example where we expect it will return the error. If for some reason any of the BigQuery impl iror the client change, we will be notified by the failure. But let me revert it, as it's harmless in both directions

stevenh (approved):
LGTM

mdelapenya:
@mtellis2 thanks for your work here. We took it over to not waste your time and make it into the next release.

We really appreciate it 🙇

PR #2531, opened 2024-05-05
Update devcontainer image

bells17:
## What does this PR do?

Update the devcontainer image to the following image: `mcr.microsoft.com/devcontainers/go:1.22-bookworm`.

## Why is it important?

Currently, we cannot open the project using devcontainer because the devcontainer image `mcr.microsoft.com/devcontainers/go:0-1.21-bullseye` cannot be found.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -2,7 +2,7 @@
 // README at: https://github.com/devcontainers/templates/tree/main/src/go
 {
 	"name": "Go",
-	"image": "mcr.microsoft.com/devcontainers/go:0-1.21-bullseye",
+	"image": "mcr.microsoft.com/devcontainers/go:1.22-bookworm",
 
 	// Features to add to the dev container. More info: https://containers.dev/features.
 	// "features": {},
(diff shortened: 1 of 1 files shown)


.devcontainer/devcontainer.json
```diff
@@ -2,7 +2,7 @@
 // README at: https://github.com/devcontainers/templates/tree/main/src/go
 {
 	"name": "Go",
-	"image": "mcr.microsoft.com/devcontainers/go:0-1.21-bullseye",
+	"image": "mcr.microsoft.com/devcontainers/go:1.22-bookworm",
```
mdelapenya, line 5: Could you use a Go 1.21 image? We set the base version as the lowest supported Go version, and we bump it whenever there is an update in the Go EOL (https://endoflife.date/go)

mdelapenya (commented):
Thanks for submitting the changes! Could you take a look at the Go version 🙏 ?

Cheers

.devcontainer/devcontainer.json
bells17: @mdelapenya Thank you for reviewing my changes.
I've changed the Go version, so could you please take another look?

mdelapenya (approved):
LGTM, thanks!

mdelapenya:
The influx error is not related: it will be fixed with #2532

PR #2532, opened 2024-05-06
fix(influxdb): include more characters in wait for log regex

mdelapenya:
## What does this PR do?
It updates the regex for InfluxDB v2, as we found the log output changed its format.

For that we have used a fixed version of the v2 image: `2.7.5-alpine`.

## Why is it important?
The CI started to fail for influxdb:latest

## Related issues

- Discovered while testing #2531


mdelapenya:
Merging, as influx tests already passing.

PR #2537, opened 2024-05-10
fix: add import to module generation

mdelapenya:
## What does this PR do?
This PR adds a missing import for "fmt" in the template that generates the module Go file.

## Why is it important?
The modules are generated, but the process fails because the code does not compile: missing import.

## Related issues

- Caused by #2267

## Follow-ups
There is no need to create a patch release, as users of the module generation tool will use it from the source code.

### DIFF
--- a/modulegen/_template/module.go.tmpl
+++ b/modulegen/_template/module.go.tmpl
@@ -2,6 +2,7 @@
 
 import (
 	"context"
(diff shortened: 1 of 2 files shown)


mdelapenya:
Merging, as the tests for the modulegen already passed.

PR #2541, opened 2024-05-17
feat(ryuk): allow to configure ryuk timeouts using env variables

fcx-mrogez:
## What does this PR do?

* Add the possibility to use environment variables to configure ryuk timeouts (reconnection and connection).
* Update unit tests
* Update documentation
* Fix up a minor typo in tests
 
## Why is it important?

Default timeout are not suitable in all situations :
* tests taking longer than 10s gets killed before completing (and may make also the following test fail from what I've observed)
* if image building / pulling takes more than 1m to complete, ryuk will timeout and make the test fail.

Configuring these timeouts is currently done via property file which must be located in user home.
This is not very convenient in CI context (need to create proper property file in runner home).
For some common configuration settings, such as TESTCONTAINERS_RYUK_{DISABLE,VERBOSE} it is possible to use env variable which is much more convenient. So we extend this logic to ryuk timeouts.

## Related issues

Couldn't find any.

### DIFF
--- a/docs/features/configuration.md
+++ b/docs/features/configuration.md
@@ -50,8 +50,8 @@ Please read more about customizing images in the [Image name substitution](image
 1. If your environment already implements automatic cleanup of containers after the execution,
 but does not allow starting privileged containers, you can turn off the Ryuk container by setting
 `TESTCONTAINERS_RYUK_DISABLED` **environment variable** to `true`.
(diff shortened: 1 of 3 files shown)


mdelapenya (approved):
LGTM, thanks!

mdelapenya:
Thanks for improving the behaviour in the library! Much appreciated 👏

PR #2551, opened 2024-05-23
Fix network accessor for port-forwarding feature

JulienBreux:
## What does this PR do?

This pull-request fixes a bug in the port-fowarding functionality.
There is confusion between the use of the network name and the network identifier.
In order to correct the problem, a new function for selecting the network by name has been added.

### DIFF
--- a/internal/core/network/network.go
+++ b/internal/core/network/network.go
@@ -10,8 +10,24 @@ import (
 	"github.com/testcontainers/testcontainers-go/internal/core"
 )
 
+type filter string
+
+const (
+	idFilter   filter = "id"
+	nameFilter filter = "name"
+)
+
 // Get returns a network by its ID.
 func Get(ctx context.Context, id string) (types.NetworkResource, error) {
+	return get(ctx, idFilter, id)
+}
+
+// GetByName returns a network by its name.
+func GetByName(ctx context.Context, name string) (types.NetworkResource, error) {
+	return get(ctx, nameFilter, name)
+}
+
+func get(ctx context.Context, f filter, value string) (types.NetworkResource, error) {
 	var nw types.NetworkResource // initialize to the zero value
 
 	cli, err := core.NewClient(ctx)
@@ -21,7 +37,7 @@ func Get(ctx context.Context, id string) (types.NetworkResource, error) {
 	defer cli.Close()
 
 	filters := filters.NewArgs()
-	filters.Add("id", id)
+	filters.Add(string(f), value)
 
 	list, err := cli.NetworkList(ctx, types.NetworkListOptions{Filters: filters})
 	if err != nil {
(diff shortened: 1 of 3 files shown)


JulienBreux:
Note: another proposal would be `network.Get(ctx, FilterID, value)` and `network.Get(ctx, FilterName, value)`

JulienBreux:
@mdelapenya could you help me? 
Sorry for direct ping dear friend!
But it's a blocker for my TS-Go usage 🥹

kiview:
Hey @JulienBreux, could you please elaborate in what kind of behavior this issue manifests? Or how it can be reproduced? Asking, because I want to align with our implementation in other languages.

JulienBreux:
> Hey @JulienBreux, could you please elaborate in what kind of behavior this issue manifests? Or how it can be reproduced? Asking, because I want to align with our implementation in other languages.

Sure! If you read this PR, you'll see that in the code there is a confusion between "name" and "id" of network.
I just fix this because, always I want to use this feature with a specific custom network, I've the error "network not found".
Normal behaviour, because the code search the first network "by name" and put to a `network.Get` by ID.

IDK if I'm clear.
So to reproduce, easy:

1) Create a network
2) Create a generic container using this network
2.1) Configure the "ExposeHostPorts" field

Network not found!

port_forwarding.go
```diff
@@ -55,7 +55,7 @@ func exposeHostPorts(ctx context.Context, req *ContainerRequest, p ...int) (Cont
 	opts := []ContainerCustomizer{}
 	if len(req.Networks) > 0 {
 		// get the first network of the container to connect the SSHD container to it.
-		nw, err := network.Get(ctx, sshdFirstNetwork)
+		nw, err := network.GetByName(ctx, sshdFirstNetwork)
```
mdelapenya, line 58: Oh good catch, the first network is a network name, not the ID

internal/core/network/network.go
```diff
@@ -21,15 +37,15 @@ func Get(ctx context.Context, id string) (types.NetworkResource, error) {
 	defer cli.Close()
 
 	filters := filters.NewArgs()
-	filters.Add("id", id)
+	filters.Add(string(f), value)
```
mdelapenya, line 40: Do we need the custom type just to cast here?

JulienBreux: You got me 😊 
Not necessary!
I'll modify you're right 😄

mdelapenya (approved):
LGTM, thanks!

JulienBreux:
@mdelapenya in your own opinion, this feature is for a next patch, for example "v0.31.1" or for a new major version "v0.32.0" ?
Thanks to you, and my apologies for the direct (force) touch. 😊

mdelapenya:
I'm usually doing a monthly minor release, so it will probably land into the next minor.

PR #2536, opened 2024-05-08
fix: Rename TC_HOST environment variable to TESTCONTAINERS_HOST_OVERRIDE

hanapedia:
## What does this PR do?
Rename `TC_HOST` environment variable to `TESTCONTAINERS_HOST_OVERRIDE` to align with other testcontainers projects.

## Why is it important?
The changes in this PR will align how to override the host of the docker daemon where the container port is exposed with other testcontainers projects and the current `testcontainers-go` documentation.

In other testcontainers projects, `TESTCONTAINERS_HOST_OVERRIDE` environment variable is used to override host of the docker daemon where the container port is exposed. `testcontainers-go` mentions this environment variable in the documentation but it is currently not implemented. Instead, `testcontainers-go` uses `TC_HOST`, which is the old envrionment variable used in other projects for the same purpose. 

## Related issues
- Relates #1373 
- Closes #2209
- Relates #2522 

- 

## Follow-ups
The changes in this PR will deprecate `TC_HOST` environment variable.

PR #2555, opened 2024-05-28
Docs: Update documentation for TESTCONTAINERS_HOST_OVERRIDE

david-kalmakoff:
## What does this PR do?

I changed the documentation for the `TESTCONTAINERS_HOST_OVERRIDE` environment variable which is not used, instead the `TC_HOST` is used.

## Why is it important?

If someone is following the documentation they will not have the desired results.

## Related issues

### DIFF
--- a/docs/system_requirements/ci/dind_patterns.md
+++ b/docs/system_requirements/ci/dind_patterns.md
@@ -35,9 +35,9 @@ Where:
 
 
 !!! warning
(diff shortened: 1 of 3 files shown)


mdelapenya:
Hey @david-kalmakoff, thanks for submitting this PR. I think it overlaps with https://github.com/testcontainers/testcontainers-go/pull/2536, which I think is more correct in order to keep the env vars with a consistent prefix.

Wdyt if we close this one in favour of the other one?

PR #2560, opened 2024-05-30
feat: add vearch module

zhanghexian:
## What does this PR do?

add vearch modules to generate dependence!

## Why is it important?
To add vearch in langchaingo

PR #2568, opened 2024-06-07
chore: pin vearch version

mdelapenya:
## What does this PR do?
It pins the version of the Vearch module, to avoid using latest.

## Why is it important?
For some reason I noticed it during the review but did not commented about it. Changing it is important for reproducibility and not receive a wrong version of the module each time it's used.

## Related issues

- Caused by #2560


mdelapenya:
@zhanghexian please take a look. I forgot to change this in your original PR.

Cheers!

mdelapenya:
FTR I'm updating the modules catalog here: https://github.com/testcontainers/community-module-registry/pull/51

PR #2577, opened 2024-06-12
fix: prepend substitutors for built images

### LINKED ISSUE #2554: [Enhancement]: Allow temporarily disabling `TESTCONTAINERS_HUB_IMAGE_NAME_PREFIX` via `ImageSubstitutor`
### Proposal

When using a custom implementation of an `ImageSubstitutor` while also having set `TESTCONTAINERS_HUB_IMAGE_NAME_PREFIX`, the prefix will be prepended if testcontainers decides that the image name still needs a substitution before starting the container. This is adequate behavior for all the cases where there should be a default registry to fall back to if none of the prior `ImageSubstitutor`s set one. But it poses a problem when the image under test was built and tagged locally and has not been published to a registry yet.
It would be great if an `ImageSubstitutor` could have the option disable the prefixing mechanism.

mdelapenya:
## What does this PR do?
This PR considers the case where an image can be built from a Dockerfile and receives Image substitutors, applying them to build the complete image name after the build.

## Why is it important?
Built images did not receive the image substitutors

## Related issues

- Closes #2554

### DIFF
--- a/container.go
+++ b/container.go
@@ -375,6 +375,20 @@ func (c *ContainerRequest) BuildOptions() (types.ImageBuildOptions, error) {
 
 	// make sure the first tag is the one defined in the ContainerRequest
 	tag := fmt.Sprintf("%s:%s", c.GetRepo(), c.GetTag())
+
+	// apply substitutors to the built image
+	for _, is := range c.ImageSubstitutors {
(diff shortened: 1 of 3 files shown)


mdelapenya:
Even though the original issue was closed, I think this is needed in order to honor substitutors when building images.

PR #2584, opened 2024-06-14
chore(deps): bump github.com/docker/docker from v25.0.5 to v26.1.4

thaJeztah:
## What does this PR do?

Updating the docker dependency to the current release.

- rename variables that shadowed imports
- Update docker and dependencies
- Remove uses of deprecated API types
- Deprecate network.WithCheckDuplicate(), which is automatically set on older API versions
- Deprecate NetworkCreate.CheckDuplicate
- Remove uses of network.WithCheckDuplicate()

## Related issues

- closes https://github.com/testcontainers/testcontainers-go/issues/2583

-

PR #2586, opened 2024-06-14
docs: update contributing guide

mdelapenya:
Render URL: https://deploy-preview-2586--testcontainers-go.netlify.app/contributing/

## What does this PR do?
This PR updates the contributing guide including how to update the codebase when the core go.mod is updated.

It also include info about passing golangci-lint in the codebase (core and submodules).

Finally it updates the makefile to use simpler names.

## Why is it important?
Everytime the core go.mod is updated, the changes must be propagated to all the downstream modules.

## Related issues

- Relates #2584

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -8,7 +8,7 @@
     * follow the style, naming and structure conventions of the rest of the project.
     * make commits atomic and easy to merge.
     * when updating documentation, please see [our guidance for documentation contributions](contributing_docs.md).
-    * when updating the `go.mod` file, please run `make tidy-all` to ensure all modules are updated. Please be aware that the lint stage could fail if this is not done.
+    * when updating the `go.mod` file, please run `make tidy-all` to ensure all modules are updated. It will run `golangci-lint` with the configuration set in the root directory of the project. Please be aware that the lint stage could fail if this is not done.
     * apply format running `make lint`
         * For examples: `make -C examples lint`
         * For modules: `make -C modules lint`
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -8,5 +8,9 @@
     * follow the style, naming and structure conventions of the rest of the project.
     * make commits atomic and easy to merge.
     * when updating documentation, please see [our guidance for documentation contributions](contributing_docs.md).
-    * apply format running `go fmt`, or the shell script `./scripts/checks.sh`
+    * when updating the `go.mod` file, please run `make tidy-all` to ensure all modules are updated. Please be aware that the lint stage could fail if this is not done.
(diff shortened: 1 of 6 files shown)


docs/contributing.md
```diff
@@ -8,5 +8,9 @@
     * follow the style, naming and structure conventions of the rest of the project.
     * make commits atomic and easy to merge.
     * when updating documentation, please see [our guidance for documentation contributions](contributing_docs.md).
-    * apply format running `go fmt`, or the shell script `./scripts/checks.sh`
+    * when updating the `go.mod` file, please run `make tidy-all` to ensure all modules are updated. Please be aware that the lint stage could fail if this is not done.
```
mdelapenya, line 11:
```suggestion
    * when updating the `go.mod` file, please run `make tidy-all` to ensure all modules are updated. It will run `golangci-lint` with the configuration set in the root directory of the project. Please be aware that the lint stage could fail if this is not done.
```

docs/contributing.md
```diff
@@ -8,7 +8,7 @@
     * follow the style, naming and structure conventions of the rest of the project.
     * make commits atomic and easy to merge.
     * when updating documentation, please see [our guidance for documentation contributions](contributing_docs.md).
-    * when updating the `go.mod` file, please run `make tidy-all` to ensure all modules are updated. Please be aware that the lint stage could fail if this is not done.
+    * when updating the `go.mod` file, please run `make tidy-all` to ensure all modules are updated. It will run `golangci-lint` with the configuration set in the root directory of the project. Please be aware that the lint stage could fail if this is not done.
```
thaJeztah, line 11: Not a blocker, but I notice the use of "please" and future tense; 

- Using "please" is generally discouraged; see here for some more on that https://github.com/docker/cli/pull/5040#issuecomment-2113053344
- In tech docs it's encouraged to use present time  (In this case "It runs ... "); tech docs tends to be in the here and now; if you do this, then this happens (then this "will happen" is both lengthier as well as gives the impression "eventually ... ")

thaJeztah (approved):
LGTM (but didn't test locally 😅), thanks!

docs/contributing.md
mdelapenya: TIL

I can change that in a follow-up for all the docs. Wdyt?

thaJeztah: Definitely! Not important at all. Just sharing some bits I learned from the experts over the Years.

The funny quote in that case was _"There's no future in tech-writing"_

Me: "that sounds really doomy; sure there must be!?"

And then she explained it was about "future tense" 😂😂🤣

PR #2589, opened 2024-06-15
network: WithCheckDuplicate: don't set CheckDuplicate value

thaJeztah:
## What does this PR do?

Setting this value is a no-opt, as it's automatically set by the client, and the field will be removed in the next release of the docker module, so let's already remove it.

updates 6e086e8de13a97106bbb29a5a8aee7b0d28fe963, which deprecated this function.

## Related issues

- follow-up to https://github.com/testcontainers/testcontainers-go/pull/2584


mdelapenya (approved):
LGTM, thanks!

PR #2594, opened 2024-06-18
Add support for userns remapping

jan-walther:
## What does this PR do?

This PR adds the configuration values `ryuk.userns` / `TESTCONTAINERS_RYUK_USER_NS`, which will cause the ryuk container to run in the host user namespace instead of the default one. This is required because Ryuk mounts the Docker socket into the container and if this isn't done it will receive a permission denied error when trying to access it.

## Why is it important?

Enable scenarios where the user namespace has been remapped.

## Related issues

N/A

## How to test this PR

Automatically testing this change is difficult because it requires changing the dockerd configuration.

- Configure the Docker daemon to use `--userns-remap` according to this guide: https://docs.docker.com/engine/security/userns-remap/
- Set `TESTCONTAINERS_RYUK_USER_NS` to `true`
- Run Ryuk


mdelapenya:
Hi @jan-walther thanks for submitting this PR. We are discussing about it internally and in the meantime we'd like to know more about your use case, like in which situations you remap the userns, so we can have a much better understanding in problem space.

Thanks in advance!

jan-walther:
Hi, sorry for the late reply! We're running multiple Docker daemons on a single machine which requires us to use `--userns-remap`.

mdelapenya:
I'd like to summon @thaJeztah for a better understanding of `userns` remapping 🙏

PR #2598, opened 2024-06-19
fix: allow compose files and readers to be used together

BoKleynen:
## What does this PR do?

Both `WithStackFiles` and `WithStackReaders` assign to the `Paths` field of the `composeStackOptions` which would result in the last option to overwrite all earlier options. 

The documentation specifies that it's possible to use both files and readers.

This PR aligns the implementation with the documentation by appending to the `Paths` field instead of overwriting it.

## Why is it important?

Align the implementation with what one would expect from the documentation

## How to test this PR

I've included a test that includes both a file and a reader

### DIFF
--- a/modules/compose/compose_api.go
+++ b/modules/compose/compose_api.go
@@ -22,8 +22,8 @@ import (
 	"github.com/docker/docker/client"
 	"golang.org/x/sync/errgroup"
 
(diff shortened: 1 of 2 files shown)


mdelapenya (commented):
LGTM, thanks for the quick fix!

PR #2601, opened 2024-06-22
chore: improve log handling when container is stopping

ash2k:
## What does this PR do?

I'm getting an error when my container cleanly stops:

> container log error: EOF. Stopping log consumer: Headers out of sync

Please see individual commits. I tried to tidy up a bit, I hope this is fine.

## Why is it important?

The error is a red herring and distracts users of the library.

## Related issues

-

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -757,17 +757,18 @@ func (c *DockerContainer) startLogProduction(ctx context.Context, opts ...LogPro
 			h := make([]byte, 8)
 			_, err := io.ReadFull(r, h)
 			if err != nil {
-				if errors.Is(err, net.ErrClosed) {
+				switch {
+				case errors.Is(err, net.ErrClosed):
 					now := time.Now()
 					since = fmt.Sprintf("%d.%09d", now.Unix(), int64(now.Nanosecond()))
 					goto BEGIN
-				}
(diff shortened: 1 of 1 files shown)


docker.go
```diff
@@ -752,57 +753,59 @@ func (c *DockerContainer) startLogProduction(ctx context.Context, opts ...LogPro
...
-						continue
-					}
+			}
+			h := make([]byte, 8)
+			_, err := io.ReadFull(r, h)
+			if err != nil {
+				switch {
+				case err == io.EOF:
+					// No more logs coming
+				case errors.Is(err, net.ErrClosed):
```
mdelapenya, line 763: Neat, thanks!

mdelapenya:
@ash2k do you think it's possible to write a test replicating the scenario where the container is stopped and the error is not present anymore?

ash2k:
I'm not sure how to test this. Start a container, emit logs, check that there is no error written to stderr? Seems too complicated just to test if a log line is written. Are there any similar tests?

ash2k:
@mdelapenya Friendly ping. I'd like to get this merged as I'd like to send a few more PRs and ideally don't want to have more than one in flight.

mdelapenya (approved):
LGTM, thanks!

PR #2608, opened 2024-06-27
chore: test cleanups

ash2k:
## What does this PR do?

I was debugging [failures in CI](https://github.com/testcontainers/testcontainers-go/actions/runs/9675487368/job/26693089053#step:10:2551) and found a few issues in code. See individual commits. I think the failure was not a real log reading failure - the container just failed to start for an **unexpected** reason but the test didn't catch it because it didn't check the error message.

Then, panic is a red herring too because the returned container is `nil`.

## Why is it important?

We want clear test signals that help pinpoint the bugs, we don't want hard to understand failures.

## Related issues

-


mdelapenya (approved):
LGTM, thanks!

mdelapenya:
The tests failing for the rootless mode are not related. Merging

PR #2610, opened 2024-06-27
feat: prepare modules for the new API

mdelapenya:
- **feat: define new API for modules using Run**
- **chore: update module generator**
- **docs: update module generation**
- **fixup: modulegen**
- **docs: update module pages**

## What does this PR do?
This PR updates all the modules, the module generator and the current docs to use the new API:

- deprecates `RunContainer` in favor of `Run`
- the `Run` function needs the Docker image as second argument, which forces users to know which image is used in their tests.
- updates the Go templates to generate new modules following this new API

## Why is it important?
Prepare for a more consistent API that makes evident which image is used.

## Follow-ups
Modules catalog examples will need to be udpated in consequence.


docs/modules/cassandra.md
```diff
@@ -22,13 +22,14 @@ go get github.com/testcontainers/testcontainers-go/modules/cassandra
 
 ## Module reference
 
-The Cassandra module exposes one entrypoint function to create the Cassandra container, and this function receives two parameters:
+The Cassandra module exposes one entrypoint function to create the Cassandra container, and this function receives three parameters:
 
 ```golang
-func RunContainer(ctx context.Context, opts ...testcontainers.ContainerCustomizer) (*CassandraContainer, error)
+func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustomizer) (*CassandraContainer, error)
```
thaJeztah, line 28: Silly thought / while thought / thing out loud; as this is changing the interface;

`docker run` is a combination of steps;

1. create container
    - if image is missing, pull the image (depending on pull settings)
    - once image is pulled;
    - create container again
2. start container
3. attach to the container (if in foreground)

Specifically, the "image is missing" (and what to do in that case) is something that could be relevant for testing;

- Do you want the image to be pulled?
- Perhaps you want the image to be loaded (offline tar of images)
- Perhaps even need the image to be built

Wondering if the `img string` should be / could be a functional argument that allows controlling this (in future);

- `PullImageIfMissing("foo:latest")`
- `LoadImageIfMissing("foo:latest", "/path/to/tar")`
- ...

We probably don't need all of those implemented, but it _would_ allow for those to be expanded on 🤔

mdelapenya: Before this change, we have (deprecated in this PR) a `testcontainers.WithImage` functional opt, that could leverage what you propose. We already have support for loading images from tar files, and the code will automatically pull the image if missing. So I'd see the second use case as a nice addition.

kiview: I am not sure if we have to be that considerate for this change at hand (where we want to be pragmatic and move forward to allow support for Official Modules in tc-go soon), but it is for sure something to consider more for the v1 release of tc-go.

tc-java essentially provides support for such an abstraction, in practice mostly realized through a `Future<String>` parameter. However, the resulting API needs to be on a higher abstraction level, so that it encapsulates pulling, building or loading images.

mdelapenya: > we want to be pragmatic and move forward

mdelapenya: I'm merging this, taking your ideas as follow-ups. Thanks all for the discussion here

Glup3:
will the package also be updated soon? related to #2625

PR #2611, opened 2024-06-27
chore: run rootless mode in nighlty builds

mdelapenya:
## What does this PR do?
This PR updates the nightly build that uses the TEST channel of the Docker builds, so that we are able to run the nightly build for rootful and rootless mode.

## Why is it important?
We want to detect regressions the soonest.

## Related issues

- Relates https://github.com/moby/moby/issues/48064#issuecomment-2194901075

### DIFF
--- a/.github/workflows/docker-moby-latest.yml
+++ b/.github/workflows/docker-moby-latest.yml
@@ -7,9 +7,20 @@ on:
 
 jobs:
   test_latest_moby:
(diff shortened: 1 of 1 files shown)


mdelapenya:
Merging, as there is no code related to the tests. The current failures in rootless mode are precisely the ones we want to proactively discover in this nightly build.

PR #2613, opened 2024-06-29
Log retried errors

ash2k:
## What does this PR do?

Add logging of errors that are retried.

## Why is it important?

I want to see any unexpected errors when they happen. Otherwise nothing is logged and the test appears to be stuck forever.

## Related issues

-

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -889,20 +889,25 @@ func (p *DockerProvider) BuildImage(ctx context.Context, img ImageBuildInfo) (st
 
 	var buildError error
 	var resp types.ImageBuildResponse
-	err = backoff.Retry(func() error {
-		resp, err = p.client.ImageBuild(ctx, buildOptions.Context, buildOptions)
-		if err != nil {
-			buildError = errors.Join(buildError, err)
-			if isPermanentClientError(err) {
-				return backoff.Permanent(err)
+	err = backoff.RetryNotify(
+		func() error {
+			resp, err = p.client.ImageBuild(ctx, buildOptions.Context, buildOptions)
+			if err != nil {
+				buildError = errors.Join(buildError, err)
+				if isPermanentClientError(err) {
+					return backoff.Permanent(err)
+				}
+				return err
 			}
-			Logger.Printf("Failed to build image: %s, will retry", err)
-			return err
-		}
-		defer p.Close()
+			defer p.Close()
 
-		return nil
-	}, backoff.WithContext(backoff.NewExponentialBackOff(), ctx))
+			return nil
+		},
+		backoff.WithContext(backoff.NewExponentialBackOff(), ctx),
+		func(err error, duration time.Duration) {
+			p.Logger.Printf("Failed to build image: %s, will retry", err)
+		},
+	)
 	if err != nil {
 		return "", errors.Join(buildError, err)
 	}
@@ -1015,7 +1020,7 @@ func (p *DockerProvider) CreateContainer(ctx context.Context, req ContainerReque
 			}
 
 			if modifiedTag != imageName {
-				Logger.Printf("✍🏼 Replacing image with %s. From: %s to %s\n", is.Description(), imageName, modifiedTag)
+				p.Logger.Printf("✍🏼 Replacing image with %s. From: %s to %s\n", is.Description(), imageName, modifiedTag)
 				imageName = modifiedTag
 			}
 		}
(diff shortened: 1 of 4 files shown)


docker.go
```diff
@@ -899,20 +899,25 @@ func (p *DockerProvider) BuildImage(ctx context.Context, img ImageBuildInfo) (st
...
+		func() error {
+			resp, err = p.client.ImageBuild(ctx, buildOptions.Context, buildOptions)
+			if err != nil {
+				buildError = errors.Join(buildError, err)
+				if isPermanentClientError(err) {
+					return backoff.Permanent(err)
+				}
+				return err
 			}
-			Logger.Printf("Failed to build image: %s, will retry", err)
```
ash2k, line 909: Note that if the error is logged here, the last error will be logged twice: once here and once by the caller when `backoff.RetryNotify()` returns it.

ash2k: Hence, we use the callback version that only calls the notify callback when the error will be retried.

docker.go
```diff
@@ -1192,22 +1197,31 @@ func (p *DockerProvider) findContainerByName(ctx context.Context, name string) (
...
 
-		ctr = c
-		return nil
-	}, backoff.WithContext(backoff.NewExponentialBackOff(), ctx))
+			ctr = c
+			return nil
+		},
+		backoff.WithContext(backoff.NewExponentialBackOff(), ctx),
+		func(err error, duration time.Duration) {
+			if errdefs.IsNotFound(err) {
```
ash2k, line 1219: Don't log "not found" error.

modules/clickhouse/clickhouse_test.go
```diff
@@ -278,75 +279,69 @@ func TestClickHouseWithZookeeper(t *testing.T) {
...
+				err := rows.Scan(&r.Id)
+				if err != nil {
+					return nil, err
+				}
 
-func performCRUD(conn driver.Conn) ([]Test, error) {
-	var (
-		err  error
-		rows []Test
+				res = append(res, r)
```
ash2k, line 314: Note that the previous version of this code is buggy - it reuses `res` between retries. This may lead to partial results from the failed attempt to be still in the slice.

ash2k:
Rebased

docker.go
mdelapenya: TIL

ash2k:
Tests are failing with:

```
  2024/07/02 23:11:31 Failed to pull image: Error response from daemon: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit, will retry
```

mdelapenya:
> Rebased

I'm sorry but I do not see the changes for the lifecycle.go function, adding a retry for checking all exposed ports are already available.

ash2k:
Hm, I've just checked and I think all commits are there:

- `main` branch of this repo: https://github.com/testcontainers/testcontainers-go/commits/main/
- PR branch in my fork: https://github.com/ash2k/testcontainers-go/commits/log-retriable-errors/

Am I missing something?

mdelapenya:
Oh sorry, I probably did not explain myself very well. I asked you to check (and update if neeeded) the retry added in #2606, to include the RetryNotify pattern this PR introduces 🙏

ash2k:
Ah, ok. Done!

lifecycle.go
```diff
@@ -215,33 +215,39 @@ var defaultReadinessHook = func() ContainerLifecycleHooks {
...
 						}
-					}
 
-					return nil
-				}, b)
+						return nil
+					},
+					b,
+					func(err error, duration time.Duration) {
+						dockerContainer.logger.Printf("Readiness hook: %v", err)
```
mdelapenya, line 248: Do we want to print a more consistent error here?
```suggestion
						dockerContainer.logger.Printf("All requested ports were not exposed: %v", err)
```

reaper.go
```diff
@@ -87,55 +87,53 @@ func lookUpReaperContainer(ctx context.Context, sessionID string) (*DockerContai
 	exp.MaxInterval = 5.0 * time.Second  // max interval between attempts
 	exp.MaxElapsedTime = 1 * time.Minute // max time to keep trying
 
-	var reaperContainer *DockerContainer
-	err = backoff.Retry(func() error {
-		args := []filters.KeyValuePair{
+	opts := container.ListOptions{
```
mdelapenya, line 90: Neat, better initialising it just outside the retry

reaper.go
```diff
@@ -87,55 +87,53 @@ func lookUpReaperContainer(ctx context.Context, sessionID string) (*DockerContai
...
-		r, err := containerFromDockerResponse(ctx, resp[0])
-		if err != nil {
-			return err
-		}
+			if len(resp) == 0 {
+				// reaper container not found in the running state: do not look for it again
+				return nil, nil
+			}
 
-		reaperContainer = r
```
mdelapenya, line 121: We probably missed updating the reaperContainer singleton instance

lifecycle.go
ash2k: There may be other errors too, from `inspectRawContainer(ctx)`. Not a big deal, can change, but that's why I kept it generic.

reaper.go
ash2k: This was setting the variable on line 90 that was then returned.

mdelapenya: Ah sorry I see it now that it's a RetryNotifyWithData 🤦 Sorry!

lifecycle.go
mdelapenya: Indeed, although the sole reason for calling the inspect would be checking the available exposed ports. Because the returned error would be added to the message, I think it's fine to combine both into the print.

If you mind, I'd prefer being explicit here

ash2k: No problem, updated!

mdelapenya (approved):
LGTM, thanks!

PR #2614, opened 2024-06-29
chore: remove most uses of  TestcontainersConfig and deprecated TestcontainersConfig fields.

thaJeztah:
### remove TestProviderHasConfig as it wasn't testing anything

This test was verifying that DockerProvider.Config() returned a
non-nil value, but Config() does not return a pointer, so it
would always pass (even with an empty struct).

### modules/compose: NewDockerComposeWith: don't use deprecated RyukDisabled

The TestcontainersConfig.RyukDisabled field is deprecated.
NewDockerProvider() uses ReadConfig() internally to load the config,
which always copies its value from config.Config.RyukDisabled field,
which is not deprecated.

This patch removes the extra redirection, and changes the code to
use Config.RyukDisabled instead.

### DockerProvider: don't use legacy TestcontainersConfig internally

All fields in TestcontainersConfig are deprecated, making it a
wrapper around config.Config and DockerProvider internally only
depends on config.Config. All deprecated fields contain a copy
of values available in config.Config.

However, the exported `Config()` has TestcontainersConfig as return type;
filling the deprecated fields is is currently handled by ReadConfig, but
straightforward, so we can inline that code in the `Config()` method.

This patch:

- changes the (non-exported) config field to hold a config.Config (which
  provides all the data used).
- removes internal uses of the `Config()` method; the `Config()` method
  did not provide value for internal use; the `config` field is not mutated
  (so no synchronisation needed for concurrency).
- updates the `Config()` method to handle conversion of the config.Config
  to a `TestcontainersConfig` (effectively inlining a copy of `ReadConfig`)

After this patch, the only remaining use of `ReadConfig` in testcontainers-go
itself is in tests (`TestReadConfig`); as a follow-up, we can consider
deprecating `ReadConfig` and `TestcontainersConfig`, although `DockerProvider`
may need a new `Config()` method with a different signature if its config
must be accessible for others, which would require the `ContainerProvider`
and `ReaperProvider` interfaces to be updated.

## Why is it important?

Remove uses of deprecated fields to make it easier to spot where it's still used.

## Follow-ups

consider deprecating `ReadConfig` and `TestcontainersConfig`, although `DockerProvider`
may need a new `Config()` method with a different signature if its config
must be accessible for others, which would require the `ContainerProvider`
and `ReaperProvider` interfaces to be updated.

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -866,7 +866,7 @@ type DockerProvider struct {
 	client    client.APIClient
 	host      string
 	hostCache string
-	config    TestcontainersConfig
+	config    config.Config
 }
 
 // Client gets the docker client used by the provider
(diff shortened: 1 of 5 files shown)


mdelapenya (approved):
LGTM, thanks!

PR #2626, opened 2024-07-05
Fix issues in BuildImage()

ash2k:
## What does this PR do?

Please see individual commits. The most important fixes are to not reuse the `Context` and to close it to release resources.

`BuildOptions()` calls `GetContext()` which calls `archive.TarWithOptions()` which [starts a goroutine](https://github.com/moby/moby/blob/v27.0.3/pkg/archive/archive.go#L850). This goroutine will be stuck until the reader is closed. So, current code leaks goroutines and open files when `BuildImage()` fails.

## Why is it important?

Improves reliability of image building.

## Related issues

-

### DIFF
--- a/container.go
+++ b/container.go
@@ -209,6 +209,7 @@ func (c *ContainerRequest) Validate() error {
 }
 
 // GetContext retrieve the build context for the request
+// Must be closed when no longer needed.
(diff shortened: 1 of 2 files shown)


mdelapenya (approved):
LGTM, thanks!

Issue #2640, opened 2024-07-12
[Bug]: Endpoint returns the wrong port randomly


stevenh:
### Testcontainers version

v0.32.0

### Using the latest Testcontainers version?

Yes

### Host OS

Linux

### Host arch

amd64

### Go version

1.22.5

### Docker version

```shell
Client:
 Version:           26.1.4
 API version:       1.45
 Go version:        go1.21.11
 Git commit:        5650f9b
 Built:             Wed Jun  5 11:27:57 2024
 OS/Arch:           linux/amd64
 Context:           default

Server: Docker Desktop
 Engine:
  Version:          26.1.4
  API version:      1.45 (minimum version 1.24)
  Go version:       go1.21.11
  Git commit:       de5c9cf
  Built:            Wed Jun  5 11:29:22 2024
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.33
  GitCommit:        d2d58213f83a351ca8f528a95fbd145f5654e957
 runc:
  Version:          1.1.12
  GitCommit:        v1.1.12-0-g51d5e94
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### Docker info

```shell
Client:
 Version:    26.1.4
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.14.1-desktop.1
    Path:     /usr/local/lib/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.27.1-desktop.1
    Path:     /usr/local/lib/docker/cli-plugins/docker-compose
  debug: Get a shell into any image or container (Docker Inc.)
    Version:  0.0.32
    Path:     /usr/local/lib/docker/cli-plugins/docker-debug
  dev: Docker Dev Environments (Docker Inc.)
    Version:  v0.1.2
    Path:     /usr/local/lib/docker/cli-plugins/docker-dev
  extension: Manages Docker extensions (Docker Inc.)
    Version:  v0.2.24
    Path:     /usr/local/lib/docker/cli-plugins/docker-extension
  feedback: Provide feedback, right in your terminal! (Docker Inc.)
    Version:  v1.0.5
    Path:     /usr/local/lib/docker/cli-plugins/docker-feedback
  init: Creates Docker-related starter files for your project (Docker Inc.)
    Version:  v1.2.0
[...]

PR #2641, opened 2024-07-12
fix: container.Endpoint and wait.FortHTTP to use lowest internal port

stevenh:
Ensure `Container.Endpoint` returns the lowest numbered aka first port when there is more than one exposed.

Ensure that `wait.ForHTTP` uses the lowest numbered aka first port when there is more than one exposed.

Prior to this a random port would be returned leading to unpredictable 
results.

Fixed: #2640

### DIFF
--- a/wait/http.go
+++ b/wait/http.go
@@ -76,6 +76,8 @@ func (ws *HTTPStrategy) WithStartupTimeout(timeout time.Duration) *HTTPStrategy 
 	return ws
 }
 
+// WithPort set the port to wait for.
+// Default is the lowest numbered port.
 func (ws *HTTPStrategy) WithPort(port nat.Port) *HTTPStrategy {
 	ws.Port = port
 	return ws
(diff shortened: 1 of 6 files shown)


stevenh:
Failures look like they might be down to needing ryuk updated with a version including this https://github.com/testcontainers/moby-ryuk/pull/121

mdelapenya:
Hi @stevenh do you think we should mark this as a breaking change, as we are changing the docs from `first` to `lowest`?

Other than that, LGTM

stevenh:
> Hi @stevenh do you think we should mark this as a breaking change, as we are changing the docs from `first` to `lowest`?
> 
> Other than that, LGTM

I don't think so as I believe that original intent of `first` was the `first port numerically` however this wasn't totally clear, hence the change in language disambiguate.

mdelapenya:
Wdyt if we change the title to `fix: use the lowest numbered (aka first) port when there is more than one exposed` so users read it more clear in the release notes?

stevenh:
> Wdyt if we change the title to `fix: use the lowest numbered (aka first) port when there is more than one exposed` so users read it more clear in the release notes?

Updated the title to something similar, which I hope is a little bit more concise, let me know what you think?

wait/http.go
```diff
@@ -173,38 +175,41 @@ func (ws *HTTPStrategy) WaitUntilReady(ctx context.Context, target StrategyTarge
...
-			case <-ctx.Done():
-				return fmt.Errorf("%w: %w", ctx.Err(), err)
-			case <-time.After(ws.PollInterval):
-				if err := checkTarget(ctx, target); err != nil {
-					return err
-				}
+		select {
+		case <-ctx.Done():
+			return fmt.Errorf("%w: %w", ctx.Err(), err)
+		case <-time.After(ws.PollInterval):
```
mdelapenya, line 182: So here we just wait and continue if context is not done, right? Can you help me understand the differences with checking the target inside the pollInterval case? 🙏

stevenh: Just reduced indentation, making the happy path the natural flow down the file, which typically reduces cognitive load, similar to how return errors early does.

The for loop was removed as the loop conditions were impossible, if an error was hit we return and ports is always set in the inspect output even if empty it seems.
```json
{
        "NetworkSettings": {
            "Bridge": "",
            "SandboxID": "f196e8cebe297cab0a9a99130f218aec2d93bdc43d15cd9e3f53180f7c71feb9",
            "SandboxKey": "/var/run/docker/netns/f196e8cebe29",
            "Ports": {},
        }
}
```

stevenh: Added a comment for clarity.

stevenh: If the intent was to loop until ready, then loops needs to be restored and checked for `len(ports) == 0` and the comment corrected.

mdelapenya: Thanks for the explanations!!

mdelapenya (approved):
LGTM, thanks!

PR #2642, opened 2024-07-12
chore: replace stretchr/testify with gotest.tools

thaJeztah:
### chore(deps): golang.org/x/crypto v0.24.0

align the dependency with the version used in various modules

full diff: https://github.com/golang/crypto/compare/v0.22.0...v0.24.0

### chore: prune go.mod indirect dependencies

Prune some indirect dependencies that are unused.

### chore: replace stretchr/testify with gotest.tools

gotest.tools has an ever-so-slightly smaller dependency graph; this
patch migrates most uses of testify with gotest.tools, with the
exception of the cockroachdb module, which uses stretchr's "suite",
for which there's no direct replacement.

## Related issues

-


stevenh:
I would suggest that this makes this project harder to maintain, while it might have a slightly smaller dependency graph there are far less people are familiar with using it, just checking import under 900 for gotest.tools vs over 12k for testify.

It's may also be less maintained with the last commit being 9 months ago vs last month for testify.

If anything I would recommend replacing uses of [testify/assert](https://pkg.go.dev/github.com/stretchr/testify/assert) with [testify/require](https://pkg.go.dev/github.com/stretchr/testify/require) as it prevents knock on failures which make it harder to understanding when debugging failures.

mdelapenya:
@thaJeztah I'm going to close this as we are standing with testify at the moment.

In any case, I'm open to start a GH discussion about the libraries that are preferred by our community, so we all can have a voice. Hope you understand 🙏 

Thanks anyway for your dedication to testcontainers-go 🙇

PR #2637, opened 2024-07-11
bug:Merge custom labels with default

bearrito:
## What does this PR do?

Fixes custom label merging with the default labels.

## Why is it important?

If users set custom labels via container customizes, they will expect those labels are present in the container inspection

## Related issues

- Closes #2632 

## How to test this PR

Unit tests have been added.

PR #2644, opened 2024-07-14
fix: retry on label error

stevenh:
Add system errors to list of permanent errors which fixes infinite retries when a build is attempted with an invalid label.

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -1653,6 +1653,7 @@ var permanentClientErrors = []func(error) bool{
 	errdefs.IsUnauthorized,
 	errdefs.IsForbidden,
 	errdefs.IsNotImplemented,
+	errdefs.IsSystem,
 }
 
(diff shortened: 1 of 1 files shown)


mdelapenya:
I guess you mean an invalid label when building the image, right? If so, what would be the impact of merging this and https://github.com/testcontainers/testcontainers-go/pull/2637?

stevenh:
> I guess you mean an invalid label when building the image, right? If so, what would be the impact of merging this and #2637?

Yes I mistakenly had a colon in value of the `Tag` field as I expected it to take the same format as `docker -t` so had used `Tag: "repo:version"` instead of `Tag = "version", Repo: "repo"`.

This was a persistent error, and sent the retry into an infinite loop.

#2637 is separate, no conflict there that I can see.

mdelapenya (approved):
LGTM, thanks!

PR #2646, opened 2024-07-14
perf: optimise docker authentication config lookup

stevenh:
Improve the performance of retrieving docker authentication configuration by performing requests in parallel and only performing lookups once.

For a simple double layer image with standard Google Cloud Registry configuration this reduced the time spent doing authentication config lookups from ~4.6s to ~400ms.

### DIFF
--- a/docker_auth.go
+++ b/docker_auth.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"net/url"
 	"os"
+	"sync"
 
 	"github.com/cpuguy83/dockercfg"
 	"github.com/docker/docker/api/types/registry"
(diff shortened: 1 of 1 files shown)


mdelapenya:
Thanks for adding this contribution! Could you share how you measured the lookups? I'm interested in doing some internal tests before merging this. (at first sight, it LGTM)

stevenh:
> Thanks for adding this contribution! Could you share how you measured the lookups? I'm interested in doing some internal tests before merging this. (at first sight, it LGTM)

To get accurate results I temporarily instrumented `getAuthConfigsFromDockerfile` but you can get a good idea from running a any test which builds a multi layer image from a Dockerfile.

```go
func getAuthConfigsFromDockerfile(c *ContainerRequest) map[string]registry.AuthConfig {
	start := time.Now()
	defer func() { fmt.Printf("getAuthConfigsFromDockerfile took: %s\n", time.Since(start)) }()
```

I was seeing 5 seconds to build an image which was fully cached, so I got curious and found that 90% of that time was spent in  `getAuthConfigsFromDockerfile` with each call to `GetRegistryCredentials` taking 70-400ms.

In my case I had Google GCR configured so my `~/.docker/config.json` contains:
```json
{
  "_credsStore": "desktop.exe",
  "auths": {
    "https://index.docker.io/v1/": {}
  },
  "credHelpers": {
    "asia.gcr.io": "gcloud",
    "eu.gcr.io": "gcloud",
    "gcr.io": "gcloud",
    "marketplace.gcr.io": "gcloud",
    "staging-k8s.gcr.io": "gcloud",
    "us-west1-docker.pkg.dev": "gcloud",
    "us.gcr.io": "gcloud"
  },
  "credsStore": "desktop.exe",
  "experimental": "enabled"
}
```

mdelapenya (approved):
Ok, thanks for the explanation, it was very illustrative and I could experiment myself the issue. Indeed I'm pretty sure there are some other places were parallelisation could definitively help in performance.

LGTM, thanks!

stevenh:
> Ok, thanks for the explanation, it was very illustrative and I could experiment myself the issue. Indeed I'm pretty sure there are some other places were parallelisation could definitively help in performance.
> 
> LGTM, thanks!

Cool, I'll keep an eye out 😃

PR #2647, opened 2024-07-15
feat: add custom container registry substitutor

stillya:
# Custom Registry Substitutor

## What does this PR do?
The new substitutor can be used by modules whose images are not placed on `DockerHub` by default. This way, you don't need to manually prefix your image with a custom container registry. Instead, you can configure your registry with a prefix at the configuration level using hub.image.name.prefix.

In a development environment, `hub.image.name.prefix` is commonly empty, so `CustomHubSubstitutor` will add the desired default prefix. However, in a CI environment, the default prefix will not be added, and the config prefix will be used (handled by `prependHubRegistry`).

## Usage
``` golang
func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustomizer) (*KeycloakContainer, error) {
		Image: img,
		Env: map[string]string{
			keycloakAdminUsernameEnv: defaultKeycloakAdminUsername,
			keycloakAdminPasswordEnv: defaultKeycloakAdminPassword,
		},
		ExposedPorts: []string{keycloakPort},
		Cmd:          []string{keycloakStartupCommand},
		ImageSubstitutors: []testcontainers.ImageSubstitutor{
			testcontainers.NewCustomHubSubstitutor("quay.io"),
		},
	}
}
```

### DIFF
--- a/image_substitutors_test.go
+++ b/image_substitutors_test.go
@@ -3,7 +3,51 @@ package testcontainers
 import (
 	"context"
 	"testing"
+
+	"github.com/testcontainers/testcontainers-go/internal/config"
 )
+
+func TestCustomHubSubstitutor(t *testing.T) {
+	t.Run("should substitute the image with the provided one", func(t *testing.T) {
+		s := NewCustomHubSubstitutor("quay.io")
+
+		img, err := s.Substitute("foo/foo:latest")
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		if img != "quay.io/foo/foo:latest" {
+			t.Errorf("expected quay.io/foo/foo:latest, got %s", img)
+		}
+	})
+	t.Run("should not substitute the image if it is already using the provided hub", func(t *testing.T) {
+		s := NewCustomHubSubstitutor("quay.io")
+
+		img, err := s.Substitute("quay.io/foo/foo:latest")
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		if img != "quay.io/foo/foo:latest" {
+			t.Errorf("expected quay.io/foo/foo:latest, got %s", img)
+		}
+	})
+	t.Run("should not substitute the image if hub image name prefix config exist", func(t *testing.T) {
+		t.Cleanup(config.Reset)
+		config.Reset()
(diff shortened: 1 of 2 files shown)


image_substitutors_test.go
```diff
@@ -3,8 +3,51 @@ package testcontainers
...
+			t.Fatal(err)
+		}
+
+		if img != "quay.io/keycloak/keycloak:24.0" {
+			t.Errorf("expected quay.io/keycloak/keycloak:24.0, got %s", img)
+		}
+	})
+	t.Run("should not substitute the image if hub image name prefix config exist", func(t *testing.T) {
+		config.Reset()
+		t.Setenv("TESTCONTAINERS_HUB_IMAGE_NAME_PREFIX", "registry.mycompany.com/mirror")
```
mdelapenya, line 37: Any test involving the config suffers from side effect (on my TODO to fix that), so I think we need to 1) put a config.Reset in `TestSubstituteBuiltImage` 2) call reset in a defer/t.Cleanup for the methods updating the config

stillya: Fixed. Check it again please.

mdelapenya:
@stillya thanks for submitting the PR. After discussing this on Slack, and seeing the code, I think the substitutor could make sense. I have one question regarding

> In a development environment, hub.image.name.prefix is commonly empty, so CustomHubSubstitutor will add the desired default prefix. However, in a CI environment, the default prefix will not be added, and the config prefix will be used (handled by prependHubRegistry).

Do you have issues updating the properties file on the CI, configuring the `well-known environment` in which the project must work? I think it would be possible to do:

```shell
echo "hub.image.name.prefix=quay.io" >> ~/.testcontainers.properties 
```

as part of your Github action job, jenkins job, etc.

I'd like to understand the reasons why it's not possible to do that. Of course, if it's in code, the settings are applied to everybody, that's why I consider this changeset correct. Just want verify you have a workaround in the meantime.

mdelapenya:
Responding to myself (thanks @kiview): setting the properties will apply to all containers, but the substitutor will apply to the ones of interest.

stillya:
Yep, that's the reason.

mdelapenya (approved):
LGTM, thanks!

PR #2653, opened 2024-07-17
feat: add build to life cycle hooks

stevenh:
Add pre and post build life cycle hooks so the user can see when an image build is occurring and what it's tag is.

Refactor combineContainerHooks to support additional fields automatically.

Reduce overhead of hook error handling by starting with an nil error slice and use append to allocate only if needed, in other words assume the happy path by default hence minimising allocations.

Minimise code used in hook handling by using helper methods.

Eliminate unused / shadowed parameters from defaultPreCreateHook.

GetRepo and GetTag now return a consistent UUID if the Repo or Tag field aren't set by setting their values. This was needed to enable the output of the image we're going to build in the log.


mdelapenya (approved):
LGTM, thanks!

Issue #2652, opened 2024-07-16
[Bug]: Explicitely mapped ports break default readiness hook


robinvanderstraeten-klarrio:
### Testcontainers version

0.32.0

### Using the latest Testcontainers version?

Yes

### Host OS

Linux (WSL)

### Host arch

x86_64

### Go version

1.22.3

### Docker version

```shell
Client: Docker Engine - Community
 Version:           26.0.1
 API version:       1.45
 Go version:        go1.21.9
 Git commit:        d260a54
 Built:             Thu Apr 11 10:53:21 2024
 OS/Arch:           linux/amd64
 Context:           default

Server: Docker Engine - Community
 Engine:
  Version:          26.0.1
  API version:      1.45 (minimum version 1.24)
  Go version:       go1.21.9
  Git commit:       60b9add
  Built:            Thu Apr 11 10:53:21 2024
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.31
  GitCommit:        e377cd56a71523140ca6ae87e30244719194a521
 runc:
  Version:          1.1.12
  GitCommit:        v1.1.12-0-g51d5e94
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### Docker info

```shell
Client: Docker Engine - Community
 Version:    26.0.1
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.13.1
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.26.1
    Path:     /usr/libexec/docker/cli-plugins/docker-compose
  scan: Docker Scan (Docker Inc.)
    Version:  v0.23.0
    Path:     /usr/libexec/docker/cli-plugins/docker-scan

Server:
 Containers: 7
  Running: 1
  Paused: 0
  Stopped: 6
 Images: 32
 Server Version: 26.0.1
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
[...]

PR #2658, opened 2024-07-17
fix: incorrect parsing of exposedPorts in readiness check

robinvanderstraeten-klarrio:
## What does this PR do?

If an explicit port mapping is given in the definition of a container (through `exposedPorts`), the default readiness check fails. See the related issue below for an example. This PR fixes the issue by using the `nat.ParsePortSpecs` function instead of the `nat.Port` function when parsing the `exposedPorts`.

## Why is it important?

Anyone who upgrades to v0.32.0 and has a port defined as for example `4443:4443/tcp` will run into the below issue.

## Related issues

- Closes https://github.com/testcontainers/testcontainers-go/issues/2652

## How to test this PR
I still need to add a unit test for this, any advice on how to approach this?

### DIFF
--- a/lifecycle.go
+++ b/lifecycle.go
@@ -201,20 +201,21 @@ var defaultLogConsumersHook = func(cfg *LogConsumerConfig) ContainerLifecycleHoo
 func checkPortsMapped(exposedAndMappedPorts nat.PortMap, exposedPorts []string) error {
 	portMap, _, err := nat.ParsePortSpecs(exposedPorts)
 	if err != nil {
-		return err
+		return fmt.Errorf("could not check mapped ports: %w", err)
 	}
 
-	for exposedPort, _ := range portMap {
+	for exposedPort := range portMap {
 		// having entries in exposedAndMappedPorts, where the key is the exposed port,
 		// and the value is the mapped port, means that the port has been already mapped.
 		if _, ok := exposedAndMappedPorts[exposedPort]; !ok {
 			// check if the port is mapped with the protocol (default is TCP)
-			if !strings.Contains(string(exposedPort), "/") {
-				exposedPort = nat.Port(fmt.Sprintf("%s/tcp", string(exposedPort)))
-				if _, ok := exposedAndMappedPorts[exposedPort]; !ok {
-					return fmt.Errorf("port %s is not mapped yet", exposedPort)
-				}
-			} else {
+			if strings.Contains(string(exposedPort), "/") {
+				return fmt.Errorf("port %s is not mapped yet", exposedPort)
+			}
+
+			// Port didn't have a type, default to tcp and retry.
+			exposedPort = nat.Port(fmt.Sprintf("%s/tcp", string(exposedPort)))
+			if _, ok := exposedAndMappedPorts[exposedPort]; !ok {
 				return fmt.Errorf("port %s is not mapped yet", exposedPort)
 			}
 		}
@@ -247,8 +248,7 @@ var defaultReadinessHook = func() ContainerLifecycleHooks {
 							return err
 						}
 
-						exposedAndMappedPorts := jsonRaw.NetworkSettings.Ports
-						return checkPortsMapped(exposedAndMappedPorts, dockerContainer.exposedPorts)
+						return checkPortsMapped(jsonRaw.NetworkSettings.Ports, dockerContainer.exposedPorts)
 					},
 					b,
 					func(err error, duration time.Duration) {
--- a/lifecycle.go
+++ b/lifecycle.go
@@ -198,6 +198,31 @@ var defaultLogConsumersHook = func(cfg *LogConsumerConfig) ContainerLifecycleHoo
 	}
 }
 
+func checkPortsMapped(exposedAndMappedPorts nat.PortMap, exposedPorts []string) error {
+	portMap, _, err := nat.ParsePortSpecs(exposedPorts)
+	if err != nil {
+		return err
+	}
+
+	for exposedPort, _ := range portMap {
+		// having entries in exposedAndMappedPorts, where the key is the exposed port,
+		// and the value is the mapped port, means that the port has been already mapped.
+		if _, ok := exposedAndMappedPorts[exposedPort]; !ok {
+			// check if the port is mapped with the protocol (default is TCP)
+			if !strings.Contains(string(exposedPort), "/") {
+				exposedPort = nat.Port(fmt.Sprintf("%s/tcp", string(exposedPort)))
+				if _, ok := exposedAndMappedPorts[exposedPort]; !ok {
+					return fmt.Errorf("port %s is not mapped yet", exposedPort)
+				}
+			} else {
+				return fmt.Errorf("port %s is not mapped yet", exposedPort)
+			}
+		}
+	}
+
+	return nil
+}
+
 // defaultReadinessHook is a hook that will wait for the container to be ready
 var defaultReadinessHook = func() ContainerLifecycleHooks {
 	return ContainerLifecycleHooks{
@@ -223,28 +248,7 @@ var defaultReadinessHook = func() ContainerLifecycleHooks {
 						}
 
 						exposedAndMappedPorts := jsonRaw.NetworkSettings.Ports
-						portMap, _, err := nat.ParsePortSpecs(dockerContainer.exposedPorts)
-						if err != nil {
-							return err
-						}
-
-						for exposedPort, _ := range portMap {
-							// having entries in exposedAndMappedPorts, where the key is the exposed port,
-							// and the value is the mapped port, means that the port has been already mapped.
-							if _, ok := exposedAndMappedPorts[exposedPort]; !ok {
-								// check if the port is mapped with the protocol (default is TCP)
-								if !strings.Contains(string(exposedPort), "/") {
-									exposedPort = nat.Port(fmt.Sprintf("%s/tcp", string(exposedPort)))
-									if _, ok := exposedAndMappedPorts[exposedPort]; !ok {
-										return fmt.Errorf("port %s is not mapped yet", exposedPort)
-									}
-								} else {
-									return fmt.Errorf("port %s is not mapped yet", exposedPort)
-								}
-							}
-						}
-
-						return nil
+						return checkPortsMapped(exposedAndMappedPorts, dockerContainer.exposedPorts)
 					},
 					b,
 					func(err error, duration time.Duration) {
--- a/lifecycle.go
+++ b/lifecycle.go
@@ -223,16 +223,19 @@ var defaultReadinessHook = func() ContainerLifecycleHooks {
 						}
 
 						exposedAndMappedPorts := jsonRaw.NetworkSettings.Ports
+						portMap, _, err := nat.ParsePortSpecs(dockerContainer.exposedPorts)
+						if err != nil {
+							return err
+						}
 
-						for _, exposedPort := range dockerContainer.exposedPorts {
-							portMap := nat.Port(exposedPort)
+						for exposedPort, _ := range portMap {
 							// having entries in exposedAndMappedPorts, where the key is the exposed port,
 							// and the value is the mapped port, means that the port has been already mapped.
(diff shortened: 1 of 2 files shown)


mdelapenya:
> I still need to add a unit test for this, any advice on how to approach this?

you could extract the logic to a function and unit test it creating different port instances, covering all possible cases 🤔

robinvanderstraeten-klarrio:
Hi @mdelapenya , I've added some tests. Any more work that needs to be done for this PR?

lifecycle.go
```diff
@@ -198,6 +198,31 @@ var defaultLogConsumersHook = func(cfg *LogConsumerConfig) ContainerLifecycleHoo
 	}
 }
 
+func checkPortsMapped(exposedAndMappedPorts nat.PortMap, exposedPorts []string) error {
+	portMap, _, err := nat.ParsePortSpecs(exposedPorts)
+	if err != nil {
+		return err
```
stevenh, line 204: suggestion: Could we wrap here so we know where it comes from

lifecycle.go
```diff
@@ -198,6 +198,31 @@ var defaultLogConsumersHook = func(cfg *LogConsumerConfig) ContainerLifecycleHoo
...
+		if _, ok := exposedAndMappedPorts[exposedPort]; !ok {
+			// check if the port is mapped with the protocol (default is TCP)
+			if !strings.Contains(string(exposedPort), "/") {
+				exposedPort = nat.Port(fmt.Sprintf("%s/tcp", string(exposedPort)))
+				if _, ok := exposedAndMappedPorts[exposedPort]; !ok {
+					return fmt.Errorf("port %s is not mapped yet", exposedPort)
+				}
+			} else {
+				return fmt.Errorf("port %s is not mapped yet", exposedPort)
+			}
```
stevenh, line 219: suggestion: Using early return reduces indentation and makes it easier to grok.
The creation of the port can also be simplified eliminating the `fmt.Sprintf`
```suggestion
			if strings.Contains(string(exposedPort), "/") {
				return fmt.Errorf("port %s is not mapped yet", exposedPort)
			}
			
			// Port didn't have a type, default to tcp and retry.
			exposedPort = nat.Port(exposedPort +"/tcp")
			if _, ok := exposedAndMappedPorts[exposedPort]; !ok {
				return fmt.Errorf("port %s is not mapped yet", exposedPort)
			}
```

lifecycle.go
```diff
@@ -198,6 +198,31 @@ var defaultLogConsumersHook = func(cfg *LogConsumerConfig) ContainerLifecycleHoo
 	}
 }
 
+func checkPortsMapped(exposedAndMappedPorts nat.PortMap, exposedPorts []string) error {
+	portMap, _, err := nat.ParsePortSpecs(exposedPorts)
+	if err != nil {
+		return err
+	}
+
+	for exposedPort, _ := range portMap {
```
stevenh, line 207: suggestion: No need for two val version
```suggestion
	for exposedPort := range portMap {
```

lifecycle.go
```diff
@@ -223,25 +248,7 @@ var defaultReadinessHook = func() ContainerLifecycleHooks {
...
-										return fmt.Errorf("port %s is not mapped yet", exposedPort)
-									}
-								} else {
-									return fmt.Errorf("port %s is not mapped yet", exposedPort)
-								}
-							}
-						}
-
-						return nil
+						return checkPortsMapped(exposedAndMappedPorts, dockerContainer.exposedPorts)
```
stevenh, line 251: nit: Is exposedAndMappedPorts needed now?

lifecycle_test.go
```diff
@@ -456,6 +456,83 @@ func TestMergePortBindings(t *testing.T) {
 	}
 }
 
+func TestPortMappingCheck(t *testing.T) {
+	makePortMap := func(ports []string) nat.PortMap {
```
stevenh, line 460: suggestion: changing to a variadic makes calling cleaner
```suggestion
	makePortMap := func(ports ...string) nat.PortMap {
```

lifecycle_test.go
```diff
@@ -456,6 +456,83 @@ func TestMergePortBindings(t *testing.T) {
...
+		out := make(nat.PortMap)
+		for _, port := range ports {
+			// We don't care about the actual binding in this test
+			out[nat.Port(port)] = make([]nat.PortBinding, 0)
```
stevenh, line 464: question: do we care about it being allocated or could it just be nil?

lifecycle_test.go
```diff
@@ -456,6 +456,83 @@ func TestMergePortBindings(t *testing.T) {
...
+
+	tests := []struct {
+		name                  string
+		exposedAndMappedPorts nat.PortMap
+		exposedPorts          []string
+		expectError           bool
+	}{
+		{
+			name:                  "No protocol given",
+			exposedAndMappedPorts: makePortMap([]string{"1024/tcp"}),
```
stevenh, line 477: suggestion: with the variadic this just becomes the following, more below:
```suggestion
			exposedAndMappedPorts: makePortMap("1024/tcp"),
```

lifecycle_test.go
```diff
@@ -456,6 +456,83 @@ func TestMergePortBindings(t *testing.T) {
...
+			exposedPorts:          []string{"1024"},
+			expectError:           false,
```
stevenh, line 479: nit: while explicit, it's idiomatic to not specify fields for zero values, which here reduces the definition and makes easier to read overall.
```suggestion
```

lifecycle_test.go
```diff
@@ -456,6 +456,83 @@ func TestMergePortBindings(t *testing.T) {
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := checkPortsMapped(tt.exposedAndMappedPorts, tt.exposedPorts)
+			if tt.expectError {
+				assert.Error(t, err)
+			} else {
+				assert.NoError(t, err)
+			}
```
stevenh, line 531: nit: I would use golden path flow as its easier to follow.

Either way I would recommend `require` instead of `assert`, no difference in this case but for more complex flows it eliminates knock on failures which just add noise, so good to standardise on.
```suggestion
			if tt.expectError {
				require.Error(t, err)
				return
			}
			
			require.NoError(t, err)
```

lifecycle_test.go
```diff
@@ -456,6 +456,83 @@ func TestMergePortBindings(t *testing.T) {
...
+	}
+
+	tests := []struct {
+		name                  string
+		exposedAndMappedPorts nat.PortMap
+		exposedPorts          []string
+		expectError           bool
+	}{
+		{
+			name:                  "No protocol given",
```
stevenh, line 476: nit: using test names with spaces makes it hard to find when doing a search due to space to underscore replacement.

I tend to use `-` instead and keep succinct e.g.
```suggestion
			name:                  "no-protocol",
```

stevenh, line 476: nit: using a map can make it easier to read and adds randomisation which ensures there's no inter test dependencies.
```suggestion
	tests := map[string]struct {
		exposedAndMappedPorts nat.PortMap
		exposedPorts          []string
		expectError           bool
	}{
		"no-protocol": {
```

stevenh (changes requested):

lifecycle.go
robinvanderstraeten-klarrio: https://github.com/testcontainers/testcontainers-go/pull/2658/commits/5fe2f77b1c6101b2c550f182ec4021d828ff5bd0

lifecycle.go
robinvanderstraeten-klarrio: https://github.com/testcontainers/testcontainers-go/pull/2658/commits/5fe2f77b1c6101b2c550f182ec4021d828ff5bd0

lifecycle.go
robinvanderstraeten-klarrio: https://github.com/testcontainers/testcontainers-go/pull/2658/commits/5fe2f77b1c6101b2c550f182ec4021d828ff5bd0

lifecycle.go
robinvanderstraeten-klarrio: https://github.com/testcontainers/testcontainers-go/pull/2658/commits/5fe2f77b1c6101b2c550f182ec4021d828ff5bd0

lifecycle_test.go
robinvanderstraeten-klarrio: https://github.com/testcontainers/testcontainers-go/pull/2658/commits/5fe2f77b1c6101b2c550f182ec4021d828ff5bd0

lifecycle_test.go
robinvanderstraeten-klarrio: It can indeed just be nil.
https://github.com/testcontainers/testcontainers-go/pull/2658/commits/5fe2f77b1c6101b2c550f182ec4021d828ff5bd0

lifecycle_test.go
robinvanderstraeten-klarrio: https://github.com/testcontainers/testcontainers-go/pull/2658/commits/5fe2f77b1c6101b2c550f182ec4021d828ff5bd0

lifecycle_test.go
robinvanderstraeten-klarrio: https://github.com/testcontainers/testcontainers-go/pull/2658/commits/5fe2f77b1c6101b2c550f182ec4021d828ff5bd0

lifecycle_test.go
robinvanderstraeten-klarrio: https://github.com/testcontainers/testcontainers-go/pull/2658/commits/5fe2f77b1c6101b2c550f182ec4021d828ff5bd0

lifecycle_test.go
robinvanderstraeten-klarrio: https://github.com/testcontainers/testcontainers-go/pull/2658/commits/5fe2f77b1c6101b2c550f182ec4021d828ff5bd0

robinvanderstraeten-klarrio: https://github.com/testcontainers/testcontainers-go/pull/2658/commits/5fe2f77b1c6101b2c550f182ec4021d828ff5bd0

robinvanderstraeten-klarrio:
Thanks for the suggestions @stevenh , applied them here: https://github.com/testcontainers/testcontainers-go/pull/2658/commits/5fe2f77b1c6101b2c550f182ec4021d828ff5bd0

stevenh (changes requested):
Thanks for applying the previous suggestions, few more came out of that, if you would be so kind :)

lifecycle.go
```diff
@@ -198,6 +198,32 @@ var defaultLogConsumersHook = func(cfg *LogConsumerConfig) ContainerLifecycleHoo
...
+			if strings.Contains(string(exposedPort), "/") {
+				return fmt.Errorf("port %s is not mapped yet", exposedPort)
+			}
+
+			// Port didn't have a type, default to tcp and retry.
+			exposedPort = nat.Port(fmt.Sprintf("%s/tcp", string(exposedPort)))
+			if _, ok := exposedAndMappedPorts[exposedPort]; !ok {
+				return fmt.Errorf("port %s is not mapped yet", exposedPort)
+			}
+		}
```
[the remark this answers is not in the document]
robinvanderstraeten-klarrio: https://github.com/testcontainers/testcontainers-go/pull/2658/commits/42344f6f5abb2fd6218156a402c97a2e2ed68bae

robinvanderstraeten-klarrio:
Sure @stevenh , applied them here: https://github.com/testcontainers/testcontainers-go/pull/2658/commits/42344f6f5abb2fd6218156a402c97a2e2ed68bae

stevenh (approved):
Looks good to me.

lifecycle_test.go
mdelapenya: I do not have a strong opinion about using white spaces, as it makes it easier to "write" test names. OTOH, if that makes "reading/searching" test names easier, I'm fine with it

mdelapenya (approved):
LGTM, thanks!

mdelapenya:
Thanks @stevenh for driving the review, much appreciated!

Issue #2636, opened 2024-07-10
[Bug]: Config is read too early to be modified in test code


yiftachkarkason:
### Testcontainers version

0.32.0

### Using the latest Testcontainers version?

Yes

### Host OS

MacOS

### Host arch

ARM

### Go version

1.22.5

### Docker version

```shell
Client: Docker Engine - Community
 Version:           27.0.3
 API version:       1.45 (downgraded from 1.46)
 Go version:        go1.22.4
 Git commit:        7d4bcd863a
 Built:             Fri Jun 28 14:56:30 2024
 OS/Arch:           darwin/arm64
 Context:           desktop-linux

Server: Docker Desktop 4.31.0 (153195)
 Engine:
  Version:          26.1.4
  API version:      1.45 (minimum version 1.24)
  Go version:       go1.21.11
  Git commit:       de5c9cf
  Built:            Wed Jun  5 11:29:12 2024
  OS/Arch:          linux/arm64
  Experimental:     false
 containerd:
  Version:          1.6.33
  GitCommit:        d2d58213f83a351ca8f528a95fbd145f5654e957
 runc:
  Version:          1.1.12
  GitCommit:        v1.1.12-0-g51d5e94
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### Docker info

```shell
Client: Docker Engine - Community
 Version:    27.0.3
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.14.1-desktop.1
    Path:     /Users/username/.docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.27.1-desktop.1
    Path:     /Users/username/.docker/cli-plugins/docker-compose
  debug: Get a shell into any image or container (Docker Inc.)
    Version:  0.0.32
    Path:     /Users/username/.docker/cli-plugins/docker-debug
  dev: Docker Dev Environments (Docker Inc.)
    Version:  v0.1.2
    Path:     /Users/username/.docker/cli-plugins/docker-dev
  extension: Manages Docker extensions (Docker Inc.)
    Version:  v0.2.24
    Path:     /Users/username/.docker/cli-plugins/docker-extension
  feedback: Provide feedback, right in your terminal! (Docker Inc.)
    Version:  v1.0.5
    Path:     /Users/username/.docker/cli-plugins/docker-feedback
[...]

PR #2661, opened 2024-07-19
fix: remove configuration reading from logger initialization

henrybarreto:
- fix: remove configuration reading from logger initialization

## What does this PR do?

It removes the reading for the environment variables from the logger initialization, and prints the message in a more appropriate location.

## Why is it important?

Due to the addition of a new init function for the logger initialization and Ryuk disable message, the environment variables were reading as an init function side effect. This caused variables set on the code using the package does not read property.

### DIFF
--- a/config.go
+++ b/config.go
@@ -18,6 +18,17 @@ type TestcontainersConfig struct {
 // of the TestcontainersConfig struct
 func ReadConfig() TestcontainersConfig {
 	cfg := config.Read()
+
+	if cfg.RyukDisabled {
+		ryukDisabledMessage := `
+**********************************************************************************************
+Ryuk has been disabled for the current execution. This can cause unexpected behavior in your environment.
+More on this: https://golang.testcontainers.org/features/garbage_collector/
+**********************************************************************************************`
+
+		Logger.Printf(ryukDisabledMessage)
+	}
+
 	return TestcontainersConfig{
 		Host:           cfg.Host,
 		TLSVerify:      cfg.TLSVerify,
--- a/logger.go
+++ b/logger.go
@@ -8,8 +8,6 @@ import (
 	"testing"
 
 	"github.com/docker/docker/client"
-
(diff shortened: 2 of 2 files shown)


config.go
```diff
@@ -18,6 +18,17 @@ type TestcontainersConfig struct {
 // of the TestcontainersConfig struct
 func ReadConfig() TestcontainersConfig {
 	cfg := config.Read()
+
+	if cfg.RyukDisabled {
```
mdelapenya, line 22: Will this be printed anytime that a user reads the config? As this method is public API, it would be a bit noisy seeing that message that much.

mdelapenya:
Relates #2636

config.go
henrybarreto: I, personally, don't see a problem, as this is a kind of warning message. What do you think could be a better approach?

sum-elier:
@mdelapenya we are affected by this bug and this PR would solve our issues. What's missing for you to approve and merge?

mdelapenya:
> @mdelapenya we are affected by this bug and this PR would solve our issues. What's missing for you to approve and merge?

I returned from summer holidays yesterday. I'm keeping up with the email, notifications and stuff. Will go back to it asap. OTOH I'd like to discuss more in https://github.com/testcontainers/testcontainers-go/pull/2661#discussion_r1684449902, as I'd prefer not printing out that message for each call to config.Read

stevenh:
Didn't spot this and created https://github.com/testcontainers/testcontainers-go/pull/2725 which just removes the message totally, as its opt in I would suggest the users knows what they are doing and doesn't want a warning about a conscious decision they have made.

mdelapenya:
I think we can close this one, as #2725 has been merged. @henrybarreto could you verify it with the current HEAD in main?

In the meantime I'm closing this one, so please reopen if needed. Thank you!

PR #2664, opened 2024-07-20
fix: lots of random failures

stevenh:
This is a combination PR of a large amount of fixes.

These are all in one branch for now so we can see if there are any more tests which don't succeed first time.

DO NOT MERGE!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,6 @@ TEST-*.xml
 tcvenv
 
 **/go.work
+
+# VS Code settings
+.vscode
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -8,6 +8,7 @@ linters:
     - nonamedreturns
     - testifylint
     - errcheck
+    - nolintlint
 
 linters-settings:
   errorlint:
--- a/cleanup.go
+++ b/cleanup.go
@@ -1,0 +1,107 @@
+package testcontainers
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"reflect"
+	"time"
+)
+
+// terminateOptions is a type that holds the options for terminating a container.
+type terminateOptions struct {
+	ctx     context.Context
+	timeout *time.Duration
+	volumes []string
+}
+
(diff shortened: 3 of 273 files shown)


stevenh:
> Hey @stevenh, what is the overall intent of this PR? 🙂

It's a big bug bash, with the goal of eliminating all the niggles which are causing random failures not only here but in peoples tests.

The reason these are getting through ATM is because the tests are rerunning a number of times, when they see a failure.

A lot of the issues are subtle race conditions and it would take months to commit these individually to get to a good place.

The intent is to get to a good place and then look at splitting it out.

I feel like its getting close the last piece is a combo issue which needs a new reaper image.

stevenh:
I should also mention that these issues are really hard to track down locally, the remote environment is much more reliable at triggering it seems because it's running a full suite across a lot of different setups.

kiview:
Alright, I just want to make sure that we are aligned on an approach to merge small and atomic changes, so your effort is not invested in vain. 

It might be good to sync with @mdelapenya, since we also have a bigger `v1` branch in the works with substantial changes, so I we might run into more merge problems down the road, if we don't time things appropriately.

stevenh:
Every test passing first time!!!

webstradev:
I can say, some of the fixes here will make the use of testcontainers a LOT better for us, we use it a lot in our CI and run into a lot of  failures a day causing pretty big delays for us.

mdelapenya:
WOW, thanks for this! I wonder if we could take small chunks of it in several PRs so that 1) the review process take less time, 2) the review is simpler and 3) the merge with v1 is simpler too.

Do you think we can coordinate and work in separating concerns? I'm more than happy to help in that.

mdelapenya:
@stevenh do you think we can close this one? We addressed it in small bits

stevenh:
It's just waiting on the reaper PR

mdelapenya:
Given the reaper PR was already merged, I think we can close this one.

Please reopen it if you consider there is more work to do here 🙏

PR #2666, opened 2024-07-22
feat(redpanda): Add option for arbitrary bootstrap config

oleiman:
## What does this PR do?

WithBootstrapConfig adds an arbitrary config kvp to the Redpanda container. Per the name, this config will be interpolated into the generated bootstrap config file, which is particularly useful for configs requiring a restart when otherwise applied to a running Redpanda instance.

## Why is it important?

Several Redpanda cluster config properties can be updated live but require a node restart to take effect. The simplest way to expose these to testcontainer consumers is to provide generic access prior to container startup.

An example usage case is an integration test for a particularly memory hungry Wasm Data Transform, where tuning the per core and per function memory limits in a flexible, implementation driven way may be the difference between testing or not testing that function.

## Related issues

-

### DIFF
--- a/modules/redpanda/mounts/bootstrap.yaml.tpl
+++ b/modules/redpanda/mounts/bootstrap.yaml.tpl
@@ -22,3 +22,8 @@ data_transforms_enabled: true
 {{- if .AutoCreateTopics }}
 auto_create_topics_enabled: true
 {{- end }}
+
+{{- range $key, $value := .ExtraBootstrapConfig }}
+{{ $key }}: {{ $value }}
+{{- end }}
+
--- a/modules/redpanda/options.go
+++ b/modules/redpanda/options.go
@@ -43,6 +43,10 @@ type options struct {
 	// Listeners is a list of custom listeners that can be provided to access the
 	// containers form within docker networks
 	Listeners []listener
+
+	// ExtraBootstrapConfig is a map of configs to be interpolated into the
+	// container's bootstrap.yml
+	ExtraBootstrapConfig map[string]any
 }
 
 func defaultOptions() options {
@@ -55,6 +59,7 @@ func defaultOptions() options {
 		AutoCreateTopics:                   false,
 		EnableTLS:                          false,
 		Listeners:                          make([]listener, 0),
+		ExtraBootstrapConfig:               make(map[string]any, 0),
 	}
 }
 
@@ -153,5 +158,15 @@ func WithListener(lis string) Option {
 			Port:                 portInt,
 			AuthenticationMethod: o.KafkaAuthenticationMethod,
 		})
+	}
+}
+
+// WithBootstrapConfig adds an arbitrary config kvp to the Redpanda container.
+// Per the name, this config will be interpolated into the generated bootstrap
+// config file, which is particularly useful for configs requiring a restart
+// when otherwise applied to a running Redpanda instance.
+func WithBootstrapConfig(cfg string, val any) Option {
+	return func(o *options) {
+		o.ExtraBootstrapConfig[cfg] = val
 	}
 }
--- a/modules/redpanda/redpanda_test.go
+++ b/modules/redpanda/redpanda_test.go
@@ -2,6 +2,7 @@ package redpanda_test
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
@@ -572,4 +573,59 @@ func TestRedpandaListener_NoNetwork(t *testing.T) {
 	require.Error(t, err)
 
 	require.Contains(t, err.Error(), "container must be attached to at least one network")
+}
+
+func TestRedpandaBootstrapConfig(t *testing.T) {
+	ctx := context.Background()
+
+	container, err := redpanda.RunContainer(ctx,
+		redpanda.WithEnableWasmTransform(),
+		// These configs would require a restart if applied to a live Redpanda instance
+		redpanda.WithBootstrapConfig("data_transforms_per_core_memory_reservation", 33554432),
+		redpanda.WithBootstrapConfig("data_transforms_per_function_memory_limit", 16777216),
+	)
+	require.NoError(t, err)
+
+	httpCl := &http.Client{
+		Timeout: 5 * time.Second,
+		Transport: &http.Transport{
+			ForceAttemptHTTP2:   true,
+			TLSHandshakeTimeout: 10 * time.Second,
+		},
+	}
+	adminAPIUrl, err := container.AdminAPIAddress(ctx)
+	require.NoError(t, err)
+
+	{
+		// Check that the configs reflect specified values
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, fmt.Sprintf("%s/v1/cluster_config", adminAPIUrl), nil)
+		require.NoError(t, err)
+		resp, err := httpCl.Do(req)
+		require.NoError(t, err)
+		defer resp.Body.Close()
+		require.Equal(t, http.StatusOK, resp.StatusCode)
+		var data map[string]any
+		err = json.NewDecoder(resp.Body).Decode(&data)
+		require.NoError(t, err)
+		reservation := int(data["data_transforms_per_core_memory_reservation"].(float64))
+		require.Equal(t, 33554432, reservation)
+		pf_limit := int(data["data_transforms_per_function_memory_limit"].(float64))
+		require.Equal(t, 16777216, pf_limit)
+	}
+
+	{
+		// Check that no restart is required. i.e. that the configs were applied via bootstrap config
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, fmt.Sprintf("%s/v1/cluster_config/status", adminAPIUrl), nil)
+		require.NoError(t, err)
+		resp, err := httpCl.Do(req)
+		require.NoError(t, err)
+		defer resp.Body.Close()
+		require.Equal(t, http.StatusOK, resp.StatusCode)
+		var data []map[string]any
+		err = json.NewDecoder(resp.Body).Decode(&data)
+		require.NoError(t, err)
+		require.Len(t, data, 1)
+		needs_restart := data[0]["restart"].(bool)
+		require.False(t, needs_restart)
+	}
 }
--- a/modules/redpanda/examples_test.go
+++ b/modules/redpanda/examples_test.go
@@ -17,6 +17,8 @@ func ExampleRun() {
 		redpanda.WithEnableSASL(),
 		redpanda.WithEnableKafkaAuthorization(),
 		redpanda.WithEnableWasmTransform(),
+		redpanda.WithBootstrapConfig("data_transforms_per_core_memory_reservation", 33554432),
+		redpanda.WithBootstrapConfig("data_transforms_per_function_memory_limit", 16777216),
 		redpanda.WithNewServiceAccount("superuser-1", "test"),
 		redpanda.WithNewServiceAccount("superuser-2", "test"),
 		redpanda.WithNewServiceAccount("no-superuser", "test"),
--- a/modules/redpanda/redpanda.go
+++ b/modules/redpanda/redpanda.go
@@ -281,6 +281,7 @@ func renderBootstrapConfig(settings options) ([]byte, error) {
 		KafkaAPIEnableAuthorization: settings.KafkaEnableAuthorization,
 		AutoCreateTopics:            settings.AutoCreateTopics,
 		EnableWasmTransform:         settings.EnableWasmTransform,
+		ExtraBootstrapConfig:        settings.ExtraBootstrapConfig,
 	}
 
 	tpl, err := template.New("bootstrap.yaml").Parse(bootstrapConfigTpl)
@@ -355,6 +356,7 @@ type redpandaBootstrapConfigTplParams struct {
 	KafkaAPIEnableAuthorization bool
 	AutoCreateTopics            bool
 	EnableWasmTransform         bool
+	ExtraBootstrapConfig        map[string]any
 }
 
 type redpandaConfigTplParams struct {


mdelapenya (commented):
Could you add tests for this new option? 

Once there, I think this is LGTM 

Cheers!

oleiman:
> Could you add tests for this new option? 
> 
> Once there, I think this is LGTM 
> 
> Cheers!

Absolutely. Thanks for the quick review @mdelapenya !

kiview (approved):

oleiman:
[force push](https://github.com/testcontainers/testcontainers-go/compare/c69723c2623e000eb23083414224c725f99dce45..86985f06fabbeeb4acee69865e2e3aa23cb16e2b) contents: a test

oleiman:
Hey @mdelapenya - does the test I added look ok to you?

rockwotj:
Ping @mdelapenya - I think the concerns here are addressed - thanks in advance!

mdelapenya:
I'm on PTO for this week, will take a look then 🙏

modules/redpanda/mounts/bootstrap.yaml.tpl
```diff
@@ -22,3 +22,10 @@ data_transforms_enabled: true
 {{- if .AutoCreateTopics }}
 auto_create_topics_enabled: true
 {{- end }}
+
+{{- if .ExtraBootstrapConfig }}
+{{- range $key, $value := .ExtraBootstrapConfig }}
+{{ $key }}: {{ $value }}
+{{- end }}
+{{- end }}
```
stevenh, line 30: suggestion: I don't believe you need the if here as zero iterations should be fine?
```suggestion
{{- range $key, $value := .ExtraBootstrapConfig }}
{{ $key }}: {{ $value }}
{{- end }}
```

modules/redpanda/options.go
```diff
@@ -155,3 +160,13 @@ func WithListener(lis string) Option {
 		})
 	}
 }
+
+// WithBootstrapConfig adds an arbitrary config kvp to the Redpanda container.
```
stevenh, line 164: question: This seems like it might get cumbersome to specify multiple values, is there a reason we don't just allow a string to include, which would support arbitrary levels of yaml too?

modules/redpanda/redpanda_test.go
```diff
@@ -573,3 +574,58 @@ func TestRedpandaListener_NoNetwork(t *testing.T) {
...
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, fmt.Sprintf("%s/v1/cluster_config", adminAPIUrl), nil)
+		require.NoError(t, err)
+		resp, err := httpCl.Do(req)
+		require.NoError(t, err)
+		assert.Equal(t, http.StatusOK, resp.StatusCode)
+		var data map[string]interface{}
+		err = json.NewDecoder(resp.Body).Decode(&data)
+		require.NoError(t, err)
+		reservation := int(data["data_transforms_per_core_memory_reservation"].(float64))
+		assert.Equal(t, reservation, 33554432)
```
stevenh, line 610: bug: As noted by the linter some of these parameters are reversed.

modules/redpanda/redpanda_test.go
```diff
@@ -573,3 +574,58 @@ func TestRedpandaListener_NoNetwork(t *testing.T) {
...
+		Timeout: 5 * time.Second,
+		Transport: &http.Transport{
+			ForceAttemptHTTP2:   true,
+			TLSHandshakeTimeout: 10 * time.Second,
+		},
+	}
+	adminAPIUrl, err := container.AdminAPIAddress(ctx)
+	require.NoError(t, err)
+
+	{
```
stevenh, line 599: question: Is there a need for extra indentation?

modules/redpanda/redpanda_test.go
```diff
@@ -573,3 +574,58 @@ func TestRedpandaListener_NoNetwork(t *testing.T) {
...
+		require.NoError(t, err)
+		assert.Equal(t, http.StatusOK, resp.StatusCode)
+		var data map[string]interface{}
+		err = json.NewDecoder(resp.Body).Decode(&data)
+		require.NoError(t, err)
+		reservation := int(data["data_transforms_per_core_memory_reservation"].(float64))
+		assert.Equal(t, reservation, 33554432)
+		pf_limit := int(data["data_transforms_per_function_memory_limit"].(float64))
+		assert.Equal(t, pf_limit, 16777216)
+		resp.Body.Close()
```
stevenh, line 613: bug: close needs to be directly after the error check for the request otherwise it may not be called.

modules/redpanda/redpanda_test.go
```diff
@@ -573,3 +574,58 @@ func TestRedpandaListener_NoNetwork(t *testing.T) {
...
+		resp, err := httpCl.Do(req)
+		require.NoError(t, err)
+		assert.Equal(t, http.StatusOK, resp.StatusCode)
+		var data []map[string]interface{}
+		err = json.NewDecoder(resp.Body).Decode(&data)
+		require.NoError(t, err)
+		assert.Equal(t, len(data), 1)
+		needs_restart := data[0]["restart"].(bool)
+		assert.False(t, needs_restart)
+		resp.Body.Close()
```
stevenh, line 629: bug: close needs to be directly after the error check for the request otherwise it may not be called.

modules/redpanda/redpanda_test.go
```diff
@@ -573,3 +574,58 @@ func TestRedpandaListener_NoNetwork(t *testing.T) {
...
+	adminAPIUrl, err := container.AdminAPIAddress(ctx)
+	require.NoError(t, err)
+
+	{
+		// Check that the configs reflect specified values
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, fmt.Sprintf("%s/v1/cluster_config", adminAPIUrl), nil)
+		require.NoError(t, err)
+		resp, err := httpCl.Do(req)
+		require.NoError(t, err)
+		assert.Equal(t, http.StatusOK, resp.StatusCode)
```
stevenh, line 605: suggestion: use `require` instead of assert, as if this fails everything else below will as well, but it will just be unhelpful.

modules/redpanda/redpanda_test.go
```diff
@@ -573,3 +574,58 @@ func TestRedpandaListener_NoNetwork(t *testing.T) {
...
+		resp.Body.Close()
+	}
+
+	{
+		// Check that no restart is required. i.e. that the configs were applied via bootstrap config
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, fmt.Sprintf("%s/v1/cluster_config/status", adminAPIUrl), nil)
+		require.NoError(t, err)
+		resp, err := httpCl.Do(req)
+		require.NoError(t, err)
+		assert.Equal(t, http.StatusOK, resp.StatusCode)
```
stevenh, line 622: suggestion: use `require` instead of assert, as if this fails everything else below will as well, but it will just be unhelpful more below.

stevenh (changes requested):
Quick pass to help keep this moving along, some issues and questions.

modules/redpanda/redpanda_test.go
oleiman: > as noted by the linter
ah, totally missed that. thanks!

oleiman:
[force push](https://github.com/testcontainers/testcontainers-go/compare/86985f06fabbeeb4acee69865e2e3aa23cb16e2b..cc28e932195b43113dabc2eef009f68f57171f1d) some CR feedback:
- linter errors
- template cruft
- better testify usage

modules/redpanda/redpanda_test.go
oleiman: not sure I understand the bug here, but I've moved `close` closer to the last use of `resp.Body`.

modules/redpanda/options.go
mdelapenya: I like having a map because readers will identify the key-value pairs very easily. Of course you can do it with a multiline string, but I see using a map more "typed". OTOH if you all agree the string is fine, I'm fine too

modules/redpanda/redpanda_test.go
stevenh: To clarify, we always want the body to be closed so if you want to leverage defer:
```go
res, err := http.Get(requestURL)
require.NoError(t, err)
defer resp.Body.Close()

// Decode body and check for errors.
```

modules/redpanda/options.go
stevenh: The reason I asked was yaml is typically structured, so not all values are at the base level like the example, [here's full config example](https://docs.redpanda.com/current/reference/console/config/) so if if none route elements might be needed this approach would be limited.

Someone who knows redpanda should comment if structured config for bootstrap could be needed?

rockwotj: Our cluster config and thus bootstrap yaml file is always a flat key value pair - nesting is not allowed

stevenh: Thanks for confirming

modules/redpanda/redpanda_test.go
```diff
@@ -573,3 +574,58 @@ func TestRedpandaListener_NoNetwork(t *testing.T) {
...
+		// Check that the configs reflect specified values
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, fmt.Sprintf("%s/v1/cluster_config", adminAPIUrl), nil)
+		require.NoError(t, err)
+		resp, err := httpCl.Do(req)
+		require.NoError(t, err)
+		require.Equal(t, http.StatusOK, resp.StatusCode)
+		var data map[string]interface{}
+		err = json.NewDecoder(resp.Body).Decode(&data)
+		require.NoError(t, err)
+		resp.Body.Close()
```
stevenh, line 609: To clarify my previous comment about the location of `Body.Close()`, here's an example which ensures that it's always closed.

While its not so important for tests, its a really good habit to get into as it ensure that underlying transport connections can be reused, hope that helps.
```suggestion
		require.NoError(t, err)
		defer resp.Body.Close()
		
		require.Equal(t, http.StatusOK, resp.StatusCode)
		var data map[string]interface{}
		err = json.NewDecoder(resp.Body).Decode(&data)
		require.NoError(t, err)
```

modules/redpanda/redpanda_test.go
```diff
@@ -573,3 +574,58 @@ func TestRedpandaListener_NoNetwork(t *testing.T) {
...
+		// Check that no restart is required. i.e. that the configs were applied via bootstrap config
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, fmt.Sprintf("%s/v1/cluster_config/status", adminAPIUrl), nil)
+		require.NoError(t, err)
+		resp, err := httpCl.Do(req)
+		require.NoError(t, err)
+		require.Equal(t, http.StatusOK, resp.StatusCode)
+		var data []map[string]interface{}
+		err = json.NewDecoder(resp.Body).Decode(&data)
+		require.NoError(t, err)
+		resp.Body.Close()
```
stevenh, line 626: See above for how to ensure that this is always closed.

modules/redpanda/redpanda_test.go
```diff
@@ -573,3 +574,58 @@ func TestRedpandaListener_NoNetwork(t *testing.T) {
...
+	require.NoError(t, err)
+
+	{
+		// Check that the configs reflect specified values
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, fmt.Sprintf("%s/v1/cluster_config", adminAPIUrl), nil)
+		require.NoError(t, err)
+		resp, err := httpCl.Do(req)
+		require.NoError(t, err)
+		require.Equal(t, http.StatusOK, resp.StatusCode)
+		var data map[string]interface{}
```
stevenh, line 606: nit: we can use any here.
```suggestion
		var data map[string]any
```

stevenh (changes requested):
Thanks for the previous changes, have clarified my previous commend about Body.Close and provided an example, so hope that helps.

modules/redpanda/redpanda_test.go
oleiman: thanks for clarifying

oleiman:
[force push](https://github.com/testcontainers/testcontainers-go/compare/cc28e932195b43113dabc2eef009f68f57171f1d..5899844fc0bb1015fb46621c53b4e4723f0a99b0) contents:
- correct response body closure in tests
- `s/interface{}/any/`

modules/redpanda/redpanda_test.go
oleiman: not an outright need, just limiting identifier scope for my own convenience.

oleiman:
Thanks for the reviews @stevenh - I think we've addressed all of your comments; please lmk if you have any lingering concerns 🙂

modules/redpanda/redpanda_test.go
```diff
@@ -573,3 +574,58 @@ func TestRedpandaListener_NoNetwork(t *testing.T) {
...
+		require.NoError(t, err)
+		resp, err := httpCl.Do(req)
+		require.NoError(t, err)
+		defer resp.Body.Close()
+		require.Equal(t, http.StatusOK, resp.StatusCode)
+		var data map[string]any
+		err = json.NewDecoder(resp.Body).Decode(&data)
+		require.NoError(t, err)
+		reservation := int(data["data_transforms_per_core_memory_reservation"].(float64))
+		assert.Equal(t, 33554432, reservation)
```
stevenh, line 611: Still a few asserts.

modules/redpanda/redpanda_test.go
```diff
@@ -573,3 +574,58 @@ func TestRedpandaListener_NoNetwork(t *testing.T) {
...
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, fmt.Sprintf("%s/v1/cluster_config/status", adminAPIUrl), nil)
+		require.NoError(t, err)
+		resp, err := httpCl.Do(req)
+		require.NoError(t, err)
+		defer resp.Body.Close()
+		require.Equal(t, http.StatusOK, resp.StatusCode)
+		var data []map[string]any
+		err = json.NewDecoder(resp.Body).Decode(&data)
+		require.NoError(t, err)
+		assert.Len(t, data, 1)
```
stevenh, line 627: suggestion: This is actually a good example of why require is preferable to assert.

If the length here was zero the assert would mark the test as failed, but would allow it continue, the following line would then panic as it indexed into an empty slice.

stevenh (changes requested):
Just a few asserts remaining if you would be so kind.

As you can see from the example, they aren't an issue until a test actually fails, then they can result it quite hard to understand output as you typically end up with compound errors.

modules/redpanda/redpanda_test.go
oleiman: good catch, yeah. need to have it beaten into my brain that `assert` is the non-halting version...

modules/redpanda/redpanda_test.go
oleiman: I don't see a reason to halt here since the error is non-fatal with respect to subsequent checks, but I'll happily to defer to idiomatic usage.

mdelapenya (approved):
LGTM, thanks all for your work here 🙇

stevenh (approved):

modules/redpanda/mounts/bootstrap.yaml.tpl
```diff
@@ -22,3 +22,8 @@ data_transforms_enabled: true
 {{- if .AutoCreateTopics }}
 auto_create_topics_enabled: true
 {{- end }}
+
+{{- range $key, $value := .ExtraBootstrapConfig }}
```
mdelapenya, line 26: Question: how will redpanda behave if users pass incorrect config key-value pairs here?

This is not related to the code itself but to the experience using the module

oleiman:
[rebase main](https://github.com/testcontainers/testcontainers-go/compare/0e1db02a08ebaedb7c180d0b031e4f39aadf383a..98389fe8efa5ba473a2bb194a6cf6cb68a975a57)

modules/redpanda/mounts/bootstrap.yaml.tpl
oleiman: Sorry, missed this. In general, redpanda will simply reject any unrecognized config (the key) and halt. I'm fairly certain the same applies to ill-formed values. For example, a value outside the range of a bounded config. Broker logs should reflect this in detail.

So the experience configuring the module this way should be quite similar to editing a Redpanda cluster config file before firing up Redpanda manually.

mdelapenya: Thanks for the explanation. Once the CI passes, I'll merge this PR

mdelapenya (approved):
LGTM, thanks!

PR #2679, opened 2024-08-04
fix(compose): fix compose destroys right after Up

kezhenxu94:
## What does this PR do?

The `NewDockerCompose.Up` sends `termSignal` as soon as the function finishes, this causes the compose project to be destroyed and future usages of it is impossible, this PR only sends `termSignal` when there is any error when invoking `Up` method, otherwise do not send terminal signal.

## Why is it important?

Without this patch, the docker compose API is useless, without this patch it just starts the docker compose project and then immediately destroys it, no container can be used after calling `Up` function.

## Related issues

- 

## How to test this PR

It can be easily reproduced by reverting the fix in the `compose_api.go` and run the `compose_api_test.go`, an error will be thrown:

```
 Container 415c1584-bbc5-4984-95ee-51aac9e7e8a4-api-nginx-1  Healthy
    compose_api_test.go:724: 
                Error Trace:    /Users/kezhenxu94/workspace/testcontainers-go/modules/compose/compose_api_test.go:724
                Error:          Received unexpected error:
                                Error response from daemon: No such container: 03b490452d68941c348d83cb4a8f2f26ad3068c3f45c46f77436cf2430006bd1
                Test:           TestDockerComposeUp
                Messages:       Ports
```

while the container must exist as it is what we aim to start up

### DIFF
--- a/modules/compose/compose_api.go
+++ b/modules/compose/compose_api.go
@@ -327,20 +327,26 @@ func (d *dockerCompose) Up(ctx context.Context, opts ...StackUpOption) error {
 		return err
 	}
 
+	var termSignal chan bool
 	if d.reaper != nil {
-		for _, n := range d.networks {
-			termSignal, err := d.reaper.Connect()
-			if err != nil {
-				return fmt.Errorf("failed to connect to reaper: %w", err)
-			}
-			n.SetTerminationSignal(termSignal)
+		termSignal, err = d.reaper.Connect()
 
-			// Cleanup on error, otherwise set termSignal to nil before successful return.
-			defer func() {
-				if termSignal != nil {
-					termSignal <- true
-				}
-			}()
+		// Cleanup on error
+		defer func() {
+			if err != nil {
+				termSignal <- true
+			}
+		}()
+
+		if err != nil {
+			err = fmt.Errorf("failed to connect to reaper: %w", err)
+			return err
+		}
+	}
+
+	if termSignal != nil {
+		for _, n := range d.networks {
+			n.SetTerminationSignal(termSignal)
 		}
 	}
 
@@ -350,24 +356,15 @@ func (d *dockerCompose) Up(ctx context.Context, opts ...StackUpOption) error {
 		// we are going to connect each container to the reaper
 		srv := srv
 		errGrpContainers.Go(func() error {
-			dc, err := d.lookupContainer(errGrpCtx, srv.Name)
+			var dc *testcontainers.DockerContainer
+
+			dc, err = d.lookupContainer(errGrpCtx, srv.Name)
 			if err != nil {
 				return err
 			}
 
-			if d.reaper != nil {
-				termSignal, err := d.reaper.Connect()
-				if err != nil {
-					return fmt.Errorf("failed to connect to reaper: %w", err)
-				}
+			if termSignal != nil {
 				dc.SetTerminationSignal(termSignal)
-
-				// Cleanup on error, otherwise set termSignal to nil before successful return.
-				defer func() {
-					if termSignal != nil {
-						termSignal <- true
-					}
-				}()
 			}
 
 			d.containersLock.Lock()
@@ -379,12 +376,12 @@ func (d *dockerCompose) Up(ctx context.Context, opts ...StackUpOption) error {
 	}
 
 	// wait here for the containers lookup to finish
-	if err := errGrpContainers.Wait(); err != nil {
+	if err = errGrpContainers.Wait(); err != nil {
 		return err
 	}
 
 	if len(d.waitStrategies) == 0 {
-		return nil
+		return err
 	}
 
 	errGrpWait, errGrpCtx := errgroup.WithContext(ctx)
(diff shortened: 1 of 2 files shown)


modules/compose/compose_api.go
```diff
@@ -267,12 +267,10 @@ func (d *dockerCompose) Down(ctx context.Context, opts ...StackDownOption) error
 	return d.composeService.Down(ctx, d.name, options.DownOptions)
 }
 
-func (d *dockerCompose) Up(ctx context.Context, opts ...StackUpOption) error {
+func (d *dockerCompose) Up(ctx context.Context, opts ...StackUpOption) (err error) {
```
mdelapenya, line 270: Is there a way to avoid naked returns?

kezhenxu94: > Is there a way to avoid naked returns?

If we really want to avoid that, we have to check `err != nil` or not in every single `return` statement, and if `err != nil` cleanup by sending a `termSignal`, this is more error prune in my opinion, what do you think?

kezhenxu94:
> I have one question regarding my example project for compose: https://github.com/mdelapenya/testcontainers-go-compose
> 
> Can you elaborate a repro snippet based on that project? 🙏

You can replace the entire file `main_test.go` with the following:


```go
package main

import (
	"context"
	"fmt"
	"log"
	"path/filepath"
	"time"

	tccompose "github.com/testcontainers/testcontainers-go/modules/compose"
)

func ExampleCompose() {
	compose, err := tccompose.NewDockerCompose(filepath.Join("testdata", "docker-compose.yml"))
	if err != nil {
		log.Fatal(err)
	}

	defer func() {
		if err := compose.Down(context.Background(),
			tccompose.RemoveOrphans(true), tccompose.RemoveImagesLocal); err != nil {
			log.Fatal(err)
		}
	}()

	ctx, cancel := context.WithCancel(context.Background())
	defer cancel()

	err = compose.Up(ctx, tccompose.Wait(true))
	if err != nil {
		log.Fatal(err)
	}

	serviceNames := compose.Services()
	fmt.Println(serviceNames)

	time.Sleep(20 * time.Second)
	mysqlContainer, err := compose.ServiceContainer(context.Background(), "mysql")
	fmt.Println(err)
	_, err = mysqlContainer.Inspect(ctx)
	fmt.Println(err) // I expect there is a container for mysql service, which is not the case
	// Output:
	// [mysql nginx]
	// <nil>
	// <nil>
}
```

and run `go test main_test.go`, it prints something like:

```
❯ go test main_test.go
time="2024-08-05T22:37:54+08:00" level=warning msg="/private/tmp/testcontainers-go-compose/testdata/docker-compose.yml: `version` is obsolete"
 Network 6a8a23a0-541c-4182-bb0e-56e95c57a62a_default  Creating
 Network 6a8a23a0-541c-4182-bb0e-56e95c57a62a_default  Created
 Container 6a8a23a0-541c-4182-bb0e-56e95c57a62a-nginx-1  Creating
 Container 6a8a23a0-541c-4182-bb0e-56e95c57a62a-mysql-1  Creating
 Container 6a8a23a0-541c-4182-bb0e-56e95c57a62a-nginx-1  Created
 Container 6a8a23a0-541c-4182-bb0e-56e95c57a62a-mysql-1  Created
 Container 6a8a23a0-541c-4182-bb0e-56e95c57a62a-nginx-1  Starting
 Container 6a8a23a0-541c-4182-bb0e-56e95c57a62a-mysql-1  Starting
 Container 6a8a23a0-541c-4182-bb0e-56e95c57a62a-mysql-1  Started
 Container 6a8a23a0-541c-4182-bb0e-56e95c57a62a-nginx-1  Started
 Container 6a8a23a0-541c-4182-bb0e-56e95c57a62a-mysql-1  Waiting
 Container 6a8a23a0-541c-4182-bb0e-56e95c57a62a-nginx-1  Waiting
 Container 6a8a23a0-541c-4182-bb0e-56e95c57a62a-nginx-1  Healthy
 Container 6a8a23a0-541c-4182-bb0e-56e95c57a62a-mysql-1  Healthy
--- FAIL: ExampleCompose (21.97s)
got:
[mysql nginx]
<nil>
Error response from daemon: No such container: e3a60e7c7a174de3d1dfbc947c91d677236647b1deed2ac81607e2861272eaba
want:
[mysql nginx]
<nil>
<nil>
FAIL
FAIL    command-line-arguments  22.675s
FAIL
```

modules/compose/compose_api.go
kezhenxu94: I refactor the code to avoid naked return, let me know how you like it 😄

stevenh:
There's actually a more in depth fix for this in https://github.com/testcontainers/testcontainers-go/pull/2664

stevenh:
I've extracted the compose fix from the bug bash and its here https://github.com/testcontainers/testcontainers-go/pull/2722 let me know if that fixes the issues.

kezhenxu94:
> I've extracted the compose fix from the bug bash and its here https://github.com/testcontainers/testcontainers-go/pull/2722 let me know if that fixes the issues.

You either pasted a wrong lint or it doesn't fix my issue as it's a chore issue "Remove unused parameters from private functions."

stevenh:
> > I've extracted the compose fix from the bug bash and its here #2722 let me know if that fixes the issues.
> 
> Are you sure a chore issue "Remove unused parameters from private functions." also fix a bug?

Sorry @kezhenxu94 right link but wrong cherry-pick commit id included, have fixed that now.

mdelapenya:
@kezhenxu94 could you please verify if the main branch, which contains #2722, works for you? 🙏 

If so, let's close this one, thanks!

mdelapenya:
Thanks @kezhenxu94 for taking the time to contribute to the project, and for verifying the fix. Very much appreciated 🙇

Issue #2206, opened 2024-02-12
[Enhancement]: Kafka Module: graceful shutdown


JordanP:
### Proposal

Currently, the Kafka module spawns a container that doesn't properly handle signals (term/int). The following (next) unit test reproduces the issue:
```go
func TestKafkaGracefulShutdown(t *testing.T) {
	ctx := context.Background()
	kafkaContainer, err := RunContainer(ctx, WithClusterID("kraftCluster"), testcontainers.WithImage("confluentinc/confluent-local:7.5.0"))
	if err != nil {
		t.Fatal(err)
	}

	done := make(chan struct{})
	go func() {
		// Wait 60 seconds, after that the pod will be forcefully terminated.
		timeout := 60 * time.Second
		kafkaContainer.Stop(context.Background(), &timeout)
		done <- struct{}{}
	}()
	select {
	case <-done:
	case <-time.After(50 * time.Second): // 50 secs should be plenty for Kafka to properly handle the signal
		t.Fatalf("Kafka did not gracefully exit in due time")
	}
}
```

I tried to replace `/etc/confluent/docker/launch` with `exec /etc/confluent/docker/launch` and `starterScript` with `exec starterScript` but that didn't do the trick.

JordanP:
Interestingly, using `confluentinc/cp-kafka:7.5.3` and `exec /etc/confluent/docker/launch` and `exec " + starterScript` work just fine.

mdelapenya:
Summoning @eddumelendez as our Kafka expert 🙏

eddumelendez:
It could be a difference between both images cp-kafka and confluent-local images. You can raise an issue [here](https://github.com/confluentinc/kafka-images). I also found this [old issue](https://github.com/confluentinc/kafka-images/issues/30) but it is just for cp-kafka

JordanP:
Do you remember/know why this Kafka module uses the `confluent-local` variant and not the (more widely used / less experimental ?) `cp-kafka` ?

eddumelendez:
I think the same implementation will work with `confluentinc/cp-kafka` but only in zookeeperless (raft) mode. IIRC, in go we decided to start with confluent-local as it is declared to be used for development workflows.

mabrarov:
Hi,

I believe at least part of this issue will be resolve with https://github.com/confluentinc/kafka-images/pull/401. I'll try the test from description of this issue with fixed confluentinc/confluent-local image and report here the result.

Thank you.

[...]

PR #2683, opened 2024-08-05
Use new apache/kafka-native docker image for kafka module

CTrando:
## What does this PR do?

Uses the new apache/kafka-native docker image instead of the confluent local docker image. The kafka native docker image is much faster (startup time is maybe a second once the image is downloaded). 

Trying to see if there is a desire from the community to use this. I can also add support for authentication mechanisms like SASL_PLAINTEXT if desired. 

## Related issues

-


modules/kafka/kafka.go
```diff
@@ -174,34 +152,4 @@ func configureControllerQuorumVoters(req *testcontainers.GenericContainerRequest
 
 		req.Env["KAFKA_CONTROLLER_QUORUM_VOTERS"] = fmt.Sprintf("1@%s:9094", host)
 	}
-	// }
-}
-
-// validateKRaftVersion validates if the image version is compatible with KRaft mode,
-// which is available since version 7.0.0.
-func validateKRaftVersion(fqName string) error {
```
CTrando, line 182: every apache/kafka-native image supports kraft

modules/kafka/kafka_test.go
```diff
@@ -87,41 +81,3 @@ func TestKafka(t *testing.T) {
...
-// - The BROKER:// protocol is using the hostname of the Kafka container
-func assertAdvertisedListeners(t *testing.T, container testcontainers.Container) {
-	inspect, err := container.Inspect(context.Background())
-	if err != nil {
-		t.Fatal(err)
-	}
-
-	hostname := inspect.Config.Hostname
-
-	code, r, err := container.Exec(context.Background(), []string{"cat", "/usr/sbin/testcontainers_start.sh"})
```
CTrando, line 110: this script doesn't exist anymore in the kafka image - also I didn't think it was necessary to check advertisted listeners, considering if it's set incorrectly, the other tests would fail because kafka wouldn't work in the first place

modules/kafka/kafka_test.go
```diff
@@ -17,7 +15,7 @@ func TestKafka(t *testing.T) {
 
 	ctx := context.Background()
 
-	kafkaContainer, err := kafka.Run(ctx, "confluentinc/confluent-local:7.5.0", kafka.WithClusterID("kraftCluster"))
+	kafkaContainer, err := kafka.Run(ctx, "apache/kafka-native:3.8.0", kafka.WithClusterID("kraftCluster"))
```
CTrando, line 18: TODO: need some more tests on the client side to demonstrate this works + add tests for setting host port

mdelapenya:
@CTrando thanks for opening this PR, and sorry for getting to it that late.

As I'm not a Kafka expert, I wonder if we want to provide a different flavour of kafka for the native image to avoid changing the existing behaviours.

@eddumelendez wdyt?

mabrarov:
Hi,

I believe this pull request can be modified to match [Kafka Module](https://java.testcontainers.org/modules/kafka/) of [Testcontainers for Java](https://java.testcontainers.org/) (which supports [apache/kafka-native](https://hub.docker.com/r/apache/kafka-native/) image):
1. https://github.com/testcontainers/testcontainers-java/blob/main/modules/kafka/src/main/java/org/testcontainers/kafka/KafkaContainer.java
1. https://github.com/testcontainers/testcontainers-java/blob/main/modules/kafka/src/main/java/org/testcontainers/kafka/KafkaHelper.java

Thank you.

mabrarov:
Hi @CTrando,

FYI, this can be helpful / connected:

1. https://github.com/testcontainers/testcontainers-go/issues/2206#issuecomment-2887726469 (why apache/kafka-native cannot be a choice for some teams).
1. https://github.com/mabrarov/testcontainers-go-kafka-2206/pull/1 (similar approach can be used in this PR).

Thank you.

mabrarov:
Hi,

FYI, support of both bitnami/kafka:4.0 and apache/kafka-native:latest images is implemented in https://github.com/mabrarov/testcontainers-go-kafka-2206/pull/2. Feel free to use that PR in this one if find it useful.

Thank you.

PR #2686, opened 2024-08-06
fix: port forwarding race condition

stevenh:
Fix port forwarding race condition by waiting for the exposed port which was causing random test failures.

Make errors and parameters more consumable while there.

### DIFF
--- a/port_forwarding.go
+++ b/port_forwarding.go
@@ -13,6 +13,7 @@ import (
 	"golang.org/x/crypto/ssh"
 
 	"github.com/testcontainers/testcontainers-go/internal/core/network"
+	"github.com/testcontainers/testcontainers-go/wait"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


mdelapenya (commented):
LGTM, although I'd like to understand the reasons to change the error messages. Do you see a non consistent pattern here?

stevenh:
> LGTM, although I'd like to understand the reasons to change the error messages. Do you see a non consistent pattern here?

Yep when you wrap having multiple `failed to` prefixes just adds noise, for example compare:
`failed to wait: failed to get the network: name conflict`

`wait: get network: name conflict`

With wrapping my rule is to describe the action which errored, which typically is a human readable version of the function call.

mdelapenya (approved):
LGTM, thanks!

PR #2687, opened 2024-08-06
test: add retry on system error test

stevenh:
Add a test to cover the retry on system error test.

Update test to use empty context to prevent knock on failures being logged confusing the output and check the error returned.


mdelapenya (approved):
LGTM, thanks!

PR #2689, opened 2024-08-06
test: fix image label test

stevenh:
Fix test for image label setting which could fail if the volume already existed.


mdelapenya (approved):
LGTM, thanks!

PR #2691, opened 2024-08-06
feat(wait): skip internal host port check

stevenh:
Add the ability to skip the internal host port check when performing as HostPortStrategy check. This is useful when a container does bind to the internal port until additional conditions are met.

### DIFF
--- a/wait/host_port.go
+++ b/wait/host_port.go
@@ -28,9 +28,15 @@ type HostPortStrategy struct {
 	// all WaitStrategies should have a startupTimeout to avoid waiting infinitely
 	timeout      *time.Duration
 	PollInterval time.Duration
+
+	// skipInternalCheck is a flag to skip the internal check, which is useful when
+	// a shell is not available in the container or when the container doesn't bind
+	// the port internally until additional conditions are met.
+	skipInternalCheck bool
 }
 
-// NewHostPortStrategy constructs a default host port strategy
+// NewHostPortStrategy constructs a default host port strategy that waits for the given
+// port to be exposed. The default startup timeout is 60 seconds.
 func NewHostPortStrategy(port nat.Port) *HostPortStrategy {
 	return &HostPortStrategy{
 		Port:         port,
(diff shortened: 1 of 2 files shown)


wait/host_port.go
```diff
@@ -28,28 +28,44 @@ type HostPortStrategy struct {
...
 func NewHostPortStrategy(port nat.Port) *HostPortStrategy {
 	return &HostPortStrategy{
 		Port:         port,
 		PollInterval: defaultPollInterval(),
 	}
 }
 
+// ForExposedPortOnly returns a host port strategy that waits for the given port
+// to be exposed only, it does not wait for the port to be bound inside the container.
+func ForExposedPortOnly(port nat.Port) *HostPortStrategy {
```
mdelapenya, line 49: I could see the value in adding a new constructor function, but I'd prefer consistency across the different wait strategies using the fluid builder pattern, something like this:

```go
  wait.ForListeningPort().SkipInternalCheck() // or any other relevant method name
```

I think this would be more consistent. Thoughts?

stevenh: Personally I find that needlessly verbose for a common option, so opted to follow the existing example set by `ForListeningPort` and `ForExposedPort` which admittedly only configure the port.

It could also be a bit confusing because the use case for this is when the port isn't actually listening.

What do you recon?

mdelapenya: We can probably find a better name:
- ForExternalPort, using the comments you added for the current `Only` function
- ForNotBoundPort
- ForPort
- ...

stevenh: Agreed the `Only` does seem unnecessary, how about just `ForExposedPort` as that matches the existing docker terminology so makes it clear what its looking to achieve?

mdelapenya: Mmm we already have the `ForExposedPort` that's why I'd add the fluid builder option to the strategy. In v1, we could go with functional options instead of the builder pattern for wait strategies.

stevenh: Yep forgot I added Only for that reason and naming things is hard :) 

Think I'm gonna do a 180, should we do the builder `SkipInternalCheck` and then review the approach for v1?

stevenh: Went for `WithoutInternalCheck` as the `SkipInternalCheck` conflicts with the field name.

wait/host_port.go
```diff
@@ -28,9 +28,15 @@ type HostPortStrategy struct {
 	// all WaitStrategies should have a startupTimeout to avoid waiting infinitely
 	timeout      *time.Duration
 	PollInterval time.Duration
+
+	// SkipInternalCheck is a flag to skip the internal check, which is useful when
+	// a shell is not available in the container or when the container doesn't bind
+	// the port internally until additional conditions are met.
+	SkipInternalCheck bool
```
mdelapenya, line 35: NIT: do we neet an exposed field here? Not a blocker though

mdelapenya (dismissed):
LGTM! will merge once the comment on private Vs public fields is addressed.

Thanks agains for your hard work

wait/host_port.go
stevenh: Nope was just following the example from Port and PollInterval, I guess that's something to review in v1 too

mdelapenya:
Given we changed the field to be private, we can declare the new function as `SkipInternalCheck`.  I'm doing it by myself including docs, so I will send a few commits on top of yours.

Thanks

mdelapenya:
@stevenh the CI passed, so if my changes are good to you, then we can merge, thanks!

stevenh (commented):
Looks good to me!

PR #2692, opened 2024-08-06
fix(redpanda): race condition on port check

stevenh:
Use SkipInternalCheck to wait for ports to be bound externally only  before trying to get the mapped port. This fixes a race condition in the setup of the container which was causing random test failures.

### DIFF
--- a/modules/redpanda/redpanda.go
+++ b/modules/redpanda/redpanda.go
@@ -86,6 +86,13 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
 				"--smp=1",
 				"--memory=1G",
 			},
+			WaitingFor: wait.ForAll(
+				// Wait for the ports to be exposed only as the container needs configuration
+				// before it will bind to the ports and be ready to serve requests.
+				wait.ForListeningPort(defaultKafkaAPIPort).SkipInternalCheck(),
+				wait.ForListeningPort(defaultAdminAPIPort).SkipInternalCheck(),
+				wait.ForListeningPort(defaultSchemaRegistryPort).SkipInternalCheck(),
+			),
 		},
 		Started: true,
 	}
@@ -119,7 +126,7 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
 
 	// 4. Register extra kafka listeners if provided, network aliases will be
 	// set
-	if err := registerListeners(ctx, settings, req); err != nil {
+	if err := registerListeners(settings, req); err != nil {
 		return nil, fmt.Errorf("failed to register listeners: %w", err)
 	}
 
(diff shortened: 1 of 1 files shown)


modules/redpanda/redpanda.go
```diff
@@ -86,6 +86,13 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
 				"--smp=1",
 				"--memory=1G",
 			},
+			WaitingFor: wait.ForAll(
+				// Wait for the ports to be exposed only as the container needs configuration
+				// before it will bind to the ports and be ready to serve requests.
+				wait.ForExposedPortOnly(defaultKafkaAPIPort),
+				wait.ForExposedPortOnly(defaultAdminAPIPort),
+				wait.ForExposedPortOnly(defaultSchemaRegistryPort),
```
mdelapenya, line 94: Failing because that function does not exist. See my comment here https://github.com/testcontainers/testcontainers-go/pull/2691#discussion_r1706718746

stevenh: Ah yes, depends on #2691

modules/redpanda/redpanda.go
```diff
@@ -200,11 +207,13 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
...
 
-	// 8. Wait until Redpanda is ready to serve requests
+	// 8. Wait until Redpanda is ready to serve requests.
 	err = wait.ForAll(
-		wait.ForListeningPort(defaultKafkaAPIPort),
-		wait.ForLog("Successfully started Redpanda!").WithPollInterval(100*time.Millisecond)).
-		WaitUntilReady(ctx, container)
+		wait.NewHostPortStrategy(defaultKafkaAPIPort),
+		wait.NewHostPortStrategy(defaultAdminAPIPort),
+		wait.NewHostPortStrategy(defaultSchemaRegistryPort),
```
mdelapenya, line 214: Better like this:
```suggestion
		wait.ForListeningPort(defaultKafkaAPIPort),
		wait.ForListeningPort(defaultAdminAPIPort),
		wait.ForListeningPort(defaultSchemaRegistryPort),
```

mdelapenya: I think we need to update this block too 🙏

mdelapenya (approved):
LGTM, thanks!

@weeco I'm going to merge this as is, to make progress, although I'd like to get your feedback if possible, as a redpanda contributor, in an eventual issue if you find any concern with this PR: in the end, it only adds stronger wait conditions to mark the container as ready, so I think it's safe to merge.

Thanks in advance

weeco:
That looks fine to me, thanks for the fix

PR #2693, opened 2024-08-06
fix(milvus): racy container setup

stevenh:
Fix race in container setup caused by copying the files into the container manually.

While here update tests to use require so they are easier to follow.


mdelapenya (approved):
LGTM, thanks!

Issue #2633, opened 2024-07-09
[Bug]: Version 0.32.0 seems to ignore WithStartupTimeout


KenjiTakahashi:
### Testcontainers version

0.32.0

### Using the latest Testcontainers version?

Yes

### Host OS

Linux

### Host arch

x86

### Go version

1.22

### Docker version

```shell
.
```

### Docker info

```shell
.
```

### What happened?

After updating to `0.32.0`, we started getting errors like this:
```
can not create container: failed to start container: all exposed ports, [127.0.0.1:35533:35533/tcp], were not mapped in 5s: port 127.0.0.1:35533:35533/tcp is not mapped yet
```

Our setup code is like this:
```go
req := testcontainers.ContainerRequest{
    Image:        "docker.elastic.co/elasticsearch/elasticsearch:8.14.2",
    ExposedPorts: []string{fmt.Sprintf("127.0.0.1:%[1]d:%[1]d/tcp", port)},
    WaitingFor: wait.ForHTTP("/_cluster/health").
        WithStartupTimeout(6 * time.Minute).
        WithPort(internalPort).
        WithResponseMatcher(func(body io.Reader) bool {
            response, err := io.ReadAll(body)
            if err != nil {
                return false
            }
            return strings.Contains(string(response), "green")
        }),
    AutoRemove: true,
}

container, err := testcontainers.GenericContainer(ctx, testcontainers.GenericContainerRequest{
    ContainerRequest: req,
    Started:          true,
})
if err != nil {
    return nil, nil, fmt.Errorf("can not create container: %w", err)
}
```

We've tried changing the value of `WithStartupTimeout`, but the error always says `5s`.

Version `0.31.0` still works as expected.

mdelapenya:
It's weird because the wait package has not been modified in this release, and looking at the commits I cannot see any one yet that could affect that part: https://github.com/testcontainers/testcontainers-go/compare/v0.31.0...v0.32.0

Can you verify if this issue is present with the elasticsearch module too?

I'm in Gophercon, but will double check this later

Thanks for the report!

andsleonardo:
I'm facing the same error after updating to v0.32.0.

[...]

PR #2694, opened 2024-08-06
fix(redis): port race

stevenh:
Fix race condition on port availability by switching to port instead of log check.

### DIFF
--- a/modules/redis/redis.go
+++ b/modules/redis/redis.go
@@ -54,7 +54,7 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
 	req := testcontainers.ContainerRequest{
 		Image:        img,
 		ExposedPorts: []string{"6379/tcp"},
-		WaitingFor:   wait.ForLog("* Ready to accept connections"),
+		WaitingFor:   wait.ForListeningPort("6379/tcp"),
 	}
 
 	genericContainerReq := testcontainers.GenericContainerRequest{


modules/redis/redis.go
```diff
@@ -54,7 +54,7 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
 	req := testcontainers.ContainerRequest{
 		Image:        img,
 		ExposedPorts: []string{"6379/tcp"},
-		WaitingFor:   wait.ForLog("* Ready to accept connections"),
+		WaitingFor:   wait.ForListeningPort("6379/tcp"),
```
mdelapenya, line 57: I haven't seen any issue with this wait strategy, which is widely use here and in the java module.

Could you bring more light about the motivation for the change?

stevenh: With all the iterations on the test suite I've had a few cases where test failed due to redis not accept incoming requests.

I didn't dig in fully as with this change, I never saw a reoccurrence.  It could be redis reports the log entry early?

mdelapenya: I'd keep this wait strategy as is, as it's been used for years. The PR is so simple that we can go back to it whenever it's needed. In https://github.com/testcontainers/testcontainers-go/issues/2633 we added a wait for all available ports, but it contains a bug, reported in https://github.com/testcontainers/testcontainers-go/issues/2633

I'd close this one because of that

stevenh: Sure will keep an eye for a failure.

stevenh:
Closing until we get another reproduction of this issue.

PR #2696, opened 2024-08-06
fix(kafka): port race on start

stevenh:
Fix the port race on start up by explicitly waiting for the port to be exposed. Once this is successful run the wait, this prevents the wait always being run due to way hooks are processed.


mdelapenya (approved):
LGTM, thanks!

bruteforce1414:
the same error with the postgres

bruteforce1414:
container.MappedPort don't work correctly

stevenh:
> container.MappedPort doesn't work correctly

Could you clarify why you think MappedPort doesn't work correctly?

stevenh:
> the same error with the postgres

Postgres doesn't have a post configuration step so not sure this can be the same issue.

Can you provide a reproduction example and the error you're seeing?

PR #2697, opened 2024-08-06
fix: logging restart

stevenh:
Fix being able to stop and restart a container by moving the stop processing from PreTerminates to PostStops hooks.

Fix log consumer being duplicated when restarting.

Fix race condition in log consumer test, cleaning up to flow to validate correctly.

### DIFF
--- a/lifecycle.go
+++ b/lifecycle.go
@@ -165,26 +165,25 @@ var defaultCopyFileToContainerHook = func(files []ContainerFile) ContainerLifecy
 var defaultLogConsumersHook = func(cfg *LogConsumerConfig) ContainerLifecycleHooks {
 	return ContainerLifecycleHooks{
 		PostStarts: []ContainerHook{
-			// first post-start hook is to produce logs and start log consumers
+			// Produce logs sending details to the log consumers.
+			// See combineContainerHooks for the order of execution.
 			func(ctx context.Context, c Container) error {
-				dockerContainer := c.(*DockerContainer)
-
-				if cfg == nil {
+				if cfg == nil || len(cfg.Consumers) == 0 {
 					return nil
 				}
 
+				dockerContainer := c.(*DockerContainer)
+				dockerContainer.consumers = dockerContainer.consumers[:0]
 				for _, consumer := range cfg.Consumers {
 					dockerContainer.followOutput(consumer)
 				}
 
-				if len(cfg.Consumers) > 0 {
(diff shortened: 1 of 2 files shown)


lifecycle.go
```diff
@@ -165,26 +165,25 @@ var defaultCopyFileToContainerHook = func(files []ContainerFile) ContainerLifecy
...
 			func(ctx context.Context, c Container) error {
-				dockerContainer := c.(*DockerContainer)
-
-				if cfg == nil {
+				if cfg == nil || len(cfg.Consumers) == 0 {
 					return nil
 				}
 
+				dockerContainer := c.(*DockerContainer)
+				dockerContainer.consumers = dockerContainer.consumers[:0]
```
mdelapenya, line 176: I'm sorry but I do not understand this line: we are updating the container's consumers with them? Is it a typo with the config's consumers?

stevenh: Consumers are only added via this postStart, this is needed to correctly handle start, stop, start cycle otherwise we would have stale consumers on second call to start.

logconsumer_test.go
```diff
@@ -641,21 +641,47 @@ func Test_MultiContainerLogConsumer_CancelledContext(t *testing.T) {
 
 type FooLogConsumer struct {
```
mdelapenya, line 642: Given we are moving the `Hello from Docker!` assertion to the FooLogConsumer methods, I think we need to document it to avoid future mis-usages of that test struct. Do you think we can rename it to a meaningful name, such as `HelloWorldLogConsumer`?

```suggestion
// FooLogConsumer test log consumer that accepts logs from the "hello-world" Docker image, which
// prints out the "Hello from Docker!" log message.
type FooLogConsumer struct {
```

lifecycle.go
```diff
@@ -165,26 +165,25 @@ var defaultCopyFileToContainerHook = func(files []ContainerFile) ContainerLifecy
...
-				if len(cfg.Consumers) > 0 {
-					return dockerContainer.startLogProduction(ctx, cfg.Opts...)
-				}
-				return nil
+				return dockerContainer.startLogProduction(ctx, cfg.Opts...)
 			},
 		},
-		PreTerminates: []ContainerHook{
-			// first pre-terminate hook is to stop the log production
+		PostStops: []ContainerHook{
```
mdelapenya, line 184: Neat! Better executed in this hook, thanks for the fix

PR #2698, opened 2024-08-06
fix(wait): data race in test

stevenh:
Fix the data race on access to health in TestWaitForHealthSucceeds.

### DIFF
--- a/wait/health_test.go
+++ b/wait/health_test.go
@@ -3,6 +3,7 @@ package wait
 import (
 	"context"
 	"io"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


wait/health_test.go
```diff
@@ -15,41 +16,54 @@ import (
...
+	// Return a copy to prevent data race.
+	state := *st.state
+
+	return &state, nil
+}
+
+func (st *healthStrategyTarget) setState(health *types.Health) {
+	st.mtx.Lock()
+	st.state.Health = health
+	st.mtx.Unlock()
```
mdelapenya, line 60: Safer?
```suggestion
	defer st.mtx.Unlock()
	st.state.Health = health
```

mdelapenya (commented):
Just added one comment. Once addressed, LGTM, thanks!

wait/health_test.go
stevenh: Yer, old habit from when defer had more overhead.

mdelapenya (approved):

PR #2699, opened 2024-08-06
test(mongodb): fix replica test failures

stevenh:
Fix mongodb replica test failures due to client using the internal container addresses when direct connection is not requested.

Fix invalid use of log.Fatal instead of tt.Fatal which was causing failures to not be output correctly due to os.Exit.


mdelapenya (approved):
LGTM, thanks!

PR #2705, opened 2024-08-08
fix(registry): compatibility with WSL

stevenh:
Use local IP instead of localhost to access registry container to address failures under WSL where the docker daemon can't access localhost.

Refactor tests to use helpers and require reducing code and simplifying the flow.

Add a default image which can be used be consumers.

Add HostAddress which returns host : port compatible with WSL.

Add SetDockerAuthConfig and DockerAuthConfig methods that can be used to easily configure authentication via DOCKER_AUTH_CONFIG environment variable.

Fix container clean up which was missed in a number of case.

### DIFF
--- a/modules/registry/examples_test.go
+++ b/modules/registry/examples_test.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"log"
-	"os"
 	"path/filepath"
 
 	"github.com/testcontainers/testcontainers-go"
@@ -39,9 +38,10 @@ func ExampleRun() {
 }
 
 func ExampleRun_withAuthentication() {
+	ctx := context.Background()
 	// htpasswdFile {
 	registryContainer, err := registry.Run(
-		context.Background(),
+		ctx,
 		"registry:2.8.3",
 		registry.WithHtpasswdFile(filepath.Join("testdata", "auth", "htpasswd")),
 		registry.WithData(filepath.Join("testdata", "data")),
@@ -51,33 +51,21 @@ func ExampleRun_withAuthentication() {
 		log.Fatalf("failed to start container: %s", err)
 	}
 	defer func() {
-		if err := registryContainer.Terminate(context.Background()); err != nil {
+		if err := registryContainer.Terminate(ctx); err != nil {
 			log.Fatalf("failed to terminate container: %s", err) // nolint:gocritic
 		}
 	}()
 
-	registryPort, err := registryContainer.MappedPort(context.Background(), "5000/tcp")
+	registryHost, err := registryContainer.HostAddress(ctx)
 	if err != nil {
-		log.Fatalf("failed to get mapped port: %s", err) // nolint:gocritic
+		log.Fatalf("failed to get host: %s", err) // nolint:gocritic
 	}
-	strPort := registryPort.Port()
 
-	previousAuthConfig := os.Getenv("DOCKER_AUTH_CONFIG")
-
-	// make sure the Docker Auth credentials are set
-	// using the same as in the Docker Registry
-	// testuser:testpassword
-	os.Setenv("DOCKER_AUTH_CONFIG", `{
-		"auths": {
-			"localhost:`+strPort+`": { "username": "testuser", "password": "testpassword", "auth": "dGVzdHVzZXI6dGVzdHBhc3N3b3Jk" }
-		},
-		"credsStore": "desktop"
-	}`)
-	defer func() {
-		// reset the original state after the example.
-		os.Unsetenv("DOCKER_AUTH_CONFIG")
-		os.Setenv("DOCKER_AUTH_CONFIG", previousAuthConfig)
-	}()
+	_, cleanup, err := registry.SetDockerAuthConfig(registryHost, "testuser", "testpassword")
+	if err != nil {
+		log.Fatalf("failed to set docker auth config: %s", err) // nolint:gocritic
+	}
+	defer cleanup()
 
 	// build a custom redis image from the private registry,
 	// using RegistryName of the container as the registry.
@@ -87,7 +75,7 @@ func ExampleRun_withAuthentication() {
 			FromDockerfile: testcontainers.FromDockerfile{
 				Context: filepath.Join("testdata", "redis"),
 				BuildArgs: map[string]*string{
-					"REGISTRY_PORT": &strPort,
+					"REGISTRY_HOST": &registryHost,
 				},
 				PrintBuildLog: true,
 			},
@@ -118,9 +106,10 @@ func ExampleRun_withAuthentication() {
 }
 
 func ExampleRun_pushImage() {
+	ctx := context.Background()
 	registryContainer, err := registry.Run(
-		context.Background(),
-		"registry:2.8.3",
+		ctx,
+		registry.DefaultImage,
 		registry.WithHtpasswdFile(filepath.Join("testdata", "auth", "htpasswd")),
 		registry.WithData(filepath.Join("testdata", "data")),
 	)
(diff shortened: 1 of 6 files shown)


modules/registry/registry.go
```diff
@@ -24,17 +35,56 @@ type RegistryContainer struct {
...
 	}
 
-	return fmt.Sprintf("http://%s:%s", ipAddress, port.Port()), nil
+	return net.JoinHostPort(host, port.Port()), nil
+}
+
+// localAddress returns the local address of the machine
+// which can be used to connect to the local registry.
+// This avoids the issues with localhost on WSL.
+func localAddress(ctx context.Context) (string, error) {
```
mdelapenya, line 72: Do you think that https://github.com/testcontainers/testcontainers-go/pull/2403 has something to do here? It will apply to all containers

modules/registry/examples_test.go
```diff
@@ -39,9 +38,10 @@ func ExampleRun() {
 }
 
 func ExampleRun_withAuthentication() {
+	ctx := context.Background()
 	// htpasswdFile {
```
mdelapenya, line 42: Let's include the creation of the ctx in the embedded docs:
```suggestion
	// htpasswdFile {
	ctx := context.Background()
```

modules/registry/registry.go
```diff
@@ -203,3 +253,67 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
...
+	previousAuthConfig := os.Getenv("DOCKER_AUTH_CONFIG")
+	os.Setenv("DOCKER_AUTH_CONFIG", auth)
+
+	cleanup = func() {
+		if previousAuthConfig == "" {
+			os.Unsetenv("DOCKER_AUTH_CONFIG")
+			return
+		}
+		os.Setenv("DOCKER_AUTH_CONFIG", previousAuthConfig)
+	}
```
mdelapenya, line 276: Neat!

modules/registry/registry.go
```diff
@@ -203,3 +253,67 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
 
 	return c, nil
 }
+
+// SetDockerAuthConfig sets the DOCKER_AUTH_CONFIG environment variable with
+// authentication for with the given host, username and password.
+// It returns the base64 encoded credentials and a function to reset back
+// to the previous state.
+func SetDockerAuthConfig(host, username, password string, additional ...string) (creds string, cleanup func(), err error) { //nolint:nonamedreturns // Adds context to the return values.
```
mdelapenya, line 261: Do we need the nolint for no-named-returns here? 
```suggestion
func SetDockerAuthConfig(host, username, password string, additional ...string) (creds string, cleanup func(), err error) { // Adds context to the return values.
```

modules/registry/registry.go
```diff
@@ -203,3 +253,67 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
...
+		os.Setenv("DOCKER_AUTH_CONFIG", previousAuthConfig)
+	}
+
+	return creds, cleanup, nil
+}
+
+// DockerAuthConfig returns the base64 encoded credentials and the auth config
+// for the provided details. It also accepts additional host, username and password
+// triples to add more auth configurations.
+func DockerAuthConfig(host, username, password string, additional ...string) (creds, authConfig string, err error) { //nolint:nonamedreturns // Adds context to the return values.
```
mdelapenya, line 284: Same here about the need of nonamed returns:
```suggestion
func DockerAuthConfig(host, username, password string, additional ...string) (creds, authConfig string, err error) { // Adds context to the return values.
```

mdelapenya, line 284: Just to understand this better, I guess this public function gets the auth config struct from the input data, right? It's useful for getting the "raw" config so that the tests can inject it in the test env (using `t.Setenv`). Also used by the module to set the env (using `os`).

Is this assumption correct?

mdelapenya (commented):
Added some comments, this is almost ready to be merged, although I'd like to discuss more with the team about the credentialStore being pinned to "desktop".

In any case, thanks in advance for this fix!

modules/registry/registry.go
stevenh: This is really just a bug in WSL, there's no reason it shouldn't work it just doesn't.

modules/registry/registry.go
stevenh: You're right I'm used to a more strict lint set.

We should discuss changes to the linters and config going forward, as there's a lot of code smells which could be caught by the linter, however would need a clean up pass first.

mdelapenya: Indeed, happy to discuss about that 😊

stevenh: Correct it allows for `os` and `testing.T` flows.

My thinking was I didn't want to expose methods that took `testing.T` directly.

That said we could keep these private for now and see if there is a demand for them moving forward?

stevenh:
Depends on https://github.com/testcontainers/testcontainers-go/pull/2727

modules/registry/registry.go
stevenh: I just realised I made it public because the tests are in `registry_test` package I could change that to just `registry`, what's the preferred cause?

mdelapenya: Thanks for the explanation 🙇 

If we make it private, then we need to change the test scope of the registry_test.go file, from `registry_test` to `registry`.  I'd prefer keeping the test scope so let's keep it public, and maybe a comment could be helpful. Wdyt about an internal package for this kind of things?

modules/registry/registry.go
```diff
@@ -203,3 +253,67 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
...
+
+	cfg := dockercfg.Config{
+		AuthConfigs: map[string]dockercfg.AuthConfig{
+			host: {
+				Username: username,
+				Password: password,
+				Auth:     creds,
+			},
+		},
+		CredentialsStore: "desktop",
```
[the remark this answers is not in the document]
stevenh: I have tested without and still seems to work, but I'd be interested if anyone knows of any hidden issues that might create?

modules/registry/registry.go
stevenh: That could work let me think on it.

modules/registry/registry.go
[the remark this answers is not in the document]
kiview: I am fairly convinced, the `CredentialsStore` key here would not have any effect (unless providing meta information to Docker downstream?), since the actual credentials are already provided (right?). This should normally indicate to Docker, after reading the `config.json`, that it needs to load credentials using a desktop credential store helper. But this situation does not apply to this part of the code, right?

So removing it sounds right.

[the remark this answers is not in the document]
stevenh: Agreed, have removed and refactored the new methods a little bit to simplify given the original returns aren't needed.

modules/registry/registry.go
```diff
@@ -203,3 +253,55 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
 
 	return c, nil
 }
+
+// SetDockerAuthConfig sets the DOCKER_AUTH_CONFIG environment variable with
+// authentication for with the given host, username and password sets.
+// It returns a function to reset the environment back to the previous state.
+func SetDockerAuthConfig(host, username, password string, additional ...string) (func(), error) {
```
mdelapenya, line 260: I'm going to take this PR and add the docs so you don't have to deal with that burden. Will push commits on top of these ones

docs/modules/registry.md
```diff
@@ -39,6 +39,17 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
 - `string`, the Docker image to use.
 - `testcontainers.ContainerCustomizer`, a variadic argument for passing options.
 
+### Docker Auth Config
```
mdelapenya, line 42: @stevenh could you take a look? 🙏

docs/modules/registry.md
```diff
@@ -39,6 +39,17 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
...
+
+The module exposes a way to set the Docker Auth Config for the Registry container, thanks to the `SetDockerAuthConfig` function.
+This is useful when you need to pull images from a private registry. It basically sets the `DOCKER_AUTH_CONFIG` environment variable
+with authentication for with the given host, username and password sets. It returns a function to reset the environment back to the previous state,
```
stevenh, line 46: Looks like a typo here, think the with should be removed

docs/modules/registry.md
```diff
@@ -39,6 +39,17 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
...
+which is helpful when you need to reset the environment after a test.
+
+On the same hand, the module also exposes a way to get the Docker Auth Config for the Registry container, thanks to the `DockerAuthConfig` function.
```
stevenh, line 49: This is actually a helper to set, it constructs the config for the given details, including calculating the Auth value for the username and password.

stevenh (commented):
Needs a few tweaks

docs/modules/registry.md
mdelapenya: Actually copied from the function comments. Fixed in both places

docs/modules/registry.md
mdelapenya: Could you suggest the right words here? 🙏

docs/modules/registry.md
```diff
@@ -39,6 +39,17 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
...
 
+### Docker Auth Config
+
+The module exposes a way to set the Docker Auth Config for the Registry container, thanks to the `SetDockerAuthConfig` function.
+This is useful when you need to pull images from a private registry. It basically sets the `DOCKER_AUTH_CONFIG` environment variable
+with authentication for the given host, username and password sets. It returns a function to reset the environment back to the previous state,
+which is helpful when you need to reset the environment after a test.
+
+On the same hand, the module also exposes a way to get the Docker Auth Config for the Registry container, thanks to the `DockerAuthConfig` helper function.
+This function returns a map of `AuthConfigs` including base64 encoded Auth field for the provided details.
```
stevenh, line 50:
```suggestion
On the same hand, the module also exposes a way to build a Docker Auth Config for the Registry container, thanks to the `DockerAuthConfig` helper function.
This function returns a map of `AuthConfigs` including base64 encoded Auth field for the provided details.
```

docs/modules/registry.md
stevenh: Suggestion below, just a simple tweak from `get the` to `build a`.

mdelapenya (approved):
LGTM, thanks!

PR #2706, opened 2024-08-08
fix: authentication tests on WSL

stevenh:
Fix docker auth tests to run on WSL by using the local IP instead of localhost, which can be unreliable.

While there refactor the tests to use a helper instead of hand coding each environment setup.

Remove duplicate logging about terminating container as default hooks already provide this.

### DIFF
--- a/docker_auth_test.go
+++ b/docker_auth_test.go
@@ -3,7 +3,9 @@ package testcontainers
 import (
 	"context"
 	_ "embed"
+	"encoding/base64"
 	"fmt"
+	"net"
 	"os"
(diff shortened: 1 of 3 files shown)


mdelapenya:
I think this is a duplicate of #2705, isn't it?

stevenh:
> I think this is a duplicate of #2705, isn't it?

Same issue different location, happy to merge the two if you want.

mdelapenya (approved):
LGTM, in a follow up, let's reuse what #2705 added for the localhost IP.

Thanks!

PR #2707, opened 2024-08-08
chore(mysql): add missing error check in example

stevenh:
Add a missing error check in the mysql example for connection setup.

### DIFF
--- a/modules/mysql/examples_test.go
+++ b/modules/mysql/examples_test.go
@@ -68,7 +68,7 @@ func ExampleRun_connect() {
 
 	connectionString, err := mysqlContainer.ConnectionString(ctx)
 	if err != nil {
-		log.Fatalf("failed to get connection string: %s", err)
+		log.Fatalf("failed to get connection string: %s", err) // nolint:gocritic
 	}
 
 	db, err := sql.Open("mysql", connectionString)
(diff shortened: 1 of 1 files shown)


modules/mysql/examples_test.go
```diff
@@ -66,7 +66,10 @@ func ExampleRun_connect() {
 		}
 	}()
 
-	connectionString, _ := mysqlContainer.ConnectionString(ctx)
+	connectionString, err := mysqlContainer.ConnectionString(ctx)
+	if err != nil {
+		log.Fatalf("failed to get connection string: %s", err)
```
mdelapenya, line 71:
```suggestion
		log.Fatalf("failed to get connection string: %s", err) // nolint:gocritic
```

stevenh: FYI: This gets reworked in a latter PR as all the examples are broken

mdelapenya: Broken in what manner? Is the lint complaining?

stevenh: If an error occurs clean up doesn't happen as `log.Fatalf` issues an `os.Exit(1)` so the defers don't run.

mdelapenya: Ah oh yes, sure, thanks for the explanation. You're very right 🙇

PR #2708, opened 2024-08-08
chore: improve lifecycle errors

stevenh:
Improve the lifecycle errors with wrapping so that its easier to determine the cause of an error.


mdelapenya (approved):
LGTM, thanks!

PR #2711, opened 2024-08-08
chore(registry): disable build log

stevenh:
Disable build log so its easier to see where failures occur.


mdelapenya (approved):
LGTM, thanks!

PR #2712, opened 2024-08-08
fix(compose): remove test volumes

stevenh:
Remove test volumes so that we don't leave orphaned resources after tests have run.

Remove really unused code as there is no need for it.

Add missing error checks on pipe setup.

### DIFF
--- a/modules/compose/compose_api.go
+++ b/modules/compose/compose_api.go
@@ -32,14 +32,6 @@ func (f stackUpOptionFunc) applyToStackUp(o *stackUpOptions) {
 	f(o)
 }
 
-//nolint:unused
-type stackDownOptionFunc func(do *api.DownOptions)
-
-//nolint:unused
-func (f stackDownOptionFunc) applyToStackDown(do *api.DownOptions) {
-	f(do)
-}
-
 // RunServices is comparable to 'docker compose run' as it only creates a subset of containers
 // instead of all services defined by the project
 func RunServices(serviceNames ...string) StackUpOption {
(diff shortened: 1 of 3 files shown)


modules/compose/compose_api.go
```diff
@@ -32,14 +32,6 @@ func (f stackUpOptionFunc) applyToStackUp(o *stackUpOptions) {
...
 }
 
-//nolint:unused
-type stackDownOptionFunc func(do *api.DownOptions)
-
-//nolint:unused
-func (f stackDownOptionFunc) applyToStackDown(do *api.DownOptions) {
-	f(do)
-}
-
```
mdelapenya, line 42: In line L561, I'd also add the testcontainers labels to all the created volumes, like this:

```go
	for key, n := range proj.Volumes {
		n.Labels = map[string]string{
			api.ProjectLabel: proj.Name,
			api.VolumeLabel:  n.Name,
			api.VersionLabel: api.ComposeVersion,
		}

		for k, label := range testcontainers.GenericLabels() {
			n.Labels[k] = label
		}

		proj.Volumes[key] = n
	}
```

stevenh: I'll do that in a follow up PR if you don't mind as there's some pending changes to how that works which will make it easier.

mdelapenya: SGTM, thanks

modules/compose/compose_local.go
```diff
@@ -307,14 +307,26 @@ func execute(
 		cmd.Env = append(cmd.Env, key+"="+value)
 	}
 
-	stdoutIn, _ := cmd.StdoutPipe()
-	stderrIn, _ := cmd.StderrPipe()
+	stdoutIn, err := cmd.StdoutPipe()
```
mdelapenya, line 310: Fine to add this, although the local compose flavour is deprecated.

mdelapenya (approved):
LGTM, thanks!

PR #2715, opened 2024-08-08
fix: filter context errors from log production

stevenh:
Don't return context errors from log production as it's confusing to callers which cancel, making it harder to understand any actual errors that occurred.


mdelapenya (approved):
LGTM

PR #2716, opened 2024-08-08
chore(wait): log test timeout

stevenh:
Increase the log test timeout to avoid random test failures on slow machines.

Convert to require to make tests easier to read / understand.


mdelapenya (approved):
LGTM, thanks!

PR #2717, opened 2024-08-08
fix: skip unhealthy

stevenh:
Pass the right variable to skip unhealthy function in tests, so it actually functions.


mdelapenya (approved):
LGTM intensifies

PR #2718, opened 2024-08-08
fix: readiness hook back off

stevenh:
Fix the definition of readiness hook back off which had the wrong MaxElapsedTime resulting in some flaky tests.

### DIFF
--- a/lifecycle.go
+++ b/lifecycle.go
@@ -233,15 +233,15 @@ var defaultReadinessHook = func() ContainerLifecycleHooks {
 			func(ctx context.Context, c Container) error {
 				// wait until all the exposed ports are mapped:
 				// it will be ready when all the exposed ports are mapped,
-				// checking every 50ms, up to 5s, and failing if all the
-				// exposed ports are not mapped in that time.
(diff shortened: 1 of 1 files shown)


lifecycle.go
```diff
@@ -233,15 +233,15 @@ var defaultReadinessHook = func() ContainerLifecycleHooks {
...
+				// exposed ports are not mapped in 5s.
 				dockerContainer := c.(*DockerContainer)
 
 				b := backoff.NewExponentialBackOff()
 
 				b.InitialInterval = 50 * time.Millisecond
-				b.MaxElapsedTime = 1 * time.Second
-				b.MaxInterval = 5 * time.Second
+				b.MaxElapsedTime = 5 * time.Second
+				b.MaxInterval = time.Duration(float64(time.Second) * backoff.DefaultRandomizationFactor)
```
mdelapenya, line 244: Do you think this will fix https://github.com/testcontainers/testcontainers-go/issues/2633?

stevenh: It might, but the exposed port check using a fixed timeout could still be an issue.

PR #2719, opened 2024-08-08
chore: increase timeout values

stevenh:
Increase timeout values from microseconds to milliseconds as the former can lead to failures depending on hardware.


mdelapenya (approved):
LGTM, thanks

PR #2720, opened 2024-08-08
chore: improve error wrapping in some docker container methods and compose module

stevenh:
Improve error wrapping to help detect failures in compose and wait tests.

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -508,12 +508,12 @@ func (c *DockerContainer) Exec(ctx context.Context, cmd []string, options ...tce
 
 	response, err := cli.ContainerExecCreate(ctx, c.ID, processOptions.ExecConfig)
 	if err != nil {
-		return 0, nil, err
+		return 0, nil, fmt.Errorf("container exec create: %w", err)
 	}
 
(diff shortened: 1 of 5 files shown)


options_test.go
```diff
@@ -97,7 +97,7 @@ func TestWithLogConsumers(t *testing.T) {
 	// we expect an error because the MySQL environment variables are not set
 	// but this is expected because we just want to test the log consumer
 	require.Error(t, err)
-	require.Contains(t, err.Error(), "container exited with code 1")
+	require.Contains(t, err.Error(), "start container: container exited with code 1")
```
mdelapenya, line 100:
```suggestion
	require.Contains(t, err.Error(), "start container: started hook: wait until ready: container exited with code 1")
```

stevenh: This is using `Contains` so it only has to check for base error we're expecting. This specifically avoids checking any wrapping that might happen as that may change making the test fragile, is as close to `errors.Is` as we can get.

mdelapenya (approved):
LGTM, thanks!

PR #2721, opened 2024-08-08
chore: remove unused parameters

stevenh:
Remove unused parameters from private functions.


mdelapenya (approved):
LGTM, thanks!

IIRC I started with functions with the same signature, but it probably does not matter at all

PR #2722, opened 2024-08-08
fix(compose): container locking

stevenh:
Fix container locking which was duplicated and held for the duration of the concurrent operation so preventing any concurrency.

### DIFF
--- a/modules/compose/compose_api.go
+++ b/modules/compose/compose_api.go
@@ -192,8 +192,8 @@ type dockerCompose struct {
 	// only one strategy can be added to a service, to use multiple use wait.ForAll(...)
 	waitStrategies map[string]wait.Strategy
 
-	// used to synchronise writes to the containers map
-	containersLock sync.RWMutex
+	// Used to synchronise writes to the containers.
+	containersLock sync.Mutex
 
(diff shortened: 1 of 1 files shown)


modules/compose/compose_api.go
```diff
@@ -474,15 +465,14 @@ func (d *dockerCompose) lookupContainer(ctx context.Context, svcName string) (*t
 
 	ctr.SetProvider(dockerProvider)
 
+	d.containersLock.Lock()
```
mdelapenya, line 468: Totally right, makes it more sense to lock inside the lookup method. Good catch!

mdelapenya (approved):
LGTM, thanks!

mdelapenya (changes requested):
BTW, could you take a look at the test errors? they seem narrowed to the compose module, so they seem legit

stevenh:
> BTW, could you take a look at the test errors? they seem narrowed to the compose module, so they seem legit

Yep, there was other fixes in later versions which we're key for this one. I've back ported those, and will check the tests once they complete.

mdelapenya (approved):
LGTM, thanks!

PR #2723, opened 2024-08-08
chore: fix nginx request failures in tests

stevenh:
Fix tests which make http requests to nginx container by switching to http startup check, which avoids issues on nginx being able to respond to requests.


mdelapenya:
Could you elaborate a bit more on the issues experienced here? I see the value in the HTTP call instead of the listening port, but I'd like to understand more the failures here.

mdelapenya (approved):
LGTM, thanks!

PR #2724, opened 2024-08-08
fix(elasticsearch): wait for

stevenh:
Wait for the HTTP port to be available to prevent random failures when the container isn't fully started and returns 503 errors.

### DIFF
--- a/modules/elasticsearch/elasticsearch.go
+++ b/modules/elasticsearch/elasticsearch.go
@@ -2,6 +2,8 @@ package elasticsearch
 
 import (
 	"context"
+	"crypto/tls"
+	"crypto/x509"
 	"fmt"
 	"io"
 	"os"
@@ -15,6 +17,7 @@ const (
 	defaultTCPPort      = "9300"
 	defaultPassword     = "changeme"
 	defaultUsername     = "elastic"
+	defaultCaCertPath   = "/usr/share/elasticsearch/config/certs/http_ca.crt"
 	minimalImageVersion = "7.9.2"
 )
 
@@ -32,7 +35,7 @@ type ElasticsearchContainer struct {
 }
 
 // Deprecated: use Run instead
-// RunContainer creates an instance of the Couchbase container type
+// RunContainer creates an instance of the Elasticsearch container type
 func RunContainer(ctx context.Context, opts ...testcontainers.ContainerCustomizer) (*ElasticsearchContainer, error) {
 	return Run(ctx, "docker.elastic.co/elasticsearch/elasticsearch:7.9.2", opts...)
 }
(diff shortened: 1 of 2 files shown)


modules/elasticsearch/elasticsearch.go
```diff
@@ -111,6 +112,32 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
 	return esContainer, nil
 }
 
+// defaultWaitFor sets the default waiting strategy based on settings.
+func defaultWaitFor(settings *Options, req *testcontainers.ContainerRequest) {
+	waitHTTP := wait.ForHTTP("/").WithPort(defaultHTTPPort)
+	if settings.CACert != nil {
+		caCertPool := x509.NewCertPool()
+		caCertPool.AppendCertsFromPEM(settings.CACert)
+		waitHTTP = waitHTTP.WithTLS(true, &tls.Config{RootCAs: caCertPool})
```
mdelapenya, line 121: There is no test for this block. Could you please add one? 🙏

mdelapenya (commented):
Apart from the missing test, LGTM. Once added, I think this is ready to go.

Thanks!

stevenh:
> WithTLS

This is actually a bug, I missed the circular dependency between the ready and wait conditions.

There's two options:
1. Just use insecure for SSL check
2. Refactor to a wait on the file using https://github.com/testcontainers/testcontainers-go/pull/2731

stevenh:
@mdelapenya rebased on top of the new `wait.ForFile` so ready fore review now.

modules/elasticsearch/elasticsearch.go
```diff
@@ -110,6 +100,63 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
...
+		cw := &certWriter{
+			options:  options,
+			certPool: x509.NewCertPool(),
+		}
+
+		waitHTTP = waitHTTP.
+			WithTLS(true, &tls.Config{RootCAs: cw.certPool})
+		if options.Password != "" || options.Username != "" {
+			waitHTTP = waitHTTP.WithBasicAuth(options.Username, options.Password)
+		}
```
mdelapenya, line 142: Is this executed in both parts of the if/else: here and in https://github.com/testcontainers/testcontainers-go/pull/2724/files#diff-cef74d75f306334365eb9db88b5a03e2342780f1e3fee93ae21954d0bd7078b0R146-R148?

modules/elasticsearch/elasticsearch.go
```diff
@@ -110,6 +100,63 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
...
+		strategies = append(strategies, wait.ForFile(defaultCaCertPath).WithMatcher(cw.Read))
```
mdelapenya, line 143: BTW I enjoyed a lot understanding the behaviour of the wait.ForFile and this matcher represented by simply reading the file after copying it.

Really smart and super useful 👏

mdelapenya (dismissed):
There is just one comment regarding adding the basic auth to the wait strategy twice.

modules/elasticsearch/elasticsearch.go
stevenh: Good catch

mdelapenya (approved):
LGTM, thanks!

Issue #2701, opened 2024-08-07
[Bug]: Unable to disable Ryuk in 0.0.32


jonchammer:
### Testcontainers version

0.0.32

### Using the latest Testcontainers version?

Yes

### Host OS

MacOS

### Host arch

ARM

### Go version

1.22

### Docker version

```shell
Client:
 Version:           27.1.1
 API version:       1.46
 Go version:        go1.21.12
 Git commit:        6312585
 Built:             Tue Jul 23 19:54:12 2024
 OS/Arch:           darwin/arm64
 Context:           desktop-linux

Server: Docker Desktop 4.33.0 (160616)
 Engine:
  Version:          27.1.1
  API version:      1.46 (minimum version 1.24)
  Go version:       go1.21.12
  Git commit:       cc13f95
  Built:            Tue Jul 23 19:57:14 2024
  OS/Arch:          linux/arm64
  Experimental:     false
 containerd:
  Version:          1.7.19
  GitCommit:        2bf793ef6dc9a18e00cb12efb64355c2c9d5eb41
 runc:
  Version:          1.7.19
  GitCommit:        v1.1.13-0-g58aa920
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### Docker info

```shell
Client:
 Version:    27.1.1
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.16.1-desktop.1
    Path:     /Users/jch002j/.docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.29.1-desktop.1
    Path:     /Users/jch002j/.docker/cli-plugins/docker-compose
  debug: Get a shell into any image or container (Docker Inc.)
    Version:  0.0.34
    Path:     /Users/jch002j/.docker/cli-plugins/docker-debug
  desktop: Docker Desktop commands (Alpha) (Docker Inc.)
    Version:  v0.0.14
    Path:     /Users/jch002j/.docker/cli-plugins/docker-desktop
  dev: Docker Dev Environments (Docker Inc.)
    Version:  v0.1.2
    Path:     /Users/jch002j/.docker/cli-plugins/docker-dev
  extension: Manages Docker extensions (Docker Inc.)
    Version:  v0.2.25
    Path:     /Users/jch002j/.docker/cli-plugins/docker-extension
[...]

Issue #2609, opened 2024-06-27
[Bug]: TESTCONTAINERS_HOST_OVERRIDE is not working


emfomy:
### Testcontainers version

0.31.0

### Using the latest Testcontainers version?

Yes

### Host OS

N/A

### Host arch

N/A

### Go version

N/A

### Docker version

```shell
N/A
```

### Docker info

```shell
N/A
```

### What happened?

According to https://golang.testcontainers.org/system_requirements/rancher/ and https://golang.testcontainers.org/features/configuration/, we should be able to use `TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE` and `TESTCONTAINERS_HOST_OVERRIDE`.

However, these environment variables are not parsed in https://github.com/testcontainers/testcontainers-go/blob/v0.31.0/internal/config/config.go#L69

mdelapenya:
Hi @emfomy thanks for opening this issue. Just to let you know, there is an ongoing PR addressing this: https://github.com/testcontainers/testcontainers-go/issues/2609

stevenh:
This is fixed by two different pending PR's https://github.com/testcontainers/testcontainers-go/pull/2661 and https://github.com/testcontainers/testcontainers-go/pull/2725

mdelapenya:
@emfomy could you verify if the current HEAD main resolved this issues? Thanks!

emfomy:
It does solve the issue! Thanks!

PR #2725, opened 2024-08-08
fix: config via environment

stevenh:
Fix the ability to set the configuration of testcontainers using an environment variable set by a caller before the use of testcontainers.

This removes the warning about disabling the reaper which is actually safe when done correctly.

Also optimise verbose check while there.

Fixes: #2701 #2636 #2609

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -20,18 +20,71 @@ var (
 
 // testcontainersConfig {
 
-// Config represents the configuration for Testcontainers
+// Config represents the configuration for Testcontainers.
+// User values are read from ~/.testcontainers.properties file which can be overridden
+// using the specified environment variables. For more information, see [Custom Configuration].
+//
+// The Ryuk prefixed fields controls the [Garbage Collector] feature, which ensures that
+// resources are cleaned up after the test execution.
+//
+// [Garbage Collector]: https://golang.testcontainers.org/features/garbage_collector/
+// [Custom Configuration]: https://golang.testcontainers.org/features/configuration/
 type Config struct {
-	Host                    string        `properties:"docker.host,default="`
-	TLSVerify               int           `properties:"docker.tls.verify,default=0"`
-	CertPath                string        `properties:"docker.cert.path,default="`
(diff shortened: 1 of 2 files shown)


mdelapenya:
@stevenh we probably want to still log the message but not that early. I naively added it to the init method without considering the impact for users not being able to set the logger at the right time.

The current requirements for this would be:
- allow users set their logger. At the moment, each container request supports passing a logger, otherwise, the default one will be used.
- log the Ryuk-disabled message, using user's logger.

For a follow-up redesign (v1) we probably want to include the logger as part of the setup (see your comment in https://github.com/testcontainers/testcontainers-go/issues/2701#issuecomment-2276560744)

stevenh:
> @stevenh we probably want to still log the message but not that early. I naively added it to the init method without considering the impact for users not being able to set the logger at the right time.
> 
> The current requirements for this would be:
> 
> * allow users set their logger. At the moment, each container request supports passing a logger, otherwise, the default one will be used.
> * log the Ryuk-disabled message, using user's logger.
> 
> For a follow-up redesign (v1) we probably want to include the logger as part of the setup (see your comment in [#2701 (comment)](https://github.com/testcontainers/testcontainers-go/issues/2701#issuecomment-2276560744))

I'd actually push back on that and suggest we don't log, here's why:

If someone has chosen to disable the reaper it will be a very conscious decision, and in doing so they are actively taking responsibility. While the user can't protect leaked resources in code for a test timeout we can document that and the extra noise makes it harder to see issues when they occur.

So to help unblock, until we can revisit for v1, I've pushed an update which improves the pkgsite docs, does that work?

mdelapenya:
> If someone has chosen to disable the reaper it will be a very conscious decision

Indeed. Let's do as you propose here without the log. I know we have commented this internally, @kiview, and we want to log the message not using STDOUT but a user-defined logger. Let's keep this in mind for v1.

mdelapenya:
> I've pushed an update which improves the pkgsite docs, does that work?

Not only the pkgsite docs but the docs site: https://deploy-preview-2725--testcontainers-go.netlify.app/features/configuration/ 👏

PR #2727, opened 2024-08-09
fix!: docker authentication setup

stevenh:
Check and return errors in the process of determining authentication configs so that unexpected failures don't occur latter in the process of build an image or creating a container.

BuildOptions will now return an empty result on error instead of an incomplete one, to ensure that consumers don't use partial data.

Fix builds with different config or override environments failing when the authentication configuration changes, which was introduced by #2646.

Report errors from GetRegistryCredentials calls to avoid unexpected failures latter on in the authentication process.

Split out the functionality to read a Dockerfile from an io.Reader into ExtractImagesFromReader, as required when processing from a tar archive.

Deprecated function ContainerRequest.GetAuthConfigs will now panic if an error occurs, so that callers understand that an failure occured.

BREAKING CHANGE Add support for determining the required authentication in when building an image from a `ContextArchive`, this requires `ContextArchive` support `io.Seeker`.

### DIFF
--- a/container.go
+++ b/container.go
@@ -1,6 +1,7 @@
 package testcontainers
 
 import (
+	"archive/tar"
 	"context"
 	"errors"
 	"fmt"
@@ -10,6 +11,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/cpuguy83/dockercfg"
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/container"
 	"github.com/docker/docker/api/types/network"
@@ -85,7 +87,7 @@ type ImageBuildInfo interface {
 // rather than using a pre-built one
 type FromDockerfile struct {
 	Context        string                         // the path to the context of the docker build
-	ContextArchive io.Reader                      // the tar archive file to send to docker that contains the build context
+	ContextArchive io.ReadSeeker                  // the tar archive file to send to docker that contains the build context
 	Dockerfile     string                         // the path from the context to the Dockerfile for the image, defaults to "Dockerfile"
 	Repo           string                         // the repo label for image, defaults to UUID
 	Tag            string                         // the tag label for image, defaults to UUID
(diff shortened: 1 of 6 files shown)


container.go
```diff
@@ -85,7 +87,7 @@ type ImageBuildInfo interface {
 // rather than using a pre-built one
 type FromDockerfile struct {
 	Context        string                         // the path to the context of the docker build
-	ContextArchive io.Reader                      // the tar archive file to send to docker that contains the build context
+	ContextArchive io.ReadSeeker                  // the tar archive file to send to docker that contains the build context
```
mdelapenya, line 90: I guess we are using a Seeker here for allowing more performant retries [here](https://github.com/testcontainers/testcontainers-go/pull/2727/files#diff-14bacfb63a209323729af295c64c9bb978c84e810d18e3e7b7ae66d9b3cd7acaR357), right? Let's be clear in the release notes about the motivation for this change. Could you write a meaningful and short proposal for it 🙏 ?

container.go
```diff
@@ -305,30 +307,90 @@ func (c *ContainerRequest) GetTag() string {
...
+	// Get the auth configs once for all images as it can be a time-consuming operation.
+	configs, err := getDockerAuthConfigs()
+	if err != nil {
+		return nil, err
 	}
 
 	authConfigs := map[string]registry.AuthConfig{}
 	for _, image := range images {
-		registry, authConfig, err := DockerImageAuth(context.Background(), image)
+		registry, authConfig, err := dockerImageAuth(context.Background(), image, configs)
```
mdelapenya, line 380: Here `configs` is obtained from L373 (`configs, err := getDockerAuthConfigs()`), and inside the `dockerImageAuth` function we are going to call it again (docker_auth.go:L41) if configs is nil. Reading the code from `getDockerAuthConfigs` it only can be nil when an error occurs. So I guess we are doing this because we want to force a nil value for configs in `docker_auth.go:L29`.

I can see it, although a bit complex. I'm not blocking it, although I'd like to give it a second round of thoughts. I'm not opposed to it so if we cannot find it it's fine.

container.go
```diff
@@ -378,7 +443,7 @@ func (c *ContainerRequest) BuildOptions() (types.ImageBuildOptions, error) {
 	for _, is := range c.ImageSubstitutors {
 		modifiedTag, err := is.Substitute(tag)
 		if err != nil {
-			return buildOptions, fmt.Errorf("failed to substitute image %s with %s: %w", tag, is.Description(), err)
+			return types.ImageBuildOptions{}, fmt.Errorf("failed to substitute image %s with %s: %w", tag, is.Description(), err)
```
mdelapenya, line 446: Cool, returning zero values here is fine. Thanks!

container.go
mdelapenya: Idea, wdyt: we can move the specialised case for nil configs to the place it's needed, which is just one, making the general case simpler.

docker_auth.go
```diff
@@ -80,24 +95,93 @@ func defaultRegistry(ctx context.Context) string {
...
-// authConfig represents the details of the auth config for a registry.
-type authConfig struct {
+// authConfigResult is a result looking up auth details for key.
+type authConfigResult struct {
 	key string
 	cfg registry.AuthConfig
+	err error
+}
+
+// credentialsCache is a cache for registry credentials.
```
mdelapenya, line 105: In v1 there will be an `auth` package in the root level. Do you think we can move this struct and methods to `internal/core/auth` so the forward-port is easier? 🤔

mdelapenya (commented):
LGTM although I'd like to discuss a bit on some part of the code. PLMK your thoughts on my comments 🙏

container.go
stevenh: Not more performance, we need to be able to rewind to do multiple passes.

container.go
stevenh: The reason for the handling `nil` in `dockerImageAuth` is to support the call from `DockerImageAuth`. I did that to avoid replicating default registry and image registry lookup, happy to move if you think that's better?

Here's an example of what it would look like.
```go
// DockerImageAuth returns the auth config for the given Docker image, extracting first its Docker registry.
// Finally, it will use the credential helpers to extract the information from the docker config file
// for that registry, if it exists.
func DockerImageAuth(ctx context.Context, image string) (string, registry.AuthConfig, error) {
	configs, err := getDockerAuthConfigs()
	if err != nil {
		reg := core.ExtractRegistry(image, defaultRegistryFn(ctx))
		return reg, registry.AuthConfig{}, err
	}

	return dockerImageAuth(ctx, image, configs)
}

// dockerImageAuth returns the auth config for the given Docker image.
// If configs is nil it will load it, which is useful as loading can
// be a time consuming operation.
func dockerImageAuth(ctx context.Context, image string, configs map[string]registry.AuthConfig) (string, registry.AuthConfig, error) {
	defaultRegistry := defaultRegistryFn(ctx)
	reg := core.ExtractRegistry(image, defaultRegistry)

	if cfg, ok := getRegistryAuth(reg, configs); ok {
		return reg, cfg, nil
	}

	return reg, registry.AuthConfig{}, dockercfg.ErrCredentialsNotFound
}
```

docker_auth.go
stevenh: I'd have to look into this some more, would suggest we move at a later date if we do.

container.go
```diff
@@ -305,30 +307,90 @@ func (c *ContainerRequest) GetTag() string {
...
-// Deprecated: Testcontainers will detect registry credentials automatically, and it will be removed in the next major release
-// GetAuthConfigs returns the auth configs to be able to pull from an authenticated docker registry
+// Deprecated: Testcontainers will detect registry credentials automatically, and it will be removed in the next major release.
+// GetAuthConfigs returns the auth configs to be able to pull from an authenticated docker registry.
+// Panics if an error occurs.
 func (c *ContainerRequest) GetAuthConfigs() map[string]registry.AuthConfig {
-	return getAuthConfigsFromDockerfile(c)
+	auth, err := getAuthConfigsFromDockerfile(c)
+	if err != nil {
+		panic(fmt.Sprintf("failed to get auth configs from Dockerfile: %v", err))
```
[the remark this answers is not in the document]
stevenh: Not sure I understand, could you clarify?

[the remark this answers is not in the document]
mdelapenya: Yeah, this method will be removed in the next major version, so in this change I'd move the panic to the `getAuthConfigsFromDockerfile` method. Or maybe I'm wrong and we want to panic here because it's deprecated to inform that something will eventually change 🤔 Is that a pattern for deprecations that I'm not aware of? Just curious to learn about that.

container.go
mdelapenya: I see this code simpler. Can we add unit tests for `getRegistryAuth` as they do not exist atm, or do you prefer me doing that in a separate PR?

container.go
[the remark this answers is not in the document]
stevenh: It only needs to panic here because the internal method now returns an error which this method can't handle without an API breaking change.

We could ignore like before, but thought it better to have the user of deprecated method know there was an issue.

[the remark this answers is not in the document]
mdelapenya: I see, thanks for the clarification. Given the comment explicitly mentions it, lgtm

mdelapenya (approved):
LGTM, thanks again for your contributions 🙇

PR #2728, opened 2024-08-09
fix(reaper): refactor to allow retries and fix races

stevenh:
Refactor how the reaper is created to allow for proper retries of temporary errors such as container not found issues during startup or shutdown races.

This eliminates the use of sync.Once which wasn't solving the problem at hand and replaces it with a singleton spawner with locked access.

Wrap reaper errors so we can determine the cause of failures more easily.

Fix race condition in port wait when loading from container by always waiting for the port first.

Remove unnecessary use of buffering and invalid retry logic in reaper connection handling which could never recover correctly from a partial read.

Move the reaper creation just before connections are established in compose to ensure its still running when the Connect calls are made.

Previously the reaper was requested in NewDockerComposeWith which means it could have already shutdown before connections are made during the later sections of Up if the startup took over 1 minute.

This was causing consistent failures for:
TestDockerComposeAPIWithWaitLogStrategy

Ensure that resource labels are correct so that resources aren't reaped when the reaper is disabled by excluding session id when reaper is disabled.

Error when creating a reaper when the config says it's disabled so that we avoid hard to debug issues because a reaper is running when it shouldn't be.

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -183,7 +183,7 @@ func (c *DockerContainer) Inspect(ctx context.Context) (*types.ContainerJSON, er
 func (c *DockerContainer) MappedPort(ctx context.Context, port nat.Port) (nat.Port, error) {
 	inspect, err := c.Inspect(ctx)
 	if err != nil {
-		return "", err
+		return "", fmt.Errorf("inspect: %w", err)
 	}
 	if inspect.ContainerJSONBase.HostConfig.NetworkMode == "host" {
 		return port, nil
@@ -204,7 +204,7 @@ func (c *DockerContainer) MappedPort(ctx context.Context, port nat.Port) (nat.Po
 		return nat.NewPort(k.Proto(), p[0].HostPort)
 	}
 
-	return "", errors.New("port not found")
+	return "", errdefs.NotFound(fmt.Errorf("port %q not found", port))
 }
 
 // Deprecated: use c.Inspect(ctx).NetworkSettings.Ports instead.
@@ -980,9 +980,7 @@ func (p *DockerProvider) BuildImage(ctx context.Context, img ImageBuildInfo) (st
 }
 
 // CreateContainer fulfils a request for a container without starting it
-func (p *DockerProvider) CreateContainer(ctx context.Context, req ContainerRequest) (Container, error) {
-	var err error
-
+func (p *DockerProvider) CreateContainer(ctx context.Context, req ContainerRequest) (con Container, err error) { //nolint:nonamedreturns // Needed for error checking.
 	// defer the close of the Docker client connection the soonest
 	defer p.Close()
 
@@ -1027,22 +1025,23 @@ func (p *DockerProvider) CreateContainer(ctx context.Context, req ContainerReque
 	// the reaper does not need to start a reaper for itself
 	isReaperContainer := strings.HasSuffix(imageName, config.ReaperDefaultImage)
 	if !p.config.RyukDisabled && !isReaperContainer {
-		r, err := reuseOrCreateReaper(context.WithValue(ctx, core.DockerHostContextKey, p.host), core.SessionID(), p)
+		r, err := spawner.reaper(context.WithValue(ctx, core.DockerHostContextKey, p.host), core.SessionID(), p)
 		if err != nil {
-			return nil, fmt.Errorf("%w: creating reaper failed", err)
+			return nil, fmt.Errorf("reaper: %w", err)
 		}
-		termSignal, err = r.Connect()
-		if err != nil {
-			return nil, fmt.Errorf("%w: connecting to reaper failed", err)
-		}
-	}
 
-	// Cleanup on error, otherwise set termSignal to nil before successful return.
-	defer func() {
-		if termSignal != nil {
-			termSignal <- true
+		termSignal, err := r.Connect()
+		if err != nil {
+			return nil, fmt.Errorf("reaper connect: %w", err)
 		}
-	}()
+
+		// Cleanup on error.
+		defer func() {
+			if err != nil {
+				termSignal <- true
+			}
+		}()
+	}
 
 	if err = req.Validate(); err != nil {
 		return nil, err
@@ -1108,10 +1107,9 @@ func (p *DockerProvider) CreateContainer(ctx context.Context, req ContainerReque
 	}
 
 	if !isReaperContainer {
-		// add the labels that the reaper will use to terminate the container to the request
-		for k, v := range core.DefaultLabels(core.SessionID()) {
-			req.Labels[k] = v
-		}
+		// Add the labels that identify this as a testcontainers container and
+		// allow the reaper to terminate it if requested.
+		AddGenericLabels(req.Labels)
 	}
 
 	dockerInput := &container.Config{
@@ -1205,9 +1203,6 @@ func (p *DockerProvider) CreateContainer(ctx context.Context, req ContainerReque
 		return nil, err
 	}
 
-	// Disable cleanup on success
-	termSignal = nil
-
 	return c, nil
 }
 
@@ -1256,7 +1251,7 @@ func (p *DockerProvider) waitContainerCreation(ctx context.Context, name string)
 	)
 }
 
-func (p *DockerProvider) ReuseOrCreateContainer(ctx context.Context, req ContainerRequest) (Container, error) {
+func (p *DockerProvider) ReuseOrCreateContainer(ctx context.Context, req ContainerRequest) (con Container, err error) { //nolint:nonamedreturns // Needed for error check.
 	c, err := p.findContainerByName(ctx, req.Name)
 	if err != nil {
 		return nil, err
@@ -1279,14 +1274,22 @@ func (p *DockerProvider) ReuseOrCreateContainer(ctx context.Context, req Contain
 
 	var termSignal chan bool
 	if !p.config.RyukDisabled {
-		r, err := reuseOrCreateReaper(context.WithValue(ctx, core.DockerHostContextKey, p.host), sessionID, p)
+		r, err := spawner.reaper(context.WithValue(ctx, core.DockerHostContextKey, p.host), sessionID, p)
 		if err != nil {
 			return nil, fmt.Errorf("reaper: %w", err)
 		}
-		termSignal, err = r.Connect()
+
+		termSignal, err := r.Connect()
 		if err != nil {
-			return nil, fmt.Errorf("%w: connecting to reaper failed", err)
+			return nil, fmt.Errorf("reaper connect: %w", err)
 		}
+
+		// Cleanup on error.
+		defer func() {
+			if err != nil {
+				termSignal <- true
+			}
+		}()
 	}
 
 	// default hooks include logger hook and pre-create hook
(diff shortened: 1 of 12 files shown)


reaper.go
```diff
@@ -61,32 +85,68 @@ func reaperContainerNameFromSessionID(sessionID string) string {
...
+}
+
+// cleanupLocked terminates the reaper container if set.
+// It must be called with the lock held.
+func (r *reaperSpawner) cleanupLocked() error {
+	if r.instance == nil {
+		return nil
+	}
+
+	err := TerminateContainer(r.instance.container)
```
mdelapenya, line 133: There is no `TerminateContainer` function. Probably caused by extracting the PR from the bug bash one.

reaper.go
```diff
@@ -97,159 +157,212 @@ func lookUpReaperContainer(ctx context.Context, sessionID string) (*DockerContai
...
+	return nil
+}
+
+// retryError returns a permanent error if the error is not considered retryable.
+func (r *reaperSpawner) retryError(err error) error {
+	var timeout interface {
+		Timeout() bool
+	}
+	switch {
+	case isCleanupSafe(err),
```
mdelapenya, line 218: isCleanupSafe function is not defined.

mdelapenya (commented):
I added a commit on top to satisfy the usage of several missing functions (see comments).

Other than than, I merged the branch with current main and resolved conflicts.

docker.go
```diff
@@ -944,9 +944,7 @@ func (p *DockerProvider) BuildImage(ctx context.Context, img ImageBuildInfo) (st
 }
 
 // CreateContainer fulfils a request for a container without starting it
-func (p *DockerProvider) CreateContainer(ctx context.Context, req ContainerRequest) (Container, error) {
-	var err error
-
+func (p *DockerProvider) CreateContainer(ctx context.Context, req ContainerRequest) (con Container, err error) { //nolint:nonamedreturns // Needed for error checking.
```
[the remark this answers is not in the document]
stevenh: This doesn't use naked returns, it names the variables to it can use deferred error handling, as per comment.

docker.go
```diff
@@ -1220,7 +1215,7 @@ func (p *DockerProvider) waitContainerCreation(ctx context.Context, name string)
 	)
 }
 
-func (p *DockerProvider) ReuseOrCreateContainer(ctx context.Context, req ContainerRequest) (Container, error) {
+func (p *DockerProvider) ReuseOrCreateContainer(ctx context.Context, req ContainerRequest) (con Container, err error) { //nolint:nonamedreturns // Needed for error check.
```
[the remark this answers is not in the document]
stevenh: As mentioned, no naked returns, error handling.

stevenh:
@mdelapenya reminder this is dependent on #2728 which is why it has missing components still.

I'll rebase once #2738 is merged.

stevenh:
Rebased and picking up again, hopefully wont require many fixes. Will convert from draft when ready.

mdelapenya (commented):
I only added a few super minor comments, other than that, my only concern is about the implications of these changes with the reuse mode we are proposing in #2768 

I'm fine merging this, although I'd like to discuss with you about reuse more in depth, as it could apply to the reaper code as well.

reaper.go
```diff
@@ -97,159 +157,213 @@ func lookUpReaperContainer(ctx context.Context, sessionID string) (*DockerContai
...
-	// 1. if the reaper instance has been already created, return it
-	if reaperInstance != nil {
-		// Verify this instance is still running by checking state.
-		// Can't use Container.IsRunning because the bool is not updated when Reaper is terminated
-		state, err := reaperInstance.container.State(ctx)
-		if err != nil {
-			if !errdefs.IsNotFound(err) {
-				return nil, err
+// isRunning returns an error if the container is not running.
+func (r *reaperSpawner) isRunning(ctx context.Context, container Container) error {
```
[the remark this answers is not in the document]
stevenh: Made this change across the board, there were three other cases.

reaper_test.go
```diff
@@ -131,59 +117,55 @@ func TestContainerStartsWithoutTheReaper(t *testing.T) {
...
-	}
+	ctx := context.Background()
+	sessionID := core.SessionID()
+	reaperContainer, err := spawner.lookupContainer(ctx, sessionID)
+	require.NoError(t, err)
+	require.NotNil(t, reaperContainer)
 }
 
-func TestContainerStartsWithTheReaper(t *testing.T) {
-	config.Reset() // reset the config using the internal method to avoid the sync.Once
```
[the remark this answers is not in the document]
stevenh: I'm struggling to understand this, possible typo, lets discuss 😄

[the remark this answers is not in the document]
mdelapenya: No worries, I mentioned this as a reminder that, in general, I added config.Reset to many places as a means to write tests more easily, while in reality it causes more trouble than it solves. I wanted to provide context on that. That's why I marked it as offtopic, because I'll tackle that refactor of the config tests later on.

mdelapenya:
@stevenh is there anything I can do to move forward with this PR?

generic.go
```diff
@@ -101,7 +101,17 @@ type GenericProvider interface {
...
+// GenericLabels returns a map of labels that can be used to identify resources
+// created by this library. This includes the standard LabelSessionID if the
+// reaper is enabled, otherwise this is excluded to prevent resources being
+// incorrectly reaped.
 func GenericLabels() map[string]string {
 	return core.DefaultLabels(core.SessionID())
 }
+
+// AddGenericLabels adds the generic labels to target.
+func AddGenericLabels(target map[string]string) {
```
mdelapenya, line 113: I think we can leverage the `core.MergeCustomLabels(dst, src)` function addedin #2775

internal/core/labels.go
```diff
@@ -17,12 +18,26 @@ const (
...
+
+	if !config.Read().RyukDisabled {
+		labels[LabelSessionID] = sessionID
+	}
+
+	return labels
+}
+
+// AddDefaultLabels adds the default labels for sessionID to target.
+func AddDefaultLabels(sessionID string, target map[string]string) {
```
mdelapenya, line 38: Same here: #2775 added a very similar method to merge dst and src labels

internal/core/labels.go
```diff
@@ -17,12 +18,26 @@ const (
...
-		LabelSessionID: sessionID,
-		LabelVersion:   internal.Version,
+	labels := map[string]string{
+		LabelBase:    "true",
+		LabelLang:    "go",
+		LabelVersion: internal.Version,
+	}
+
+	if !config.Read().RyukDisabled {
+		labels[LabelSessionID] = sessionID
```
mdelapenya, line 31: I'd not remove the session ID, as this label could be used in the backend to connect test sessions with remote VM executions.

reaper.go
```diff
@@ -34,9 +37,22 @@ const (
 var (
 	// Deprecated: it has been replaced by an internal value
 	ReaperDefaultImage = config.ReaperDefaultImage
-	reaperInstance     *Reaper // We would like to create reaper only once
-	reaperMutex        sync.Mutex
-	reaperOnce         sync.Once
+
+	// reaperPort is the port that the reaper listens on.
+	reaperPort = nat.Port("8080/tcp")
```
mdelapenya, line 42: For this refactor, I'd make this port configurable with the `RYUK_PORT` env var. See moby-ryuk docs: https://github.com/testcontainers/moby-ryuk?tab=readme-ov-file#ryuk-configuration

generic.go
stevenh: We can't use that as its designed to add custom and we're adding default, so it would error.

internal/core/labels.go
stevenh: They actually do kind of opposite jobs, this adds defaults not custom so we can't reuse unfortunately.

internal/core/labels.go
stevenh: We have to exclude it otherwise the reaper would attempt to clean it when the configuration say specifically not to.

stevenh: switched to a dedicated reap label.

mdelapenya:
There are failing tests for the reaper. Could you take a look?

stevenh:
> There are failing tests for the reaper. Could you take a look?

Thanks for the poke I was expecting some issues here, but hadn't got round to checking yet.

erikmansson:
Any progress on this? We're having issues with flaky tests and would love for this to get merged.

stevenh:
ryuk side was merged last week, just fixing a few compatibility issue with tc-java so once that's done, this is next on the list @erikmansson

modules/compose/compose_api.go
```diff
@@ -401,7 +426,11 @@ func (d *dockerCompose) Up(ctx context.Context, opts ...StackUpOption) error {
 		})
 	}
 
-	return errGrpWait.Wait()
+	if err := errGrpWait.Wait(); err != nil {
```
mdelapenya, line 429: question: I think this change is for readability, or I'm missing something else?

reaper.go
```diff
@@ -97,159 +171,214 @@ func lookUpReaperContainer(ctx context.Context, sessionID string) (*DockerContai
...
 	}, nil
 }
 
-// newReaper creates a Reaper with a sessionID to identify containers and a
-// provider to use. Do not call this directly, use reuseOrCreateReaper instead.
-func newReaper(ctx context.Context, sessionID string, provider ReaperProvider) (*Reaper, error) {
+// newReaper creates a connected Reaper with a sessionID to identify containers
+// and a provider to use.
+//
+// Do not call this directly, use reuseOrCreateReaper instead.
```
mdelapenya, line 370: I think this comment is outdated, as we don't have that method any more
```suggestion
```

reaper_test.go
```diff
@@ -102,21 +84,26 @@ func createContainerRequest(customize func(ContainerRequest) ContainerRequest) C
...
 
-	return customize(req)
+	return req
 }
 
-func TestContainerStartsWithoutTheReaper(t *testing.T) {
-	config.Reset() // reset the config using the internal method to avoid the sync.Once
-	tcConfig := config.Read()
-	if !tcConfig.RyukDisabled {
-		t.Skip("Ryuk is enabled, skipping test")
```
mdelapenya, line 117: question: We have a dedicated pipeline that runs all the tests for the code with the Reaper disabled, so there are tests that we skip in there. For that, we set the env var in the GH workflow: please see ci-test-go.yaml file:

```yaml
TESTCONTAINERS_RYUK_DISABLED: "${{ inputs.ryuk-disabled }}"
```

Do we still need to skip those tests?

mdelapenya (commented):
Just a few comments, but LGTM.

Once addressed, specially those regarding the t.Skip when the reaper is disabled on GH actions, we can move on and merge it.

Thanks as always for your hard work with the library 🙇

modules/compose/compose_api.go
stevenh: Nope that was just me forgetting to add the wrap.

reaper_test.go
stevenh: With these changes a single test run can validate with and without the reaper being abled, so there's no need to skip tests anymore, see `func TestContainer`.

There might still benefit of a full test run with the reaper disabled to catch issues in modules, but we could consider removing that as it would cut the number of tests in half, reducing the time to validate. Something to discuss and include in a separate PR.

mdelapenya: Yeah, our reaper-off pipeline runs only for the core. Not sure if it would be desired to run the modules without the reaper as a second step in the same pipeline 🤔

stevenh: The compose module is the only one which springs to mind, it has a `TestDockerComposeAPI_WithoutReaper` and `TestDockerComposeAPI_WithReaper`.

mdelapenya (approved):
LGTM, thanks!

I'd do the refactor of the compose tests in a follow-up, so we get this PR in, and we can start a release.

PR #2739, opened 2024-08-19
chore: avoid using of log.Fatal in main in tests

stevenh:
Remove use of log fatal which prevents defers and correct order of Write and WriteHeader.


mdelapenya (approved):
LGTM, thanks!

I'm changing the title to `chore` as this is not user-facing, but internal to the tests of the project.

PR #2741, opened 2024-08-21
fix: check if the discovered docker socket responds

### LINKED ISSUE #2622: [Bug]: Docker address confusion
### Testcontainers version

9cc2598f23c7

Yes

### Host OS

macOS

### Host arch

arm64

### Go version

1.22.4

### Docker version

```shell
 Version:           27.0.2
 API version:       1.44 (downgraded from 1.46)
 Go version:        go1.22.4
 Git commit:        912c1ddf8a
 Built:             Wed Jun 26 18:39:48 2024
 Context:           rancher-desktop

Server:
 Engine:
  Version:          25.0.5
  API version:      1.44 (minimum version 1.24)
  Go version:       go1.21.10
  Git commit:       e63daec8672d77ac0b2b5c262ef525c7cf17fd20
  Built:            Sun May 12 07:25:43 2024
  OS/Arch:          linux/arm64
  Experimental:     false
 containerd:
  Version:          v1.7.10
  GitCommit:        4e1fe7492b9df85914c389d1f15a3ceedbb280ac
 runc:
  Version:          1.1.12
  GitCommit:        51d5e94601ceffbbd85688df1c928ecccbfa4685
 docker-init:
  Version:          0.19.0
  GitCommit:
```

### Docker info

```shell
 Version:    27.0.2
 Context:    rancher-desktop
 Debug Mode: false
 Plugins:
    Version:  v0.14.1
    Path:     /Users/mike/.docker/cli-plugins/docker-buildx
    Version:  v2.27.1
    Path:     /Users/mike/.docker/cli-plugins/docker-compose

Server:
 Containers: 6
  Running: 0
  Paused: 0
  Stopped: 6
 Images: 58
 Server Version: 25.0.5
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 4e1fe7492b9df85914c389d1f15a3ceedbb280ac
 runc version: 51d5e94601ceffbbd85688df1c928ecccbfa4685
 init version: 
 Security Options:
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 6.6.30-0-virt
 Operating System: Alpine Linux v3.19
 OSType: linux
 Architecture: aarch64
 CPUs: 9
 Total Memory: 15.6GiB
 Name: lima-rancher-desktop
 ID: d04c0365-7614-447f-960d-353ba1523b4f
 Debug Mode: false
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false
```

### What happened?

I have `DOCKER_HOST=unix:///Users/mike/.rd/docker.sock` to use Rancher Desktop that I have running. I also have Test Containers Desktop installed but it's not running. But there is a config file that it created:

```shell
$ cat ~/.testcontainers.properties
testcontainers.reuse.enable = true
tc.host = tcp://127.0.0.1:61153
docker.host = tcp://127.0.0.1:61153
```

When I run the test it runs for some time, does a few test containers things but then fails with an error

> saving images saving images Cannot connect to the Docker daemon at tcp://127.0.0.1:61153. Is the docker daemon running?

It fails on `k3s.LoadImages()`. Looking at the log output, `Resolved Docker Host` is from Test Containers Desktop and `Resolved Docker Socket Path` is from the default Docker context. They don't match each other at all.

The right thing to do is probably to use the selected Docker context?

```shell
$ docker context list
NAME                DESCRIPTION                               DOCKER ENDPOINT                      ERROR
default             Current DOCKER_HOST based configuration   unix:///var/run/docker.sock
rancher-desktop *   Rancher Desktop moby context              unix:///Users/mike/.rd/docker.sock
```

### Relevant log output

```shell
2024/07/03 13:41:44 github.com/testcontainers/testcontainers-go - Connected to docker: 
  Server Version: 25.0.5
  API Version: 1.44
  Operating System: Alpine Linux v3.19
  Total Memory: 15970 MB
  Resolved Docker Host: tcp://127.0.0.1:61153
  Resolved Docker Socket Path: /var/run/docker.sock
  Test SessionID: 6e4dba3df3ad66d1e9850235cab48c0932ccf548532d0ca1aa0deafa462e7c1c
  Test ProcessID: c4deea13-769c-4a5a-9827-d7c4cf5007ee
```

### Additional information

_No response_

mdelapenya:
## What does this PR do?
This PR adds a check for each discovered Docker host in order to verify if a vanilla Docker client can talk to that host.

## Why is it important?
It will remove the confusion when a testcontainers.properties file exists and the Testcontainers Desktop (TCD) app has been used in the past, adding a `tc.host` entry in the properties. In the case TCD is not running, and the container runtime is running (i.e. Docker Desktop), then the latter should be used. Before this fix, the code will cause the discovery algorithm to get the tc.host entry first (it simply exists) and use it, even though it's not responding.

With this fix we'll fail if that Docker host is not accessible, continuing with the Docker host resolution.

## Related issues

- Closes #2622

### DIFF
--- a/internal/core/docker_host.go
+++ b/internal/core/docker_host.go
@@ -116,6 +116,22 @@ func extractDockerHost(ctx context.Context) string {
 			continue
 		}
 
+		// check if the docker host responds to an Info request
+		// we are going to use the default Docker client to check if the host is reachable
+		// which will avoid recursive calls to this function
+		cli, err := client.NewClientWithOpts(client.WithHost(dockerHost))
+		if err != nil {
+			outerErr = fmt.Errorf("%w: %w", outerErr, err)
+			continue
+		}
+		defer cli.Close()
+
+		_, err = cli.Info(ctx)
+		if err != nil {
+			outerErr = fmt.Errorf("%w: %w", outerErr, err)
+			continue
+		}
+
 		return dockerHost
 	}
 
@@ -141,8 +157,7 @@ func extractDockerSocket(ctx context.Context) string {
 // and receiving an instance of the Docker API client interface.
 // This internal method is handy for testing purposes, passing a mock type simulating the desired behaviour.
 func extractDockerSocketFromClient(ctx context.Context, cli client.APIClient) string {
-	// check that the socket is not a tcp or unix socket,
-	// and that it is reachable.
+	// check that the socket is not a tcp or unix socket
 	checkDockerSocketFn := func(socket string) string {
 		// this use case will cover the case when the docker host is a tcp socket
 		if strings.HasPrefix(socket, TCPSchema) {
@@ -151,11 +166,6 @@ func extractDockerSocketFromClient(ctx context.Context, cli client.APIClient) st
 
 		if strings.HasPrefix(socket, DockerSocketSchema) {
 			return strings.Replace(socket, DockerSocketSchema, "", 1)
-		}
-
-		_, err := cli.Info(ctx)
-		if err != nil {
-			return DockerSocketPath
 		}
 
 		return socket
--- a/internal/core/docker_host.go
+++ b/internal/core/docker_host.go
@@ -130,22 +130,29 @@ func extractDockerHost(ctx context.Context) (string, error) {
 		rootlessDockerSocketPath,
 	}
 
-	outerErr := ErrSocketNotFound
+	var errs []error
 	for _, dockerHostFn := range dockerHostFns {
 		dockerHost, err := dockerHostFn(ctx)
 		if err != nil {
-			outerErr = fmt.Errorf("%w: %w", outerErr, err)
+			if !isHostNotFound(err) {
+				errs = append(errs, err)
+			}
 			continue
 		}
 
 		if err = defaultCallbackCheckFn(ctx, dockerHost); err != nil {
+			errs = append(errs, fmt.Errorf("check host %q: %w", dockerHost, err))
 			continue
 		}
 
 		return dockerHost, nil
 	}
 
-	return "", outerErr
+	if len(errs) > 0 {
+		return "", errors.Join(errs...)
+	}
+
+	return "", ErrSocketNotFound
 }
 
 // extractDockerSocket Extracts the docker socket from the different alternatives, without caching the result.
@@ -211,6 +218,26 @@ func extractDockerSocketFromClient(ctx context.Context, cli client.APIClient) st
 	}
 
 	return checkDockerSocketFn(dockerHost)
+}
+
+// isHostNotSet returns true if the error is related to the Docker host
+// not being set, false otherwise.
+func isHostNotFound(err error) bool {
+	switch {
+	case errors.Is(err, ErrTestcontainersHostNotSetInProperties),
+		errors.Is(err, ErrDockerHostNotSet),
+		errors.Is(err, ErrDockerSocketNotSetInContext),
+		errors.Is(err, ErrDockerSocketNotSetInProperties),
+		errors.Is(err, ErrSocketNotFoundInPath),
+		errors.Is(err, ErrDockerSocketNotSetInProperties),
+		errors.Is(err, ErrXDGRuntimeDirNotSet),
+		errors.Is(err, ErrRootlessDockerNotFoundHomeRunDir),
+		errors.Is(err, ErrRootlessDockerNotFoundHomeDesktopDir),
+		errors.Is(err, ErrRootlessDockerNotFoundRunDir):
+		return true
+	default:
+		return false
+	}
 }
 
 // dockerHostFromEnv returns the docker host from the DOCKER_HOST environment variable, if it's not empty
--- a/internal/core/docker_host.go
+++ b/internal/core/docker_host.go
@@ -59,7 +59,7 @@ func DefaultGatewayIP() (string, error) {
 // defaultCallbackCheckFn Use a vanilla Docker client to check if the Docker host is reachable.
 // It will avoid recursive calls to this function.
 var defaultCallbackCheckFn = func(ctx context.Context, host string) error {
-	cli, err := client.NewClientWithOpts(client.FromEnv, client.WithHost(host))
+	cli, err := client.NewClientWithOpts(client.FromEnv, client.WithHost(host), client.WithAPIVersionNegotiation())
 	if err != nil {
 		return err
 	}
--- a/internal/core/docker_host.go
+++ b/internal/core/docker_host.go
@@ -116,22 +116,6 @@ func extractDockerHost(ctx context.Context) string {
 			continue
 		}
 
-		// check if the docker host responds to an Info request
-		// we are going to use the default Docker client to check if the host is reachable
-		// which will avoid recursive calls to this function
-		cli, err := client.NewClientWithOpts(client.WithHost(dockerHost))
-		if err != nil {
-			outerErr = fmt.Errorf("%w: %w", outerErr, err)
-			continue
-		}
-		defer cli.Close()
-
-		_, err = cli.Info(ctx)
-		if err != nil {
-			outerErr = fmt.Errorf("%w: %w", outerErr, err)
-			continue
-		}
-
 		return dockerHost
 	}
 
@@ -157,7 +141,8 @@ func extractDockerSocket(ctx context.Context) string {
 // and receiving an instance of the Docker API client interface.
 // This internal method is handy for testing purposes, passing a mock type simulating the desired behaviour.
 func extractDockerSocketFromClient(ctx context.Context, cli client.APIClient) string {
-	// check that the socket is not a tcp or unix socket
+	// check that the socket is not a tcp or unix socket,
+	// and that it is reachable.
 	checkDockerSocketFn := func(socket string) string {
 		// this use case will cover the case when the docker host is a tcp socket
 		if strings.HasPrefix(socket, TCPSchema) {
@@ -166,6 +151,11 @@ func extractDockerSocketFromClient(ctx context.Context, cli client.APIClient) st
 
 		if strings.HasPrefix(socket, DockerSocketSchema) {
 			return strings.Replace(socket, DockerSocketSchema, "", 1)
+		}
+
+		_, err := cli.Info(ctx)
+		if err != nil {
+			return DockerSocketPath
 		}
 
 		return socket
--- a/internal/core/docker_host.go
+++ b/internal/core/docker_host.go
@@ -83,10 +83,15 @@ var defaultCallbackCheckFn = func(ctx context.Context, host string) error {
 //  4. Docker host from the default docker socket path, without the unix schema.
 //  5. Docker host from the "docker.host" property in the ~/.testcontainers.properties file.
 //  6. Rootless docker socket path.
-//  7. Else, the default Docker socket including schema will be returned.
+//  7. Else, because the Docker host is not set, it panics.
 func ExtractDockerHost(ctx context.Context) string {
 	dockerHostOnce.Do(func() {
-		dockerHostCache = extractDockerHost(ctx)
+		cache, err := extractDockerHost(ctx)
+		if err != nil {
+			panic(err)
+		}
+
+		dockerHostCache = cache
 	})
 
 	return dockerHostCache
@@ -115,7 +120,7 @@ func ExtractDockerSocket(ctx context.Context) string {
 
 // extractDockerHost Extracts the docker host from the different alternatives, without caching the result.
 // This internal method is handy for testing purposes.
-func extractDockerHost(ctx context.Context) string {
+func extractDockerHost(ctx context.Context) (string, error) {
 	dockerHostFns := []func(context.Context) (string, error){
 		testcontainersHostFromProperties,
 		dockerHostFromEnv,
@@ -137,11 +142,10 @@ func extractDockerHost(ctx context.Context) string {
 			continue
 		}
 
-		return dockerHost
+		return dockerHost, nil
 	}
 
-	// We are not supporting Windows containers at the moment
-	return DockerSocketPathWithSchema
+	return "", outerErr
 }
 
 // extractDockerSocket Extracts the docker socket from the different alternatives, without caching the result.
@@ -200,7 +204,10 @@ func extractDockerSocketFromClient(ctx context.Context, cli client.APIClient) st
 		return DockerSocketPath
 	}
 
-	dockerHost := extractDockerHost(ctx)
+	dockerHost, err := extractDockerHost(ctx)
+	if err != nil {
+		panic(err) // Docker host is required to get the Docker socket
+	}
 
 	return checkDockerSocketFn(dockerHost)
 }
--- a/internal/core/docker_host.go
+++ b/internal/core/docker_host.go
@@ -133,9 +133,7 @@ func extractDockerHost(ctx context.Context) string {
 			continue
 		}
 
-		err = defaultCallbackCheckFn(ctx, dockerHost)
-		if err != nil {
-			outerErr = fmt.Errorf("%w: %w", outerErr, err)
+		if err = defaultCallbackCheckFn(ctx, dockerHost); err != nil {
 			continue
 		}
 
--- a/internal/core/docker_host.go
+++ b/internal/core/docker_host.go
@@ -116,22 +116,6 @@ func extractDockerHost(ctx context.Context) string {
 			continue
 		}
 
-		// check if the docker host responds to an Info request
-		// we are going to use the default Docker client to check if the host is reachable
-		// which will avoid recursive calls to this function
-		cli, err := client.NewClientWithOpts(client.WithHost(dockerHost))
-		if err != nil {
-			outerErr = fmt.Errorf("%w: %w", outerErr, err)
-			continue
-		}
-		defer cli.Close()
-
-		_, err = cli.Info(ctx)
-		if err != nil {
-			outerErr = fmt.Errorf("%w: %w", outerErr, err)
-			continue
-		}
-
 		return dockerHost
 	}
 
--- a/internal/core/docker_host.go
+++ b/internal/core/docker_host.go
@@ -56,6 +56,23 @@ func DefaultGatewayIP() (string, error) {
 	return ip, nil
 }
 
+// Use a vanilla Docker client to check if the Docker host is reachable.
+// It will avoid recursive calls to this function.
+var defaultCallbackCheckFn = func(host string) error {
+	cli, err := client.NewClientWithOpts(client.FromEnv, client.WithHost(host))
+	if err != nil {
+		return err
+	}
+
+	_, err = cli.Info(context.Background())
+	if err != nil {
+		return err
+	}
+	defer cli.Close()
+
+	return nil
+}
+
 // ExtractDockerHost Extracts the docker host from the different alternatives, caching the result to avoid unnecessary
 // calculations. Use this function to get the actual Docker host. This function does not consider Windows containers at the moment.
 // The possible alternatives are:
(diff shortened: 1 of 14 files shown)


mdelapenya:
@ash2k could you double check this is fixing #2622 ? Thanks!

internal/core/docker_host.go
```diff
@@ -56,6 +56,23 @@ func DefaultGatewayIP() (string, error) {
 	return ip, nil
 }
 
+// Use a vanilla Docker client to check if the Docker host is reachable.
+// It will avoid recursive calls to this function.
+var defaultCallbackCheckFn = func(host string) error {
```
stevenh, line 61: suggestion: I don't think we need this to be a var.
```suggestion
func defaultCallbackCheck(host string) error {
```

internal/core/docker_host.go
```diff
@@ -67,9 +84,16 @@ func DefaultGatewayIP() (string, error) {
...
 //  7. Else, the default Docker socket including schema will be returned.
-func ExtractDockerHost(ctx context.Context) string {
+func ExtractDockerHost(ctx context.Context, callbackChecks ...func(host string) error) string {
 	dockerHostOnce.Do(func() {
-		dockerHostCache = extractDockerHost(ctx)
+		// the tests can pass a custom callback check to avoid the Docker client creation
+		// and to avoid changing the code base, passing zero callbacks will mean the default one,
+		// which checks that the host is reachable.
+		if len(callbackChecks) == 0 {
+			callbackChecks = append(callbackChecks, defaultCallbackCheckFn)
```
stevenh, line 93: suggestion: if the above is correct then we need to:
```suggestion
			callbackChecks = append(callbackChecks, defaultCallbackCheck)
```

stevenh (changes requested):
Seems like this might have some edge cases, so have asked a few questions.

internal/core/docker_host.go
```diff
@@ -67,9 +84,16 @@ func DefaultGatewayIP() (string, error) {
 //  5. Docker host from the "docker.host" property in the ~/.testcontainers.properties file.
 //  6. Rootless docker socket path.
 //  7. Else, the default Docker socket including schema will be returned.
-func ExtractDockerHost(ctx context.Context) string {
+func ExtractDockerHost(ctx context.Context, callbackChecks ...func(host string) error) string {
```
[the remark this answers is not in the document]
mdelapenya: This code is internal, so won't be used by the client code, that's why I decided to go with a variadic arg and not change the APIs that much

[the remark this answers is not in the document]
stevenh: In which case could we not just force the check vs making it optional, as I couldn't see an place where it was passed in and not the default?

internal/core/docker_host.go
```diff
@@ -109,21 +133,30 @@ func extractDockerHost(ctx context.Context) string {
...
 		dockerHost, err := dockerHostFn(ctx)
 		if err != nil {
 			outerErr = fmt.Errorf("%w: %w", outerErr, err)
 			continue
 		}
 
+		for _, callbackCheck := range callbackChecks {
+			err = callbackCheck(dockerHost)
+			if err != nil {
+				outerErr = fmt.Errorf("%w: %w", outerErr, err)
```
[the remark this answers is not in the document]
mdelapenya: Mmmm you are right. I perfectly remember using that error **bug** while debugging the tests 🤔 

The error contains the paths checked during the lookup of the docker socket. I'd log the error, if not nil, before each return in that function. Thoughts?

internal/core/docker_host.go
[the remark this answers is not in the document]
mdelapenya: It's used in the tests (see comment [here](https://github.com/testcontainers/testcontainers-go/pull/2741/files#diff-a8ee20e4d0804d67968f2493bb05b3ad5265f5ce9456d6d438fbb4f91d18023eR89-R91)):
- https://github.com/testcontainers/testcontainers-go/pull/2741/files#diff-1a24c2a3e6705e9b915bf8fde90c9c3688fc68edc7bafd0528d8db6b3a5fc0dfR6
- https://github.com/testcontainers/testcontainers-go/pull/2741/files#diff-1a24c2a3e6705e9b915bf8fde90c9c3688fc68edc7bafd0528d8db6b3a5fc0dfR447
- https://github.com/testcontainers/testcontainers-go/pull/2741/files#diff-1a24c2a3e6705e9b915bf8fde90c9c3688fc68edc7bafd0528d8db6b3a5fc0dfR90
- https://github.com/testcontainers/testcontainers-go/pull/2741/files#diff-1a24c2a3e6705e9b915bf8fde90c9c3688fc68edc7bafd0528d8db6b3a5fc0dfR63

internal/core/docker_host.go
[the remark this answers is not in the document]
mdelapenya: What are your thoughts about logging the alternate paths used to discover the docker socket? Does it bring value, or none at all?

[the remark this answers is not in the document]
stevenh: What happens if there are no valid paths, shouldn't the user get an error then?

internal/core/docker_host.go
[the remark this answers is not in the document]
stevenh: If its just to enable edge case testing in a very limited amount of cases you could use a variable based function.

This would simplify all round, no argument changes and the loop would be simply:
```go
	for _, dockerHostFn := range dockerHostFns {
		dockerHost, err := dockerHostFn(ctx)
		if err != nil {
			continue
		}

		if err = hostCheck(dockerHost); err != nil {
			continue
		}
	}
```

Then add a little test helper:
```go
// customHostCheck overrides hostCheck used by extractDockerHost for
// the duration of the test.
func customHostCheck(t *testing.T, fn func(host string) error) {
	t.Helper()

	old := hostCheck
	hostCheck= fn
	t.Cleanup(func() {
		hostCheck = oldHostCheck
	})
}
```

```go
	t.Run("Docker Host is extracted just once", func(t *testing.T) {
		expected := "/path/to/docker.sock"
		t.Setenv("DOCKER_HOST", expected)
 		customHostCheck(t, testCallbackCheckPassing)

		host := ExtractDockerHost(context.Background())
		require.Equal(t, expected, host)
	})
```
I commonly use that pattern to override `time.Now()` when running tests.

internal/core/docker_host.go
stevenh: Ignore this if the structure change above makes sense, but would rename as per example to something like `hostCheck`.

internal/core/docker_host.go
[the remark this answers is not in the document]
mdelapenya: We are defaulting to the OS-dependant default docker socket

internal/core/docker_host.go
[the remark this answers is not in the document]
mdelapenya: Ok, so we create just a default hostCheck (the current default callback check function) which we override in tests, right?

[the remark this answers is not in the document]
mdelapenya: @stevenh I addressed this in https://github.com/testcontainers/testcontainers-go/pull/2741/commits/e8eeff38be456f9d531c449dbf5fac6272cff88c

I think we only need to deal with the error 🤔 Because the `ExtractDockerHost` function is internal, I think it's fine to make it return an error, or because testcontainers-go _needs_ a valid socket path, I think we could panic here. 🤔 

I'm not that familiar with how idiomatic is to panic in a library, so I'd ask for advice here.

internal/core/docker_host.go
```diff
@@ -116,14 +133,20 @@ func extractDockerHost(ctx context.Context) string {
 			continue
 		}
 
+		err = defaultCallbackCheckFn(ctx, dockerHost)
+		if err != nil {
+			outerErr = fmt.Errorf("%w: %w", outerErr, err)
```
stevenh, line 138: nit: I would eliminate the assignment to outerErr as if this is for debug we can do that without it, given its not used.
```suggestion
		if err = defaultCallbackCheckFn(ctx, dockerHost); err != nil {
```

stevenh (approved):
Looks good, just one nit, if you're so inclined.

internal/core/docker_host.go
mdelapenya: Oh good catch! I dragged that from the previous state.

About the outerErr, wdyt on https://github.com/testcontainers/testcontainers-go/pull/2741#discussion_r1727202446?

internal/core/docker_host.go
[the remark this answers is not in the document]
stevenh: Yer I would absolutely return an error if we can't find a valid docker host.

internal/core/docker_host.go
```diff
@@ -133,17 +138,14 @@ func extractDockerHost(ctx context.Context) string {
 			continue
 		}
 
-		err = defaultCallbackCheckFn(ctx, dockerHost)
-		if err != nil {
-			outerErr = fmt.Errorf("%w: %w", outerErr, err)
+		if err = defaultCallbackCheckFn(ctx, dockerHost); err != nil {
```
stevenh, line 141: suggestion: As this is now returning an error we can make it more useful by filtering out all the not set errors.

Would be nice if all the errors were related so we could use and errors.As which would work without having to add each case which is fragile in the case that more options are set but, this should work for now.
```go
// isHostNotSet returns true if the error is related to the Docker host
// not being set, false otherwise.
func isHostNotFound(err error) bool {
	switch {
	case errors.Is(err, ErrTestcontainersHostNotSetInProperties),
		errors.Is(err, ErrDockerHostNotSet),
		errors.Is(err, ErrDockerSocketNotSetInContext),
		errors.Is(err, ErrDockerSocketNotSetInProperties),
		errors.Is(err, ErrSocketNotFoundInPath),
		errors.Is(err, ErrDockerSocketNotSetInProperties),
		errors.Is(err, ErrXDGRuntimeDirNotSet),
		errors.Is(err, ErrRootlessDockerNotFoundHomeRunDir),
		errors.Is(err, ErrRootlessDockerNotFoundHomeDesktopDir),
		errors.Is(err, ErrRootlessDockerNotFoundRunDir):
		return true
	default:
		return false
	}
}

// extractDockerHost Extracts the docker host from the different alternatives, without caching the result.
// This internal method is handy for testing purposes.
func extractDockerHost(ctx context.Context) (string, error) {
	dockerHostFns := []func(context.Context) (string, error){
		testcontainersHostFromProperties,
		dockerHostFromEnv,
		dockerHostFromContext,
		dockerSocketPath,
		dockerHostFromProperties,
		rootlessDockerSocketPath,
	}

	var errs []error
	for _, dockerHostFn := range dockerHostFns {
		dockerHost, err := dockerHostFn(ctx)
		if err != nil {
			if !isHostNotFound(err) {
				errs = append(errs, err)
			}
			continue
		}

		if err = defaultCallbackCheckFn(ctx, dockerHost); err != nil {
			errs = append(errs, fmt.Errorf("check host %q: %w", dockerHost, err))
			continue
		}

		return dockerHost, nil
	}

	if len(errs) > 0 {
		return "", errors.Join(errs...)
	}

	return "", ErrSocketNotFound
}

File: internal/core/docker_host.go [hunk 7]
```diff
@@ -202,7 +204,10 @@ func extractDockerSocketFromClient(ctx context.Context, cli client.APIClient) st
 		return DockerSocketPath
 	}
 
-	dockerHost := extractDockerHost(ctx)
+	dockerHost, err := extractDockerHost(ctx)
+	if err != nil {
+		panic(err) // Docker host is required to get the Docker socket
```
stevenh, line 209: bug: This results in `ExtractDockerSocket` panic which is used in other non internal functions including its namesake `ExtractDockerSocket` none of which mention they can panic.

internal/core/docker_host_test.go
```diff
@@ -462,10 +462,17 @@ func TestExtractDockerSocketFromClient(t *testing.T) {
 
 		mockCallbackCheck(t, testCallbackCheckError)
 
+		// check that panic is raised
+		defer func() {
+			if r := recover(); r == nil {
+				t.Errorf("The code did not panic")
+			}
+		}()
+
```
stevenh, line 471: suggestion: require has a Panics helper, reset of the test will need removing.
```suggestion
		require.Panics(t, func() {
			extractDockerSocketFromClient(ctx, mockCli{OS: "Ubuntu"})
		})
```

stevenh (changes requested):

internal/core/docker_host.go
mdelapenya: I like it, thanks for this suggestion. I'm adding it asap

internal/core/docker_host.go
mdelapenya: I added comments to the functions, could you please check that? 🙏

internal/core/docker_host.go
mdelapenya: I think we already addressed this. Resolving.

internal/core/docker_host.go
```diff
@@ -56,6 +56,23 @@ func DefaultGatewayIP() (string, error) {
 	return ip, nil
 }
 
+// defaultCallbackCheckFn Use a vanilla Docker client to check if the Docker host is reachable.
+// It will avoid recursive calls to this function.
+var defaultCallbackCheckFn = func(ctx context.Context, host string) error {
```
stevenh, line 61: nit: should we change the name of this so it better describes its new behaviour, its not a default any more.
```suggestion
// dockerHostCheck validates that host is reachable.
var dockerHostCheck = func(ctx context.Context, host string) error {
```

internal/core/docker_host.go
```diff
@@ -56,6 +56,23 @@ func DefaultGatewayIP() (string, error) {
...
+	cli, err := client.NewClientWithOpts(client.FromEnv, client.WithHost(host), client.WithAPIVersionNegotiation())
+	if err != nil {
+		return err
```
stevenh, line 64: suggestion: wrap the error, so we can understand the source.
```suggestion
		return fmt.Errorf("new client: %w", err)
```

internal/core/docker_host.go
```diff
@@ -56,6 +56,23 @@ func DefaultGatewayIP() (string, error) {
...
+	}
+
+	_, err = cli.Info(ctx)
+	if err != nil {
+		return err
```
stevenh, line 69:
```suggestion
		return fmt.Errorf("host info: %w", err)
```

internal/core/docker_host.go
```diff
@@ -56,6 +56,23 @@ func DefaultGatewayIP() (string, error) {
 	return ip, nil
 }
 
+// defaultCallbackCheckFn Use a vanilla Docker client to check if the Docker host is reachable.
+// It will avoid recursive calls to this function.
+var defaultCallbackCheckFn = func(ctx context.Context, host string) error {
+	cli, err := client.NewClientWithOpts(client.FromEnv, client.WithHost(host), client.WithAPIVersionNegotiation())
+	if err != nil {
+		return err
+	}
```
stevenh, line 65: bug: close defer needs to be earlier
```suggestion
	}
	defer cli.Close()
```

internal/core/docker_host.go
```diff
@@ -56,6 +56,23 @@ func DefaultGatewayIP() (string, error) {
...
+
+	_, err = cli.Info(ctx)
+	if err != nil {
+		return err
+	}
+	defer cli.Close()
```
stevenh, line 71:
```suggestion
```

internal/core/docker_host.go
```diff
@@ -66,10 +83,15 @@ func DefaultGatewayIP() (string, error) {
 //  4. Docker host from the default docker socket path, without the unix schema.
 //  5. Docker host from the "docker.host" property in the ~/.testcontainers.properties file.
 //  6. Rootless docker socket path.
-//  7. Else, the default Docker socket including schema will be returned.
+//  7. Else, because the Docker host is not set, it panics.
```
stevenh, line 86: question: Change the function name to MustExtractDockerHost to indicate this can panic?

internal/core/docker_host.go
```diff
@@ -87,7 +109,7 @@ func ExtractDockerHost(ctx context.Context) string {
 //  5. If the socket contains the unix schema, the schema is removed (e.g. unix:///var/run/docker.sock -> /var/run/docker.sock)
 //  6. Else, the default location of the docker socket is used (/var/run/docker.sock)
 //
-// In any case, if the docker socket schema is "tcp://", the default docker socket path will be returned.
+// It panics if a Docker client cannot be created, or the Docker host cannot be discovered.
```
stevenh, line 112: question: as above MustExtractDockerSocket?

More below too.

internal/core/docker_host.go
```diff
@@ -179,11 +212,33 @@ func extractDockerSocketFromClient(ctx context.Context, cli client.APIClient) st
...
+	if err != nil {
+		panic(err) // Docker host is required to get the Docker socket
+	}
 
 	return checkDockerSocketFn(dockerHost)
 }
 
+// isHostNotSet returns true if the error is related to the Docker host
+// not being set, false otherwise.
+func isHostNotFound(err error) bool {
```
stevenh, line 225: suggestion: this is more not set than than not found.
```suggestion
func isHostNotSet(err error) bool {
```

internal/core/docker_host_test.go
```diff
@@ -40,24 +41,44 @@ var resetSocketOverrideFn = func() {
 	os.Setenv("TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE", originalDockerSocketOverride)
 }
 
+var testCallbackCheckPassing = func(_ context.Context, _ string) error {
+	return nil
+}
+
+var testCallbackCheckError = func(_ context.Context, _ string) error {
+	return fmt.Errorf("could not check the Docker host")
+}
```
stevenh, line 50: suggestion: these don't need to be vars as they are never assigned to and functions are first class.
```suggestion
func testCallbackCheckPassing(_ context.Context, _ string) error {
	return nil
}

func testCallbackCheckError (_ context.Context, _ string) error {
	return fmt.Errorf("could not check the Docker host")
}
```

internal/core/docker_host_test.go
```diff
@@ -409,7 +446,28 @@ func TestExtractDockerSocketFromClient(t *testing.T) {
...
+
+		mockCallbackCheck(t, testCallbackCheckError)
+
+		var socket string
+		require.Panics(t, func() {
+			socket = extractDockerSocketFromClient(ctx, mockCli{OS: "Ubuntu"})
+		})
+
+		// the default Docker host without schema is returned because we cannot any other host
+		require.Equal(t, "", socket)
```
stevenh, line 470: suggestion: socket check is redundant here as the panics would have already failed.
```suggestion
		require.Panics(t, func() {
			extractDockerSocketFromClient(ctx, mockCli{OS: "Ubuntu"})
		})
```

stevenh (changes requested):
I'm wondering if should make a breaking change to either return errors all the way or rename exposed to be "Must" functions, as even with it mentioned in the docs its easy to miss, resulting in unexpected failures.

internal/core/docker_host.go
mdelapenya: Done

internal/core/docker_host.go
mdelapenya: Done

internal/core/docker_host.go
mdelapenya: Makes total sense. I updated the name and deprecated the public API that retrieved the docker socket: See `testcontainers.go:ExtractDockerSocket`

internal/core/docker_rootless_test.go
```diff
@@ -178,14 +178,8 @@ func TestRootlessDockerSocketPath(t *testing.T) {
 		setupRootlessNotFound(t)
 
 		socketPath, err := rootlessDockerSocketPath(context.Background())
-		require.ErrorIs(t, err, ErrRootlessDockerNotFound)
+		require.Error(t, err)
```
stevenh, line 181: suggestion: shouldnt this be:
```suggestion
		require.ErrorIs(t, err, ErrRootlessDockerNotFoundXDGRuntimeDir)
```

testcontainers.go
```diff
@@ -18,9 +23,9 @@ import (
 //  5. If the socket contains the unix schema, the schema is removed (e.g. unix:///var/run/docker.sock -> /var/run/docker.sock)
 //  6. Else, the default location of the docker socket is used (/var/run/docker.sock)
 //
-// In any case, if the docker socket schema is "tcp://", the default docker socket path will be returned.
-func ExtractDockerSocket() string {
-	return core.ExtractDockerSocket(context.Background())
+// It panics if a Docker client cannot be created, or the Docker host cannot be discovered.
+func MustExtractDockerSocket() string {
```
stevenh, line 27: suggestion: as this is new should we pass in the ctx?

stevenh (commented):
Thanks for going the extra mile on this, feels must better to make it explicit.

Just a couple of minor things to consider

stevenh (approved):
LGTM

PR #2744, opened 2024-08-22
break: use a docker image type for modules

mdelapenya:
## What does this PR do?
It adds a new type alias for string to represent Docker image names, which will be used in the signature of all the Run functions in the modules.

All calls to Run with an untyped string will not break, but users of the lib passing a typed string will have to change it to a `testcontainers.DockerImage`.

## Why is it important?
It will allow providing support for the compatible-for feature in testcontainers-java, among other Docker image validations.

### DIFF
--- a/image.go
+++ b/image.go
@@ -4,6 +4,21 @@ import (
 	"context"
 )
 
+// DockerImage represents a Docker image as a string.
+// Valid examples:
+// - "alpine:3.12"
+// - "docker.io/nginx:latest"
+// - "my-registry.local:5000/my-image:my-tag"
+type DockerImage string
+
+func NewDockerImage(image string) DockerImage {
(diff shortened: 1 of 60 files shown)


image.go
```diff
@@ -4,6 +4,21 @@ import (
 	"context"
 )
 
+// DockerImage represents a Docker image as a string.
+// Valid examples:
+// - "alpine:3.12"
+// - "docker.io/nginx:latest"
+// - "my-registry.local:5000/my-image:my-tag"
```
lanwen, line 11: Would be `sha256:abc...` form valid as well? Could you add example to the doc?

mdelapenya:
I'm going to close this PoC, as I'm not working on it for the moment.

PR #2747, opened 2024-08-26
Fix trailing slash on Image Prefix

### LINKED ISSUE #2746: [Bug]: TESTCONTAINERS_HUB_IMAGE_NAME_PREFIX trailing slash not handled properly
### Testcontainers version

0.33.0

Yes

### Host OS

MacOS

### Host arch

x86

### Go version

1.23

### Docker version

```shell
Docker version 27.1.1, build 6312585
```

### Docker info

```shell
N/A
```

### What happened?

If we add a trailing slash to `TESTCONTAINERS_HUB_IMAGE_NAME_PREFIX`, which we do by convention, Test Containers doesn't handle this very well. All containers end up having something like i.e.:`<custom_url>//<image>` which causes it to fail.

### Relevant log output

_No response_

### Additional information

_No response_

driverpt:
## What does this PR do?
Fixes Trailing Slash on Image Prefix.

## Why is it important?
Because we need to manually remove the Trailing Slash specifically for Go Projects. Where we want consistency between Testcontainers Java and Go

## Related issues
- Closes #2746 

## How to test this PR
Unit Test Created

### DIFF
--- a/docker_test.go
+++ b/docker_test.go
@@ -2280,3 +2280,31 @@ func TestDockerProvider_attemptToPullImage_retries(t *testing.T) {
 		})
 	}
 }
+
+func TestCustomPrefixTrailingSlashIsProperlyRemovedIfPresent(t *testing.T) {
+	hubPrefixWithTrailingSlash := "public.ecr.aws/"
+	dockerImage := "amazonlinux/amazonlinux:2023"
+
+	ctx := context.Background()
+	req := ContainerRequest{
+		Image:             dockerImage,
+		ImageSubstitutors: []ImageSubstitutor{newPrependHubRegistry(hubPrefixWithTrailingSlash)},
+	}
+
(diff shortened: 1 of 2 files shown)


mdelapenya:
Hi @driverpt thanks for contributing the fix! I'm curious about the use case: from what I read from your message, you have projects in Java and in Go, using testcontainers, right? Can you share more about the configuration used in both? I guess you're setting the hub prefix in the properties file, and because it's shared across languages, it fails for Go. Is this assumption correct?

mdelapenya:
Yeah, sure, I'm not questioning the fix: I added the test myself with the current head and I saw it failing, so thanks for contributing the test too!

I'm just interested in how testcontainers is used across a set multi-language projects.

mdelapenya (approved):
LGTM, thanks!

PR #2758, opened 2024-08-31
feat(compose): select services via profiles

ngrash:
## What does this PR do?
This PR adds the option to use [Compose profiles](https://docs.docker.com/compose/profiles) to enable/disable services based on the profiles defined in the compose file:

```golang
func WithProfiles(profiles ...string) ComposeStackOption
```

The current behavior is not changed unless `WithProfiles` is used.
## Why is it important?
We would like to use our existing Compose files with `testcontainers-go/modules/compose` to set up test environments locally and in our pipelines.

However, our workflow relies on Compose profiles, which currently makes the use of `testcontainers-go/modules/compose` impossible, as only services without profiles are enabled by default.

Our workaround is to set up the test environment via the Compose CLI, but we think this could be well automated if `testcontainers-go` supported Compose profiles.

## Related issues
- Relates testcontainers/testcontainers-java#5041 where as a workaround, `withOptions("--profiles foo")` is suggested. This is not possible with `testcontainers-go` as it relies directly on the Go module and does not wrap the CLI.
## How to test this PR
To only run the tests for this individual feature, you can run `go test -v -run TestDockerComposeAPIWithProfiles` in `modules/compose`.

In addition to the unit tests, reviewers can try their own Compose files with profiles:
```golang
func ExampleWithProfiles() error {
    ctx := context.Background()

    filesOpt := compose.WithStackFiles("compose.yaml")
    profilesOpt := compose.WithProfiles("test", "dev")
    stack, err := compose.NewDockerComposeWith(filesOpt, profilesOpt)
    if err != nil {
       return fmt.Errorf("new docker compose: %w", err)
    }
    fmt.Println("stack up")
    if err := stack.Up(ctx); err != nil {
       return fmt.Errorf("stack up: %w", err)
    }
    fmt.Println("stack down")
    if err := stack.Down(ctx); err != nil {
       return fmt.Errorf("stack down: %w", err)
    }
    return nil
}
```

## Follow-ups
From my perspective, this feature is self-contained.

If this gets merged and released, we will adapt our workflow to use `testcontainers-go/modules/compose` for setting up test environments. This might generate more insights and follow-ups.

### DIFF
--- a/modules/compose/compose.go
+++ b/modules/compose/compose.go
@@ -32,6 +32,7 @@ type composeStackOptions struct {
 	Paths          []string
 	temporaryPaths map[string]bool
 	Logger         testcontainers.Logging
+	Profiles       []string
 }
 
 type ComposeStackOption interface {
@@ -116,6 +117,11 @@ func WithStackReaders(readers ...io.Reader) ComposeStackOption {
 	return ComposeStackReaders(readers)
 }
 
+// WithProfiles allows to enable/disable services based on the profiles defined in the compose file.
+func WithProfiles(profiles ...string) ComposeStackOption {
+	return ComposeProfiles(profiles)
+}
+
 func NewDockerCompose(filePaths ...string) (*dockerCompose, error) {
 	return NewDockerComposeWith(WithStackFiles(filePaths...))
 }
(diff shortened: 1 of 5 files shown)


mdelapenya (approved):
LGTM, thanks for adding this! The changes are easy to review and the test is always welcomed.

Issue #2765, opened 2024-09-02
[Enhancement]: replace manual mocks


stevenh:
### Proposal

Replace all manual mocks with generated ones, as started by https://github.com/testcontainers/testcontainers-go/pull/2731

mdelapenya:
We must make sure there is a build command to do this 🙏

mdelapenya:
sure, the local build must do exactly the same as in the CI, so that developers can replicate what the CI does. So:
- adding a build command
- the CI uses a command to validate (simply checking the git state??), that can be used by developers to understand if they missed anything

stevenh:
Yer this would be an example of github action to validate:
```yaml
      - name: Install dependencies
        run: |
          go install github.com/vektra/mockery/v2@v2.42.2

      - name: Validate go mod / generate
        run: |
          go generate ./...
          go mod tidy
          git --no-pager diff && [[ 0 -eq $(git status --porcelain | wc -l) ]]
```

mdelapenya:
Sounds good to me. I'm labelling this as a quick-win

jasonyunicorn:
Has anyone stepped up to work on this?  If not, I would like to volunteer as a first-time contributor to this repo!

stevenh:
Yep I have some work in progress on this but ran into some additional inconsistencies. Need to get some time to split it out

PR #2774, opened 2024-09-08
ci: add generate for mocks

stevenh:
Add generation for mocks, which can be run using make generate or go generate ./...

Correct output-format for golangci-lint as githubs-actions has been replaced by colored-line-numbers.

Add pre-commit target to help check code is valid before pushing.

Implements: #2765

### DIFF
--- a/.github/workflows/ci-test-go.yml
+++ b/.github/workflows/ci-test-go.yml
@@ -85,13 +85,25 @@ jobs:
           #           takes precedence over all other caching options.
           skip-cache: true
 
+      - name: generate
+        if: ${{ inputs.platform == 'ubuntu-latest' }}
+        working-directory: ./${{ inputs.project-directory }}
+        shell: bash
+        run: |
+          make generate
+          git --no-pager diff && [[ 0 -eq $(git status --porcelain | wc -l) ]]
+
       - name: modVerify
         working-directory: ./${{ inputs.project-directory }}
         run: go mod verify
 
       - name: modTidy
+        if: ${{ inputs.platform == 'ubuntu-latest' }}
         working-directory: ./${{ inputs.project-directory }}
-        run: make tidy
+        shell: bash
+        run: |
+          make tidy
+          git --no-pager diff && [[ 0 -eq $(git status --porcelain | wc -l) ]]
 
       - name: ensure compilation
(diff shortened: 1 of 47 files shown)


.github/workflows/ci-test-go.yml
```diff
@@ -85,6 +85,10 @@ jobs:
           #           takes precedence over all other caching options.
           skip-cache: true
 
+      - name: generate
+        working-directory: ./${{ inputs.project-directory }}
+        run: make generate
```
mdelapenya, line 90: I think we want to verify there are no modified files after this step

mdelapenya (commented):
LGTM, just a comment regarding checking if there are modified files on CI

.github/workflows/ci-test-go.yml
stevenh: Yep, saw you had mod tidy in there and assumed it was doing that, but on checking it's not, so added to both.

.github/workflows/ci-test-go.yml
```diff
@@ -85,13 +85,23 @@ jobs:
...
         working-directory: ./${{ inputs.project-directory }}
         run: go mod verify
 
       - name: modTidy
         working-directory: ./${{ inputs.project-directory }}
-        run: make tidy
+        shell: bash
+        run: |
+          make tidy
+          git --no-pager diff && [[ 0 -eq $(git status --porcelain | wc -l) ]]
```
mdelapenya, line 104: Neat! This will avoid any mistakes on CI. Thanks!

mdelapenya (approved):
LGTM, thanks!

Waiting for the CI to merge.

As a follow-up, we are going to replace the custom mocks from the wait strategies with this approach, right?

mdelapenya (dismissed):

stevenh:
> LGTM, thanks!
> 
> Waiting for the CI to merge.
> 
> As a follow-up, we are going to replace the custom mocks from the wait strategies with this approach, right?

Yep already have some [work in progress here](https://github.com/testcontainers/testcontainers-go/pull/2774)

stevenh:
Unrelated failure, should be good now @mdelapenya

mdelapenya (approved):

PR #2777, opened 2024-09-11
feat(mongodb): Wait for mongodb module with a replicaset to finish

smgt:
## What does this PR do?

This will add a test for writing to the mongodb module replica set so we can catch the error `(NotWritablePrimary) not primary`. It also adds a new wait strategy when using `mongodb.WithReplicaSet()` that waits for replica set status to be available.

## Why is it important?

We get flaky tests in our CI (seems to be present on slow hardware) and they fail sometimes since we try to write to a mongodb container which isn't ready.

## How to test this PR

There are tests added to test for this error. Tests will randomly fail when the wait strategy in `WithReplicaSet` is removed.

### DIFF
--- a/modules/mongodb/mongodb.go
+++ b/modules/mongodb/mongodb.go
@@ -3,6 +3,7 @@ package mongodb
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/testcontainers/testcontainers-go"
 	"github.com/testcontainers/testcontainers-go/wait"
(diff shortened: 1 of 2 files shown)


modules/mongodb/mongodb.go
```diff
@@ -87,6 +88,10 @@ func WithPassword(password string) testcontainers.CustomizeRequestOption {
 func WithReplicaSet(replSetName string) testcontainers.CustomizeRequestOption {
 	return func(req *testcontainers.GenericContainerRequest) error {
 		req.Cmd = append(req.Cmd, "--replSet", replSetName)
+		req.WaitingFor = wait.ForAll(
+			wait.ForExec(eval("rs.status().ok")),
+			wait.ForListeningPort("27017/tcp"),
+		).WithDeadline(60 * time.Second)
```
stevenh, line 94: question: is the intent here to replace the existing or to merge? Also the order seems reversed I would expect the listen check before the exec?
```suggestion
		req.WaitingFor = wait.ForAll(
			req.WaitingFor,
			wait.ForExec(eval("rs.status().ok")),
		).WithDeadline(60 * time.Second)
```

smgt: That is much smarter! I imagine that the best solution is to extend the default strategies since this should basically be a extra wait strategy that is applied after the other ones. I will update the PR and thanks for the solution! 😁

smgt:
@mdelapenya 

I've updated the PR now so it uses testify's require module.

mdelapenya (approved):
LGTM, thanks! Waiting for the CI to pass for merging it.

Great job!

stevenh (approved):

PR #2778, opened 2024-09-12
fix: handle 127 error code for podman compatibility

vchandela:
## What does this PR do?

- Handle 127 error code for podman compatibility.
- After merging this, we still need to disable ryuk for tests to work with Podman. Raised an issue for this: https://github.com/testcontainers/testcontainers-go/issues/2781

## Why is it important?

- Docker has a bug where it overrides exit code = 127 with exit code = 126. However, Podman handles these error codes separately.
- While testing etcd module, my container was returning exit code = 127 on Podman. test-containers.go didn't know what to do with 127 error code as it follows the moby project so, it was stuck in an infinite loop.
- Although, the container was up but the code was panicking.

## Related issues

- Some relevant PRs in other projects:
  - https://github.com/containers/podman/issues/367
  - https://github.com/moby/moby/issues/45795

## How to test this PR

- Output for `podman exec -ti <containerid> /bin/sh -c "echo test"; echo $?`

```
podman exec -ti 3604e691a9efbaa916bdeaa365f7481a04aca79722c430d3cc285b671215cfec /bin/sh -c "echo test"; echo $?
Error: crun: executable file `/bin/sh` not found in $PATH: No such file or directory: OCI runtime attempted to invoke a command that was not found
127
```

### DIFF
--- a/wait/host_port.go
+++ b/wait/host_port.go
@@ -24,8 +24,10 @@ var (
 	_ StrategyTimeout = (*HostPortStrategy)(nil)
 )
 
-var errShellNotExecutable = errors.New("/bin/sh command not executable")
-var errShellNotFound = errors.New("/bin/sh command not found")
+var (
+	errShellNotExecutable = errors.New("/bin/sh command not executable")
+	errShellNotFound      = errors.New("/bin/sh command not found")
+)
 
 type HostPortStrategy struct {
 	// Port is a string containing port number and protocol in the format "80/tcp"
@@ -167,7 +169,7 @@ func (hp *HostPortStrategy) WaitUntilReady(ctx context.Context, target StrategyT
 		default:
 			return fmt.Errorf("internal check: %w", err)
 		}
-	}	
+	}
 
 	return nil
 }
--- a/wait/host_port.go
+++ b/wait/host_port.go
@@ -13,6 +13,13 @@ import (
 	"github.com/docker/go-connections/nat"
 )
 
+type exitStatus = int
+
+const (
+	exitEaccess     exitStatus = 126 // container cmd can't be invoked (permission denied)
+	exitCmdNotFound exitStatus = 127 // container cmd not found/does not exist or invalid bind-mount
+)
+
 // Implement interface
 var (
 	_ Strategy        = (*HostPortStrategy)(nil)
@@ -20,6 +27,7 @@ var (
 )
 
 var errShellNotExecutable = errors.New("/bin/sh command not executable")
+var errShellNotFound = errors.New("/bin/sh command not found")
 
 type HostPortStrategy struct {
 	// Port is a string containing port number and protocol in the format "80/tcp"
@@ -207,10 +215,16 @@ func internalCheck(ctx context.Context, internalPort nat.Port, target StrategyTa
 			return fmt.Errorf("%w, host port waiting failed", err)
 		}
 
+		/*
+			Docker maps exit code 127 to 126, while Podman treats them separately.
+			Handle both to ensure compatibility with Docker and Podman.
+		*/
 		if exitCode == 0 {
 			break
-		} else if exitCode == 126 || exitCode == 127 {
+		} else if exitCode == exitEaccess {
 			return errShellNotExecutable
+		} else if exitCode == exitCmdNotFound {
+			return errShellNotFound
 		}
 	}
 	return nil
--- a/wait/host_port.go
+++ b/wait/host_port.go
@@ -13,11 +13,9 @@ import (
 	"github.com/docker/go-connections/nat"
 )
 
-type exitStatus = int
-
 const (
-	exitEaccess     exitStatus = 126 // container cmd can't be invoked (permission denied)
-	exitCmdNotFound exitStatus = 127 // container cmd not found/does not exist or invalid bind-mount
+	exitEaccess     = 126 // container cmd can't be invoked (permission denied)
+	exitCmdNotFound = 127 // container cmd not found/does not exist or invalid bind-mount
 )
 
 // Implement interface
@@ -159,13 +157,17 @@ func (hp *HostPortStrategy) WaitUntilReady(ctx context.Context, target StrategyT
 	}
 
 	if err = internalCheck(ctx, internalPort, target); err != nil {
-		if errors.Is(errShellNotExecutable, err) {
+		switch {
+		case errors.Is(err, errShellNotExecutable):
 			log.Println("Shell not executable in container, only external port validated")
 			return nil
+		case errors.Is(err, errShellNotFound):
+			log.Println("Shell not found in container")
+			return nil
+		default:
+			return fmt.Errorf("internal check: %w", err)
 		}
-
-		return fmt.Errorf("internal check: %w", err)
-	}
+	}	
 
 	return nil
 }
@@ -215,19 +217,18 @@ func internalCheck(ctx context.Context, internalPort nat.Port, target StrategyTa
 			return fmt.Errorf("%w, host port waiting failed", err)
 		}
 
-		/*
-			Docker maps exit code 127 to 126, while Podman treats them separately.
-			Handle both to ensure compatibility with Docker and Podman.
-		*/
-		if exitCode == 0 {
-			break
-		} else if exitCode == exitEaccess {
+		// Docker has a issue which override exit code 127 to 126 due to:
+		// https://github.com/moby/moby/issues/45795
+		// Handle both to ensure compatibility with Docker and Podman for now.
+		switch exitCode {
+		case 0:
+			return nil
+		case exitEaccess:
 			return errShellNotExecutable
-		} else if exitCode == exitCmdNotFound {
+		case exitCmdNotFound:
 			return errShellNotFound
 		}
 	}
-	return nil
 }
 
 func buildInternalCheckCommand(internalPort int) string {
--- a/wait/host_port.go
+++ b/wait/host_port.go
@@ -209,7 +209,7 @@ func internalCheck(ctx context.Context, internalPort nat.Port, target StrategyTa
 
 		if exitCode == 0 {
 			break
-		} else if exitCode == 126 {
+		} else if exitCode == 126 || exitCode == 127 {
 			return errShellNotExecutable
 		}
 	}
--- a/wait/host_port_test.go
+++ b/wait/host_port_test.go
@@ -456,16 +456,12 @@ func TestHostPortStrategyFailsWhileInternalCheckingDueToUnexpectedContainerStatu
 
 func TestHostPortStrategySucceedsGivenShellIsNotInstalled(t *testing.T) {
 	listener, err := net.Listen("tcp", "localhost:0")
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err)
 	defer listener.Close()
 
 	rawPort := listener.Addr().(*net.TCPAddr).Port
 	port, err := nat.NewPort("tcp", strconv.Itoa(rawPort))
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err)
 
 	target := &MockStrategyTarget{
 		HostImpl: func(_ context.Context) (string, error) {
@@ -511,16 +507,12 @@ func TestHostPortStrategySucceedsGivenShellIsNotInstalled(t *testing.T) {
 
 func TestHostPortStrategySucceedsGivenShellIsNotFound(t *testing.T) {
 	listener, err := net.Listen("tcp", "localhost:0")
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err)
 	defer listener.Close()
 
 	rawPort := listener.Addr().(*net.TCPAddr).Port
 	port, err := nat.NewPort("tcp", strconv.Itoa(rawPort))
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err)
 
 	target := &MockStrategyTarget{
 		HostImpl: func(_ context.Context) (string, error) {
--- a/wait/host_port_test.go
+++ b/wait/host_port_test.go
@@ -1,8 +1,10 @@
 package wait
 
 import (
+	"bytes"
 	"context"
 	"io"
+	"log"
 	"net"
 	"strconv"
 	"testing"
@@ -501,8 +503,17 @@ func TestHostPortStrategySucceedsGivenShellIsNotInstalled(t *testing.T) {
 		WithStartupTimeout(5 * time.Second).
 		WithPollInterval(100 * time.Millisecond)
 
+	oldWriter := log.Default().Writer()
+	var buf bytes.Buffer
+	log.Default().SetOutput(&buf)
+	t.Cleanup(func() {
+		log.Default().SetOutput(oldWriter)
+	})
+
 	err = wg.WaitUntilReady(context.Background(), target)
 	require.NoError(t, err)
+
+	require.Contains(t, buf.String(), "Shell not executable in container, only external port validated")
 }
 
 func TestHostPortStrategySucceedsGivenShellIsNotFound(t *testing.T) {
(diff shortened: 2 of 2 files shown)


wait/host_port.go
```diff
@@ -209,7 +209,7 @@ func internalCheck(ctx context.Context, internalPort nat.Port, target StrategyTa
 
 		if exitCode == 0 {
 			break
-		} else if exitCode == 126 {
+		} else if exitCode == 126 || exitCode == 127 {
```
stevenh, line 212: suggestion: Could we add a comment about why we have multiple values and extract them to constants so they are not magic numbers any more.
126 is also used in `host_port_test.go` too.

You can find the details of what they [represent here](https://github.com/moby/moby/blob/b3c775059314039807862defe9f531c2bca049db/daemon/errors.go#L122).

stevenh (changes requested):

wait/host_port.go
vchandela: Done.

wait/host_port.go
```diff
@@ -207,10 +215,16 @@ func internalCheck(ctx context.Context, internalPort nat.Port, target StrategyTa
...
+			Docker maps exit code 127 to 126, while Podman treats them separately.
+			Handle both to ensure compatibility with Docker and Podman.
+		*/
 		if exitCode == 0 {
 			break
-		} else if exitCode == 126 {
+		} else if exitCode == exitEaccess {
 			return errShellNotExecutable
+		} else if exitCode == exitCmdNotFound {
+			return errShellNotFound
```
stevenh, line 227: suggestion: For multiple cases like this a select is easer to follow. The return nil at the end would need to be removed too.

This won't apply cleanly because github doesn't allow suggestions to unedited code:
```suggestion
		select exitCode {
		case 0:
			return nil
		case exitEaccess:
			return errShellNotExecutable
		case exitCmdNotFound:
			return errShellNotFound
		}
```

stevenh, line 227: This would also need to be handled by callers, so we'll need additional updates.

wait/host_port.go
```diff
@@ -13,13 +13,21 @@ import (
 	"github.com/docker/go-connections/nat"
 )
 
+type exitStatus = int
+
+const (
+	exitEaccess     exitStatus = 126 // container cmd can't be invoked (permission denied)
+	exitCmdNotFound exitStatus = 127 // container cmd not found/does not exist or invalid bind-mount
+)
```
stevenh, line 21: suggestion: As we're not using the type here because its not our data, this can be removed.
```suggestion
const (
	exitEaccess             = 126 // container cmd can't be invoked (permission denied)
	exitCmdNotFound = 127 // container cmd not found/does not exist or invalid bind-mount
)
```

stevenh (changes requested):
Thanks for these improvements, some more suggestions

wait/host_port.go
vchandela: Thanks for pointing out.

wait/host_port.go
vchandela: Done.

stevenh (changes requested):
Looking good, we just need to handle the new error on return from the internalCheck function and we should be good

vchandela:
> Looking good, we just need to handle the new error on return from the internalCheck function and we should be good

do we also need to write a test in host_port_test.go for exitCode = 127?

mdelapenya:
> do we also need to write a test in host_port_test.go for exitCode = 127?

Yes please, once added, I think this PR is good to merge, thanks!

wait/host_port_test.go
```diff
@@ -508,3 +508,58 @@ func TestHostPortStrategySucceedsGivenShellIsNotInstalled(t *testing.T) {
 	err = wg.WaitUntilReady(context.Background(), target)
 	require.NoError(t, err)
 }
+
+func TestHostPortStrategySucceedsGivenShellIsNotFound(t *testing.T) {
```
mdelapenya, line 512: cool, thanks for this! Do you think it's possible to add a test for the internal function so that we verify the new error is raised?

wait/host_port_test.go
```diff
@@ -508,3 +508,58 @@ func TestHostPortStrategySucceedsGivenShellIsNotInstalled(t *testing.T) {
...
+	listener, err := net.Listen("tcp", "localhost:0")
+	if err != nil {
+		t.Fatal(err)
+	}
```
stevenh, line 516: suggestion: could we use require for all checks please, just makes the tests much cleaner, for example:
```suggestion
	require.NoError(t, err)
```

wait/host_port_test.go
```diff
@@ -508,3 +508,58 @@ func TestHostPortStrategySucceedsGivenShellIsNotInstalled(t *testing.T) {
...
+	defer listener.Close()
+
+	rawPort := listener.Addr().(*net.TCPAddr).Port
+	port, err := nat.NewPort("tcp", strconv.Itoa(rawPort))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	target := &MockStrategyTarget{
```
stevenh, line 525: suggestion: Could we use the new generated mock `mockStrategyTarget` for this, as we're in the process of moving across to it. You can see an [example in this PR](https://github.com/testcontainers/testcontainers-go/pull/2773).

wait/host_port_test.go
```diff
@@ -508,3 +508,58 @@ func TestHostPortStrategySucceedsGivenShellIsNotInstalled(t *testing.T) {
...
+		},
+	}
+
+	wg := NewHostPortStrategy("80").
+		WithStartupTimeout(5 * time.Second).
+		WithPollInterval(100 * time.Millisecond)
+
+	err = wg.WaitUntilReady(context.Background(), target)
+	require.NoError(t, err)
+}
```
stevenh, line 565: suggestion: missing new line at the end of file.

stevenh (changes requested):

kiview:
Just to double check, if https://github.com/moby/moby/issues/45795 would get fixed in the upstream, this PR would still be valid, correct?

stevenh:
> Just to double check, if [moby/moby#45795](https://github.com/moby/moby/issues/45795) would get fixed in the upstream, this PR would still be valid, correct?

Technically it should only check 127, as 126 is an access issue which should not happen for `/bin/sh` so it could hide an issue if for some reason `/bin` has the wrong permissions preventing the internal check from being successful.

wait/host_port_test.go
vchandela: done. `require` sounds much better for sure.

wait/host_port_test.go
vchandela: done.

wait/host_port_test.go
vchandela: @stevenh it would be nice if you could modify the newly added tests in your PR itself as you have context of the changes. This will help in keeping similar changes in the same PR. wdyt?

wait/host_port_test.go
vchandela: I've done it for all the test files.
We can also substitute some conditions with `require.Contains()` but maybe we can pick it up in future PRs. Otherwise, this PR will be too bulky.

wait/host_port_test.go
stevenh: I can but I'm trying to avoid a moving target and the functionality already exists, its just not in common use yet.

docker_files_test.go
```diff
@@ -19,14 +19,10 @@ func TestCopyFileToContainer(t *testing.T) {
 
 	// copyFileOnCreate {
 	absPath, err := filepath.Abs(filepath.Join(".", "testdata", "hello.sh"))
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err)
```
stevenh, line 22: suggestion: sorry the intent was to only use require for the new test, all these changes will cause a bunch of conflicts with other PR's which are inflight, so would we revert all of these and just use it for your new test.

Sorry to be be a pain, but this would cause quit a lot of conflicts.

stevenh (changes requested):
Sorry I think there was a miss understanding about my comment to use require, the intent was only for the new test, not to do it globally, which while desired could cause a lot of conflicts with other PRs that are currently in flight.

docker_files_test.go
vchandela: Cool. Made the changes.

wait/host_port_test.go
vchandela: some `{}_test.go` files in `/wait` have package = `wait_test` while others have only `wait`. What convention do we want to follow?
This causes unnecessary confusion as `MockStrategyTarget` is in package `wait` while `mockStrategyTarget` is in `wait_test`.

vchandela: I don't understand how `MockStrategyTarget` will be migrated to `mockStrategyTarget`. Could you share a sample for it?
MockStrategyTarget was quite easy to understand to be honest. I'm not too sure which example you're talking about in this PR: https://github.com/testcontainers/testcontainers-go/pull/2773

stevenh: You can see an example in the wait file tests here https://github.com/testcontainers/testcontainers-go/blob/main/wait/file_test.go

vchandela: Done.

wait/host_port.go
```diff
@@ -193,7 +203,7 @@ func externalCheck(ctx context.Context, ipAddress string, port nat.Port, target
 	}
 }
 
-func internalCheck(ctx context.Context, internalPort nat.Port, target StrategyTarget) error {
+func InternalCheck(ctx context.Context, internalPort nat.Port, target StrategyTarget) error {
```
stevenh, line 206: bug: this shouldn’t be exported it’s a private method

wait/host_port_test.go
```diff
@@ -1,18 +1,25 @@
-package wait
+package wait_test
```
stevenh, line 1: suggestion: no need to make this change. We will do that separately as it’s just adding unnecessary changes

wait/host_port_test.go
```diff
@@ -456,55 +463,166 @@ func TestHostPortStrategyFailsWhileInternalCheckingDueToUnexpectedContainerStatu
...
-	wg := NewHostPortStrategy("80").
+	wg := wait.NewHostPortStrategy("80").
 		WithStartupTimeout(5 * time.Second).
 		WithPollInterval(100 * time.Millisecond)
 
 	err = wg.WaitUntilReady(context.Background(), target)
 	require.NoError(t, err)
 }
+
+func TestInternalCheckFailsGivenShellIsNotInstalled(t *testing.T) {
```
stevenh, line 544: These are the exception to the rule 🙃sorry to mess you about. If you undo the test package change they will all just work without making the internal check public

wait/host_port_test.go
```diff
@@ -1,18 +1,25 @@
...
 	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/errdefs"
 	"github.com/docker/go-connections/nat"
 	"github.com/stretchr/testify/require"
 
 	"github.com/testcontainers/testcontainers-go/exec"
+	"github.com/testcontainers/testcontainers-go/wait"
+)
+
+var (
```
stevenh, line 21: This looks unrelated

wait/host_port_test.go
```diff
@@ -456,55 +463,166 @@ func TestHostPortStrategyFailsWhileInternalCheckingDueToUnexpectedContainerStatu
...
-			return &types.ContainerState{
-				Running: true,
-			}, nil
+			},
 		},
-		ExecImpl: func(_ context.Context, _ []string, _ ...exec.ProcessOption) (int, io.Reader, error) {
-			// This is the error that would be returned if the shell is not installed.
-			return 126, nil, nil
+	}, nil)
+	target.EXPECT().MappedPort(anyContext, nat.Port("80")).Return(port, nil)
```
stevenh, line 489: suggestion: no need to do all the tests that already in the other PR just the new tests your adding

wait/host_port_test.go
```diff
@@ -456,55 +463,166 @@ func TestHostPortStrategyFailsWhileInternalCheckingDueToUnexpectedContainerStatu
...
+					},
+				},
+			},
+		},
+	}, nil)
+	target.EXPECT().MappedPort(anyContext, nat.Port("80")).Return(port, nil)
+	target.EXPECT().State(anyContext).Return(&types.ContainerState{
+		Running: true,
+	}, nil)
+	target.EXPECT().Exec(anyContext, []string{"/bin/sh", "-c", "true && (\n\t\t\t\t\tcat /proc/net/tcp* | awk '{print $2}' | grep -i :0050 ||\n\t\t\t\t\tnc -vz -w 1 localhost 80 ||\n\t\t\t\t\t/bin/sh -c '</dev/tcp/localhost/80'\n\t\t\t\t)\n\t\t\t\t"}).Return(wait.ExitEaccess, nil, nil)
```
stevenh, line 573: suggestion: you don’t need to specify the exact args when they are complex like this just use an any of type

stevenh (changes requested):

wait/host_port_test.go
vchandela: If we don't do this, then we'll have to export `mockStrategyTarget` as it is `wait_test` while our tests were originally in `wait`. So, I moved `host_port_test.go` to `wait_test` and exported the necessary variables from `host_port.go`.

wait/host_port.go
vchandela: This is needed to avoid exporting `mockStrategyTarget `. I've explained it better here: https://github.com/testcontainers/testcontainers-go/pull/2778#discussion_r1760575891

wait/host_port_test.go
vchandela: After moving `host_port_test.go` to `wait_test`, we need to import `wait` package to get `Internal()`, `ErrPortNotFound`, etc

wait/host_port_test.go
vchandela: I'll remove this once other conversations are resolved.

wait/host_port_test.go
vchandela: makes sense. I'll incorporate it with the other changes.

wait/host_port.go
stevenh: Sorry we can't export an internal function just be support testing.

wait/host_port_test.go
stevenh: This could be testing using the end to end instead of direct call to the method.

wait/host_port_test.go
stevenh: You never want to export functions to test like this. There are two options, do a end to end which exercises these cases or a seperate file which which is in the `wait` packages.

stevenh: As it uses the logger you can capture the output in the test and assert the content. e.g.
```go
func MyTest(t *testing.T) {
	oldWriter := log.Writer()
	var buf bytes.Buffer
	log.SetWriter(&buf)
	t.Cleanup(func() {
		log.SetWriter(oldWriter)
	})

	// Run the test
	require.Contains(t, buf.String(), "Shell not found in container")
}
```

wait/host_port.go
vchandela: done.

wait/host_port_test.go
vchandela: @stevenh made the changes. PTAL. I've merged internal check tests with the existing tests.
Also, I've reverted back to `MockStrategyTarget` as it will lead to unnecessary code changes for importing certain variables and methods. Please handle `MockStrategyTarget` -> `mockStrategyTarget` for this test in your draft PR.

wait/host_port_test.go
vchandela: I've removed these tests and merged them with the existing ones.

mdelapenya:
@vchandela can you run `make lint` so that all files are gofumpt-ed? 🙏

vchandela:
> @vchandela can you run `make lint` so that all files are gofumpt-ed? 🙏

Done. Thanks for pointing out.

stevenh (approved):
Looks good to me, thank you for all your effort on the back and forward on this most appreciated.

stevenh:
@vchandela can you check into the CI failures to make sure they aren't fall out from this.

vchandela:
> @vchandela can you check into the CI failures to make sure they aren't fall out from this.

They seem unrelated to this PR. One is for K3s and another is for rabbitmq.

mdelapenya (approved):
LGTM, thanks!

PR #2600, opened 2024-06-21
 feat(postgres): use faster sql.DB instead of docker exec psql for snapshot/restore [rebased for main]

cfstras:
See description of https://github.com/testcontainers/testcontainers-go/pull/2599

This is the same PR, but rebased for main.

The original description follows, just with latest benchmarks

--- 

## What does this PR do?

The postgres module has functions for saving snapshots and restoring them later.
This feature is helpful for integration testing without having to manually clear the database.

This PR speeds up the snapshot/restore functionality by executing the necessary SQL commands via a regular `sql.DB` connection.

If the driver is not available in the test environment, or the commands fail for some other reason (SSL config, etc), it falls back to the old variant of calling `docker exec`.

## Why is it important?

`docker exec` can be unnecessarily slow, especially when using docker in a VM with Docker Desktop or colima.

On my laptop (MacBook Air M1), each call to `Container.Restore()` takes about 150~200ms.
In a small subset of 20 of our tests, where each test restores the database before starting, with creating the initial container, this brings the total time up to 12 seconds, which is quite a lot for a few small database tests.

After these changes, the same subset of tests now takes 6 seconds.

## Related issues

- Relates to https://github.com/testcontainers/testcontainers-go/pull/2199
- Relates to https://github.com/testcontainers/testcontainers-go/pull/2599

## How to test this PR

```go
cd modules/postgres
go test ./...
```

## Follow-ups

I've also considered caching the DB pool that is created to manage the databases.
After running some benchmark with this code (from branch X), I noticed it does not have a big impact on the whole test suite - my internal test suite does not change in execution time (since the time is dominated by re-connecting the actual test client, which will be terminated when the database gets deleted during the restore).
The module/postgres test suite even gets slower:

cached: https://github.com/testcontainers/testcontainers-go/compare/main...cfstras:postgres-snapshot-sql-cached-rebased-main
not cached: this PR

<details>
  <summary>testcontainers main, sql, not cached: <b>42.227 s ±  0.635</b></summary>
<pre>
➜ hyperfine "go test ./... -count 1"
Benchmark 1: go test ./... -count 1
  Time (mean ± σ):     42.227 s ±  0.635 s    [User: 1.213 s, System: 1.034 s]
  Range (min … max):   41.832 s … 44.006 s    10 runs
</pre>
</details>

<details>
  <summary>testcontainers main, sql, cached: <b>43.207 s ±  1.549</b></summary>
<pre>
➜ hyperfine "go test ./... -count 1"
[...]

PR #2783, opened 2024-09-17
fix(postgres): Apply default snapshot name if no name specified

kiview:
## What does this PR do?

Restores the original behavior of the Postgres snapshot functionality, that would use a default snapshot name, if no snapshot name was specified.

## Why is it important?

Without this change, #2600 is a breaking change, that requires user to make changes to their test code, manually specifying a snapshot name.

## Related issues

- Fixes regression from #2600

## How to test this PR

Multiple test cases for the default and custom snapshot name were added to `postgres_test.TestSnapshot()`.

### DIFF
--- a/modules/postgres/options.go
+++ b/modules/postgres/options.go
@@ -7,11 +7,13 @@ import (
 type options struct {
 	// SQLDriverName is the name of the SQL driver to use.
 	SQLDriverName string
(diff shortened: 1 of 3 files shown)


mdelapenya (approved):
LGTM thanks!

Will wait for the CI for merging it.

PR #2788, opened 2024-09-19
feat: add etcd module

mdelapenya:
## What does this PR do?
This PR adds a Go module for etcd, including clustering support. This module includes options to configure:

- nodes in the cluster
- cluster token
- ~~auto TLS for cluster-node communication~~ _I removed it because it was not critical and causing issues in tests_

I followed https://etcd.io/docs/v3.5/op-guide/clustering/ to accomplish it.

The module exposes methods to talk to the the client and the peer endpoints.

## Why is it important?
A new module for etcd, which could unblock adoption in other technologies (gofiber's etcd storage)


modules/etcd/cmd_test.go
```diff
@@ -0,0 +1,116 @@
...
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := configureCMD(tt.settings)
+			require.Equal(t, tt.want, got)
+		})
+	}
```
stevenh, line 115: suggestion: This works but is actually more code than separating it out, which also has the benefit of editors such as VS Code being able to run a single test from within the editor:
```suggestion
	t.Run("with-cluster-token", func(t *testing.T) {
		got := configureCMD(options{
			Nodes:        []string{"node1", "node2"},
			ClusterToken: "token",
		)
		want := []string{
			"etcd",
			"--name=node1",
			"--initial-advertise-peer-urls=http://node1:2380",
			"--advertise-client-urls=http://node1:2379",
			"--listen-peer-urls=http://0.0.0.0:2380",
			"--listen-client-urls=http://0.0.0.0:2379",
			"--initial-cluster-state=new",
			"--initial-cluster=node1=http://node1:2380,node2=http://node2:2380",
			"--initial-cluster-token=token",
		}
		require.Equal(t, want, got)
	})
```

modules/etcd/etcd.go
```diff
@@ -0,0 +1,266 @@
...
+
+	return clusterOpts, nil
+}
+
+// configureCMD configures the etcd command line arguments, based on the settings provided,
+// in order to create a cluster or a single-node instance.
+func configureCMD(settings options) []string {
+	cmds := []string{"etcd"}
+
+	scheme := "http"
```
stevenh, line 152: question: This looks like a constant as its hard coded everywhere else?

modules/etcd/options.go
```diff
@@ -0,0 +1,89 @@
...
+	ClusterToken   string
+	AdditionalArgs []string
+	MountDataDir   bool
+}
+
+func defaultOptions() options {
+	return options{
+		currentNode:    0,
+		Nodes:          []string{},
+		AdditionalArgs: []string{},
```
stevenh, line 20: suggestion: all these can use default values.
```suggestion
```

modules/etcd/options.go
```diff
@@ -0,0 +1,89 @@
...
+		ClusterToken:   DefaultClusterToken,
+	}
+}
+
+// Compiler check to ensure that Option implements the testcontainers.ContainerCustomizer interface.
+var _ testcontainers.ContainerCustomizer = (Option)(nil)
+
+// Option is an option for the Redpanda container.
```
stevenh, line 28: bug: cut and paste error

modules/etcd/etcd.go
```diff
@@ -0,0 +1,266 @@
...
+		c = &EtcdContainer{Container: container}
+	}
+
+	if err != nil {
+		return c, fmt.Errorf("generic container: %w", err)
+	}
+
+	// only the first node creates the cluster
+	if settings.currentNode == 0 {
+		for i := 1; i < len(settings.Nodes); i++ {
```
stevenh, line 93: suggestion: range over Nodes:
```suggestion
		for i := range settings.Nodes {
```

modules/etcd/etcd.go
```diff
@@ -0,0 +1,266 @@
...
+			clusterOpts = append(clusterOpts, withCurrentNode(i)) // move to the next node
+			clusterNode, err := Run(ctx, req.Image, clusterOpts...)
+			if err != nil {
+				return nil, fmt.Errorf("run cluster node: %w", err)
```
stevenh, line 97: bug: always return c so the caller can clean up, is there any benefit to adding the node too?
```suggestion
				return c, fmt.Errorf("run cluster node %d: %w", i, err)
```

modules/etcd/etcd.go
```diff
@@ -0,0 +1,266 @@
...
+			apply(&settings)
+		}
+		if err := opt.Customize(&genericContainerReq); err != nil {
+			return nil, err
+		}
+	}
+
+	// configure data dir
+	if settings.MountDataDir {
+		genericContainerReq.LifecycleHooks = append(genericContainerReq.LifecycleHooks, testcontainers.ContainerLifecycleHooks{
```
stevenh, line 51: suggestion: have this handled by option.

modules/etcd/options.go
```diff
@@ -0,0 +1,89 @@
...
+	return func(o *options) {
+		o.MountDataDir = true
+	}
+}
+
+// WithNodes is an option to set the nodes of the etcd cluster.
+// If only one node is specified, the cluster will be a single-node cluster.
+func WithNodes(nodes ...string) Option {
+	return func(o *options) {
+		o.Nodes = nodes
```
stevenh, line 56: bug: the code above seems to rely on specific naming but this doesn't enforce, should this be a count?

modules/etcd/options.go
```diff
@@ -0,0 +1,89 @@
+package etcd
+
+import (
+	"github.com/testcontainers/testcontainers-go"
+)
+
+type options struct {
```
stevenh, line 7: question: This looks like an anti pattern resulting the processing of options twice, once to set values on a struct and then once to check the values in the struct and perform further actions.

Could each option perform needed actions?

modules/etcd/etcd.go
```diff
@@ -0,0 +1,266 @@
...
+		return "", err
+	}
+
+	return fmt.Sprintf("http://%s:%s", host, port.Port()), nil
+}
+
+// MustClientEndpoint returns the client endpoint for the etcd container.
+// For a cluster, it returns the client endpoint of the first node.
+// This method panics if an error occurs.
+func (c *EtcdContainer) MustClientEndpoint(ctx context.Context) string {
```
stevenh, line 207: question: given these could easily fail do we want to expose `Must` versions?

modules/etcd/etcd.go
```diff
@@ -0,0 +1,266 @@
...
+
+	if err != nil {
+		return c, fmt.Errorf("generic container: %w", err)
+	}
+
+	// only the first node creates the cluster
+	if settings.currentNode == 0 {
+		for i := 1; i < len(settings.Nodes); i++ {
+			clusterOpts = append(clusterOpts, withCurrentNode(i)) // move to the next node
+			clusterNode, err := Run(ctx, req.Image, clusterOpts...)
```
stevenh, line 95: bug: while this will likely result in the correct behaviour I don't think having multiple withCurrentNode on clustOptions is the intent.
```suggestion
			clusterNode, err := Run(ctx, req.Image, append(clusterOpts, withCurrentNode(i))...)
```

stevenh (changes requested):

modules/etcd/etcd.go
mdelapenya: We must start the loop in the second node (first one is the main node in the cluster), therefore the child nodes are added starting from i = 1

modules/etcd/etcd.go
mdelapenya: I'm not sure if I follow: you mean exposing them could lead to more errors?

modules/etcd/options.go
mdelapenya: As discussed offline, we could make the options struct receive the container request, so that it's possible to update both in the functional options pattern:
- the container request, which is the way to modify raw containers
- the options, to modify custom behaviours for the given technology.

modules/etcd/etcd.go
stevenh: Ah missed that

modules/etcd/options.go
stevenh: Building on that options need not be a separate struct they could operate on the container struct itself.

modules/etcd/etcd.go
stevenh: Must is typically resolved for cases which only happen due coding bug, for example [regexp.MustCompile](https://pkg.go.dev/regexp#MustCompile), you have to give it a bad regexp for it to fail. These can fail due to runtime conditions, which in turn would panic a whole app or test run which isn't desirable.

I would only include the non-must versions to prevent the user shooting themselves in the foot by using them.

mdelapenya: Understood.
> These can fail due to runtime conditions, which in turn would panic a whole app or test run which isn't desirable.

modules/etcd/options.go
mdelapenya: Mmmm so you would store the options in the container struct?

stevenh: I would consider it, depends on the use case.

For example if you had an option to enable https, the value is something that would be needed each time one of the endpoint methods were called, I'd put that on the container struct as a private.

modules/etcd/etcd.go
```diff
@@ -0,0 +1,241 @@
...
+	if settings.currentNode == 0 {
+		for i := 1; i < len(settings.Nodes); i++ {
+			// move to the next node
+			clusterNode, err := Run(ctx, req.Image, append(clusterOpts, withCurrentNode(i))...)
+			if err != nil {
+				// return the parent cluster node and the error, so the caller can clean up.
+				return c, fmt.Errorf("run cluster node: %w", err)
+			}
+
+			c.Nodes = append(c.Nodes, clusterNode)
```
stevenh, line 85: suggestion: `clusterNode` could also be valid here, requiring clean up.
```suggestion
			clusterNode, err := Run(ctx, req.Image, append(clusterOpts, withCurrentNode(i))...)
			if clusterNode != nil {
				c.Nodes = append(c.Nodes, clusterNode)
			}
			if err != nil {
				// return the parent cluster node and the error, so the caller can clean up.
				return c, fmt.Errorf("run cluster node: %w", err)
			}
```

modules/etcd/etcd.go
```diff
@@ -0,0 +1,241 @@
...
+
+	"github.com/testcontainers/testcontainers-go"
+	tcnetwork "github.com/testcontainers/testcontainers-go/network"
+)
+
+const (
+	ClientPort          = "2379"
+	PeerPort            = "2380"
+	DataDir             = "/data.etcd"
+	DefaultClusterToken = "mys3cr3ttok3n"
```
stevenh, line 16: question: do these need to be exported? I always try to keep the API surface area as small as possible.

stevenh (changes requested):

modules/etcd/etcd.go
```diff
@@ -0,0 +1,249 @@
...
+	scheme              = "http"
+)
+
+// EtcdContainer represents the etcd container type used in the module
+type EtcdContainer struct {
+	testcontainers.Container
+	ClusterToken string
+	Nodes        []*EtcdContainer
```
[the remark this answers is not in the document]
mdelapenya: Indeed, I made it public to allow users accessing the nodes, but it's probably fine to hide them and only expose the client/peer URLs.

👍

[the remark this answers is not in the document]
mdelapenya: We do have a test that iterates through the nodes to verify the are members in the cluster. I think we can have another test file using the same `etcd` package in order to test private functionality like this one. Thoughts?

modules/etcd/etcd.go
mdelapenya: Originally, the Nodes field represents all the child nodes, not all cluster nodes, therefore the initial node is not part of that list. I added tests for describing those use cases.

Maybe worth it to rename that field to childNodes (private)?

modules/etcd/options.go
mdelapenya: Given we are going to work on a PoC refactoring the options pattern in modules, do you think we can move this to that follow-up?

stevenh (changes requested):
Agreed to delay the refactor to the follow up PR to avoid blocking.

modules/etcd/etcd.go
```diff
@@ -0,0 +1,243 @@
...
+			"--advertise-client-urls=" + scheme + "://" + settings.nodeNames[settings.currentNode] + ":" + clientPort,
+			"--listen-peer-urls=" + scheme + "://0.0.0.0:" + peerPort,
+			"--listen-client-urls=" + scheme + "://0.0.0.0:" + clientPort,
+			"--initial-cluster-state=new",
+		}
+
+		clusterStateValues := []string{}
+		for _, node := range settings.nodeNames {
+			clusterStateValues = append(clusterStateValues, node+"="+scheme+"://"+node+":"+peerPort)
+		}
```
stevenh, line 154: suggestion: this could be preallocated
```suggestion
		clusterStateValues := make([]string, len(settings.nodeNames))
		for i, node := range settings.nodeNames {
			clusterStateValues[i] = node+"="+scheme+"://"+node+":"+peerPort)
		}
```

modules/etcd/etcd.go
```diff
@@ -0,0 +1,243 @@
...
+
+	return clusterOpts, nil
+}
+
+// configureCMD configures the etcd command line arguments, based on the settings provided,
+// in order to create a cluster or a single-node instance.
+func configureCMD(settings options) []string {
+	cmds := []string{"etcd"}
+
+	if len(settings.nodeNames) < 1 {
```
stevenh, line 139: question: do we want to allow a cluster of 1 node, doesn't really make sense?

modules/etcd/options.go
```diff
@@ -0,0 +1,107 @@
...
+			},
+		})
+	}
+}
+
+// WithNodes is an option to set the nodes of the etcd cluster.
+// If only one node is specified, the cluster will be a single-node cluster.
+func WithNodes(nodes ...string) Option {
+	return func(o *options) {
+		o.nodeNames = nodes
```
stevenh, line 74: suggestion: If we don't want to allow single node cluster then we could use:
```suggestion
func WithNodes(node1, node2, nodes...string) Option {
	return func(o *options) {
		o.nodeNames = append([]string{node1, node2}, nodes...)
```

stevenh (changes requested):

modules/etcd/options.go
mdelapenya: I like the idea!

modules/etcd/etcd.go
```diff
@@ -0,0 +1,269 @@
...
+	}
+
+	// configure CMD with the nodes
+	genericContainerReq.Cmd = configureCMD(settings)
+
+	if settings.clusterNetwork != nil {
+		// apply the network to the current node
+		err := tcnetwork.WithNetwork([]string{settings.nodeNames[settings.currentNode]}, settings.clusterNetwork)(&genericContainerReq)
+		if err != nil {
+			return nil, fmt.Errorf("with network: %w", err)
```
stevenh, line 89: bug: this could orphan the network, see comments around best effort.

modules/etcd/etcd.go
```diff
@@ -0,0 +1,269 @@
...
+	}
+
+	if err := c.Container.Terminate(ctx); err != nil {
+		return fmt.Errorf("terminate cluster node: %w", err)
+	}
+
+	// remove the cluster network if it was created, but only for the first node
+	// we could check if the current node is the first one (index 0),
+	// and/or check that there are no child nodes
+	if c.opts.clusterNetwork != nil && c.opts.currentNode == 0 {
```
stevenh, line 46: Nice catch!

modules/etcd/etcd.go
mdelapenya: I think it could be the case you want a single etcd instance

modules/etcd/etcd.go
```diff
@@ -0,0 +1,269 @@
...
+
+// Terminate terminates the etcd container, its child nodes, and the network in which the cluster is running
+// to communicate between the nodes.
+func (c *EtcdContainer) Terminate(ctx context.Context) error {
+	// child nodes has no other children
+	for i, child := range c.childNodes {
+		if err := child.Terminate(ctx); err != nil {
+			return fmt.Errorf("terminate child node(%d): %w", i, err)
+		}
+	}
```
stevenh, line 37: bug: this could leave other resources orphaned, I would use a best effort.
```suggestion
	var errs []error
	for i, child := range c.childNodes {
		if err := child.Terminate(ctx); err != nil {
			errs = append(errs, fmt.Errorf("terminate child node(%d): %w", i, err))
		}
	}
```

modules/etcd/etcd.go
```diff
@@ -0,0 +1,269 @@
...
+	// remove the cluster network if it was created, but only for the first node
+	// we could check if the current node is the first one (index 0),
+	// and/or check that there are no child nodes
+	if c.opts.clusterNetwork != nil && c.opts.currentNode == 0 {
+		if err := c.opts.clusterNetwork.Remove(ctx); err != nil {
+			return fmt.Errorf("remove cluster network: %w", err)
+		}
+	}
+
+	return nil
```
stevenh, line 52: suggestion: with the best effort this would be:
```suggestion
	return errors.Join(errs...)
```

modules/etcd/etcd.go
```diff
@@ -0,0 +1,269 @@
...
+	// child nodes has no other children
+	for i, child := range c.childNodes {
+		if err := child.Terminate(ctx); err != nil {
+			return fmt.Errorf("terminate child node(%d): %w", i, err)
+		}
+	}
+
+	if err := c.Container.Terminate(ctx); err != nil {
+		return fmt.Errorf("terminate cluster node: %w", err)
+	}
```
stevenh, line 41: bug: use best effort and check for nil to allow this to clean up in all cases including only network creation success.
```suggestion
	if c.Container != nil {
		if err := c.Container.Terminate(ctx); err != nil {
			errs = append(errs, fmt.Errorf("terminate container: %w", err))
		}
	}
```
```

File: modules/etcd/etcd.go [hunk 23]
```diff
@@ -0,0 +1,243 @@
...
+	if container != nil {
+		c = &EtcdContainer{Container: container}
+	}
+
+	if err != nil {
+		return c, fmt.Errorf("generic container: %w", err)
+	}
+
+	// only the first node creates the cluster
+	if settings.currentNode == 0 {
```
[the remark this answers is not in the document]
mdelapenya: do you mean having a run function only for the first node?

modules/etcd/etcd.go
```diff
@@ -0,0 +1,269 @@
...
+	if err := c.Container.Terminate(ctx); err != nil {
+		return fmt.Errorf("terminate cluster node: %w", err)
+	}
+
+	// remove the cluster network if it was created, but only for the first node
+	// we could check if the current node is the first one (index 0),
+	// and/or check that there are no child nodes
+	if c.opts.clusterNetwork != nil && c.opts.currentNode == 0 {
+		if err := c.opts.clusterNetwork.Remove(ctx); err != nil {
+			return fmt.Errorf("remove cluster network: %w", err)
```
stevenh, line 48: bug: use best effort.
```suggestion
			errs = append(errs, fmt.Errorf("remove cluster network: %w", err))
```

modules/etcd/etcd_unit_test.go
```diff
@@ -0,0 +1,103 @@
...
+			node2: "etcd-2",
+			nodes: []string{},
+		},
+		{
+			name:  "3-nodes",
+			node1: "etcd-1",
+			node2: "etcd-2",
+			nodes: []string{"etcd-3"},
+		},
+	}
```
stevenh, line 48: suggestion: if you extract the main body to a function you can use inline test which is significantly cleaner and the benefit of being able to run individual cases in editors e.g.
```suggestion
	t.Run("2-nodes", testCluster(t, "etcd-1", "etcd-1", "etcd-3"))
	t.Run("3-nodes", testCluster(t, "etcd-1", "etcd-1", "etcd-3"))
```

The Run call will look something like:
```go
ctr, err := Run(ctx, "gcr.io/etcd-development/etcd:v3.5.14", WithNodes(nodes[0], nodes[1], nodes[2:]...), WithClusterToken(clusterToken))
```

stevenh (changes requested):
Making some good progress, nice catch on the network clean up.

modules/etcd/etcd.go
stevenh: Single instance yes but not single node cluster?

modules/etcd/etcd.go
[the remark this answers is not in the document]
stevenh: Yer I have some ideas on this, I would suggest we do as a follow up PR to optimise the flow.

modules/etcd/etcd.go
mdelapenya: I don't know the semantical difference between being a one-node cluster Vs being one instance (e.g. one elasticsearch instance is the same as one-node elasticsearch cluster).

In our case, the code just detects there is one node/instance, and it returns a single container. When the user requests a cluster (more than one nodes), then we return a parent container with children. So I do not see the difference here.

I could picture the Run function returning just one container and an eventual RunCluster one that separates concerns. But this is the purpose of the `WithNodes` option, right?

modules/etcd/etcd_unit_test.go
mdelapenya: I agree that inlining is cleaner 👍 OTOH my IDE allows my to run subtests even with this test case layout. I'm using VSCode. What IDE do you use?

mdelapenya: > "2-nodes", testCluster(t, "etcd-1", "etcd-1", "etcd-3")

Is this on purpose a bug? declaring a two-node cluster with 3 parameters? This makes me think if we need to validate the node names are unique 🤔 Wdyt?

mdelapenya: Added in https://github.com/testcontainers/testcontainers-go/pull/2788/commits/f52460d030715fd67baecee875f9bcac10141e35

modules/etcd/etcd.go
stevenh: Yer I don't see the need for a with cluster, but what I saw was all the additional command line options complexity for cluster setup which could be avoided for a single node. I believe the only reason etcd supports a single node cluster is so you can add nodes at a later date, which doesn't seem like a scenario you would need for testing.

stevenh: suggestion: With the change to WithNodes, this can be changed as we can no longer create a single node cluster.
```suggestion
	if len(settings.nodeNames) == 0 {
```

mdelapenya: >  I believe the only reason etcd supports a single node cluster is so you can add nodes at a later date
Indeed. When I started the feature it was to support two scenarios:
- one etcd node
- multiple etcd nodes in a cluster

No more

stevenh: You raise a good point, do we even need cluster support, should we just keep a patch about and add it later if there is demand, that avoids the hassle of deprecation?

modules/etcd/etcd_unit_test.go
stevenh: VSCode for me you can only select subtests when they are explicitly named, the for loop prevents you running just the subtest for example: `TestRunClusterMultipleNodes/2-nodes` and not just the top level test `TestRunClusterMultipleNodes`.

stevenh: That example was just a typo

modules/etcd/etcd.go
stevenh: I think so, I might be wrong. I guess someone might want to test their app worked correctly if they killed a cluster node, but that would be an edge case and need additional work to expose the nodes themselves.

modules/etcd/etcd.go
```diff
@@ -0,0 +1,285 @@
...
+	c.opts = settings
+
+	return c, nil
+}
+
+// configureCluster configures the cluster settings, ensuring that the cluster is properly configured with the necessary network and options,
+// avoiding duplicate application of options to be passed to the successive nodes.
+func configureCluster(ctx context.Context, settings *options, opts []testcontainers.ContainerCustomizer) ([]testcontainers.ContainerCustomizer, error) {
+	var clusterOpts []testcontainers.ContainerCustomizer
+	if len(settings.nodeNames) <= 1 {
```
stevenh, line 136: suggestion: use zero now we have made the change to WithNodes
```suggestion
	if len(settings.nodeNames) == 0 {
```

modules/etcd/etcd.go
```diff
@@ -0,0 +1,285 @@
...
+		endpoints = append(endpoints, endpoint)
+	}
+
+	return endpoints, nil
+}
+
+// PeerEndpoint returns the peer endpoint for the etcd container, and an error if any.
+// For a cluster, it returns the peer endpoint of the first node.
+func (c *EtcdContainer) PeerEndpoint(ctx context.Context) (string, error) {
+	var errs []error
```
stevenh, line 251: suggestion: you don't want best effort here, it was only need for the resource clean up as you want to try to remove all resources, in this case if any request fails the URL is unusable

stevenh (changes requested):
Just a few little things then I think we're there.

modules/etcd/etcd.go
mdelapenya: @eddumelendez @kiview what are your thoughts about feature parity across modules in the libraries? Should we start simple with no cluster support, or just keep what this PR does?

modules/etcd/etcd.go
```diff
@@ -0,0 +1,279 @@
...
+		err := tcnetwork.WithNetwork([]string{settings.nodeNames[settings.currentNode]}, settings.clusterNetwork)(&genericContainerReq)
+		if err != nil {
+			return c, fmt.Errorf("with network: %w", err)
+		}
+	}
+
+	container, err := testcontainers.GenericContainer(ctx, genericContainerReq)
+	if container != nil {
+		c.Container = container
+	}
```
stevenh, line 107: suggestion: we don't need the check any more
```suggestion
	c.Container = container
```

modules/etcd/etcd.go
```diff
@@ -0,0 +1,279 @@
...
+				// return the parent cluster node and the error, so the caller can clean up.
+				return c, fmt.Errorf("run cluster node: %w", err)
+			}
+
+			c.childNodes = append(c.childNodes, childNode)
+		}
+	}
+
+	c.opts = settings
+
```
stevenh, line 128: suggestion: not needed any more
```suggestion
```

modules/etcd/etcd.go
```diff
@@ -0,0 +1,279 @@
...
+	clusterOpts, err := configureCluster(ctx, &settings, opts)
+	if err != nil {
+		return nil, fmt.Errorf("configure cluster: %w", err)
+	}
+
+	// configure CMD with the nodes
+	genericContainerReq.Cmd = configureCMD(settings)
+
+	// Initialise the etcd container with the current settings.
```
stevenh, line 90: Nice comment, as this wouldn't be obvious to all.

modules/etcd/etcd.go
```diff
@@ -0,0 +1,243 @@
...
+
+	return cmds
+}
+
+// ClientEndpoint returns the client endpoint for the etcd container, and an error if any.
+// For a cluster, it returns the client endpoint of the first node.
+func (c *EtcdContainer) ClientEndpoint(ctx context.Context) (string, error) {
+	host, err := c.Host(ctx)
+	if err != nil {
+		return "", err
```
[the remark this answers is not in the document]
stevenh: suggestion: As we might have not been on this same page as to what I meant, this is what I was referring to with wrapping, more below.
```suggestion
		return "", fmt.Errorf("host: %w", err)
```

stevenh (changes requested):

modules/etcd/etcd.go
stevenh: It kind of does, but the changes already made mean that it doesn't make sense any more.

The original would have looked like the following, we only needed the `if` because we want to return a `nil` if `container` was `nil` so `CleanupContainer` works. This not the case anymore because we have other clean up we need to handle, so we always return an initialised `EtcContainer` and have the overridden `Terminate` method handle the if `c.Container` check.
```go
	var c *EtcdContainer
	if container != nil {
		c = &EtcdContainer{Container: container}
	}
```

Actually you could just make it:
```go
	c.Container, err = testcontainers.GenericContainer(ctx, genericContainerReq)
```

stevenh (approved):
I would make that one last change to the internal run call, but it's a nit so approving anyway if you want to do in a follow up.

modules/etcd/etcd.go
mdelapenya: oh I now know what you mean; I think it's more meaningful and readable for eventual readers to have the container-nil-check in the Run function than in the terminate method: we could add the check in the latter too for being defensive, but having in the "constructor" function makes it more understandable.

Wdyt?

stevenh: I think its more confusing having the `if` check given `c.Container` is `nil` by default so has to be handled by `Terminate` anyway, so I would just use:
```go
	if c.Container, err = testcontainers.GenericContainer(ctx, genericContainerReq); err != nil {
		return c, fmt.Errorf("generic container: %w", err)
	}
```

But not a blocker.

mdelapenya: Well, it actually does not matter, as the c.Container will be nil anyway. Changing it as you suggested, thanks!

stevenh (approved):
LGTM

PR #2790, opened 2024-09-21
fix: parallel containers clean race

stevenh:
Simply the logic in parallel containers, eliminating a clean up race condition where multiple clean ups on the same container could occur at the same time.


mdelapenya (approved):
LGTM, thanks for adding clarity and simplicity to that parallel block. 🙇

PR #2791, opened 2024-09-21
fix: container logging deadlocks

stevenh:
Refactor container log handling simplifying the logic fixing various issues with error handling and race conditions between the complex combinations of multiple channels that have been causing random deadlocks in tests.

The new version has simple for loop with an inter call to ContainerLogs and stdcopy.StdCopy leveraging an adapter between io.Writer and LogConsumer.

This could be used to easily expose separate stdout and stderr handlers.

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -5,7 +5,6 @@ import (
 	"bufio"
 	"context"
 	"encoding/base64"
-	"encoding/binary"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -17,7 +16,6 @@ import (
 	"path/filepath"
 	"regexp"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/cenkalti/backoff/v4"
@@ -30,6 +28,7 @@ import (
 	"github.com/docker/docker/client"
 	"github.com/docker/docker/errdefs"
 	"github.com/docker/docker/pkg/jsonmessage"
+	"github.com/docker/docker/pkg/stdcopy"
 	"github.com/docker/go-connections/nat"
 	"github.com/moby/term"
 	specs "github.com/opencontainers/image-spec/specs-go/v1"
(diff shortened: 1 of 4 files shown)


docker.go
```diff
@@ -690,6 +694,29 @@ func (c *DockerContainer) copyToContainer(ctx context.Context, fileContent func(
 	return nil
 }
 
+// logConsumerWriter is a writer that writes to a LogConsumer.
+type logConsumerWriter struct {
+	log       Log
+	consumers []LogConsumer
```
mdelapenya, line 700: Offtopic: In the v1 branch I have a LogConsumer that is a Multi log consumer (https://github.com/testcontainers/testcontainers-go/commit/9de581c05a541927a825c1b304e22ca818e4178f#diff-7a89ef1d0fc02a6398aad1d81b64b68f375822b4de5b15458f812daacf93595dR21), which would simplify this. I did not backport it from v1 to main to avoid another breaking change. Although it could be convenient if we see it useful.

docker.go
```diff
@@ -707,124 +734,94 @@ func (c *DockerContainer) StartLogProducer(ctx context.Context, opts ...LogProdu
...
+		// Retry from the last log received.
+		now := time.Now()
+		options.Since = fmt.Sprintf("%d.%09d", now.Unix(), int64(now.Nanosecond()))
+	}
+}
 
-			// a map of the log type --> int representation in the header, notice the first is blank, this is stdin, but the go docker client doesn't allow following that in logs
-			logTypes := []string{"", StdoutLog, StderrLog}
+// copyLogs copies logs from the container to stdout and stderr.
+func (c *DockerContainer) copyLogs(ctx context.Context, options container.LogsOptions, stdout, stderr io.Writer) error {
```
mdelapenya, line 815: NIT: we could pass the options as the last argument, as a convention

docker.go
```diff
@@ -707,124 +734,94 @@ func (c *DockerContainer) StartLogProducer(ctx context.Context, opts ...LogProdu
...
+	rc, err := c.provider.client.ContainerLogs(ctx, c.GetContainerID(), options)
```
mdelapenya, line 816: Question: do we need to close the docker client here too? I know there is an open discussion about this topic on how to treat the docker client more consistent (make it internal infra?), just want to make sure we do not leak connections here.

docker_test.go
```diff
@@ -240,6 +240,14 @@ func TestContainerReturnItsContainerID(t *testing.T) {
 	}
 }
 
+type logger struct {
```
mdelapenya, line 243: NIT: rename to make it clear it's a log consumer
```suggestion
type testLogConsumer struct {
```

mdelapenya (commented):
LGTM, just a few comments, none of them blockers.

docker.go
stevenh: We should consider eliminating the current Accept for something more idiomatic, separate readers for stdout and stderr, then you can just use [io.MultiReader](https://pkg.go.dev/io#MultiReader)

docker.go
stevenh: I'm thinking we're actually creating connection churn by having close in the code base, have added that a discussion point.

mdelapenya (approved):
I think the code is much simpler to read, thanks!

LGTM

PR #2792, opened 2024-09-21
fix: container timeout test

stevenh:
Reduce the timeout for TestContainerCreationTimesOutWithHttp to reduce the likely hood of failure.


mdelapenya (approved):
LGTM thanks!

PR #2793, opened 2024-09-21
fix(registry): wait for

stevenh:
Switch registry containers to wait for http response to ensure the container is ready to serve requests, which was causing random test failures.

Also remove unnecessary context creation.


mdelapenya (approved):
LGTM thanks! 🙇🏻

PR #2794, opened 2024-09-21
fix(redpanda): wait for

stevenh:
Wait for the admin interface to response to HTTP to avoid failures in configuring the instance when its not fully ready.

### DIFF
--- a/modules/redpanda/redpanda.go
+++ b/modules/redpanda/redpanda.go
@@ -62,7 +62,7 @@ func RunContainer(ctx context.Context, opts ...testcontainers.ContainerCustomize
 func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustomizer) (*Container, error) {
 	tmpDir, err := os.MkdirTemp("", "redpanda")
 	if err != nil {
-		return nil, fmt.Errorf("failed to create directory: %w", err)
+		return nil, fmt.Errorf("create temporary directory: %w", err)
 	}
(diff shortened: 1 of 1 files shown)


mdelapenya (approved):
LGTM, thanks!

modules/redpanda/redpanda.go
```diff
@@ -192,34 +192,53 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
...
+	err = ctr.CopyToContainer(ctx, nodeConfig, filepath.Join(redpandaDir, "redpanda.yaml"), 0o600)
 	if err != nil {
-		return c, fmt.Errorf("failed to copy redpanda.yaml into container: %w", err)
+		return c, fmt.Errorf("copy to container: %w", err)
 	}
 
 	// 8. Wait until Redpanda is ready to serve requests.
+	waitHTTP := wait.ForHTTP(defaultAdminAPIPort).
+		WithStatusCodeMatcher(func(status int) bool {
+			return status == http.StatusNotFound
```
mdelapenya, line 217: Is it status not found?

stevenh: We're making a request to the root which doesn't have a handler, so yer this looks weird but it is right.

mdelapenya: Can we add a comment about that? 🙏

mdelapenya (approved):
LGTM, thanks!

PR #2809, opened 2024-10-02
feat(ryuk): make listen address of exposed port configurable

romainlaurent:
## What does this PR do?

Configurable listen address for the ExposedPorts on Ryuk container.

## Why is it important?

We run our containers from a gitlab runner where there is a network policy which forbid containers exposing port to listen on 0.0.0.0 .

## Related issues

The unit tests should cover the use cases

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -60,6 +60,12 @@ type Config struct {
 	// Environment variable: TESTCONTAINERS_RYUK_DISABLED
 	RyukDisabled bool `properties:"ryuk.disabled,default=false"`
 
+	// RyukExternalAddress is the address of the external Garbage Collector container.
+	// If set, Testcontainers will not attempt to start its own Ryuk container.
+	//
+	// Environment variable: TESTCONTAINERS_RYUK_EXTERNAL_ADDRESS
+	RyukExternalAddress string `properties:"ryuk.external.address,default="`
+
 	// RyukPrivileged is a flag to enable or disable the privileged mode for the Garbage Collector container.
 	// Setting this to true will run the Garbage Collector container in privileged mode.
 	//
@@ -114,6 +120,11 @@ func read() Config {
 		ryukDisabledEnv := os.Getenv("TESTCONTAINERS_RYUK_DISABLED")
 		if parseBool(ryukDisabledEnv) {
 			config.RyukDisabled = ryukDisabledEnv == "true"
+		}
+
+		ryukExternalAddress := os.Getenv("TESTCONTAINERS_RYUK_EXTERNAL_ADDRESS")
+		if ryukExternalAddress != "" {
+			config.RyukExternalAddress = ryukExternalAddress
 		}
 
 		hubImageNamePrefix := os.Getenv("TESTCONTAINERS_HUB_IMAGE_NAME_PREFIX")
--- a/reaper.go
+++ b/reaper.go
@@ -204,22 +204,10 @@ func reuseReaperContainer(ctx context.Context, sessionID string, provider Reaper
 
 	Logger.Printf("⏳ Waiting for Reaper port to be ready")
 
-	var containerJson *types.ContainerJSON
-
-	if containerJson, err = reaperContainer.Inspect(ctx); err != nil {
-		return nil, fmt.Errorf("failed to inspect reaper container %s: %w", reaperContainer.ID[:8], err)
-	}
-
-	if containerJson != nil && containerJson.NetworkSettings != nil {
-		for port := range containerJson.NetworkSettings.Ports {
-			err := wait.ForListeningPort(port).
-				WithPollInterval(100*time.Millisecond).
-				WaitUntilReady(ctx, reaperContainer)
-			if err != nil {
-				return nil, fmt.Errorf("failed waiting for reaper container %s port %s/%s to be ready: %w",
-					reaperContainer.ID[:8], port.Proto(), port.Port(), err)
-			}
-		}
+	err = wait.ForLog("Started").WaitUntilReady(ctx, reaperContainer)
+	if err != nil {
+		return nil, fmt.Errorf("failed waiting for reaper container %s to be ready: %w",
+			reaperContainer.ID[:8], err)
 	}
 
 	return &Reaper{
@@ -249,7 +237,7 @@ func newReaper(ctx context.Context, sessionID string, provider ReaperProvider) (
 		ExposedPorts: []string{string(listeningPort)},
 		Labels:       core.DefaultLabels(sessionID),
 		Privileged:   tcConfig.RyukPrivileged,
-		WaitingFor:   wait.ForListeningPort(listeningPort),
+		WaitingFor:   wait.ForLog("Started"),
 		Name:         reaperContainerNameFromSessionID(sessionID),
 		HostConfigModifier: func(hc *container.HostConfig) {
 			hc.AutoRemove = true
--- a/reaper.go
+++ b/reaper.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"math/rand"
 	"net"
+	"os"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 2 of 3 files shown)


stevenh (changes requested):
Nice addition, thanks for the PR.

I've made some initial suggestions but we should wait for https://github.com/testcontainers/testcontainers-go/pull/2728 to be merged before completing as there's going to be some conflicts and it will be easier to fix here.

We should also look to test that the reaper works as expected using a custom port so a test for that should be added, best to wait for the above PR to land first as that changes quite a bit in that area.

internal/config/config.go
```diff
@@ -126,6 +136,11 @@ func read() Config {
 			config.RyukPrivileged = ryukPrivilegedEnv == "true"
 		}
 
+		ryukListenAddress := os.Getenv("TESTCONTAINERS_RYUK_LISTEN_ADDRESS")
+		if parseIP(ryukListenAddress) {
```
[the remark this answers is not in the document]
romainlaurent: First thanks for the quick review! 
I agreed to handle the error, do I panic to not change the function signature?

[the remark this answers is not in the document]
stevenh: I think the knock on changes to return an error might be significant, so I'd suggest panic for now and raise an issue to track the function signature change. @mdelapenya thoughts?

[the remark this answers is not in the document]
mdelapenya: Agreed in panicking here, as the function is just used for the config. If there is a non-valid bool in the value, makes sense to raise the error.

romainlaurent:
Thanks for your comments!
Waiting patiently for #2728 to be merged

internal/config/config.go
```diff
@@ -126,6 +136,11 @@ func read() Config {
 			config.RyukPrivileged = ryukPrivilegedEnv == "true"
 		}
 
+		ryukAddress := os.Getenv("TESTCONTAINERS_RYUK_ADDRESS")
+		if ryukAddress != "" {
```
stevenh, line 140: nit: we could fold this to be a little more idiomatic
```suggestion
		if ryukAddress := os.Getenv("TESTCONTAINERS_RYUK_ADDRESS"); ryukAddress != "" {
```

internal/config/config_test.go
```diff
@@ -140,6 +142,15 @@ func TestReadTCConfig(t *testing.T) {
 		assert.Equal(t, expected, config)
 	})
 
+	t.Run("Invalid IP address", func(t *testing.T) {
```
stevenh, line 145: suggestion: to my previous set of comments, were working towards simple names with no spaces:
```suggestion
	t.Run("invalid-address", func(t *testing.T) {
```

internal/config/config_test.go
```diff
@@ -140,6 +142,15 @@ func TestReadTCConfig(t *testing.T) {
...
+		t.Setenv("TESTCONTAINERS_RYUK_ADDRESS", "invalid")
+		t.Cleanup(func() {
+			os.Unsetenv("TESTCONTAINERS_RYUK_ADDRESS")
+		})
```
stevenh, line 149: bug: this isn't needed [t.Setenv](https://pkg.go.dev/testing#T.Setenv) already adds a clean up function.
```suggestion
```

reaper.go
```diff
@@ -240,13 +240,18 @@ func newReaper(ctx context.Context, sessionID string, provider ReaperProvider) (
...
 	}
 
-	listeningPort := nat.Port("8080/tcp")
+	listeningPort := nat.Port(config.ReaperDefaultPort)
 
 	tcConfig := provider.Config().Config
 
+	exposedPort := string(listeningPort)
+	if tcConfig.RyukAddress != "" {
+		exposedPort = net.JoinHostPort(tcConfig.RyukAddress, "") + ":" + exposedPort
```
stevenh, line 249: bug: pretty sure this would result in a double colon, I would expect the following, did I miss something?
```suggestion
	exposedPort := config.ReaperDefaultPort
	if tcConfig.RyukAddress != "" {
		exposedPort = net.JoinHostPort(tcConfig.RyukAddress, exposedPort)
```

Also note that the other PR has port configurable using the env var `RYUK_PORT`, not sure if that will land, but just making you aware.

reaper_test.go
```diff
@@ -408,6 +408,22 @@ func Test_NewReaper(t *testing.T) {
 				"TESTCONTAINERS_HUB_IMAGE_NAME_PREFIX": "registry.mycompany.com/mirror",
 			},
 		},
+		{
+			name: "Reaper with a custom listen address",
```
stevenh, line 412: suggestion: simple test names
```suggestion
			name: "custom-address",
```

internal/config/config.go
```diff
@@ -168,3 +183,12 @@ func parseBool(input string) bool {
 	_, err := strconv.ParseBool(input)
 	return err == nil
 }
+
+func parseIP(input string) string {
+	ip := net.ParseIP(input)
+	if ip == nil {
+		panic("invalid IP address: " + input)
```
stevenh, line 190: suggestion: quote so users can see issues like trailing white space
```suggestion
		panic(fmt.Sprintf("invalid IP address: %q", input))
```

stevenh (changes requested):
Thanks for the quick changes, second round...

stevenh:
@romainlaurent not sure you wanted to delete this branch?

romainlaurent:
Indeed, a branch rename had more impact than expected. 
Unfortunately this approach haven't fixed my issue.
The problem I got is my tests are running in gitlab CI with docker executor.
It makes my Ryuk container running alongside the one launching the tests .
I'll propose a different PR where I override the Reaper.Endpoint field [here](https://github.com/testcontainers/testcontainers-go/blob/7c536674f43c91a4dab859aa1aa2e8c23c9f17ae/reaper.go#L332)
With an env variable such as TESTCONTAINERS_RYUK_USE_NAME_ENDPOINT=false ?
What do you think about it?

mdelapenya:
@romainlaurent You probably read this about Gitlab CI, right? https://golang.testcontainers.org/system_requirements/ci/gitlab_ci/

Isn't it working for you?

romainlaurent:
@mdelapenya thanks for your answer!
Yes I have but it's not enough, there is strict network policies on the runner which forbid traffic on container listening on 0.0.0.0:<port>
I got this log error
`failed waiting for reaper container 8fdba937 port tcp/8080 to be ready: dial tcp 172.17.0.1:33074: connect: network is unreachable: creating reaper failed`
Maybe by changing only [here](https://github.com/testcontainers/testcontainers-go/blob/7c536674f43c91a4dab859aa1aa2e8c23c9f17ae/reaper.go#L252)
`wait.ForListeningPort(listeningPort)` by `wait.ForLog("Started")`

reaper.go
```diff
@@ -249,7 +249,7 @@ func newReaper(ctx context.Context, sessionID string, provider ReaperProvider) (
 		ExposedPorts: []string{string(listeningPort)},
 		Labels:       core.DefaultLabels(sessionID),
 		Privileged:   tcConfig.RyukPrivileged,
-		WaitingFor:   wait.ForListeningPort(listeningPort),
+		WaitingFor:   wait.ForLog("Started"),
```
stevenh, line 240: bug: If the listening for check doesn't work then the reaper wont be accessible so this is not a good fix.

reaper.go
```diff
@@ -204,22 +204,10 @@ func reuseReaperContainer(ctx context.Context, sessionID string, provider Reaper
...
-		for port := range containerJson.NetworkSettings.Ports {
-			err := wait.ForListeningPort(port).
-				WithPollInterval(100*time.Millisecond).
-				WaitUntilReady(ctx, reaperContainer)
-			if err != nil {
-				return nil, fmt.Errorf("failed waiting for reaper container %s port %s/%s to be ready: %w",
-					reaperContainer.ID[:8], port.Proto(), port.Port(), err)
-			}
-		}
+	err = wait.ForLog("Started").WaitUntilReady(ctx, reaperContainer)
```
stevenh, line 207: bug: If the listening for doesn't work the container isn't accessible so wont work.

stevenh (changes requested):
I don't think this is a workable fix I'm afraid as if `ForListeningPort` doesn't work then the container wont be accessible.

reaper.go
romainlaurent: It could work by adding TESTCONTAINERS_RYUK_ENDPOINT_OVERRIDE_BY_NAME=true, allowing us to contact the container directly through its name. 
I'm not sure it's the most elegant solution, so I'm open to any alternatives you think would be better.

romainlaurent:
After a lot of try, I couldn't find a quick and elegant way to fix my problem. As it's specific at our gitlab runner configuration, I'll close this PR.
Thanks to @stevenh @mdelapenya for your time and the tool you've made, it's really great!

stevenh:
@romainlaurent would you mind creating a issue to track this as I think its something we should look to fix.

PR #2810, opened 2024-10-02
feat: support reading Docker host from the current Docker context

mdelapenya:
- **feat: support reading Docker context**
- **chore: move read Docker config to the core**
- **chore: read docker config more consistently**
- **chore: extract to a parse function to detect remote docker hosts**
- **docs: add Docker context to docs**

## What does this PR do?
This PR adds a new step in the Docekr host discovery algorithm, extracting the Docker host from the current Docker context. To achieve that, we are copying code from the docker/cli project,
as we don't want to receive all the dependencies from that project (please see `docker_context.go` file). In the end, it just does the following:

- list all contexts from the default location: docker config path / context / meta / contextA / meta.json
- get the current context from the docker config, located at docker config path / config.json
- read the meta.json file for the current context, extracting the host endpoint
- use that host endpoint for the Docker Host.

NOTE: The default docker config path lives in `$HOME/.docker`.

## Why is it important?
It will simplify the experience using different container runtimes, as this change means that with just setting the current context, testcontainers-go will use that container runtime.


internal/core/docker_config.go
```diff
@@ -0,0 +1,31 @@
...
+	"os"
+
+	"github.com/cpuguy83/dockercfg"
+)
+
+// ReadDockerConfig returns the docker config file. It will internally check, in this particular order:
+// 1. the DOCKER_AUTH_CONFIG environment variable, unmarshalling it into a dockercfg.Config
+// 2. the DOCKER_CONFIG environment variable, as the path to the config file
+// 3. else it will load the default config file, which is ~/.docker/config.json
+func ReadDockerConfig() (*dockercfg.Config, error) {
```
stevenh, line 15: nit: can we just use `DockerConfig` I try to avoid `Get`, `Read` etc when its not needed to keep things clean.
```suggestion
// DockerConfig returns the docker config file. It will internally check, in this particular order:
// 1. the DOCKER_AUTH_CONFIG environment variable, unmarshalling it into a dockercfg.Config
// 2. the DOCKER_CONFIG environment variable, as the path to the config file
// 3. else it will load the default config file, which is ~/.docker/config.json
func DockerConfig() (*dockercfg.Config, error) {
```

docs/features/configuration.md
```diff
@@ -75,17 +75,19 @@ See [Docker environment variables](https://docs.docker.com/engine/reference/comm
 
 3. Read the Go context for the **DOCKER_HOST** key. E.g. `ctx.Value("DOCKER_HOST")`. This is used internally for the library to pass the Docker host to the resource reaper.
 
-4. Read the default Docker socket path, without the unix schema. E.g. `/var/run/docker.sock`
+4. Read the host endpoint for the current Docker context in the Docker configuration file. E.g. `~/.docker/config.json`.
 
-5. Read the **docker.host** property in the `~/.testcontainers.properties` file. E.g. `docker.host=tcp://my.docker.host:1234`
+5. Read the default Docker socket path, without the unix schema. E.g. `/var/run/docker.sock`
```
stevenh, line 80: suggestion: I think we need to clarify this as I'm not sure I understand where it is getting the socket from?

internal/core/docker_context.go
```diff
@@ -0,0 +1,328 @@
...
+	// envOverrideContext is the name of the environment variable that can be
+	// used to override the context to use. If set, it overrides the context
+	// that's set in the CLI's configuration file, but takes no effect if the
+	// "DOCKER_HOST" env-var is set (which takes precedence.
+	envOverrideContext = "DOCKER_CONTEXT"
+	// envOverrideConfigDir is the name of the environment variable that can be
+	// used to override the location of the client configuration files (~/.docker).
+	//
+	// It takes priority over the default, but can be overridden by the "--config"
+	// command line option.
```
stevenh, line 33: question: not sure if this is valid in our case?

internal/core/docker_context.go
```diff
@@ -0,0 +1,328 @@
...
+		}
+	}
+	return home
+}
+
+// configurationDir returns the directory the configuration file is stored in
+func configurationDir() string {
+	configDir := os.Getenv(envOverrideConfigDir)
+	if configDir == "" {
+		configDir = filepath.Join(getHomeDir(), configFileDir)
```
stevenh, line 269: suggestion: just return, avoiding the unnecessary assignment and arguably clearer to the reader.
```suggestion
		return filepath.Join(getHomeDir(), configFileDir)
```

internal/core/docker_context.go
```diff
@@ -0,0 +1,328 @@
...
+	for _, m := range md {
+		if m.Name == currentContext {
+			ep, ok := m.Endpoints[dockerEndpoint].(endpointMeta)
+			if ok {
+				return ep.Host, nil
+			}
+		}
+	}
+
+	return "", fmt.Errorf("no Docker host found in the current context")
```
stevenh, line 301: suggestion: avoid the overhead of `fmt.Errorf` when not needed.
```suggestion
	return "", errors.New("no Docker host found in the current context")
```

internal/core/docker_context_test.go
```diff
@@ -0,0 +1,55 @@
...
+//   - meta
+//   - context$i
+//   - meta.json
+//
+// The config.json file contains the current context, and the meta.json files contain the metadata for each context.
+// It generates the specified number of contexts, setting the current context to the one specified by currentContextIndex.
+func setupDockerContexts(t *testing.T, currentContextIndex int, contextsCount int) {
+	configDir := filepath.Join(getHomeDir(), configFileDir)
+
+	err := createTmpDir(configDir)
```
stevenh, line 26: question: looks like this is just `os.MkdirAll`, could we use [T.TempDir](https://pkg.go.dev/testing#T.TempDir) which automatically cleans up instead and wouldn't trash the users config, if that understanding is correct?

internal/core/docker_context_test.go
```diff
@@ -0,0 +1,55 @@
...
+// - $HOME/.docker
+//   - config.json
+//   - contexts
+//   - meta
+//   - context$i
+//   - meta.json
+//
+// The config.json file contains the current context, and the meta.json files contain the metadata for each context.
+// It generates the specified number of contexts, setting the current context to the one specified by currentContextIndex.
+func setupDockerContexts(t *testing.T, currentContextIndex int, contextsCount int) {
```
stevenh, line 23: question: this looks like it could trash the user configuration?

internal/core/docker_host.go
```diff
@@ -262,6 +266,16 @@ func dockerHostFromContext(ctx context.Context) (string, error) {
...
 }
 
+// dockerHostFromContext returns the docker host from the Go context, if it's not empty
+func dockerHostFromDockerContext(ctx context.Context) (string, error) {
+	dockerHost, err := GetDockerHostFromCurrentContext()
+	if err == nil {
+		return dockerHost, nil
+	}
+
+	return "", ErrDockerSocketNotSetInDockerContext
```
stevenh, line 276: question: Is this discarding useful information from the actual error?

internal/core/docker_host_test.go
```diff
@@ -154,6 +154,21 @@ func TestExtractDockerHost(t *testing.T) {
 		require.Equal(t, DockerSocketSchema+"/this/is/a/sample.sock", host)
 	})
 
+	t.Run("Docker Host from Docker context", func(tt *testing.T) {
```
stevenh, line 157: suggest: simple naming no spaces, more below.

internal/core/testdata/.docker/config.json
```diff
@@ -0,0 +1,8 @@
+{
+    "auths": {
+            "https://index.docker.io/v1/": {},
+            "https://example.com": {},
+            "https://my.private.registry": {}
+    },
+    "credsStore": "desktop"
+}
```
stevenh, line 8: suggestion: missing new line

stevenh (changes requested):
quick first pass with some question,.

internal/core/docker_context.go
mdelapenya: Indeed. Copied using a sync.Once, but I removed it for simplicity. Changed

internal/core/docker_context_test.go
mdelapenya: It's used in tests, and it's always used in conjunction with t.SetEnv("HOME"). But you're totally right that, in general, not only here, there are unit tests that need to be refactored and leverage tmp dirs more and more to avoid exactly that: messing up with the current user conf.

internal/core/docker_host.go
mdelapenya: Indeed. Wdyt about this order: return "", errors.Join(ErrDockerSocketNotSetInDockerContext, err)?
```suggestion
	return "", errors.Join(ErrDockerSocketNotSetInDockerContext, err)
```

internal/core/docker_context_test.go
mdelapenya: See my comment here: https://github.com/testcontainers/testcontainers-go/pull/2810#discussion_r1786128082

Wdyt if we dedicate a follow-up PR for simplifying the **unit** tests for the socket path resolution? Some would apply for reading the tc config.

internal/core/docker_host.go
stevenh: I think that would be confusing, I would detect a specific "not found" error from the call and only translate that if needed. If that's not needed I would just return a wrapped straight wrapped error.

internal/core/docker_context_test.go
stevenh: If this is expected to override the home dir perhaps the quickest thing would be to have this function do that so its very clear and can't be accidentally used without that and destroy someone config.

internal/core/docker_context_test.go
kiview: Definitely in favor of avoiding interacting with the actual home dir, if possible.

mdelapenya: I'd work on that refinement in a separate PR, to not mix things up.

internal/core/docker_host.go
mdelapenya: Wdyt https://github.com/testcontainers/testcontainers-go/pull/2810/commits/a38a34591248c2d0205eb1a6ef60ef1686f25520 ?

internal/core/docker_context_test.go
mdelapenya: See my comment https://github.com/testcontainers/testcontainers-go/pull/2810#discussion_r1799036874 about refining that in a follow-up PR

internal/config/config.go
```diff
@@ -89,6 +89,65 @@ type Config struct {
...
+
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return applyEnvironmentConfiguration(config)
+	}
+
+	tcProp := filepath.Join(home, ".testcontainers.properties")
+	// Init from a file, ignore if it doesn't exist, which is the case for most users.
+	// The properties library will return the default values for the struct.
+	props, err := properties.LoadFiles([]string{tcProp}, properties.UTF8, true)
```
stevenh, line 104: bug: looks like this could fail due to bad config and not inform the user, making them think their config is being used but it isn't. Explicitly check of `fs.ErrNotExist`?

internal/config/config.go
```diff
@@ -89,6 +89,65 @@ type Config struct {
...
+// defaultConfig
+func defaultConfig() Config {
+	config := Config{}
+
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return applyEnvironmentConfiguration(config)
+	}
+
+	tcProp := filepath.Join(home, ".testcontainers.properties")
```
stevenh, line 101: suggestion: constant for the file name.

internal/config/config.go
```diff
@@ -89,6 +89,65 @@ type Config struct {
 
 // }
 
+// defaultConfig
+func defaultConfig() Config {
```
stevenh, line 93: bug: callers can't distinguish success and error, which could lead to unexpected behaviour.

internal/config/config.go
```diff
@@ -89,6 +89,65 @@ type Config struct {
...
+	tcProp := filepath.Join(home, ".testcontainers.properties")
+	// Init from a file, ignore if it doesn't exist, which is the case for most users.
+	// The properties library will return the default values for the struct.
+	props, err := properties.LoadFiles([]string{tcProp}, properties.UTF8, true)
+	if err != nil {
+		return applyEnvironmentConfiguration(config)
+	}
+
+	if err := props.Decode(&config); err != nil {
+		fmt.Printf("invalid testcontainers properties file, returning an empty Testcontainers configuration: %v\n", err)
```
stevenh, line 110: suggestion: return an error.

internal/config/config.go
```diff
@@ -89,6 +89,65 @@ type Config struct {
...
+		return applyEnvironmentConfiguration(config)
+	}
+
+	return config
+}
+
+func applyEnvironmentConfiguration(config Config) Config {
```
stevenh, line 117: bug: callers can't distinguish actual success from hidden failure e.g. a bad boolean or duration.

internal/config/config.go
```diff
@@ -108,58 +167,7 @@ func Reset() {
...
-		return applyEnvironmentConfiguration(config)
-	}
-
-	if err := properties.Decode(&config); err != nil {
-		fmt.Printf("invalid testcontainers properties file, returning an empty Testcontainers configuration: %v\n", err)
-		return applyEnvironmentConfiguration(config)
-	}
+	config := defaultConfig()
 
 	return applyEnvironmentConfiguration(config)
```
stevenh, line 172: bug: this is already applied

internal/core/docker_context.go
```diff
@@ -0,0 +1,328 @@
...
+	"runtime"
+
+	"github.com/docker/docker/client"
+	"github.com/docker/docker/errdefs"
+)
+
+const (
+	// defaultContextName is the name reserved for the default context (config & env based)
+	defaultContextName = "default"
+	// envOverrideContext is the name of the environment variable that can be
```
stevenh, line 24: nit: I find adding a blank line between variables with comments easier to read.

internal/core/docker_context.go
```diff
@@ -0,0 +1,328 @@
...
+		return metadata{}, err
+	}
+
+	var untyped untypedContextMetadata
+	r := metadata{
+		Endpoints: make(map[string]any),
+	}
+
+	if err := json.Unmarshal(bytes, &untyped); err != nil {
+		return metadata{}, fmt.Errorf("parsing %s: %w", fileName, err)
```
stevenh, line 137: suggestion: use a unique wrap between this and the other errors below so we can identify where the failure occurred.

internal/core/docker_context.go
```diff
@@ -0,0 +1,328 @@
...
+func listRecursivelyMetadataDirs(root string) ([]string, error) {
+	fis, err := os.ReadDir(root)
+	if err != nil {
+		return nil, err
+	}
+
+	var result []string
+	for _, fi := range fis {
+		if fi.IsDir() {
+			if isContextDir(filepath.Join(root, fi.Name())) {
```
stevenh, line 199: question: Is this a duplicate check as if we start with root and fi is a directory it should not be possible that this could ever fail?

internal/core/docker_context.go
```diff
@@ -0,0 +1,328 @@
...
+
+	typed := getter()
+	if err := json.Unmarshal(payload, typed); err != nil {
+		return nil, err
+	}
+
+	return reflect.ValueOf(typed).Elem().Interface(), nil
+}
+
+// getHomeDir returns the home directory of the current user with the help of
```
stevenh, line 238: suggestion: replace with `os.UserHomeDir`, I suspect this is copy from a time when that didn't exist.

internal/core/docker_context.go
```diff
@@ -0,0 +1,328 @@
+package core
```
stevenh, line 1: question: can we check this is all required, as it looks like some types don't seem to be referenced.

internal/core/docker_host.go
```diff
@@ -19,13 +19,14 @@ type dockerHostContext string
...
 
 var (
-	ErrDockerHostNotSet               = errors.New("DOCKER_HOST is not set")
-	ErrDockerSocketOverrideNotSet     = errors.New("TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE is not set")
-	ErrDockerSocketNotSetInContext    = errors.New("socket not set in context")
-	ErrDockerSocketNotSetInProperties = errors.New("socket not set in ~/.testcontainers.properties")
-	ErrNoUnixSchema                   = errors.New("URL schema is not unix")
-	ErrSocketNotFound                 = errors.New("socket not found")
-	ErrSocketNotFoundInPath           = errors.New("docker socket not found in " + DockerSocketPath)
+	ErrDockerHostNotSet                  = errors.New("DOCKER_HOST is not set")
```
stevenh, line 22: question: do these all need to be exported?

internal/core/docker_host.go
```diff
@@ -262,6 +270,16 @@ func dockerHostFromContext(ctx context.Context) (string, error) {
 	return "", ErrDockerSocketNotSetInContext
 }
 
+// dockerHostFromContext returns the docker host from the Go context, if it's not empty
+func dockerHostFromDockerContext(ctx context.Context) (string, error) {
```
stevenh, line 274: question: is this method needed, seems like its just enforcing that the returned string is blank on error which I would expected to be the case anyway?

stevenh (changes requested):

docker.go
```diff
@@ -1593,13 +1566,7 @@ func (p *DockerProvider) GetNetwork(ctx context.Context, req NetworkRequest) (ne
...
 
 func (p *DockerProvider) GetGatewayIP(ctx context.Context) (string, error) {
-	// Use a default network as defined in the DockerProvider
-	defaultNetwork, err := p.ensureDefaultNetwork(ctx)
-	if err != nil {
-		return "", fmt.Errorf("ensure default network: %w", err)
-	}
-
-	nw, err := p.GetNetwork(ctx, NetworkRequest{Name: defaultNetwork})
+	nw, err := p.GetNetwork(ctx, NetworkRequest{Name: Bridge})
```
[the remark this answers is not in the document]
mdelapenya: Nope, the podman **engine** will retrieve the default bridge network, that's why it's fine to remove all the logic we had for calculating the default network.

```
➜ testcontainers-go (context-support) ✔ podman network ls
NETWORK ID    NAME        DRIVER
2f259bab93aa  podman      bridge
➜ testcontainers-go (context-support) ✔ docker network ls
NETWORK ID     NAME      DRIVER    SCOPE
2f259bab93aa   bridge    bridge    local
➜ testcontainers-go (context-support) ✔ 
```

docker_test.go
```diff
@@ -450,15 +431,16 @@ func TestContainerCreationWithName(t *testing.T) {
...
 			Image: nginxAlpineImage,
 			ExposedPorts: []string{
 				nginxDefaultPort,
 			},
 			WaitingFor: wait.ForHTTP("/").WithPort(nginxDefaultPort),
 			Name:       creationName,
-			Networks:   []string{"bridge"},
+			// no network means it will be connected to the default bridge network
+			// of any given container runtime
+			Networks: []string{},
```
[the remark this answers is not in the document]
mdelapenya: I'll keep the comment, only

docker_test.go
```diff
@@ -474,12 +456,7 @@ func TestContainerCreationWithName(t *testing.T) {
 	require.NoError(t, err)
 	require.Lenf(t, networks, 1, "Expected networks 1. Got '%d'.", len(networks))
 	network := networks[0]
-	switch providerType {
-	case ProviderDocker:
-		assert.Equalf(t, Bridge, network, "Expected network name '%s'. Got '%s'.", Bridge, network)
-	case ProviderPodman:
-		assert.Equalf(t, Podman, network, "Expected network name '%s'. Got '%s'.", Podman, network)
-	}
+	assert.Equalf(t, Bridge, network, "Expected network name '%s'. Got '%s'.", Bridge, network)
```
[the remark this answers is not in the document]
mdelapenya: Same as above

I'll update it to use require

docs/features/configuration.md
```diff
@@ -81,17 +81,19 @@ See [Docker environment variables](https://docs.docker.com/engine/reference/comm
 
 3. Read the Go context for the **DOCKER_HOST** key. E.g. `ctx.Value("DOCKER_HOST")`. This is used internally for the library to pass the Docker host to the resource reaper.
 
-4. Read the default Docker socket path, without the unix schema. E.g. `/var/run/docker.sock`
+4. Read the host endpoint for the current Docker context in the Docker configuration file. E.g. `~/.docker/config.json`.
```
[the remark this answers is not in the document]
mdelapenya: From https://docs.podman.io/en/v4.3/markdown/podman-login.1.html

> Podman will then use any existing credentials found in $HOME/.docker/config.json. If those credentials are not present, Podman will create ${XDG_RUNTIME_DIR}/containers/auth.json (if the file does not exist) and will then store the username and password from STDIN as a base64 encoded string in it.

docs/system_requirements/docker.md
```diff
@@ -9,3 +9,71 @@ However, these are not actively tested in the main development workflow, so not
...
+
+### Orbstack
+
+Orbstack creates its own Docker context when it is installed. This context is called `orbstack`. You can set this as the active context by running:
+
+```sh
+docker context use orbstack
+```
+
+### Podman
```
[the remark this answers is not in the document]
mdelapenya: You're right, I'm updating the command to use just podman.

docker.go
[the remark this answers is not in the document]
stevenh: But this is filtering for `Name` = bridge not type `Driver` so essentially its running:
```shell
docker network ls -f name=bridge
```
Does that work on podman?

internal/config/config.go
mdelapenya: Do you think we can move to a separate issue for bootstrapping the config in a more consistent manner?

stevenh: That depends, is this moving code or introducing it, if the latter I would suggest this is fixed in this PR so we don't introduce new issues, all be it temporarily.

internal/core/docker_context.go
mdelapenya: It's using `os.UserHomeDir` already, with a corner case where the home is empty in not-windows 🤔 
```go
// [pkg/homedir.Get]: https://pkg.go.dev/github.com/docker/docker@v26.1.4+incompatible/pkg/homedir#Get
func getHomeDir() string {
	home, _ := os.UserHomeDir()
	if home == "" && runtime.GOOS != "windows" {
		if u, err := user.Current(); err == nil {
			return u.HomeDir
		}
	}
	return home
}
```

internal/config/config.go
mdelapenya: The magiconair library already checks that if the ignoreMissing param is set to true: https://github.com/magiconair/properties/blob/c9a06e8f8f0164e4e16c0d5c4793cbed4ac90264/load.go#L93-L102

```go 
// LoadFile reads a file into a Properties struct.
// If IgnoreMissing is true then a missing file will not be
// reported as error.
func (l *Loader) LoadFile(filename string) (*Properties, error) {
	data, err := ioutil.ReadFile(filename)
	if err != nil {
		if l.IgnoreMissing && os.IsNotExist(err) {
			LogPrintf("properties: %s not found. skipping", filename)
			return NewProperties(), nil
		}
		return nil, err
	}
	return l.loadBytes(data, l.Encoding)
}
```

internal/config/config.go
mdelapenya: Following the code, if no if-branch is eval to true, then the config won't receive the env config. I think we could move the apply call to the defaultConfig function, so that it only happens there.

internal/core/docker_host.go
mdelapenya: Probably not. Given they are internal, I think we can rename all of them

internal/core/docker_host.go
mdelapenya: I think this method is basically passing an unused context to satisfy the "contract" in the above slice of functions. I think we can remove it and simply define an anonymous function for that.

stevenh (changes requested):

internal/config/config.go
stevenh: Agreed, hiding errors is bad and replacing a config, even one with errors, is even worse as the user is unlikely to notice and will be confused that things are not behaving correctly.

internal/core/docker_context.go
stevenh: I'll open my eyes next time 🤣

docker.go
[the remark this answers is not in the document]
mdelapenya: yes, I verified this PR against podman, rancher desktop, colima, orbstack, docker, and testcontainers cloud

docker_test.go
```diff
@@ -1701,7 +1639,7 @@ func TestContainerWithNoUserID(t *testing.T) {
 func TestGetGatewayIP(t *testing.T) {
 	// When using docker compose with DinD mode, and using host port or http wait strategy
 	// It's need to invoke GetGatewayIP for get the host
-	provider, err := providerType.GetProvider(WithLogger(TestLogger(t)))
+	provider, err := ProviderDocker.GetProvider(WithLogger(TestLogger(t)))
```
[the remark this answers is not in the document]
mdelapenya: I didn't want to rename here in this PR, but if we see related, I can deprecate the existing `DockerProvider` type and use `RuntimeProvider` instead.

internal/config/config.go
mdelapenya: He issue here is that this is called inside a `sync.Once`, should we panic if the config is incorrect?

internal/config/config.go
mdelapenya: Resolved in f4783a7bbe02fc582b58b0080f45c7f9bf1b01e2

internal/config/config.go
mdelapenya: Resolved in f4783a7bbe02fc582b58b0080f45c7f9bf1b01e2

internal/core/docker_context.go
mdelapenya: The only exposed function in this file is `GetDockerHostFromCurrentContext`, which starts the call to the types and functions here.

docs/system_requirements/using_colima.md
```diff
@@ -1,47 +0,0 @@
-# Using Colima with Docker
```
tmc, line 1: why remove this file?

douglascamata:
Would love to see this moving forward. The underlying issue that this fixes when using Colima is too annoying and has been overlooked for way too long.

douglascamata:
Sure, @mdelapenya. I think the issue is related to Docker context management and the location of the Colima managed Docker socket. 

Basically a lot of it has been reported in #2952. **In my case**, my tests that make use of `testcontainers` only work after I create a soft link of Colima's `docker.sock` (which is put at `$HOME/.config/colima/default/docker.sock` by Colima's `homebrew`formula) at `/var/run/docker.sock` by running `sudo ln -sf $HOME/.config/colima/default/docker.sock /var/run/docker.sock`. I'm using testcontaienrs-go version `v0.34.0`.

For me, trying to set the `DOCKER_HOST` env var like `export DOCKER_HOST="$(docker context inspect -f='{{.Endpoints.docker.Host}}')"` or `export DOCKER_HOST="unix://${HOME}/.config/colima/default/docker.sock"` does not work. With this option I get some weird errors creating networks and also some weird messages regarding the reaper container. I'm not sure if relevant, but I can try to reproduce one more time and grab the error message if you want.

douglascamata:
Come on, guys! This is almost having its first birthday. Will we see a solution one day?

douglascamata:
We all appreciate the time and effort @mdelapenya has put into getting this fixed. I think it deserves some attention from maintainers. If the solution here isn't ideal, we need a proper one.

mdelapenya:
Hi @douglascamata we're working on a docker/go-sdk that automatically does this in a transparent manner: https://github.com/docker/go-sdk/tree/main/client, testable examples in the repo and in [here](https://github.com/mdelapenya/go-sdk-examples).

It's part of my work as a maintainer in both projects to slowly integrate the new SDK into testcontainers-go, so I'm keeping this PR open until that happens, which 🤞 will happen soon.

jmhodges:
Any update on this one? If there's some blockers, I'm happy to also take a look.

mdelapenya:
> Any update on this one? If there's some blockers, I'm happy to also take a look.

@jmhodges we are going to consume https://github.com/docker/go-sdk as base runtime, which already supports that.

Issue #2808, opened 2024-10-02
[Enhancement]: use testify's require instead of assert.


mdelapenya:
### Proposal

> TL;DR: use `require` when a failure should stop the test completely, ensuring that subsequent steps are not executed with invalid assumptions. Use `assert` when we want to evaluate all assertions, even if some fail, to provide more comprehensive feedback on test results.

Why? 

By using `require` for crucial checks, we can avoid misleading error messages from failed assertions that come later in the test after an essential condition has already failed. This makes the root cause of the failure easier to diagnose, as the test execution halts as soon as the condition fails, preventing further unnecessary execution of test logic that depends on the success of the failed condition. This can save time, especially in long or complex tests, where subsequent steps may not be relevant after a failure.

`assert`, on the other hand, might allow tests to continue, resulting in multiple failures and potentially obscuring the real cause.

AshutoshKD:
Hey! I'd like to work on this

AshutoshKD:
Should the proposed change to be done on all the test files?

mdelapenya:
@AshutoshKD thanks for stepping up! Yes, you can start with a draft PR with the files you consider important, and we can iterate on it

JoelLau:
hi @mdelapenya & @AshutoshKD , i found a few more places that could benefit from this change. let me know what you think!

https://github.com/testcontainers/testcontainers-go/pull/2814

nolandseigler:
hi @mdelapenya & @AshutoshKD, I found another instance of the issue. I tried to add the "chore" label as the template suggests but I don't seem to have the access. I look forward to hearing from you. Thanks!

https://github.com/testcontainers/testcontainers-go/pull/3252

PR #2812, opened 2024-10-03
feat: add require for critical checks

AshutoshKD:
## What does this PR do?

Using require for essential checks to ensure that tests do not proceed under invalid assumptions, while assert can be used for less critical checks to gather more information about potential issues.

**Suggested Usage of assert and require :-**

**Critical Checks (Use require):**
Use require to check critical conditions that must be true for the test to proceed.
If a require condition fails, the test stops immediately. This prevents any further code from running, which could lead to misleading error messages. 

**Non-Critical Checks (Use assert):**
Use assert to evaluate conditions that are important but not critical for the test's overall success.
If an assert condition fails, the test continues to run, allowing multiple assertions to be evaluated.

## Why is it important?

It is important to use require for critical conditions so that if a critical check fails, the test execution halts immediately. This prevents the test from continuing to run subsequent assertions, which may rely on the validity of the failed condition, thereby avoiding misleading error messages and ensuring that only relevant failures are reported. This approach enhances the clarity of test results and facilitates quicker identification of the root cause of issues.

## Related issues

- https://github.com/testcontainers/testcontainers-go/issues/2808

### DIFF
--- a/container_test.go
+++ b/container_test.go
@@ -320,7 +320,7 @@ func TestCustomLabelsImage(t *testing.T) {
 	ctr, err := testcontainers.GenericContainer(ctx, req)
 
 	require.NoError(t, err)
-	t.Cleanup(func() { assert.NoError(t, ctr.Terminate(ctx)) })
+	t.Cleanup(func() { require.NoError(t, ctr.Terminate(ctx)) })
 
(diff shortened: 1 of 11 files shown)


mdelapenya (approved):
LGTM, thanks!

PR #2814, opened 2024-10-05
test: continue tests where possible

JoelLau:
## What does this PR do?

- replaced `assert` with `require` where possible

## Why is it important?

- increase the amount of information in each test by continuing execution
    - NOTE: i've only made changes where i think assertions can be made in parallel (in theory), where failing an assertion will not lead to an error state in subsequent tests

## Related issues

- Relates #2808

### DIFF
--- a/container_test.go
+++ b/container_test.go
@@ -357,8 +357,8 @@ func TestCustomLabelsBuildOptionsModifier(t *testing.T) {
 
 	ctrJSON, err := ctr.Inspect(ctx)
 	require.NoError(t, err)
-	require.Equal(t, myLabelValue, ctrJSON.Config.Labels[myLabelName])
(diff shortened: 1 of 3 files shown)


stevenh (changes requested):
I'm actually not a fan of this approach as it just tends to result in a lot of noise which creates little to no value, better to just fail on the first error.

mdelapenya:
Exactly, the intent is just the opposite, and use `require` over `assert` if/when possible.

stevenh:
Given this is only replacing `require` with `assert` which is moving in the wrong way, I'm going to close this as it will be easier to create a new one moving in the right direction.

JoelLau:
my bad, idk how i got it the wrong way around. thanks for looking at this and sorry for the noise

stevenh:
Thank you for the effort @JoelLau, its great to have community engagement, that's that open source is all about 🥇

PR #2825, opened 2024-10-15
feat: add yugabytedb module

henripqt:
## What does this PR do?

This PR introduces the yugabytedb go module 

## Why is it important?

There is currently no official yugabytedb test container module for golang

## Related issues

No related issue

## How to test this PR

```sh
cd modules/yugabytedb && make test
```


modules/yugabytedb/options.go
```diff
@@ -0,0 +1,53 @@
+package yugabytedb
+
+import (
+	"github.com/testcontainers/testcontainers-go"
+)
+
+// WithYSQLDatabaseName sets the initial database name for the yugabyteDB container.
+func WithYSQLDatabaseName(ysqlDBName string) testcontainers.CustomizeRequestOption {
```
stevenh, line 8: question: this seems overly verbose, do we need YSQL qualification, similar with the other functional options.
```suggestion
// WithDatabaseName sets the initial database name for the yugabyteDB container.
func WithDatabaseName(ysqlDBName string) testcontainers.CustomizeRequestOption {
```

modules/yugabytedb/yugabytedb.go
```diff
@@ -0,0 +1,246 @@
...
+//	connStr, err := y.YSQLConnectionString(ctx, "sslmode=disable")
+//	if err != nil {
+//		return err
+//	}
+//
+// The above code will return a connection string with the sslmode set to disable.
+func (y *YugabyteDBContainer) YSQLConnectionString(ctx context.Context, args ...string) (string, error) {
+	host, err := y.Host(ctx)
+	if err != nil {
+		return "", err
```
stevenh, line 170: suggestion: can we wrap the error so its clear to the user the source, more below
```suggestion
		return "", fmt.Errorf("host: %w", err)
```

modules/yugabytedb/yugabytedb.go
```diff
@@ -0,0 +1,246 @@
...
+	for _, env := range inspect.Config.Env {
+		parts := strings.SplitN(env, "=", 2)
+		switch parts[0] {
+		case ycqlKeyspaceEnv:
+			keyspace = parts[1]
+		case ycqlUserNameEnv:
+			user = parts[1]
+		case ycqlPasswordEnv:
+			password = parts[1]
+		}
```
stevenh, line 235: suggestion: use [strings.Cut](https://pkg.go.dev/strings#Cut) to validate, SplitN can return nil which would fail.
```suggestion
		key, val, found := strings.Cut(env, "=")
		if !found {
			continue
		}
		switch key {
		case ycqlKeyspaceEnv:
			keyspace = val
		case ycqlUserNameEnv:
			user = val
		case ycqlPasswordEnv:
			password = val
		}
```

modules/yugabytedb/yugabytedb.go
```diff
@@ -0,0 +1,246 @@
...
+	}
+
+	cfg.Hosts = append(cfg.Hosts, net.JoinHostPort(host, mappedPort.Port()))
+	cfg.Keyspace = keyspace
+	cfg.Authenticator = gocql.PasswordAuthenticator{
+		Username: user,
+		Password: password,
+	}
```
stevenh, line 243: question: what happens if these were not found above or are blank?

modules/yugabytedb/yugabytedb_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+		db, err := sql.Open("postgres", ysqlConnStr)
+		require.NoError(t, err)
+		require.NotNil(t, db)
+
+		err = db.Ping()
+		require.NoError(t, err)
+	})
+
+	t.Run("Run with custom options", func(t *testing.T) {
```
stevenh, line 44: suggestion: avoid spaces in names as it prevents a simple search from finding the test, more below.
```suggestion
	t.Run("custom-options", func(t *testing.T) {
```

modules/yugabytedb/yugabytedb_test.go
```diff
@@ -0,0 +1,110 @@
...
+		ctx := context.Background()
+		opts := []testcontainers.ContainerCustomizer{
+			yugabytedb.WithYSQLDatabaseName("custom-db"),
+			yugabytedb.WithYSQLDatabaseUser("custom-user"),
+			yugabytedb.WithYSQLDatabasePassword("custom-password"),
+		}
+
+		ctr, err := yugabytedb.Run(ctx, "yugabytedb/yugabyte:2024.1.3.0-b105", opts...)
```
stevenh, line 52: suggestion: as this isnt being dynamically calculated use the literal to simplify.
```suggestion
		ctr, err := yugabytedb.Run(ctx, "yugabytedb/yugabyte:2024.1.3.0-b105",
			yugabytedb.WithYSQLDatabaseName("custom-db"),
			yugabytedb.WithYSQLDatabaseUser("custom-user"),
			yugabytedb.WithYSQLDatabasePassword("custom-password"),
		)
```

modules/yugabytedb/yugabytedb_test.go
```diff
@@ -0,0 +1,110 @@
...
+		ctr, err := yugabytedb.Run(ctx, "yugabytedb/yugabyte:2024.1.3.0-b105")
+		testcontainers.CleanupContainer(t, ctr)
+		require.NoError(t, err)
+
+		cluster := gocql.NewCluster()
+		ctr.YCQLConfigureClusterConfig(ctx, cluster)
+
+		session, err := cluster.CreateSession()
+		require.NoError(t, err)
+		t.Cleanup(func() { session.Close() })
```
stevenh, line 88: suggestion: call directly and check error:
```suggestion
		require.NoError(t, session.Close())
```

modules/yugabytedb/yugabytedb_test.go
```diff
@@ -0,0 +1,110 @@
...
+	})
+
+	t.Run("Run with custom options", func(t *testing.T) {
+		ctx := context.Background()
+		opts := []testcontainers.ContainerCustomizer{
```
stevenh, line 93: suggestion: use literal as above.

modules/yugabytedb/examples_test.go
```diff
@@ -0,0 +1,112 @@
...
+		return
+	}
+
+	db, err := sql.Open("postgres", connStr)
+	if err != nil {
+		log.Printf("failed to open connection: %s", err)
+		return
+	}
+
+	defer db.Close()
```
stevenh, line 80: bug: missing `Output` for validation.

modules/yugabytedb/examples_test.go
```diff
@@ -0,0 +1,112 @@
...
+	cluster := gocql.NewCluster()
+	yugabytedbContainer.YCQLConfigureClusterConfig(ctx, cluster)
+
+	session, err := cluster.CreateSession()
+	if err != nil {
+		log.Printf("failed to create session: %s", err)
+		return
+	}
+
+	defer session.Close()
```
stevenh, line 111: bug: missing `Output` for validation.

stevenh (changes requested):
Thanks for the PR, here's some initial feedback

modules/yugabytedb/yugabytedb.go
henripqt: We always have default values set. The only way we could get in a situation where these are empty is if the user explicitly does so by using one of the provided options. 
I'm happy to still check for this situation and return a dedicated error though

modules/yugabytedb/yugabytedb_test.go
henripqt: Attended ✅

modules/yugabytedb/examples_test.go
henripqt: Attended ✅

modules/yugabytedb/examples_test.go
henripqt: Attended ✅

modules/yugabytedb/yugabytedb.go
henripqt: Attended ✅

modules/yugabytedb/options.go
henripqt: Attended ✅

modules/yugabytedb/options.go
```diff
@@ -0,0 +1,53 @@
+package yugabytedb
+
+import (
+	"github.com/testcontainers/testcontainers-go"
+)
+
+// WithDatabaseName sets the initial database name for the yugabyteDB container.
+func WithDatabaseName(ysqlDBName string) testcontainers.CustomizeRequestOption {
```
stevenh, line 8: Thanks, I think that's much more readable now 👍

modules/yugabytedb/yugabytedb.go
```diff
@@ -0,0 +1,249 @@
...
+	ysqlDatabaseUserEnv     = "YSQL_USER"
+	ysqlDatabasePasswordEnv = "YSQL_PASSWORD"
+
+	ysqlDatabaseName     = "yugabyte"
+	ysqlDatabaseUser     = "yugabyte"
+	ysqlDatabasePassword = "yugabyte"
+)
+
+// YugabyteDBContainer represents the yugabyteDB container type used in the module
+type YugabyteDBContainer struct {
```
stevenh, line 44: suggestion: not sure if this is part of the generation template, if so we should consider changing that too in a separate PR, but this stutters in use e.g. `yugabytedb.YugabyteDBContainer` we should simplify to avoid that:
```suggestion
// Container represents the yugabyteDB container type used in the module.
type Container struct {
```

modules/yugabytedb/yugabytedb.go
stevenh: That sounds like a good idea if the resulting cfg is invalid if they are blank, if blank is valid then I think we can leave as is, what do you think @henripqt ?

modules/yugabytedb/yugabytedb_test.go
stevenh: I think I saw this fixed in a previous version, did it get lost? Another below too, as we only need a t.Cleanup if the test is going to continue.

modules/yugabytedb/yugabytedb.go
```diff
@@ -0,0 +1,249 @@
...
+}
+
+// Run creates an instance of the yugabyteDB container type and automatically starts it.
+// A default configuration is used for the container, but it can be customized using the
+// provided options.
+// When using default configuration values it is recommended to use the provided
+// [*YugabyteDBContainer.YSQLConnectionString] and [*YugabyteDBContainer.YCQLConfigureClusterConfig]
+// methods to use the container in their respective clients.
+//
+// # Example using YSQL dialect:
```
stevenh, line 59: question: I'm always nervous about examples in comments as they can be broken or go stale without people noticing. We also have actual examples, so would be tempted to say lets remove these. 

@mdelapenya what do you think?

stevenh (changes requested):

modules/yugabytedb/yugabytedb_test.go
henripqt: I've reverted it as the `Close` method of the `gocpl.Session` does not return an error

modules/yugabytedb/yugabytedb.go
henripqt: This is indeed part of the auto genrated template. Though I agree with you something like `yugabytedb.Container` makes more sense and seems more idiomatic.

modules/yugabytedb/yugabytedb.go
henripqt: I've just tested it out using the following which results in a valid session therefore we should not error out on empty configuration values 👍 

```go
func ExampleYugabyteDBContainer_YCQLConfigureClusterConfig() {
	ctx := context.Background()

	yugabytedbContainer, err := yugabytedb.Run(
		ctx,
		"yugabytedb/yugabyte:2024.1.3.0-b105",
		yugabytedb.WithKeyspace(""),  // setting empty keyspace
		yugabytedb.WithUser(""),          // setting empty user
		yugabytedb.WithPassword(""),  // setting empty pwd
	)

	if err != nil {
		log.Printf("failed to start container: %s", err)
		return
	}

	defer func() {
		if err := testcontainers.TerminateContainer(yugabytedbContainer); err != nil {
			log.Printf("failed to terminate container: %s", err)
		}
	}()

	cluster := gocql.NewCluster()
	yugabytedbContainer.YCQLConfigureClusterConfig(ctx, cluster)

	session, err := cluster.CreateSession()
	if err != nil {
		log.Printf("failed to create session: %s", err)
		return
	}

	defer session.Close()

	var i int
	if err := session.Query(`
		SELECT COUNT(*) 
		FROM system_schema.keyspaces 
		WHERE keyspace_name = 'system' // since we don't have a custom keyspace we target the system one 
	`).Scan(&i); err != nil {
		log.Printf("failed to scan row: %s", err)
		return
	}

	fmt.Println(i)
	// Output: 1
}
```

modules/yugabytedb/options.go
henripqt: You're welcome :) Happy to receive your feedbacks :)

modules/yugabytedb/yugabytedb.go
henripqt: Looked around and it seems that this is indeed not a common practice in the repository. Happy to remove if you want.

modules/yugabytedb/yugabytedb.go
mdelapenya: Indeed. It's changed in v1: all modules will have a `Container` struct instead of an `XXXContainer`. Changing that would be a big BC that we should postpone for v1

modules/yugabytedb/yugabytedb.go
mdelapenya: I'd prefer having a testable example demonstrating how to use it. Wdyt?

modules/yugabytedb/yugabytedb.go
```diff
@@ -0,0 +1,249 @@
...
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"strings"
+
+	"github.com/testcontainers/testcontainers-go"
+	"github.com/testcontainers/testcontainers-go/wait"
+	"github.com/yugabyte/gocql"
```
mdelapenya, line 11: Nit: can you verify golangci-lint is not complaining about the order of the imports? I have a feeling that it will 🤞

modules/yugabytedb/yugabytedb_test.go
```diff
@@ -0,0 +1,108 @@
...
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/yugabyte/gocql"
+
+	_ "github.com/lib/pq"
+	"github.com/testcontainers/testcontainers-go"
+	"github.com/testcontainers/testcontainers-go/modules/yugabytedb"
```
mdelapenya, line 15: Nit: same here about the imports and golangci-lint

docs/modules/yugabytedb.md
```diff
@@ -0,0 +1,76 @@
...
+
+When starting the yugabyteDB container, you can pass options in a variadic way to configure it.
+
+#### Image
+
+If you need to set a different yugabyteDB Docker image, you can set a valid Docker image as the second argument in the `Run` function.
+E.g. `Run(context.Background(), "yugabytedb/yugabyte")`.
+
+{% include "../features/common_functional_options.md" %}
+
```
mdelapenya, line 51: Docs: Could you add docs for the options we are adding in this module? WithDatabaseName, WithDatabaseUser, etc.

We have examples in the MySQL/Postgres modules as reference

modules/yugabytedb/yugabytedb.go
stevenh: Agreed, looks like we have examples which cover these. If there are any missing just add concrete examples and remvoe these from the comments.

modules/yugabytedb/yugabytedb.go
henripqt: Alright I'll adjust this one to Container instead then and I'll open up a new PR to adjust the generator

modules/yugabytedb/yugabytedb.go
henripqt: ✅ Attended

modules/yugabytedb/yugabytedb.go
henripqt: ✅ Attended

docs/modules/yugabytedb.md
henripqt: ✅ Attended

modules/yugabytedb/yugabytedb_test.go
henripqt: I've ran the linter but no issues on my side

modules/yugabytedb/yugabytedb.go
henripqt: @mdelapenya as promised here's the PR https://github.com/testcontainers/testcontainers-go/pull/2831/files
I also fixed an issue related to missing imports

modules/yugabytedb/yugabytedb_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+		db, err := sql.Open("postgres", ysqlConnStr)
+		require.NoError(t, err)
+		require.NotNil(t, db)
+
+		err = db.Ping()
+		require.NoError(t, err)
+	})
+
+	t.Run("custom-options", func(t *testing.T) {
```
mdelapenya, line 44: bug: when I run this test, it always fail, I guess because, when a database is set, the wait strategy should include a condition for the database is ready to accept connections. I _think_ this is the case, because if I debug the test, step by step, it always pass, so it gives me the sense it's a race condition.

```
    /Users/mdelapenya/sourcecode/src/github.com/testcontainers/testcontainers-go/modules/yugabytedb/yugabytedb_test.go:69:
        	Error Trace:	/Users/mdelapenya/sourcecode/src/github.com/testcontainers/testcontainers-go/modules/yugabytedb/yugabytedb_test.go:69
        	Error:      	Received unexpected error:
        	            	pq: password authentication failed for user "custom-user"
        	Test:       	TestYugabyteDB_YSQL/custom-options
```

modules/yugabytedb/yugabytedb_test.go
```diff
@@ -0,0 +1,108 @@
...
+			yugabytedb.WithKeyspace("custom-keyspace"),
+			yugabytedb.WithUser("custom-user"),
+			yugabytedb.WithPassword("custom-password"),
+		)
+
+		testcontainers.CleanupContainer(t, ctr)
+		require.NoError(t, err)
+
+		cluster := gocql.NewCluster()
+		ctr.YCQLConfigureClusterConfig(ctx, cluster)
```
mdelapenya, line 102: Ditto: handle the error
```suggestion
		require.NoError(t, ctr.YCQLConfigureClusterConfig(ctx, cluster))
```

modules/yugabytedb/yugabytedb_test.go
```diff
@@ -0,0 +1,108 @@
...
+func TestYugabyteDB_YCQL(t *testing.T) {
+	t.Run("Run", func(t *testing.T) {
+		ctx := context.Background()
+
+		ctr, err := yugabytedb.Run(ctx, "yugabytedb/yugabyte:2024.1.3.0-b105")
+		testcontainers.CleanupContainer(t, ctr)
+		require.NoError(t, err)
+
+		cluster := gocql.NewCluster()
+		ctr.YCQLConfigureClusterConfig(ctx, cluster)
```
mdelapenya, line 82: bug: handle the error
```suggestion
		require.NoError(t, ctr.YCQLConfigureClusterConfig(ctx, cluster))
```

modules/yugabytedb/yugabytedb_test.go
```diff
@@ -0,0 +1,108 @@
...
+		db, err := sql.Open("postgres", ysqlConnStr)
+		require.NoError(t, err)
+		require.NotNil(t, db)
+
+		err = db.Ping()
+		require.NoError(t, err)
+	})
+}
+
+func TestYugabyteDB_YCQL(t *testing.T) {
```
mdelapenya, line 73: Same behaviour as above: debugging each subtest makes them pass, but simply running them fails with:

```
    /Users/mdelapenya/sourcecode/src/github.com/testcontainers/testcontainers-go/modules/yugabytedb/yugabytedb_test.go:85:
        	Error Trace:	/Users/mdelapenya/sourcecode/src/github.com/testcontainers/testcontainers-go/modules/yugabytedb/yugabytedb_test.go:85
        	Error:      	Received unexpected error:
        	            	gocql: unable to create session: unable to discover protocol version: code=100; message=Provided username 'yugabyte' and/or password are incorrect
        	Test:       	TestYugabyteDB_YCQL/Run
```

modules/yugabytedb/examples_test.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	defer func() {
+		if err := testcontainers.TerminateContainer(yugabytedbContainer); err != nil {
+			log.Printf("failed to terminate container: %s", err)
+		}
+	}()
+
+	cluster := gocql.NewCluster()
+	yugabytedbContainer.YCQLConfigureClusterConfig(ctx, cluster)
```
mdelapenya, line 115: bug: we need to handle the error here

modules/yugabytedb/examples_test.go
```diff
@@ -0,0 +1,138 @@
...
+		FROM system_schema.keyspaces 
+		WHERE keyspace_name = 'yugabyte'
+	`).Scan(&i); err != nil {
+		log.Printf("failed to scan row: %s", err)
+		return
+	}
+
+	fmt.Println(i)
+
+	// Output: 1
```
mdelapenya, line 137: This test always fail for me locally, unless I debug it step-by-step, which makes me think there is a race condition at some point.

mdelapenya (commented):
Hey @henripqt, thanks for this new module! I added a few comments after testing this PR locally. It seems there is a race condition that causes the tests to fail if I run them, but they always pass if I debug them step-by-step. So my first impression is that we need a more consistent wait strategy. I have no experience with yugabyte, so I'd appreciate if you could do the research on how to fix that. Can you double check how the Java implementation is building them ([here](https://github.com/testcontainers/testcontainers-java/blob/main/modules/yugabytedb/src/main/java/org/testcontainers/containers/strategy/YugabyteDBYCQLWaitStrategy.java) and [here](https://github.com/testcontainers/testcontainers-java/blob/main/modules/yugabytedb/src/main/java/org/testcontainers/containers/strategy/YugabyteDBYSQLWaitStrategy.java))?

Thanks!

docs/modules/yugabytedb.md
```diff
@@ -49,6 +49,18 @@ E.g. `Run(context.Background(), "yugabytedb/yugabyte")`.
 
 {% include "../features/common_functional_options.md" %}
 
+#### Initial Database
+
+By default the yugabyteDB container will start with a database named `yugabyte` and the default credentials `yugabyte` and `yugabyte`.
+
+If you need to set a different database, and its credentials, you can use the `WithDatabaseName(dbName string)`, `WithDatabaseUser(dbUser string)` and `WithDatabasePassword(dbPassword string)` options.
```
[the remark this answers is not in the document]
henripqt: Yes that is correct.

modules/yugabytedb/examples_test.go
henripqt: ✅ Attended

modules/yugabytedb/yugabytedb_test.go
henripqt: Good catch. Only ran it in debug mode which never failed. I've added an additional log constraint which works on my side.

modules/yugabytedb/yugabytedb_test.go
henripqt: Good catch. Only ran it in debug mode which never failed. I've added an additional log constraint which works on my side.

modules/yugabytedb/yugabytedb.go
```diff
@@ -0,0 +1,183 @@
...
+		y.ysqlDatabaseName,
+		strings.Join(args, "&"),
+	), nil
+}
+
+// YCQLConfigureClusterConfig merges the provider [gocql.ClusterConfig] with the
+// yugabyteDB container configuration.
+// The method sets the hosts, keyspace and authenticator fields of the cluster
+// configuration to the values of the yugabyteDB container.
+func (y *Container) YCQLConfigureClusterConfig(ctx context.Context, cfg *gocql.ClusterConfig) error {
```
mdelapenya, line 138: blocker: we usually do not bundle any client/sdk into the module, to let users choose what they need to connect to the given technology. I'd recommend having this method as a helper function in the tests/examples, so people as a reference on how to bootstrap the config from the container state.

Therefore, the `yugabyte/gocql` package would be used just in the tests, not in the "production" container.

Wdyt?

mdelapenya (commented):
I left just one final comment regarding the usage of the yugabyte client package. Other than that, and after a thorough review from @stevenh (🙇), this PR LGTM, and it's ready to be merged after that discussion is resolved.

Thank you for your time, and for this new module 🙏

modules/yugabytedb/yugabytedb.go
```diff
@@ -61,6 +61,7 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
 		Cmd:   []string{"bin/yugabyted", "start", "--background=false"},
 		WaitingFor: wait.ForAll(
 			wait.ForLog("YugabyteDB Started").WithOccurrence(1),
+			wait.ForLog("Data placement constraint successfully verified").WithOccurrence(1),
```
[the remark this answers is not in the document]
henripqt: Always needed as far as my tests go

modules/yugabytedb/yugabytedb.go
henripqt: Fair enough. I agree, this could become a maintenance hell with versioning etc.. 
I'll make the changes 👍

henripqt: ✅ Attended

henripqt: @mdelapenya Regarding this point may I suggest we add something around it in the contributing guidelines ?

mdelapenya:
@henripqt I added two commits to avoid you the burden of doing it: they just run `make tidy` and `make lint` (see last two commits).

I think once the CI passes, we are good to go. Thank you!

modules/yugabytedb/examples_test.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+func ExampleContainer_newCluster() {
```
mdelapenya, line 94: test: This test is failing both locally and on the CI. Could you take a look? 🙏

henripqt: Will have a look

henripqt: This is now fixed. Forgot to add the port to the cluster addr while rushing through changes. Apologies 🙏

mdelapenya (approved):
LGTM, thanks for your patience during the review 🙇 and for improving the library with this new module 🚀

mdelapenya (approved):
LGTM, thanks!

Issue #2802, opened 2024-09-26
[Bug]: artemis random test failures


stevenh:
### Testcontainers version

head

### Using the latest Testcontainers version?

Yes

### Host OS

N/A

### Host arch

N/A

### Go version

N/A

### Docker version

```shell
N/A
```

### Docker info

```shell
N/A
```

### What happened?

Tests are [randomly failing](https://github.com/testcontainers/testcontainers-go/actions/runs/11056380354/job/30718318693?pr=2794)

### Relevant log output

```shell
FAIL TestArtemis/WithCredentials (4.94s)
  2024/09/26 16:42:35 🐳 Creating container for image docker.io/apache/activemq-artemis:2.30.0-alpine
  2024/09/26 16:42:35 ✅ Container created: 9b972e4d34b1
  2024/09/26 16:42:35 🐳 Starting container: 9b972e4d34b1
  2024/09/26 16:42:35 ✅ Container started: 9b972e4d34b1
  2024/09/26 16:42:35 ⏳ Waiting for container id 9b972e4d34b1 image: docker.io/apache/activemq-artemis:2.30.0-alpine. Waiting for: &{timeout:<nil> deadline:<nil> Strategies:[0xc0002bc870 0xc0002bc8a0]}
  2024/09/26 16:42:39 🔔 Container is ready: 9b972e4d34b1
      artemis_test.go:100: 
          	Error Trace:	/home/runner/work/testcontainers-go/testcontainers-go/modules/artemis/artemis_test.go:100
          	Error:      	Received unexpected error:
          	            	invalid command
          	Test:       	TestArtemis/WithCredentials
          	Messages:   	failed to connect
  2024/09/26 16:42:39 🐳 Stopping container: 9b972e4d34b1
  2024/09/26 16:42:40 ✅ Container stopped: 9b972e4d34b1
  2024/09/26 16:42:40 🐳 Terminating container: 9b972e4d34b1
  2024/09/26 16:42:40 🚫 Container terminated: 9b972e4d34b1

-----
FAIL TestArtemis/WithAnonymous (5.02s)
  2024/09/26 16:42:40 🐳 Creating container for image docker.io/apache/activemq-artemis:2.30.0-alpine
  2024/09/26 16:42:40 ✅ Container created: f465d5396d64
  2024/09/26 16:42:40 🐳 Starting container: f465d5396d64
  2024/09/26 16:42:40 ✅ Container started: f465d5396d64
[...]

PR #2827, opened 2024-10-16
chore: replace 'assert' with 'require'

JoelLau:
ran through the codebase - these should be the last items for #2802

## What does this PR do?
- use `require` wherever possible

## Why is it important?
- reduce noise in tests by failing early

## Related issues

- ~~Closes~~ Relates #2808
EDITED by @mdelapenya, as we do not want to close that issue yet. 🙏


mdelapenya (approved):
LGTM, thanks!

PR #2830, opened 2024-10-16
docs: go get latest (potentially untagged) module in quickstart

rollulus:
Using a go get without the version installs the latest tagged release. The documentation on the website can be more recent, potentially relying on functionality that's not in the latest tagged release yet.

## What does this PR do?

Instruct to go get @main in quickstart.

## Why is it important?

Using a go get without the version installs the latest tagged release. The documentation on the website can be more recent, potentially relying on functionality that's not in the latest tagged release yet.

## Related issues

#511

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -12,7 +12,7 @@ Please read the [system requirements](../system_requirements/) page before you s
 We use [go mod](https://blog.golang.org/using-go-modules) and you can get it installed via:
 
 ```
-go get github.com/testcontainers/testcontainers-go
+go get github.com/testcontainers/testcontainers-go@main
 ```
 
 ## 3. Spin up Redis


mdelapenya:
@stevenh you are right. Given #511 is still around, regarding the docs system and the multiverse, I'd consider it website as a best effort towards sharing knowledge about main. From the modules standpoint, we always try to add the version functions and methods appeared for a given module, which is what pkg.go.dev did pretty well. So, following on @stevenh's advice, we could add that message linking to pkg.go.dev in the quickstart guide, with the warning regarding the versioning. 

I think I can dedicate time to look at #511 again, but I'll be after the ongoing PRs regarding the reaper and the new modules.

rollulus:
Ok, got it, I think this then ends what I can do as a random visitor. Thanks.

mdelapenya:
In any case thanks @rollulus for your random visit :) I helped us in talking again about #511 

Please come again and provide any feedback you consider, specially about docs. We are doing our best in keeping them up-to-date, but many times is difficult because of the bandwidth.

Cheers!

PR #2831, opened 2024-10-17
fix: update module container struct name and missing imports

henripqt:
## What does this PR do?

Fixes the following in modulegen : 
- The generated name of the struct for the module is now `Container` leading to more idiomatic go (e.g postgres.Container)
- Missing imports in test generated files
- Removed `// run{{ $title }}Container {` from the example test as we can use the name of the Example{{$entrypoint}} instead

## Why is it important?

New generated modules will be more idiomatic and won't have to fix imports manually

## Related issues

No related issue but a request was made in this PR https://github.com/testcontainers/testcontainers-go/pull/2825#discussion_r1801841535 to name the container struct as `Container` from now on in preparation of v1

## How to test this PR

Generate a new module


mdelapenya:
Thanks for this! I'm going to add a few commits on top to avoid you suffer the burden of updating the modulegen tests 🙏

mdelapenya (approved):
LGTM, thanks!

PR #2833, opened 2024-10-18
chore: update ryuk to 0.10.2

stevenh:
Update the ryuk version so that shutdown processing is fast after a TERM signal.


mdelapenya (approved):
LGTM, thanks!

PR #2836, opened 2024-10-18
fix: enable formatter.require-f-funcs from testifylint

mmorel-35:
## What does this PR do?

Setup formatter.require-f-funcs from [testifylint](https://golangci-lint.run/usage/linters/#testifylint) linter and apply it’s recommandations .

## Why is it important?

To require f-assertions if format string is used.

Signed-off-by: Matthieu MOREL <matthieu.morel35@gmail.com>

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -32,5 +32,7 @@ linters-settings:
       - float-compare
       - go-require
     enable-all: true
+    formatter:
+      require-f-funcs: true
 run:
   timeout: 5m
--- a/container_test.go
+++ b/container_test.go
@@ -376,8 +376,7 @@ func Test_GetLogsFromFailedContainer(t *testing.T) {
 		Started:          true,
 	})
 	testcontainers.CleanupContainer(t, c)
-	require.Error(t, err)
-	require.Contains(t, err.Error(), "container exited with code 0")
+	require.ErrorContains(t, err, "container exited with code 0")
 
 	logs, logErr := c.Logs(ctx)
 	require.NoError(t, logErr)
--- a/docker_test.go
+++ b/docker_test.go
@@ -305,9 +305,9 @@ func TestContainerStateAfterTermination(t *testing.T) {
 		require.NoError(t, err)
 
 		state, err := nginx.State(ctx)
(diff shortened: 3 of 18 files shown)


mdelapenya:
What is your experience with this rule? I think we are doing good progress in throwing more meaningful error messages, which will be part of the test failure in the case require.Error fails. So I'm not very sure about including this one. Wdyt?

stevenh (commented):
I don't understand this one, could you clarify why this change is beneficial?

stevenh:
> I probably missed something because it's not the intention of the rule to reduce the readability of the test results, on the contrary. Let's check.
> 
> The rule reason is explained here https://github.com/Antonboom/testifylint?tab=readme-ov-file#formatter
> 
> There is also more to read https://github.com/Antonboom/testifylint?tab=readme-ov-file#historical-reference

Yer that's there to eliminate an embedded `fmt.Sprintf` as those messages are just raw strings it's not needed.

So its needed for this:
```go
assert.Errorf(t, err, fmt.Sprintf("test %s", test.testName)) // Bad
assert.Errorf(t, err, "test %s", test.testName) // Good
```

But not needed for this:
```go
assert.Error(t, err, "thing should error") // Not needed as no replaceable variables
```

stevenh:
> Have look over here, [Antonboom/testifylint#177 (comment)](https://github.com/Antonboom/testifylint/issues/177#issuecomment-2386908481) I asked the same question to testifylint's author

Thanks for that context, really useful.

Having read the history and looked at the code for assert / require, I believe there's an case which was overlooked. If you trace the calls you'll end up in [messageFromMsgAndArgs](https://github.com/stretchr/testify/blob/v1.9.0/assert/assertions.go#L280) which has a specific case for a single argument as can be found [here](https://github.com/stretchr/testify/blob/v1.9.0/assert/assertions.go#L284).

So `f` versions are actually only required in the case where there is more than one argument.

If that rule doesn't handle that edge case I suggest we leave it disabled as it would cause churn for no real value, I would even suggest that including a reason for things like `NoError` add little to no value so would be better removed.

stevenh:
Yes @mmorel-35 nice catch, that should be an `ErrorIs`

PR #2843, opened 2024-10-21
fix!: data races

stevenh:
Enable race checks for all tests in CI.

Fix data race when determining default network, this required making `DockerProviderOptions.DefaultNetwork` field private which is a breaking change.

Fix data race in test bufLogger.

Fix data races on log production context cancellation and context timeout not being cancelled in read loop.

BREAKING_CHANGE!


docker.go
```diff
@@ -1563,43 +1571,50 @@ func (p *DockerProvider) GetGatewayIP(ctx context.Context) (string, error) {
...
-func (p *DockerProvider) getDefaultNetwork(ctx context.Context, cli client.APIClient) (string, error) {
-	// Get list of available networks
-	networkResources, err := cli.NetworkList(ctx, network.ListOptions{})
-	if err != nil {
-		return "", err
-	}
+// ensureDefaultNetwork ensures that defaultNetwork is set and creates
+// it if it does not exist, returning its value.
+// It is safe to call this method concurrently.
+func (p *DockerProvider) ensureDefaultNetwork(ctx context.Context) (string, error) {
```
mdelapenya, line 1577: This code will be eventually removed with the docker context support (see #2810)

mdelapenya (approved):
LGTM, thanks!

PR #2844, opened 2024-10-21
fix(compose): container initialisation

### LINKED ISSUE #2667: [Bug]: Creation of DockerContainer results invalid state
### Testcontainers version

0.32.0

Yes

### Host OS

N/A

### Host arch

N/A

### Go version

N/A

### Docker version

```shell
N/A
```

### Docker info

```shell
N/A
```

### What happened?

Containers created by compose module result in an invalid state due to private fields.

### Relevant log output

_No response_

### Additional information

We should look at making all fields private and exposing constructors such as the currently private `containerFromDockerResponse`.

I tripped over this as I added an optimisation check using `isRunning` at the top of `Stop`, but this caused failures in `TestDockerComposeAPIWithStopServices` because the instance created wasn't fully initialised.

stevenh:
Fix compose to fully initialise the containers it returns. This ensures that running things like checks for running behave as expected.

Extracts the functionality to connect to reaper into a helper method so its consistent across uses.

Fix data race in daemonHost function converting it to a method to make use of encapsulation.

Export the functionality to create a container from a ContainerList response via provider.ContainerFromType.

Enforce no bare returns instead of no named returns as that was the original intention.

Fixes #2667


container.go
```diff
@@ -167,6 +167,15 @@ type ContainerRequest struct {
 	LogConsumerCfg          *LogConsumerConfig                         // define the configuration for the log producer and its log consumers to follow the logs
 }
 
+// sessionID returns the session ID for the container request.
+func (c *ContainerRequest) sessionID() string {
+	if sessionID := c.Labels[core.LabelSessionID]; sessionID != "" {
```
mdelapenya, line 172: If users pass their own sessionId value here, then the containers (and the networks) won't use the library's mechanism to determine the real sessionID (from parent go test process).

Is this intended?

mdelapenya: We checked that the user-defined labels are merged with the core labels, raising an error if the user-defined ones prepends the `org.testcontainers` prefix. We should notice that in the ContainerRequest struct's Labels field.

mdelapenya (approved):
LGTM, thanks!

PR #2845, opened 2024-10-22
fix(influxdb): Respect custom waitStrategy

marcinmilewski93:
## What does this PR do?

The purpose of this PR is to respect the provided `waitStrategy` for the InfluxDB container, rather than overriding it by default. This change allows the use of a custom `waitStrategy` when specified; if none is provided, the default `waitStrategy` will be applied, maintaining the same configuration as before this modification.

## Why is it important?

In the current implementation, it's not possible to use a custom `waitStrategy` and in my case the default strategy marks the container as ready too early. The container is still not ready to receive writes as you can see below. Customizing the `waitStrategy` for example to check the `/health` status solved this issue. Example log output when container is not ready yet:
```
2024/10/22 10:38:54 ✅ Container started: d5d6612ccd27
2024/10/22 10:38:54 ⏳ Waiting for container id d5d6612ccd27 image: influxdb:2.0.7. Waiting for: &{timeout:<nil> Log:Listening log_id=[0-9a-zA-Z_~]+ service=tcp-listener transport=http IsRegexp:true Occurrence:1 PollInterval:100ms}
2024/10/22 10:39:04 🔔 Container is ready: d5d6612ccd27
2024-10-22T10:39:23.408+07:00|2|ERROR||,,,testing-invalid|9469|goroutine:1783|internal/runner.go:93|Post "http://localhost:49155/api/v2/write?bucket=bucket1&org=influxdata&precision=ns": read tcp [::1]:43726->[::1]:49155: read: connection reset by peer
2024/10/22 10:39:24 [E]! Write error: Post "http://localhost:49155/api/v2/write?bucket=lbucket1&org=influxdata&precision=ns": read tcp [::1]:55852->[::1]:49155: read: connection reset by peer
2024/10/22 10:39:24 unable to send metrics to InfluxDB. err=Post "http://localhost:49155/api/v2/write?bucket=bucket1&org=influxdata&precision=ns": read tcp [::1]:55852->[::1]:49155: read: connection reset by peer
```

After this modification different strategy could be employed, for example, to verify if InfluxDB is `ready for queries and writes` as you can see on below health check response:
```json
{
  "name": "influxdb",
  "message": "ready for queries and writes",
  "status": "pass",
  "checks": [],
  "version": "2.0.7",
  "commit": "2a45f0c037"
}
```

## Related issues
- None

## How to test this PR

I have added a test that uses a custom `waitStrategy` to verify this behavior.


modules/influxdb/influxdb_test.go
```diff
@@ -48,6 +50,29 @@ func TestV2Container(t *testing.T) {
 	}
 }
 
+func TestWithWaitStrategy(t *testing.T) {
```
stevenh, line 53: question: This test passes without issue for me without any modifications, so I'm not sure I'm fully following as it looks you can customisation is already possible?

modules/influxdb/influxdb_test.go
```diff
@@ -48,6 +50,29 @@ func TestV2Container(t *testing.T) {
...
+	)
+	testcontainers.CleanupContainer(t, influxDbContainer)
+	require.NoError(t, err)
+
+	state, err := influxDbContainer.State(ctx)
+	require.NoError(t, err)
+
+	if !state.Running {
+		t.Fatal("InfluxDB container is not running")
+	}
```
stevenh, line 73: suggestion: we should simplify this with a require
```suggestion
	require.True(t, state.Running)
```

stevenh (changes requested):
Thanks for PR, tried reproducing but must be missing something

modules/influxdb/influxdb_test.go
marcinmilewski93: - In my case, even the current default strategy sometimes passes. This seems to depend on the machine's available resources, as it consistently fails on my low-resources Jenkins setup.

- Unfortunately, full customization is currently not possible because WaitingFor is overridden, for example, in [this line](https://github.com/testcontainers/testcontainers-go/blob/e94530939173164b80faf6dd5870410cd3caf7fe/modules/influxdb/influxdb.go#L75)

modules/influxdb/influxdb.go
```diff
@@ -57,39 +72,24 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
...
-				genericContainerReq.WaitingFor,
+				wait.ForListeningPort("8086/tcp"),
 				wait.ForLog("influxdb init process in progress..."),
 				wait.ForLog("Server shutdown completed"),
 				wait.ForLog("Opened shard"),
 			}
-			genericContainerReq.WaitingFor = wait.ForAll(strategies...)
-			hasInitDb = true
-			break
+			return wait.ForAll(strategies...)
```
mdelapenya, line 80: Question: do we want to return here, without processing/appending the wait for log added by the image version tag?

marcinmilewski93: I chose not to append it to maintain consistency with the previous implementation. However, I’m uncertain about the best approach here, as I don’t have experience with InfluxDB 1.x or `docker-entrypoint-initdb.d` files. I would appreciate input from someone more familiar with these areas to help provide a clearer solution.

modules/influxdb/influxdb_test.go
stevenh: suggestion: That last point seems to be the thing that should be fixed, and if the health endpoint is always available switching to that as the default would make a lot of sense.

In general we're trying to move away from `wait.ForLog` as its fragile.

marcinmilewski93: I checked the health check response for InfluxDB 1, and it looks like this:
```json
{
  "checks": [],
  "message": "ready for queries and writes",
  "name": "influxdb",
  "status": "pass",
  "version": "1.8.10"
}
```
Based on this, I've unified the default strategy for both versions of InfluxDB, simplifying it to two default strategies: one when configuration files are provided and one when they are not. I also removed the previously added tests since the health check verification is now covered by the standard tests.

modules/influxdb/influxdb.go
```diff
@@ -57,39 +73,23 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
...
-	if container != nil {
-		c = &InfluxDbContainer{Container: container}
-	}
-
-	if err != nil {
-		return c, fmt.Errorf("generic container: %w", err)
-	}
-
-	return c, nil
+			return strings.Contains(string(bs), "ready for queries and writes")
```
mdelapenya, line 91: I think the check for `"status": "pass",` could be more robust, but I do not have a strong opinion on which one is better.

modules/influxdb/influxdb.go
```diff
@@ -57,39 +73,23 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
...
-				genericContainerReq.WaitingFor,
+				wait.ForListeningPort("8086/tcp"),
 				wait.ForLog("influxdb init process in progress..."),
 				wait.ForLog("Server shutdown completed"),
 				wait.ForLog("Opened shard"),
 			}
-			genericContainerReq.WaitingFor = wait.ForAll(strategies...)
-			hasInitDb = true
-			break
+			return wait.ForAll(strategies...)
```
stevenh, line 81: suggestion: Testing this locally we should be able to use the http health check, but it would need to come after the first shutdown otherwise it could pass prematurely.

So the overall flow seems like it should be:
1. If init DB wait for DB init and shutdown
2. Wait for http health check to succeed.

Note the waits for 1 and 2 are combined, as well as any custom wait set by the user.

stevenh (changes requested):
Looking cleaner, did some testing locally which triggered some additional suggestions.

modules/influxdb/influxdb.go
```diff
@@ -47,8 +47,24 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
 		}
 	}
 
-	hasInitDb := false
+	if genericContainerReq.WaitingFor == nil {
+		genericContainerReq.WaitingFor = defaultWaitStrategy(genericContainerReq)
+	}
```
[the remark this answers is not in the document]
mdelapenya: I agree, I think we should append user-provided wait strategy with what we consider the default one.

modules/influxdb/influxdb.go
marcinmilewski93: I was undecided on which approach to use, so for simplicity, I used the same one as the health endpoint in `modules/openfga/openfga.go:61`. However, I agree that checking the JSON status for a pass value might be more appropriate. I’ll update it accordingly.

modules/influxdb/influxdb.go
```diff
@@ -57,39 +73,23 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
...
+			return wait.ForAll(strategies...)
 		}
 	}
 
-	if !hasInitDb {
-		if lastIndex := strings.LastIndex(genericContainerReq.Image, ":"); lastIndex != -1 {
-			tag := genericContainerReq.Image[lastIndex+1:]
-			if tag == "latest" || tag[0] == '2' {
-				genericContainerReq.WaitingFor = wait.ForLog(`Listening log_id=[0-9a-zA-Z_~]+ service=tcp-listener transport=http`).AsRegexp()
+	return wait.ForHTTP("/health").
```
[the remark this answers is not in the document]
marcinmilewski93: It looks like, except for InfluxDB 2, it’s failing for all existing tests anyway.

modules/influxdb/influxdb.go
marcinmilewski93: Alright, I’ll stick with wait.ForLog("Server shutdown completed") and then wait for the HTTP health check to pass.

modules/influxdb/influxdb.go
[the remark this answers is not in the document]
marcinmilewski93: I don’t agree with this approach for two reasons:

1. It doesn’t allow the caller to specify a fully custom waitStrategy. If, for instance, there’s a change in InfluxDB, the user wouldn’t be able to completely customize their waiting logic.
2. If the caller passes a MultiStrategy, appending could cause issues, as MultiStrategy doesn’t support nesting within another MultiStrategy—from my experience, this setup simply won’t work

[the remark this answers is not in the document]
mdelapenya: Interesting topics. About 2, we can maybe open an issue, although supporting multi-level wait strategies could be a very complex scenario not worth it to handle... or maybe it is. I'm not sure and would like to know your opinion.

About 1. You're right: once we couple the module to a given wait strategy, we are coupling the code to a given version of the technology (not only for influx, but for any module), so enabling users to change it sounds sensible to me. Thanks for bringing this thought.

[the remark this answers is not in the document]
stevenh: Good challenge, I would suggest this is the uncommon case but it would be nice to handle.

So here's an idea which I think handles all of this:
1. Set the default in the creation of `req` to the health check.
2. Have `WithInitDb` override the default with the needed changes

With this the user can still use a `CustomizeRequestOption` latter in the chain to override or write their own `WithInitDb` if that's more appropriate.

Thoughts?

modules/influxdb/influxdb.go
```diff
@@ -92,6 +65,37 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
...
+			}
+			return wait.ForAll(strategies...)
+		}
+	}
+	return waitForHttpHealth()
+}
+
+func waitForHttpHealth() *wait.HTTPStrategy {
+	return wait.ForHTTP("/health").
+		WithResponseMatcher(func(body io.Reader) bool {
```
stevenh, line 87: Nice!

modules/influxdb/influxdb.go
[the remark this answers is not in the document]
marcinmilewski93: I agree. Setting the default to the health check in the `req` creation and then having `WithInitDb` override it as needed will simplify the code. It also keeps the flexibility for users to still use a `CustomizeRequestOption` later in the chain or even write their own WithInitDb. This approach should cover all my concerns.

mdelapenya (approved):
LGTM, thanks!

stevenh (changes requested):
Almost there, just two suggestions

modules/influxdb/influxdb.go
```diff
@@ -147,9 +116,8 @@ func WithConfigFile(configFile string) testcontainers.CustomizeRequestOption {
 	}
 }
 
-// WithInitDb will copy a 'docker-entrypoint-initdb.d' directory to the container.
-// The secPath is the path to the directory on the host machine.
-// The directory will be copied to the root of the container.
+// WithInitDb returns a request customizer that initialises the database using the file `docker-entrypoint-initdb.d`
```
mdelapenya, line 119: Nit: I think the current comment is not explaining that srcPath is a dir, and it's not mentioning the addition of the wait strategy in case this option is used.

Sorry to be picky 🙏  just want to make sure the new comment completes the previous one.

mdelapenya (dismissed):
Added a final nit about a comment.

Other than that, LGTM, great job @marcinmilewski93 during the review 🙇

modules/influxdb/influxdb.go
mdelapenya: > having to have a specifically named file in a directory isn't a user friendly as it could be.

Totally agreed. Worth it to change that in a follow-up

marcinmilewski93: Okay, I’ve added additional information indicating that srcPath is a directory for now. Please take a look once again.

stevenh (approved):

mdelapenya:
Thanks everybody here for the thorough review. Great job!

PR #2846, opened 2024-10-22
fix: simplify fully-qualified image names

mdelapenya:
## What does this PR do?
This PR removes all the prefixes for `docker.io` from the images used in 1) tests for the library, 2) modules

## Why is it important?
It's better to ask for "nginx:latest" and let the engine deal with the image references than to enforce an invalid docker registry.

If we wanted to be explicit about official images, we should have used `https://index.docker.io/v1/library/nginx:latest` instead.

It also removes a bug when using Docker Desktop, as it considered those images were not accessible for the already used docker credentials.


kiview (dismissed):

container_test.go
```diff
@@ -392,11 +392,11 @@ func Test_GetLogsFromFailedContainer(t *testing.T) {
 type dockerImageSubstitutor struct{}
 
 func (s dockerImageSubstitutor) Description() string {
-	return "DockerImageSubstitutor (prepends docker.io)"
+	return "DockerImageSubstitutor (prepends registry.hub.docker.com)"
 }
 
 func (s dockerImageSubstitutor) Substitute(image string) (string, error) {
-	return "docker.io/" + image, nil
+	return "registry.hub.docker.com/library/" + image, nil
```
stevenh, line 399: question: should we have an private constant for this?

docker_files_test.go
```diff
@@ -66,7 +66,7 @@ func TestCopyFileToRunningContainer(t *testing.T) {
 
 	ctr, err := testcontainers.GenericContainer(ctx, testcontainers.GenericContainerRequest{
 		ContainerRequest: testcontainers.ContainerRequest{
-			Image: "docker.io/bash:5.2.26",
+			Image: "bash:5.2.26",
```
stevenh, line 69: question: others don't have a version, why is this one special?

modules/postgres/postgres_test.go
```diff
@@ -303,7 +303,7 @@ func TestSnapshotWithOverrides(t *testing.T) {
 
 	ctr, err := postgres.Run(
 		ctx,
-		"docker.io/postgres:16-alpine",
+		"6-alpine",
```
stevenh, line 306: bug: typo
```suggestion
		"postgres:16-alpine",
```

stevenh:
Oh if the concern is about *.io going away what about https://index.docker.io/v1/?

mdelapenya:
> Oh if the concern is about *.io going away what about https://index.docker.io/v1/?

No, because that's the correct index for the registry (Docker Hub)

docker_files_test.go
mdelapenya: No special at all, I wonder why it was not included for all the occurrences. I think we can change that in this PR

docker_files_test.go
```diff
@@ -30,7 +30,7 @@ func TestCopyFileToContainer(t *testing.T) {
 
 	ctr, err := testcontainers.GenericContainer(ctx, testcontainers.GenericContainerRequest{
 		ContainerRequest: testcontainers.ContainerRequest{
-			Image: "docker.io/bash",
+			Image: "bash:5.2.26",
```
stevenh, line 33: suggestion: we should add constant so its easy to update. We might even consider just using :latest given what it is.

docker_test.go
```diff
@@ -950,7 +950,7 @@ func TestWorkingDir(t *testing.T) {
 func ExampleDockerProvider_CreateContainer() {
 	ctx := context.Background()
 	req := ContainerRequest{
-		Image:        "docker.io/nginx:alpine",
+		Image:        "nginx:alpine",
```
stevenh, line 953: suggestion: use the existing constant nginxAlpineImage, more below

stevenh (changes requested):
Looking good, just a few suggestions while you're making changes.

docker_test.go
mdelapenya: I did not do it on purpose, as the testable examples could eventually land into a docs page, like pkg.go.dev, and the variable won't be interpolated.

stevenh (approved):

PR #2851, opened 2024-10-26
chore: use require.(No)Error(t,err) instead of t.Fatal(err)

mmorel-35:
## What does this PR do?

This uses `require.(No)Error(t,err)` instead of `t.Fatal(err)` calls

## Why is it important?

This reduce the number of lines required to make an assertion.
It also make the assertion easier to understand.

## Related issues
- Relates #2808

### DIFF
--- a/container_file_test.go
+++ b/container_file_test.go
@@ -7,6 +7,8 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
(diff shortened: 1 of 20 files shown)


mdelapenya (approved):
LGTM, thanks!

PR #2854, opened 2024-10-29
chore: use require.Len instead of assert.Len

mmorel-35:
## What does this PR do?

This uses `require.Len` instead of `assert.Len` calls when element is reused for tests.

## Why is it important?

When assert.Len is used, this is generally to check the length of a list that is generally used in at least one other test.

## Related issues
- Relates #2808


mdelapenya (approved):
LGTM, great job, as always!

PR #2857, opened 2024-10-29
chore: use require instead of t.Fatal (part 2)

mmorel-35:
## What does this PR do?

This uses `require` instead of `t.Fatal` calls

## Why is it important?

This reduce the number of lines required to make an assertion.
It also make the assertion easier to understand.

## Related issues
- Relates #2808

### DIFF
--- a/lifecycle_test.go
+++ b/lifecycle_test.go
@@ -3,7 +3,9 @@ package testcontainers
 import (
 	"bufio"
 	"context"
+	"errors"
 	"fmt"
+	"io"
 	"strings"
 	"testing"
 	"time"
@@ -914,13 +916,10 @@ func TestPrintContainerLogsOnError(t *testing.T) {
 	rd := bufio.NewReader(containerLogs)
 	for {
 		line, err := rd.ReadString('\n')
-		if err != nil {
-			if err.Error() == "EOF" {
-				break
-			}
-
-			t.Fatal("Read Error:", err)
+		if errors.Is(err, io.EOF) {
+			break
 		}
+		require.NoErrorf(t, err, "Read Error")
 
 		// the last line of the array should contain the line of interest,
 		// but we are checking all the lines to make sure that is present
--- a/logconsumer_test.go
+++ b/logconsumer_test.go
@@ -323,9 +323,8 @@ func TestContainerLogWithErrClosed(t *testing.T) {
 
 	hitNginx := func() {
 		i, _, err := dind.Exec(ctx, []string{"wget", "--spider", "localhost:" + port.Port()})
-		if err != nil || i > 0 {
-			t.Fatalf("Can't make request to nginx container from dind container")
-		}
+		require.NoError(t, err, "Can't make request to nginx container from dind container")
+		require.Zerof(t, i, "Can't make request to nginx container from dind container")
 	}
 
 	hitNginx()
@@ -340,13 +339,11 @@ func TestContainerLogWithErrClosed(t *testing.T) {
 	}
 	// Simulate a transient closed connection to the docker daemon
 	i, _, err := dind.Exec(ctx, append([]string{"iptables", "-A"}, iptableArgs...))
-	if err != nil || i > 0 {
-		t.Fatalf("Failed to close connection to dind daemon: i(%d), err %v", i, err)
-	}
+	require.NoErrorf(t, err, "Failed to close connection to dind daemon: i(%d), err %v", i, err)
+	require.Zerof(t, i, "Failed to close connection to dind daemon: i(%d), err %v", i, err)
 	i, _, err = dind.Exec(ctx, append([]string{"iptables", "-D"}, iptableArgs...))
-	if err != nil || i > 0 {
-		t.Fatalf("Failed to re-open connection to dind daemon: i(%d), err %v", i, err)
-	}
+	require.NoErrorf(t, err, "Failed to re-open connection to dind daemon: i(%d), err %v", i, err)
+	require.Zerof(t, i, "Failed to re-open connection to dind daemon: i(%d), err %v", i, err)
 	time.Sleep(time.Second * 3)
 
 	hitNginx()
--- a/modules/chroma/chroma_test.go
+++ b/modules/chroma/chroma_test.go
@@ -23,33 +23,23 @@ func TestChroma(t *testing.T) {
 		// restEndpoint {
 		restEndpoint, err := ctr.RESTEndpoint(ctx)
 		// }
-		if err != nil {
-			tt.Fatalf("failed to get REST endpoint: %s", err)
-		}
+		require.NoErrorf(tt, err, "failed to get REST endpoint")
 
 		cli := &http.Client{}
 		resp, err := cli.Get(restEndpoint + "/docs")
-		if err != nil {
-			tt.Fatalf("failed to perform GET request: %s", err)
-		}
+		require.NoErrorf(tt, err, "failed to perform GET request")
 		defer resp.Body.Close()
 
-		if resp.StatusCode != http.StatusOK {
-			tt.Fatalf("unexpected status code: %d", resp.StatusCode)
-		}
+		require.Equalf(tt, http.StatusOK, resp.StatusCode, "unexpected status code: %d", resp.StatusCode)
 	})
 
 	t.Run("GetClient", func(tt *testing.T) {
 		// restEndpoint {
 		endpoint, err := ctr.RESTEndpoint(context.Background())
-		if err != nil {
-			tt.Fatalf("failed to get REST endpoint: %s", err)
-		}
+		require.NoErrorf(tt, err, "failed to get REST endpoint")
 		chromaClient, err := chromago.NewClient(endpoint)
 		// }
-		if err != nil {
-			tt.Fatalf("failed to create client: %s", err)
-		}
+		require.NoErrorf(tt, err, "failed to create client")
 
 		hb, err := chromaClient.Heartbeat(context.TODO())
 		require.NoError(tt, err)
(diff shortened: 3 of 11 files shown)


lifecycle_test.go
```diff
@@ -919,7 +919,7 @@ func TestPrintContainerLogsOnError(t *testing.T) {
 				break
 			}
 
-			t.Fatal("Read Error:", err)
+			require.NoErrorf(t, err, "Read Error")
```
mdelapenya, line 922: bug: the err was already not nil in https://github.com/testcontainers/testcontainers-go/pull/2857/files#diff-8c1de2f5177b67b15485d779c1510b2fa4abeb3cd1e82c18d43d0c2e7871ed19R917

modules/elasticsearch/elasticsearch_test.go
```diff
@@ -108,21 +106,19 @@ func TestElasticsearch(t *testing.T) {
 
 			if tt.image != baseImage8 && err != nil {
```
mdelapenya, line 107: We probably need to restructure this if/else sequence here, as it was not already very readable before the changes, and the readability does not improve after them.

modules/elasticsearch/elasticsearch_test.go
```diff
@@ -108,21 +106,19 @@ func TestElasticsearch(t *testing.T) {
...
-					t.Fatal(err, "should access with authorised HTTP client using TLS.")
+				if tt.passwordCustomiser != nil {
+					require.NoErrorf(t, err, "should access with authorised HTTP client using TLS.")
 				}
 				if tt.passwordCustomiser == nil && err == nil {
 					// Elasticsearch 8 should return 401 Unauthorized, not an error in the request
-					if resp.StatusCode != http.StatusUnauthorized {
-						t.Fatal("expected 401 status code for unauthorised HTTP client using TLS, but got", resp.StatusCode)
-					}
+					require.Equalf(t, http.StatusUnauthorized, resp.StatusCode, "expected 401 status code for unauthorised HTTP client using TLS, but got: %s", resp.StatusCode)
```
mdelapenya, line 121: chore: use stdlib instead of 401 status code

modules/elasticsearch/elasticsearch_test.go
```diff
@@ -133,23 +129,19 @@ func TestElasticsearch(t *testing.T) {
 			// validate response
 			if resp != nil {
 				// validate Elasticsearch response
-				if resp.StatusCode != http.StatusOK {
-					t.Fatal("expected 200 status code but got", resp.StatusCode)
-				}
+				require.Equalf(t, http.StatusOK, resp.StatusCode, "expected 200 status code but got: %s", resp.StatusCode)
```
mdelapenya, line 132: ditto: use stdlib

mdelapenya (commented):
LGTM, thanks! just some minor comments

modules/elasticsearch/elasticsearch_test.go
mdelapenya: Yes, given we added the new lint rule for that, I consider it could be used here. Wdyt about using it inside error messages in tests?

logconsumer_test.go
```diff
@@ -323,9 +323,8 @@ func TestContainerLogWithErrClosed(t *testing.T) {
 
 	hitNginx := func() {
 		i, _, err := dind.Exec(ctx, []string{"wget", "--spider", "localhost:" + port.Port()})
-		if err != nil || i > 0 {
-			t.Fatalf("Can't make request to nginx container from dind container")
-		}
+		require.Zerof(t, i, "Can't make request to nginx container from dind container")
```
mdelapenya, line 326: Comment NOT related to this PR, but for the general maintenance of the library:

@stevenh wdyt of keeping the description messages in the checks? Should we keep them, or remove them. We are working in creating more and more contextual error messages, so at some point we'll have much better experience debugging test errors.

stevenh: Personally I don't think they add value in most cases, just adding more for the developer to read. This is particularly the case for error checks. 

The way I look at it is does it make it clearer to the reader why this might have failed, which this type of message I would argue doesn't as its obvious from the line above that the request didn't work.

In this particular example they checks are back to front, error check should come first.

logconsumer_test.go
```diff
@@ -340,13 +339,11 @@ func TestContainerLogWithErrClosed(t *testing.T) {
 	}
 	// Simulate a transient closed connection to the docker daemon
 	i, _, err := dind.Exec(ctx, append([]string{"iptables", "-A"}, iptableArgs...))
-	if err != nil || i > 0 {
-		t.Fatalf("Failed to close connection to dind daemon: i(%d), err %v", i, err)
-	}
+	require.Zerof(t, i, "Failed to close connection to dind daemon: i(%d), err %v", i, err)
```
stevenh, line 342: suggestion: checks are back to front, NoError should be first, more below and above.

modules/chroma/chroma_test.go
```diff
@@ -23,33 +23,23 @@ func TestChroma(t *testing.T) {
...
-		if err != nil {
-			tt.Fatalf("failed to perform GET request: %s", err)
-		}
+		require.NoErrorf(tt, err, "failed to perform GET request")
 		defer resp.Body.Close()
 
-		if resp.StatusCode != http.StatusOK {
-			tt.Fatalf("unexpected status code: %d", resp.StatusCode)
-		}
+		require.Equalf(tt, http.StatusOK, resp.StatusCode, "unexpected status code: %d", resp.StatusCode)
```
stevenh, line 33: This is an example of a message that people might think is helpful but isn't in my mind as the values are output anyway and is obvious from the code its a status check.

lifecycle_test.go
```diff
@@ -914,13 +914,10 @@ func TestPrintContainerLogsOnError(t *testing.T) {
 	rd := bufio.NewReader(containerLogs)
 	for {
 		line, err := rd.ReadString('\n')
-		if err != nil {
-			if err.Error() == "EOF" {
-				break
-			}
-
-			t.Fatal("Read Error:", err)
+		if err != nil && err.Error() == "EOF" {
```
stevenh, line 917: suggestion: use `errors.Is`
```suggestion
		if errors.Is(err, io.EOF) {
```

modules/compose/compose_builder_test.go
```diff
@@ -158,19 +154,15 @@ func writeTemplateWithSrvType(t *testing.T, templateFile string, srvType string,
 	values["ServiceType"] = srvType
 
 	output, err := os.Create(composeFile)
-	if err != nil {
-		t.Fatalf("creating output file: %s", err)
-	}
+	require.NoErrorf(t, err, "creating output file")
 	defer output.Close()
```
stevenh, line 158: bug: this can loose a write error. Not related to this PR, but lets fix while here
```go
defer func() {
    require.NoError(t, outpout.Close())
}()
```

modules/elasticsearch/elasticsearch_test.go
```diff
@@ -92,65 +92,40 @@ func TestElasticsearch(t *testing.T) {
...
 				// }
 			}
 
 			resp, err := httpClient.Do(req)
-			if resp != nil {
-				defer resp.Body.Close()
-			}
+			require.NotNil(t, resp)
+			defer resp.Body.Close()
+			require.NoError(t, err)
```
stevenh, line 105: bug: check for error first then no need to check for nil resp.
```suggestion
			require.NoError(t, err)
			defer resp.Body.Close()
```

modules/elasticsearch/elasticsearch_test.go
```diff
@@ -92,65 +92,40 @@ func TestElasticsearch(t *testing.T) {
...
+
+			if tt.image == baseImage8 && tt.passwordCustomiser == nil {
```
stevenh, line 107: suggestion: This would be easier to follow if we changed to switch on the image and had each case inside.

modules/neo4j/neo4j_test.go
```diff
@@ -118,20 +118,16 @@ func TestNeo4jWithWrongSettings(outer *testing.T) {
 		require.NoError(t, err)
 
 		errorLogs := logger.Logs()
-		if !Contains(errorLogs, `setting "some.key" with value "value1" is now overwritten with value "value2"`+"\n") ||
-			!Contains(errorLogs, `setting "some.key" with value "value2" is now overwritten with value "value3"`+"\n") {
-			t.Fatalf("expected setting overwrites to be logged")
-		}
+		require.Truef(t, Contains(errorLogs, `setting "some.key" with value "value1" is now overwritten with value "value2"`+"\n"), "expected setting overwrites to be logged")
+		require.Truef(t, Contains(errorLogs, `setting "some.key" with value "value2" is now overwritten with value "value3"`+"\n"), "expected setting overwrites to be logged")
```
stevenh, line 122: question: can we use require.Contains here?

modules/weaviate/weaviate_test.go
```diff
@@ -66,8 +66,7 @@ func TestWeaviate(t *testing.T) {
 		meta, err := client.Misc().MetaGetter().Do(ctx)
 		require.NoError(tt, err)
 
-		if meta == nil || meta.Version == "" {
-			tt.Fatal("failed to get /v1/meta response")
-		}
+		require.NotNilf(tt, meta, "failed to get /v1/meta response")
+		require.NotEqualf(tt, "", meta.Version, "failed to get /v1/meta response")
```
stevenh, line 70: suggestion: use require.NotEmpty

stevenh (changes requested):

mdelapenya (approved):
LGTM!

As a follow-up, I'd send a separate PR replacing the formatted requires, simply removing the messages, as it's now much clearer the reasons why they failed.

Thank you @mmorel-35 for your hard work cleaning up the project 🙇

stevenh (approved):
LGTM

PR #2858, opened 2024-10-30
feat: use recommended cluster settings for cockroachdb

martskins:
## What does this PR do?

This PR configures the cockroachdb container with the [settings recommended by Cockroach Labs](https://www.cockroachlabs.com/docs/stable/local-testing#use-a-local-single-node-cluster-with-in-memory-storage) for testing clusters.

## Why is it important?

It seems reasonable to follow their advise but the why for each of these settings is explained in the link above.

## Related issues

None

## How to test this PR

I did some rather naive testing by running a test suite with and without these changes and it seemed like it did work in that the the tests took less time to finish with these settings than without them. It's not clear to me whether that improvement will be there in all possible scenarios but, as I said before, I think it's reasonable to follow Cockroach Labs advise on this.

### DIFF
--- a/modules/cockroachdb/cockroachdb.go
+++ b/modules/cockroachdb/cockroachdb.go
@@ -244,17 +244,17 @@ func addTLS(ctx context.Context, container testcontainers.Container, opts option
 func setRecommendedSettings(ctx context.Context, container testcontainers.Container, opts options) error {
 	port, err := container.MappedPort(ctx, defaultSQLPort)
 	if err != nil {
-		return err
+		return fmt.Errorf("mapped port: %w", err)
 	}
 
 	host, err := container.Host(ctx)
 	if err != nil {
-		return err
+		return fmt.Errorf("host: %w", err)
 	}
 
 	db, err := sql.Open("pgx/v5", connString(opts, host, port))
 	if err != nil {
-		return err
+		return fmt.Errorf("sql.Open: %w", err)
 	}
 	defer db.Close()
 
@@ -272,7 +272,7 @@ func setRecommendedSettings(ctx context.Context, container testcontainers.Contai
 	for _, stmt := range stmts {
 		_, err = db.Exec(stmt)
 		if err != nil {
-			return err
+			return fmt.Errorf("db.Exec: %w", err)
 		}
 	}
 
--- a/modules/cockroachdb/cockroachdb.go
+++ b/modules/cockroachdb/cockroachdb.go
@@ -240,7 +240,7 @@ func addTLS(ctx context.Context, container testcontainers.Container, opts option
 }
 
 // runStatements runs the configured statements against the CockroachDB container.
-func runStatements(ctx context.Context, container testcontainers.Container, opts options) (err error) {
+func runStatements(ctx context.Context, container testcontainers.Container, opts options) (err error) { //nolint:nonamedreturns // Needed for error checking.
 	if len(opts.Statements) == 0 {
 		return nil
 	}
--- a/modules/cockroachdb/cockroachdb.go
+++ b/modules/cockroachdb/cockroachdb.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/tls"
 	"crypto/x509"
+	"database/sql"
 	"encoding/pem"
 	"fmt"
 	"net"
@@ -88,6 +89,11 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
 					PreStarts: []testcontainers.ContainerHook{
 						func(ctx context.Context, container testcontainers.Container) error {
 							return addTLS(ctx, container, o)
+						},
+					},
+					PostReadies: []testcontainers.ContainerHook{
+						func(ctx context.Context, container testcontainers.Container) error {
+							return setRecommendedSettings(ctx, container, o)
 						},
 					},
 				},
@@ -230,6 +236,46 @@ func addTLS(ctx context.Context, container testcontainers.Container, opts option
 			return err
 		}
 	}
+	return nil
+}
+
+// setRecommendedSettings applies the cluster settings recommended by cockroachlabs for testing clusters.
+// See https://www.cockroachlabs.com/docs/stable/local-testing for more information.
+func setRecommendedSettings(ctx context.Context, container testcontainers.Container, opts options) error {
+	port, err := container.MappedPort(ctx, defaultSQLPort)
+	if err != nil {
+		return err
+	}
+
+	host, err := container.Host(ctx)
+	if err != nil {
+		return err
+	}
+
+	db, err := sql.Open("pgx/v5", connString(opts, host, port))
+	if err != nil {
+		return err
+	}
+	defer db.Close()
+
+	stmts := []string{
+		"SET CLUSTER SETTING kv.range_merge.queue_interval = '50ms'",
+		"SET CLUSTER SETTING jobs.registry.interval.gc = '30s'",
+		"SET CLUSTER SETTING jobs.registry.interval.cancel = '180s'",
+		"SET CLUSTER SETTING jobs.retention_time = '15s'",
+		"SET CLUSTER SETTING sql.stats.automatic_collection.enabled = false",
+		"SET CLUSTER SETTING kv.range_split.by_load_merge_delay = '5s'",
+		`ALTER RANGE default CONFIGURE ZONE USING "gc.ttlseconds" = 600`,
+		`ALTER DATABASE system CONFIGURE ZONE USING "gc.ttlseconds" = 600`,
+	}
+
+	for _, stmt := range stmts {
+		_, err = db.Exec(stmt)
+		if err != nil {
+			return err
+		}
+	}
+
 	return nil
 }
 
--- a/modules/cockroachdb/cockroachdb.go
+++ b/modules/cockroachdb/cockroachdb.go
@@ -240,7 +240,7 @@ func addTLS(ctx context.Context, container testcontainers.Container, opts option
 }
 
 // runStatements runs the configured statements against the CockroachDB container.
-func runStatements(ctx context.Context, container testcontainers.Container, opts options) (err error) { //nolint:nonamedreturns // Needed for error checking.
+func runStatements(ctx context.Context, container testcontainers.Container, opts options) (err error) {
 	if len(opts.Statements) == 0 {
 		return nil
 	}
--- a/modules/cockroachdb/cockroachdb.go
+++ b/modules/cockroachdb/cockroachdb.go
@@ -240,7 +240,7 @@ func addTLS(ctx context.Context, container testcontainers.Container, opts option
 }
 
 // runStatements runs the configured statements against the CockroachDB container.
-func runStatements(ctx context.Context, container testcontainers.Container, opts options) error {
+func runStatements(ctx context.Context, container testcontainers.Container, opts options) (err error) {
 	if len(opts.Statements) == 0 {
 		return nil
 	}
@@ -259,7 +259,12 @@ func runStatements(ctx context.Context, container testcontainers.Container, opts
 	if err != nil {
 		return fmt.Errorf("sql.Open: %w", err)
 	}
-	defer db.Close()
+	defer func() {
+		cerr := db.Close()
+		if err == nil {
+			err = cerr
+		}
+	}()
 
 	for _, stmt := range opts.Statements {
 		if _, err = db.Exec(stmt); err != nil {
--- a/modules/cockroachdb/cockroachdb.go
+++ b/modules/cockroachdb/cockroachdb.go
@@ -262,8 +262,7 @@ func runStatements(ctx context.Context, container testcontainers.Container, opts
 	defer db.Close()
 
 	for _, stmt := range opts.Statements {
-		_, err = db.Exec(stmt)
-		if err != nil {
+		if _, err = db.Exec(stmt); err != nil {
 			return fmt.Errorf("db.Exec: %w", err)
 		}
 	}
--- a/modules/cockroachdb/cockroachdb.go
+++ b/modules/cockroachdb/cockroachdb.go
@@ -93,7 +93,7 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
 					},
 					PostReadies: []testcontainers.ContainerHook{
 						func(ctx context.Context, container testcontainers.Container) error {
-							return setRecommendedSettings(ctx, container, o)
+							return runStatements(ctx, container, o)
 						},
 					},
 				},
@@ -239,9 +239,12 @@ func addTLS(ctx context.Context, container testcontainers.Container, opts option
 	return nil
 }
 
-// setRecommendedSettings applies the cluster settings recommended by cockroachlabs for testing clusters.
-// See https://www.cockroachlabs.com/docs/stable/local-testing for more information.
-func setRecommendedSettings(ctx context.Context, container testcontainers.Container, opts options) error {
+// runStatements runs the configured statements against the CockroachDB container.
+func runStatements(ctx context.Context, container testcontainers.Container, opts options) error {
+	if len(opts.Statements) == 0 {
+		return nil
+	}
+
 	port, err := container.MappedPort(ctx, defaultSQLPort)
 	if err != nil {
 		return fmt.Errorf("mapped port: %w", err)
@@ -258,18 +261,7 @@ func setRecommendedSettings(ctx context.Context, container testcontainers.Contai
 	}
 	defer db.Close()
 
-	stmts := []string{
-		"SET CLUSTER SETTING kv.range_merge.queue_interval = '50ms'",
-		"SET CLUSTER SETTING jobs.registry.interval.gc = '30s'",
-		"SET CLUSTER SETTING jobs.registry.interval.cancel = '180s'",
-		"SET CLUSTER SETTING jobs.retention_time = '15s'",
-		"SET CLUSTER SETTING sql.stats.automatic_collection.enabled = false",
-		"SET CLUSTER SETTING kv.range_split.by_load_merge_delay = '5s'",
-		`ALTER RANGE default CONFIGURE ZONE USING "gc.ttlseconds" = 600`,
-		`ALTER DATABASE system CONFIGURE ZONE USING "gc.ttlseconds" = 600`,
-	}
-
-	for _, stmt := range stmts {
+	for _, stmt := range opts.Statements {
 		_, err = db.Exec(stmt)
 		if err != nil {
 			return fmt.Errorf("db.Exec: %w", err)
--- a/modules/cockroachdb/cockroachdb_test.go
+++ b/modules/cockroachdb/cockroachdb_test.go
@@ -28,6 +28,9 @@ func TestCockroach_NotRoot(t *testing.T) {
 		url: "postgres://test@localhost:xxxxx/defaultdb?sslmode=disable",
 		opts: []testcontainers.ContainerCustomizer{
 			cockroachdb.WithUser("test"),
+			// Do not run the default statements as the user used on this test is
+			// lacking the needed MODIFYCLUSTERSETTING privilege to run them.
+			cockroachdb.WithStatements(),
 		},
 	})
 }
@@ -38,6 +41,9 @@ func TestCockroach_Password(t *testing.T) {
 		opts: []testcontainers.ContainerCustomizer{
 			cockroachdb.WithUser("foo"),
 			cockroachdb.WithPassword("bar"),
+			// Do not run the default statements as the user used on this test is
+			// lacking the needed MODIFYCLUSTERSETTING privilege to run them.
+			cockroachdb.WithStatements(),
 		},
 	})
 }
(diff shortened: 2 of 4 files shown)


modules/cockroachdb/cockroachdb.go
```diff
@@ -233,6 +239,46 @@ func addTLS(ctx context.Context, container testcontainers.Container, opts option
...
+		"SET CLUSTER SETTING sql.stats.automatic_collection.enabled = false",
+		"SET CLUSTER SETTING kv.range_split.by_load_merge_delay = '5s'",
+		`ALTER RANGE default CONFIGURE ZONE USING "gc.ttlseconds" = 600`,
+		`ALTER DATABASE system CONFIGURE ZONE USING "gc.ttlseconds" = 600`,
+	}
+
+	for _, stmt := range stmts {
+		_, err = db.Exec(stmt)
+		if err != nil {
+			return err
```
mdelapenya, line 275: suggestions: let's wrap the error to provide more context on potential errors
```suggestion
			return fmt.Errorf("db.Exec: %w", err)
```

mdelapenya (commented):
LGTM, I left a minor comment.

I wonder if we always want to add those recommendations, or wrap them into a functional opt so users can decide whether to apply them or not 🤔

mdelapenya:
BTW I'd rename the title to `feat:`, as we are adding a new behaviour, not fixing a broken one

martskins:
> LGTM, I left a minor comment.
> 
> I wonder if we always want to add those recommendations, or wrap them into a functional opt so users can decide whether to apply them or not 🤔

Would you add one single `WithRecommendedClusterSettings` or one option for each of the settings? I'm leaning towards the former but happy to do either.

modules/cockroachdb/cockroachdb.go
```diff
@@ -233,6 +239,46 @@ func addTLS(ctx context.Context, container testcontainers.Container, opts option
 	return nil
 }
 
+// setRecommendedSettings applies the cluster settings recommended by cockroachlabs for testing clusters.
+// See https://www.cockroachlabs.com/docs/stable/local-testing for more information.
+func setRecommendedSettings(ctx context.Context, container testcontainers.Container, opts options) error {
+	port, err := container.MappedPort(ctx, defaultSQLPort)
+	if err != nil {
+		return err
```
stevenh, line 247: suggestion: lets wrap the error

modules/cockroachdb/cockroachdb.go
```diff
@@ -233,6 +239,46 @@ func addTLS(ctx context.Context, container testcontainers.Container, opts option
...
+	}
+
+	host, err := container.Host(ctx)
+	if err != nil {
+		return err
```
stevenh, line 252: suggestion: lets wrap the error

modules/cockroachdb/cockroachdb.go
```diff
@@ -233,6 +239,46 @@ func addTLS(ctx context.Context, container testcontainers.Container, opts option
...
+	}
+
+	db, err := sql.Open("pgx/v5", connString(opts, host, port))
+	if err != nil {
+		return err
```
stevenh, line 257: suggestion: lets wrap the error

modules/cockroachdb/cockroachdb.go
```diff
@@ -233,6 +239,46 @@ func addTLS(ctx context.Context, container testcontainers.Container, opts option
 	return nil
 }
 
+// setRecommendedSettings applies the cluster settings recommended by cockroachlabs for testing clusters.
```
stevenh, line 242: question: would it make sense to make this a function option e.g. `WithClusterSettings(settings ...string)` where settings defaults to the below if nil?

stevenh (changes requested):

mdelapenya:
Given all of us agree, I'd go with a single functional option for settings, as pointed out by @stevenh. What I'm not sure if we want to append the default settings to those passed by the user (I like this, although it could come with some conflicts)

stevenh:
> Given all of us agree, I'd go with a single functional option for settings, as pointed out by @stevenh. What I'm not sure if we want to append the default settings to those passed by the user (I like this, although it could come with some conflicts)

We could export a variable to be used e.g. `WithSettings(cockroachdb.ClusterDefaults...)`

modules/cockroachdb/cockroachdb.go
martskins: fixed

modules/cockroachdb/cockroachdb.go
martskins: fixed

modules/cockroachdb/cockroachdb.go
martskins: fixed

modules/cockroachdb/cockroachdb.go
martskins: fixed

martskins:
> > Given all of us agree, I'd go with a single functional option for settings, as pointed out by @stevenh. What I'm not sure if we want to append the default settings to those passed by the user (I like this, although it could come with some conflicts)
> 
> We could export a variable to be used e.g. `cockroachdb.WithSettings(cockroachdb.ClusterDefaults...)` to avoid having to deal with the conflicts.

Would you make those configurable? Thinking that if we expose that as something configurable then a struct would probably be better but it would be weird if we only add these specific settings to the struct and leave out all the other configurable stuff of the cluster.

I suppose, given that these are just statements we run after ready, we could also just expose this as a non-configurable `PostReadies` function that the user just plugs in there instead of trying to name these as settings. And if they want something else they just specific the function themselves?

mdelapenya:
I'd go simple: a `WithClusterSettings` functional opt that is able to execute them in the cluster. As @stevenh suggested, having a variable with default cluster settings seems good, so we can mention about its existence both in the `$ROOT_DIR/docs/modules/cockroachdb` page and in the functional option comments, so users are aware of its existence. Finally, this option would contribute the result to the PostReadies.

stevenh:
> I'd go simple: a `WithClusterSettings` functional opt that is able to execute them in the cluster. As @stevenh suggested, having a variable with default cluster settings seems good, so we can mention about its existence both in the `$ROOT_DIR/docs/modules/cockroachdb` page and in the functional option comments, so users are aware of its existence. Finally, this option would contribute the result to the PostReadies.

Thinking about it I wouldn't call it `WithClusterSettings` as it could be used to apply all sorts of settings, so something more generic like `WithStatements` for the functional option and then the specifics are exposed in the variable passed in.

martskins:
Just updated the PR. Let me know if that's roughly what you had in mind.

modules/cockroachdb/cockroachdb.go
```diff
@@ -233,6 +239,38 @@ func addTLS(ctx context.Context, container testcontainers.Container, opts option
...
+	host, err := container.Host(ctx)
+	if err != nil {
+		return fmt.Errorf("host: %w", err)
+	}
+
+	db, err := sql.Open("pgx/v5", connString(opts, host, port))
+	if err != nil {
+		return fmt.Errorf("sql.Open: %w", err)
+	}
+	defer db.Close()
```
stevenh, line 262: bug: lets ensure that the error from close is returned as otherwise we could miss an issue.

modules/cockroachdb/cockroachdb.go
```diff
@@ -233,6 +239,38 @@ func addTLS(ctx context.Context, container testcontainers.Container, opts option
...
+
+	for _, stmt := range opts.Statements {
+		_, err = db.Exec(stmt)
+		if err != nil {
+			return fmt.Errorf("db.Exec: %w", err)
+		}
```
stevenh, line 268: nit: fold the if
```suggestion
		if _, err = db.Exec(stmt); err != nil {
			return fmt.Errorf("db.Exec: %w", err)
		}
```

modules/cockroachdb/options.go
```diff
@@ -67,3 +68,23 @@ func WithTLS(cfg *TLSConfig) Option {
 		o.TLS = cfg
 	}
 }
+
+// ClusterDefaults are the settings recommended by Cockroach Labs for testing clusters.
+// See https://www.cockroachlabs.com/docs/stable/local-testing for more information.
+var ClusterDefaults []string = []string{
```
stevenh, line 74: suggestion: you don't need to type definition
```suggestion
var ClusterDefaults = []string{
```

modules/cockroachdb/options.go
```diff
@@ -67,3 +68,23 @@ func WithTLS(cfg *TLSConfig) Option {
...
+	"SET CLUSTER SETTING jobs.registry.interval.gc = '30s'",
+	"SET CLUSTER SETTING jobs.registry.interval.cancel = '180s'",
+	"SET CLUSTER SETTING jobs.retention_time = '15s'",
+	"SET CLUSTER SETTING sql.stats.automatic_collection.enabled = false",
+	"SET CLUSTER SETTING kv.range_split.by_load_merge_delay = '5s'",
+	`ALTER RANGE default CONFIGURE ZONE USING "gc.ttlseconds" = 600`,
+	`ALTER DATABASE system CONFIGURE ZONE USING "gc.ttlseconds" = 600`,
+}
+
+// WithStatements sets the statements to run on the CockroachDB cluster once the container is ready.
```
stevenh, line 85: suggestion: Lets add an example for this combined with ClusterDefaults.

modules/cockroachdb/options.go
```diff
@@ -67,3 +68,23 @@ func WithTLS(cfg *TLSConfig) Option {
 		o.TLS = cfg
 	}
 }
+
+// ClusterDefaults are the settings recommended by Cockroach Labs for testing clusters.
```
stevenh, line 72: suggestion: lets mention how this can be used in combination with `WithStatements`

modules/cockroachdb/options.go
```diff
@@ -3,11 +3,12 @@ package cockroachdb
...
-	User      string
-	Password  string
-	StoreSize string
-	TLS       *TLSConfig
+	Database   string
+	User       string
+	Password   string
+	StoreSize  string
+	TLS        *TLSConfig
+	Statements []string
```
stevenh, line 11: question: As we have this should we set the default so the cluster is setup with best practices?

stevenh (changes requested):
Thanks for making all these changes, we're getting close to a solid solution I think.

modules/cockroachdb/options.go
martskins: makes sense to me. It is a bit weird to require the user to have to pass `cockroachdb.WithStatements()` to disable the defaults though.

modules/cockroachdb/options.go
martskins: you mean a testable example or just mention it in the documentation comment?

mdelapenya: A testable example would be super 🙏

modules/cockroachdb/options.go
stevenh: Hmm, yer @mdelapenya what do you recon?

modules/cockroachdb/cockroachdb.go
```diff
@@ -233,6 +239,42 @@ func addTLS(ctx context.Context, container testcontainers.Container, opts option
 	return nil
 }
 
+// runStatements runs the configured statements against the CockroachDB container.
+func runStatements(ctx context.Context, container testcontainers.Container, opts options) (err error) {
```
mdelapenya, line 243: suggestion: I think we need this for the lint
```suggestion
func runStatements(ctx context.Context, container testcontainers.Container, opts options) (err error) { //nolint:nonamedreturns // Needed for error checking.
```

modules/cockroachdb/options.go
mdelapenya: Yes, default statements should include the recommended defaults from the vendor

modules/cockroachdb/options.go
martskins: added. Let me know if you want more outputs in there or if it's fine as it is.

modules/cockroachdb/cockroachdb.go
```diff
@@ -233,6 +239,42 @@ func addTLS(ctx context.Context, container testcontainers.Container, opts option
 	return nil
 }
 
+// runStatements runs the configured statements against the CockroachDB container.
+func runStatements(ctx context.Context, container testcontainers.Container, opts options) (err error) { //nolint:nonamedreturns // Needed for error checking.
```
stevenh, line 243: bug: this linter directive isn't needed as it was replaced by `nakedret` recently.

modules/cockroachdb/options.go
mdelapenya: Indeed, thanks for considering it too.

stevenh: Sounds sensible to me

stevenh (changes requested):

modules/cockroachdb/cockroachdb.go
martskins: removed

stevenh (approved):
LGTM

mdelapenya (approved):
LGTM, thank you very much and sorry if the review process took longer than expected 🙏

stevenh (changes requested):
Looks like this fails due the example not working with TLS

martskins:
> Looks like this fails due the example not working with TLS, [see here](https://github.com/testcontainers/testcontainers-go/actions/runs/11597013424/job/32292863969?pr=2858).

So I found two issues... First one was that the user you pass in the options when setting up the cluster needs to have MODIFYCLUSTERSETTING privilege to run the statements. I added a note about that in the `DefaultStatements` documentation and changed the non-root tests to not use the default statements, but not sure if that's good enough.

Second one was the one you mentioned, it seems like these default statements don't play nicely with TLS. I couldn't find any documentation from Cockroach Labs that would help me understand why this would be the case. I tried running only some of these statements but I couldn't really make it work with any combination that I tried. Should we maybe error if default statements and TLS are used?

modules/cockroachdb/cockroachdb_test.go
```diff
@@ -28,6 +28,7 @@ func TestCockroach_NotRoot(t *testing.T) {
 		url: "postgres://test@localhost:xxxxx/defaultdb?sslmode=disable",
 		opts: []testcontainers.ContainerCustomizer{
 			cockroachdb.WithUser("test"),
+      cockroachdb.WithStatements(),
```
stevenh, line 31: question: this looks like an anti pattern, could you clarify why this is desired? Same below.

modules/cockroachdb/options.go
```diff
@@ -67,3 +69,26 @@ func WithTLS(cfg *TLSConfig) Option {
...
+	"SET CLUSTER SETTING jobs.retention_time = '15s'",
+	"SET CLUSTER SETTING sql.stats.automatic_collection.enabled = false",
+	"SET CLUSTER SETTING kv.range_split.by_load_merge_delay = '5s'",
+	`ALTER RANGE default CONFIGURE ZONE USING "gc.ttlseconds" = 600`,
+	`ALTER DATABASE system CONFIGURE ZONE USING "gc.ttlseconds" = 600`,
+}
+
+// WithStatements sets the statements to run on the CockroachDB cluster once the container is ready.
+// This, in combination with DefaultStatements, can be used to configure the cluster with the settings
+// recommended by Cockroach Labs.
```
stevenh, line 89: suggestion: This is somewhat confusing as this won't be combined with `DefaultStatements` which is how I would interpret this currently. We should clarify and ensure we clearly state the default if not configured is `DefaultStatements`.

stevenh (changes requested):

modules/cockroachdb/options.go
mdelapenya: If we are always adding the default statements, then I think they must be private.

modules/cockroachdb/cockroachdb_test.go
martskins: This is what i meant by my last comment. The `test` user that is being used here doesn't have `MODIFYCLUSTERSETTING`, so it cannot run the default statements as that privilege is needed for those. The fact that we have to pass this `cockroachdb.WithStatements()` option to disable the default is indeed weird (I asked about that in one of other convos) but I suppose the other alternative would be to not have these defaults be actual defaults and let the user pass those in.

modules/cockroachdb/options.go
martskins: > If we are always adding the default statements, then I think they must be private, and probably prepended to the user-provided ones.

I see what you mean, but given the complications with TLS and users without MODIFYCLUSTERSETTING privilege I think prepending to the user-provided statements is probably going to cause a bit of pain.

Good spot on the comment. I'll hold on that change until we've established what we do about the TLS and user privileges thing though, as it might affect this.

stevenh: Private vs public for these is an interesting discussion, here's some ideas from the perspective of making the var private:
* Pro: Others can't mess with them
* Conn: Others can't correct if needed
* Conn: Can't pass them into WithStatements to customise
* Conn: Not self documenting.

modules/cockroachdb/cockroachdb_test.go
stevenh: What's needed to enable this to work by default?

martskins: you mean just for these tests? we'd need the user to have `MODIFYCLUSTERSETTING` privilege. But the problem will likely exist for users that try and set up their cluster with a non-root user anyways. Alternatively we make the defaults not defaults and just let the user add those statements if they want.

modules/cockroachdb/options.go
mdelapenya: One of the goals of any testcontainers module is to be simple, and maybe opinionated, helping users in not committing mistakes. So I'd look for the mechanism that creates this experience.

Whatever you find that matches that, it's fine for me 😊

modules/cockroachdb/cockroachdb_test.go
stevenh: Sorry, I don't think I fully understood the issue you described.

To confirm my understanding we need to disable the cluster settings here as we're using a custom user which doesn't have the required permissions?

If that's the case I would suggest just correcting the indentation and adding a comment to the line to make that clear.

stevenh:
> > Looks like this fails due the example not working with TLS, [see here](https://github.com/testcontainers/testcontainers-go/actions/runs/11597013424/job/32292863969?pr=2858).
> 
> So I found two issues... First one was that the user you pass in the options when setting up the cluster needs to have MODIFYCLUSTERSETTING privilege to run the statements. I added a note about that in the `DefaultStatements` documentation and changed the non-root tests to not use the default statements, but not sure if that's good enough.
> 
> Second one was the one you mentioned, it seems like these default statements don't play nicely with TLS. I couldn't find any documentation from Cockroach Labs that would help me understand why this would be the case. I tried running only some of these statements but I couldn't really make it work with any combination that I tried. Should we maybe error if default statements and TLS are used?

I believe TLS issue is down to how the self signed cert is created by the module in that it only supports the default user, which seems like an unnecessary restriction.

modules/cockroachdb/cockroachdb_test.go
martskins: > To confirm my understanding we need to disable the cluster settings here as we're using a custom user which doesn't have the required permissions?
correct!

I'll re-format the code and add that comment then

martskins:
> > > Looks like this fails due the example not working with TLS, [see here](https://github.com/testcontainers/testcontainers-go/actions/runs/11597013424/job/32292863969?pr=2858).
> > 
> > 
> > So I found two issues... First one was that the user you pass in the options when setting up the cluster needs to have MODIFYCLUSTERSETTING privilege to run the statements. I added a note about that in the `DefaultStatements` documentation and changed the non-root tests to not use the default statements, but not sure if that's good enough.
> > Second one was the one you mentioned, it seems like these default statements don't play nicely with TLS. I couldn't find any documentation from Cockroach Labs that would help me understand why this would be the case. I tried running only some of these statements but I couldn't really make it work with any combination that I tried. Should we maybe error if default statements and TLS are used?
> 
> I believe TLS issue is down to how the self signed cert is created by the module in that it only supports the default user, which seems like an unnecessary restriction.

I'm not sure I understand. Should it be failing in the main branch if that was the case? I haven't modified those failing tests or anything around TLS that would cause it fail. Also, commenting out the added statements seemed to make the tests pass for me.

Apologies if I'm misunderstanding something here.

stevenh:
> > I believe TLS issue is down to how the self signed cert is created by the module in that it only supports the default user, which seems like an unnecessary restriction.
> 
> I'm not sure I understand. Should it be failing in the main branch if that was the case? I haven't modified those failing tests or anything around TLS that would cause it fail. Also, commenting out the added statements seemed to make the tests pass for me.
> 
> Apologies if I'm misunderstanding something here.

I had a look into this in more details and the connections handling in this module for TLS is not clear requiring the user to jump through hoops to get it right. You've fallen foul of this in the way you're using the connection string in `runStatements`.

I've pushed change to this branch which fixes that.

martskins:
> > > I believe TLS issue is down to how the self signed cert is created by the module in that it only supports the default user, which seems like an unnecessary restriction.
> > 
> > 
> > I'm not sure I understand. Should it be failing in the main branch if that was the case? I haven't modified those failing tests or anything around TLS that would cause it fail. Also, commenting out the added statements seemed to make the tests pass for me.
> > Apologies if I'm misunderstanding something here.
> 
> I had a look into this in more detail and the connections handling in this module for TLS is not clear, it actually requires the user to jump through hoops to get it right combining a base URL with TLS setting. You've fallen foul of this in the way you're using the connection string in `runStatements`.
> 
> I've pushed a [commit](https://github.com/testcontainers/testcontainers-go/pull/2869/commits/f7040107a588f6d57812cc797cea04246bf36911) to a separate [draft PR](https://github.com/testcontainers/testcontainers-go/pull/2869) which fixes that.

Ooh that's a good find! I'm happy to close this one and go with yours then, sounds like the easier path.

stevenh:
Closing in favour of https://github.com/testcontainers/testcontainers-go/pull/2869

PR #2861, opened 2024-10-30
feat(ci): Test against the nightly master moby

vvoland:
## What does this PR do?
Changes the Github actions workflow that ran the tests against a Docker Engine from the `test` channel of download.docker.com to get the nightly build off the `master` branch.

Currently it uses the `crazy-max/ghaction-setup-docker@master` action. We plan to move to to Docker org soon.

It also adds the containerd integration to the test matrix.
Rootless is not supported yet.

## Why is it important?
The `test` channel isn't updated that often as it requires us to run the whole release pipeline.
With this change, testcontainers tests will be ran against the master branch of moby and test the containerd image store integration.

-

### DIFF
--- a/.github/workflows/docker-moby-latest.yml
+++ b/.github/workflows/docker-moby-latest.yml
@@ -10,6 +10,12 @@ jobs:
     strategy:
       matrix:
         rootless-docker: [true, false]
+        containerd-integration: [true, false]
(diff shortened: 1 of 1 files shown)


mdelapenya (dismissed):
LGTM, this is great for testing! Much appreciated!

Issue #1987, opened 2023-12-07
[Bug]: Docker Pull fails on some machines


stubents:
### Testcontainers version

v0.26.0

### Using the latest Testcontainers version?

Yes

### Host OS

MacOS

### Host arch

ARM

### Go version

go1.21.4 darwin/arm64

### Docker version

```shell
Client:
 Cloud integration: v1.0.35+desktop.5
 Version:           24.0.6
 API version:       1.43
 Go version:        go1.20.7
 Git commit:        ed223bc
 Built:             Mon Sep  4 12:28:49 2023
 OS/Arch:           darwin/arm64
 Context:           desktop-linux

Server: Docker Desktop 4.25.2 (129061)
 Engine:
  Version:          24.0.6
  API version:      1.43 (minimum version 1.12)
  Go version:       go1.20.7
  Git commit:       1a79695
  Built:            Mon Sep  4 12:31:36 2023
  OS/Arch:          linux/arm64
  Experimental:     false
 containerd:
  Version:          1.6.22
  GitCommit:        8165feabfdfe38c65b599c4993d227328c231fca
 runc:
  Version:          1.1.8
  GitCommit:        v1.1.8-0-g82f18fe
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### Docker info

```shell
Client:
 Version:    24.0.6
 Context:    desktop-linux
 Debug Mode: false
Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 17
 Server Version: 24.0.6
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
```

### What happened?

Hi,
it seems like I'm unable to pull new images with testcontainers-go on my local machine. I know that the same config works on other operating systems / docker versions / 🤷‍♂️.

I had a look into it and dockercfs says in its comment for `GetRegistryCredentials(...)`: "If the returned username string is empty, the password is an identity token."
With the following commit I made it work again: https://github.com/stubents/testcontainers-go/commit/0b6a416d3636e32f3707f740f529f3d3e9014965

However, I'm not too familiar with container registry authentication mechanisms so I can't really say if this is the right thing to do or might break other stuff...

Nevertheless, hope this helps :)

[...]

PR #2866, opened 2024-10-31
fix: docker auth for identity tokens

stevenh:
Fix docker authentication look up so it supports the case where the username is blank which means that the second returned value is an identity token not a password.


stevenh:
> LGTM, thanks!
> 
> I'm curious how you find out this bug

It was [mentioned here](https://github.com/testcontainers/testcontainers-go/issues/1987#issuecomment-2301837453) by you 😄

PR #2867, opened 2024-10-31
fix: host access port instability

stevenh:
Fix clean up of sshd container used for HostAccessPorts if the container setup fails.

Fix ensureDefaultNetwork instability by preferring the default network if available. This prevents random port forwarding test failures due to different network bindings between the containers.

Fix data race when reading exec command with Multiplexed enabled.

Fix deadlock if an error was encountered reading from a container exec command.

Refactor port forwarding tests so they can be run individually from code editors.

Add timeout to port forwarding requests so that we don't need to wait for a full test timeout to identify failures.

### DIFF
--- a/port_forwarding.go
+++ b/port_forwarding.go
@@ -105,6 +105,7 @@ func exposeHostPorts(ctx context.Context, req *ContainerRequest, ports ...int) (
 		return sshdConnectHook, fmt.Errorf("inspect sshd container: %w", err)
 	}
 
+	// TODO: remove once we have docker context support via #2810
 	sshdIP := inspect.NetworkSettings.IPAddress
 	if sshdIP == "" {
 		single := len(inspect.NetworkSettings.Networks) == 1
--- a/port_forwarding.go
+++ b/port_forwarding.go
@@ -99,10 +99,25 @@ func exposeHostPorts(ctx context.Context, req *ContainerRequest, ports ...int) (
 		return sshdConnectHook, fmt.Errorf("new sshd container: %w", err)
 	}
 
-	// IP in the first network of the container
-	sshdIP, err := sshdContainer.ContainerIP(context.Background())
+	// IP in the first network of the container.
+	inspect, err := sshdContainer.Inspect(ctx)
 	if err != nil {
-		return sshdConnectHook, fmt.Errorf("get sshd container IP: %w", err)
+		return sshdConnectHook, fmt.Errorf("inspect sshd container: %w", err)
+	}
+
+	sshdIP := inspect.NetworkSettings.IPAddress
+	if sshdIP == "" {
+		single := len(inspect.NetworkSettings.Networks) == 1
+		for name, network := range inspect.NetworkSettings.Networks {
+			if name == sshdFirstNetwork || single {
+				sshdIP = network.IPAddress
+				break
+			}
+		}
+	}
+
+	if sshdIP == "" {
+		return sshdConnectHook, errors.New("sshd container IP not found")
 	}
 
 	if req.HostConfigModifier == nil {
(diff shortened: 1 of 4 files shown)


port_forwarding_test.go
```diff
@@ -22,101 +22,95 @@ const (
...
+	t.Run("single-port", func(t *testing.T) {
+		testExposeHostPorts(t, 1, false, true)
+	})
+
+	t.Run("single-port-using-network", func(t *testing.T) {
+		testExposeHostPorts(t, 1, true, true)
+	})
+
+	t.Run("single-port-cancellation", func(t *testing.T) {
+		testExposeHostPorts(t, 1, false, false)
```
mdelapenya, line 34: question: does it make sense another test case for "true, false" (has network but no host access) or that combination is out of the scope?

port_forwarding_test.go
```diff
@@ -22,101 +22,95 @@ const (
...
 	t.Helper()
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+
 	// wgetHostInternal {
 	code, reader, err := c.Exec(
-		context.Background(),
-		[]string{"wget", "-q", "-O", "-", fmt.Sprintf("http://%s:%d", testcontainers.HostInternal, port)},
+		ctx,
+		[]string{"wget", "-q", "-O", "-", "-T", "2", fmt.Sprintf("http://%s:%d", testcontainers.HostInternal, port)},
```
mdelapenya, line 113: question: we are setting 2s timeout just because it's higher than the 1s context cancellation. Is my assumption correct?

port_forwarding.go
```diff
@@ -166,11 +181,10 @@ func exposeHostPorts(ctx context.Context, req *ContainerRequest, ports ...int) (
 func newSshdContainer(ctx context.Context, opts ...ContainerCustomizer) (*sshdContainer, error) {
 	req := GenericContainerRequest{
 		ContainerRequest: ContainerRequest{
-			Image:           sshdImage,
-			HostAccessPorts: []int{}, // empty list because it does not need any port
-			ExposedPorts:    []string{sshPort},
-			Env:             map[string]string{"PASSWORD": sshPassword},
-			WaitingFor:      wait.ForListeningPort(sshPort),
+			Image:        sshdImage,
```
mdelapenya, line 184: neat! 🙈

port_forwarding.go
```diff
@@ -99,10 +99,25 @@ func exposeHostPorts(ctx context.Context, req *ContainerRequest, ports ...int) (
...
-	// IP in the first network of the container
-	sshdIP, err := sshdContainer.ContainerIP(context.Background())
+	// IP in the first network of the container.
+	inspect, err := sshdContainer.Inspect(ctx)
 	if err != nil {
-		return sshdConnectHook, fmt.Errorf("get sshd container IP: %w", err)
+		return sshdConnectHook, fmt.Errorf("inspect sshd container: %w", err)
+	}
+
+	sshdIP := inspect.NetworkSettings.IPAddress
```
mdelapenya, line 108: follow-up: this will be simplified after we remove all code to detect the default bridge network, right? If so, let's add a TODO comment for the   code to calculate the `sshdFirstNetwork` 🙏

mdelapenya (approved):
LGTM, thanks!

Added a few questions, not blockers, so feel free to resolve them if they do not apply. We'll merge right after that.

port_forwarding_test.go
stevenh: The context passed into `Exec` doesn't limit the time a process has to run, could be a bug, so we need the timeout on `wget` to ensure it doesn't block forever. Using a larger value ensures that we'll get a context cancellation error.

mdelapenya (approved):
LGTM, thanks!

PR #2869, opened 2024-11-01
feat: use recommended cluster settings for cockroachdb

stevenh:
## What does this PR do?

This PR configures the cockroachdb container with the [settings recommended by Cockroach Labs](https://www.cockroachlabs.com/docs/stable/local-testing#use-a-local-single-node-cluster-with-in-memory-storage) for testing clusters.

## Why is it important?

It seems reasonable to follow their advise but the why for each of these settings is explained in the link above.

## Related issues

None

## How to test this PR

I did some rather naive testing by running a test suite with and without these changes and it seemed like it did work in that the the tests took less time to finish with these settings than without them. It's not clear to me whether that improvement will be there in all possible scenarios but, as I said before, I think it's reasonable to follow Cockroach Labs advise on this.


modules/cockroachdb/certs.go
```diff
@@ -65,3 +66,20 @@ func NewTLSConfig() (*TLSConfig, error) {
 		ClientKey:  clientCert.KeyBytes,
 	}, nil
 }
+
+// tlsConfig returns a [tls.Config] for options.
+func (c *TLSConfig) tlsConfig() (*tls.Config, error) {
+	keyPair, err := tls.X509KeyPair(c.ClientCert, c.ClientKey)
+	if err != nil {
+		return nil, fmt.Errorf("x509 key pair: %w", err)
```
mdelapenya, line 74: lint: I think we need to gofmt this file, as the `fmt` package is not imported. Please run `make lint` for this module.

stevenh: Looks like the auto merge messed things up.

stevenh:
@martskins could you review these changes based on top of your PR #2869 please.

modules/cockroachdb/cockroachdb_test.go
```diff
@@ -18,26 +15,28 @@ import (
...
 	})
 }
 
 func TestCockroach_Password(t *testing.T) {
 	suite.Run(t, &AuthNSuite{
-		url: "postgres://foo:bar@localhost:xxxxx/defaultdb?sslmode=disable",
 		opts: []testcontainers.ContainerCustomizer{
 			cockroachdb.WithUser("foo"),
 			cockroachdb.WithPassword("bar"),
+			// Do not run the default statements as the user used on this test is
```
mdelapenya, line 37: question: after reading this comment, is it possible to have a test using `WithUser` and with the default statements?

martskins:
> @martskins could you review these changes based on top of your PR #2869 please.

Looks good to me! 👍

modules/cockroachdb/cockroachdb_test.go
mdelapenya: So at that point, the user must provide statements to configure the `MODIFYCLUSTERSETTING` privilege for the user in order this to work, right? Should the library grant this privilege automatically in the case the `WithUser` option is passed?

martskins: I suppose that could an option, yes. Alternatively, [it looks like](https://github.com/testcontainers/testcontainers-go/blob/main/modules/cockroachdb/cockroachdb.go#L144) we don't allow using the `TLS` option, should we maybe do the same here?

mdelapenya: We can probably do it in a follow-up. Wdyt? Would you like to be involved in that?

martskins: I'm happy to be involved, but if the preferred path is to make this error if default statements are used with a non-root user then it might be better to do it in this PR so we don't ship this only to introduce a breaking change to it right after? Although I suppose the cluster would fail to start in that scenario anyways so maybe it's not that much of a breaking change.

mdelapenya (approved):
@stevenh is this a breaking change? I do not see any change in the signatures of the public API, just changes in how we build the examples on how to use it, but maybe I'm wrong.

Other than that LGTM

stevenh:
> @stevenh is this a breaking change? I do not see any change in the signatures of the public API, just changes in how we build the examples on how to use it, but maybe I'm wrong.
> 
> Other than that LGTM

The connection handling is not 100% backwards compatible as detailed in the PR message:
> This now returns a registered connection string so is no longer compatible with pgx.ParseConfig, use ConnectionConfig for this use case instead.

mdelapenya:
I think this is ready to be merge, but let's wait a bit until @eddumelendez 's question is addressed in the upstream project. That would be a good improvement if all the TC languages would benefit from this, instead of having each custom code.

eddumelendez:
In order to follow [the suggestion](https://github.com/cockroachdb/cockroach/issues/134429#issuecomment-2462512167), the wait strategy should also check if a fail was created. See https://github.com/testcontainers/testcontainers-java/pull/9505. This approach is valid for CockroachDB >= 22.1.0

stevenh:
> In order to follow [the suggestion](https://github.com/cockroachdb/cockroach/issues/134429#issuecomment-2462512167), the wait strategy should also check if a fail was created. See [testcontainers/testcontainers-java#9505](https://github.com/testcontainers/testcontainers-java/pull/9505). This approach is valid for CockroachDB >= 22.1.0

Thanks @eddumelendez I've incorporated that into refactored approach for the go CockroachDB module in https://github.com/testcontainers/testcontainers-go/pull/2883

This is still draft but I think the changes improve the flow, avoiding manually creating certs which prevents a bunch of the edge cases in the current version as well as streamlining to logic so users can use the functional options with less restrictions.

Be interested in feedback.

stevenh:
Converted this to draft while we discuss the different approaches

stevenh:
Merged in https://github.com/testcontainers/testcontainers-go/pull/2883

PR #2876, opened 2024-11-05
chore(ci): run lint in a separate build before running the tests

mdelapenya:
## What does this PR do?
It extracts the golangci-lint step to a separate job, so that it can be called separately before the tests.

To detect the changes, we have created a shell script, `scripts/changed-modules.sh`. This script receives one single env var as input parameter, `ALL_CHANGED_FILES`, which on CI will be provided by a Github action, https://github.com/tj-actions/changed-files, that puts in there a list of the modified files, comparing the current PR changeset with the parent branch (main). We can tune this up to always check the latest `main` branch (open to discussion here).

The script will compare all existing modules (looking up all the `go.mod` files of interest, no test files) with the modified files, building an array of modified modules. At the moment there is modified file in the core, in other words the entire root directory without modulegen/modules/examples, all the modules will be included in the build. There are some exclusions for the code: files in the `.devcontainer, .vscode, docs, scripts` will not cause all modules will be included. So it could be the case that no modules could end up in the build if only files in those exclusions are changed. In that case, we are going to conditionally skip the lint, test and upload-to-sonar jobs.

Finally, for running the tests, we are going to keep the matrix for running them in multiple Go versions and OSs, with the following rules:
- any Go module will run the tests for the two most recent Go releases (core, modulegen, modules and examples).
- the core module, modules and examples will run the tests on `ubuntu-latest`.
- the modulegen module will run the tests on `ubuntu-latest`, `macos-latest` and `windows-latest`.
- the jobs for reaper-off and rootless-mode will run the tests for the two most recent Go releases, on `ubuntu-latest, only.

BTW I was able to iterate faster on this at the moment I discovered the power of https://github.com/nektos/act. Fantastic tool!

## Why is it important?
There are two major goals:
- Do not run the lint until it's too late in the build.
- Reduce the feedback loop when contributing to modules. At the moment, a module contributor has to wait for all the core changes to be built and pass the tests. With this approach, if the PR touches just one module, then that module will be added to the workflow. 

As a drawback, in my opinion, is that two different builds with two different changesets won't run the same set of tests, so they are not deterministic. I think this is a minor detail that would be satisfied with the improved experience for module contributors at CI time. For core maintainers, we will probably run all the modules per PR, but that could be expected to verify we are not breaking the downstream modules.

## How to test this PR
The shell script is the key part, and it's explained in the comments for the script. It requires running the script with the required environment variables for the different use cases:

 1. A Go file from the core module is modified:
    ALL_CHANGED_FILES="examples/nginx/go.mod examples/foo/a.txt a/b/c/d/a.go" ./scripts/changed-modules.sh
    The output should be: all modules.

 2. A file from a module in the modules dir is modified:
    ALL_CHANGED_FILES="modules/nginx/go.mod" ./scripts/changed-modules.sh
    The output should be: just the modules/nginx module.

 3. A file from a module in the examples dir is modified:
    ALL_CHANGED_FILES="examples/nginx/go.mod" ./scripts/changed-modules.sh
    The output should be: just the examples/nginx module.

 4. A Go file from the modulegen dir is modified:
    ALL_CHANGED_FILES="modulegen/a.go" ./scripts/changed-modules.sh
    The output should be: just the modulegen module.

 5. A non-Go file from the core dir is modified:
    ALL_CHANGED_FILES="README.md" ./scripts/changed-modules.sh
    The output should be: all modules.

 6. A file from two modules in the modules dir are modified:
    ALL_CHANGED_FILES="modules/nginx/go.mod modules/localstack/go.mod" ./scripts/changed-modules.sh
    The output should be: the modules/nginx and modules/localstack modules.

 7. Files from the excluded dirs are modified:
    ALL_CHANGED_FILES="docs/a.md .vscode/a.json .devcontainer/a.json scripts/a.sh" ./scripts/changed-modules.sh
    The output should be: no modules.

## Follow-ups
**IMPORTANT!:** We need to update the branch settings so that the new checks are used to allow the merge.


.github/workflows/ci.yml
```diff
@@ -1,46 +1,69 @@
...
+    outputs:
+      modules: ${{ steps.set-modified-modules.outputs.modules }}
+      modules_count: ${{ steps.set-modified-modules-count.outputs.modules_count }}
+    steps:
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4
+
+      - id: changed-files
+        name: Get changed files
+        uses: tj-actions/changed-files@v45
```
v1v, line 25: What do you think to pinning the commit? 

```suggestion
        uses: tj-actions/changed-files@4edd678ac3f81e2dc578756871e4d00c19191daf # v45.0.4
```

mdelapenya: Done, thanks for this! 🙏

scripts/changed-modules.sh
```diff
@@ -0,0 +1,114 @@
+#!/usr/bin/env bash
```
stevenh, line 1: suggestion: set the options to detect issues e.g `set -euxo pipefail` if you want to be really strict.

.github/workflows/ci-lint-go.yml
```diff
@@ -0,0 +1,63 @@
...
+        with:
+          # Optional: version of golangci-lint to use in form of v1.2 or v1.2.3 or `latest` to use the latest version
+          version: v1.61.0
+          # Optional: working directory, useful for monorepos
+          working-directory: ${{ inputs.project-directory }}
+          # Optional: golangci-lint command line arguments.
+          args: --verbose
+          # Optional: if set to true then the all caching functionality will be complete disabled,
+          #           takes precedence over all other caching options.
+          skip-cache: true
```
stevenh, line 45: question: why set this to true?

.github/workflows/ci-lint-go.yml
```diff
@@ -0,0 +1,63 @@
...
+
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@971e284b6050e8a5849b72094c50ab08da042db8 # v6.1.1
+        with:
+          # Optional: version of golangci-lint to use in form of v1.2 or v1.2.3 or `latest` to use the latest version
+          version: v1.61.0
+          # Optional: working directory, useful for monorepos
+          working-directory: ${{ inputs.project-directory }}
+          # Optional: golangci-lint command line arguments.
+          args: --verbose
```
stevenh, line 42: suggestion: remove verbose as it just adds noise when looking at the output

.github/workflows/ci-lint-go.yml
```diff
@@ -0,0 +1,63 @@
...
+  # pull-requests: read
+
+jobs:
+  lint-go-project:
+    name: "lint: ${{ inputs.project-directory }}"
+    runs-on: 'ubuntu-latest'
+    continue-on-error: false
+    steps:
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4
```
stevenh, line 25: suggestion: can we detail the specific version not just the major, so we can easily see when we want to update

.github/workflows/ci-lint-go.yml
```diff
@@ -0,0 +1,63 @@
...
+
+      - name: Set up Go
+        uses: actions/setup-go@0c52d547c9bc32b1aa3301fd7a9cb496313a4491 # v5
```
stevenh, line 28: suggestion: can we detail the specific version not just the major, so we can easily see when we want to update

.github/workflows/ci-lint-go.yml
```diff
@@ -0,0 +1,63 @@
...
+        with:
+            go-version-file: "${{ inputs.project-directory == '' && '.' || inputs.project-directory }}/go.mod"
+            cache-dependency-path: "${{ inputs.project-directory == '' && '.' || inputs.project-directory }}/go.sum"
+        id: go
+
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@971e284b6050e8a5849b72094c50ab08da042db8 # v6.1.1
+        with:
+          # Optional: version of golangci-lint to use in form of v1.2 or v1.2.3 or `latest` to use the latest version
+          version: v1.61.0
```
stevenh, line 38: suggestion: update to v1.63.4?

.github/workflows/ci-lint-go.yml
```diff
@@ -0,0 +1,63 @@
...
+      - name: generate
+        working-directory: ./${{ inputs.project-directory }}
+        shell: bash
+        run: |
+          make generate
+          git --no-pager diff && [[ 0 -eq $(git status --porcelain | wc -l) ]]
+
+      - name: modVerify
+        working-directory: ./${{ inputs.project-directory }}
+        run: go mod verify
```
stevenh, line 56: question: is this really needed as CI is clean each time it starts?

.github/workflows/ci-lint-go.yml
```diff
@@ -0,0 +1,63 @@
...
+permissions:
+  contents: read
+  # Optional: allow read access to pull request. Use with `only-new-issues` option.
+  # pull-requests: read
+
+jobs:
+  lint-go-project:
+    name: "lint: ${{ inputs.project-directory }}"
+    runs-on: 'ubuntu-latest'
+    continue-on-error: false
```
stevenh, line 22: suggestion: this is the default so we can remove

.github/workflows/ci-lint-go.yml
```diff
@@ -0,0 +1,63 @@
...
+
+      - name: Set up Go
+        uses: actions/setup-go@0c52d547c9bc32b1aa3301fd7a9cb496313a4491 # v5
+        with:
+            go-version-file: "${{ inputs.project-directory == '' && '.' || inputs.project-directory }}/go.mod"
+            cache-dependency-path: "${{ inputs.project-directory == '' && '.' || inputs.project-directory }}/go.sum"
+        id: go
+
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@971e284b6050e8a5849b72094c50ab08da042db8 # v6.1.1
```
stevenh, line 35: suggestion: update to v6.2.0

stevenh (changes requested):
A few suggestions / questions

.github/workflows/ci-lint-go.yml
mdelapenya: This is already set to true in the main branch 🤔

.github/workflows/ci-lint-go.yml
mdelapenya: It's already set in the main branch, but because this PR is about the lint workflow, seems correct to bump it here. Doing it now, thanks!

.github/workflows/ci-lint-go.yml
mdelapenya: I'm removing it elsewhere but in the windows workers for the Docker infra

.github/workflows/ci-lint-go.yml
stevenh: I suspect it might be hang over from quite a few versions back where `actions/setup-go` and `golangci/golangci-lint-action` both used to cache package and build, but this was fixed in v4.

I would suggest removing it as it can have a significant impact on the time taken to run the linter, especially as we've been enabling more linting recently.

stevenh (changes requested):
Clarified why I think skip-cache was likely set and suggest we remove for now, thoughts?

stevenh:
Not sure if the expected test that's pending is a side effect of this change @mdelapenya?

image
image

mdelapenya:
> Not sure if the expected test that's pending is a side effect of this change @mdelapenya?

This is because we have a setting in the GH repo, as code in the `.github/settings.yml` file, with the minimal set of checks that are required to mark the PR as mergeable.

With a dynamic set of checks, I'm not sure if we need this setting anymore, or at least, look for a final node in the pipeline that is green. Do you agree that we could skip it and find the solution in a follow-up?

stevenh:
You can remove it from the required check for now, but should follow up.

stevenh (approved):
We can follow with require checks changes if needed.

mdelapenya:
I created https://github.com/testcontainers/testcontainers-go/issues/2968 to track the GH checks revamp

PR #2881, opened 2024-11-07
fix!: port forwarding clean up and make private

stevenh:
Ensure that when the port forward container is shutdown all forwarding is also stopped by switching to a context to handle cancellation correctly.

Ensure that port forwarding connection attempts are cancelled correctly when the container context is cancelled, previously these could wait for a longer timeout.

Don't cancel port forwarding when the context used to start a container is cancelled as this is unexpected behaviour as its a non-explicit use of the context after the function has returned.

Ensure forwarding clean up errors are reported on container stop and termination.

Ensure port forwarding connections are cleaned up if one of the requested ports reports an error on setup.

Port forwarding test style clean up.

Update sshd module to the latest version.

BREAKING_CHANGE: make port forwarding type and constructor private.


mdelapenya (approved):
LGTM, thanks!

PR #2883, opened 2024-11-08
refactor(cockroachdb): to use request driven options

stevenh:
Refactor cockroachdb module to to use request driven options simplifying the flow.

Clean up tests eliminating the use of suite which significantly speeds up test runs.


modules/cockroachdb/options.go
```diff
@@ -1,180 +1,106 @@
...
-// WithDatabase sets the name of the database to use.
-func WithDatabase(database string) Option {
-	return func(o *options) {
-		o.Database = database
-	}
-}
+// WithStoreSize sets the amount of available [in-memory storage].
+//
+// [in-memory storage]: https://www.cockroachlabs.com/docs/stable/cockroach-start#store
+func WithStoreSize(size string) testcontainers.CustomizeRequestOption {
```
mdelapenya, line 62: docs: let's update the docs/modules/cockroachdb.md file with the new options

stevenh: I thought I got them all, did I miss some?

mdelapenya: For some reason, I missed them in the review. I was probably looking at one commit, not the entire change set. My fault

stevenh:
This should be split out into components if we want to go down this route.

wait/tls_test.go
```diff
@@ -0,0 +1,146 @@
...
+
+const (
+	serverName         = "127.0.0.1"
+	caFilename         = "/tmp/ca.pem"
+	clientCertFilename = "/tmp/cert.crt"
+	clientKeyFilename  = "/tmp/cert.key"
+)
+
+var (
+	//go:embed testdata/cert.crt
```
mdelapenya, line 31: question: Have you considered using https://github.com/mdelapenya/tlscert instead?

wait/walk_test.go
```diff
@@ -0,0 +1,87 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/testcontainers/testcontainers-go"
+	"github.com/testcontainers/testcontainers-go/wait"
+)
+
+func TestWalk(t *testing.T) {
+	req := testcontainers.ContainerRequest{
```
mdelapenya, line 14: suggestion: I think we do not need the req, just the WaitingFor field

wait/walk.go
```diff
@@ -0,0 +1,73 @@
...
+
+	if err := visit(*root); err != nil {
+		if errors.Is(err, VisitRemove) {
+			*root = nil
+		}
+
+		return err
+	}
+
+	if s, ok := (*root).(*MultiStrategy); ok {
```
mdelapenya, line 54: question: shouldn't we check first if the root is a multi-strategy, to iterate through the child elements?

wait/walk_test.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	t.Run("walk", func(t *testing.T) {
+		var count int
+		err := wait.Walk(&req.WaitingFor, func(_ wait.Strategy) error {
+			count++
+			return nil
+		})
+		require.NoError(t, err)
+		require.Equal(t, 5, count)
```
mdelapenya, line 31: question: because the root has no logic, apart from being a "container" for the children strategies, should we count 4 here, therefore affecting the processing? I think this is related to my previous question about the root node being a multi.

wait/walk_test.go
```diff
@@ -0,0 +1,87 @@
...
+				return wait.VisitRemove
+			}
+
+			return nil
+		})
+		require.NoError(t, err)
+		require.Equal(t, 5, count)
+		require.Equal(t, 2, matched)
+
+		count = 0
```
mdelapenya, line 59: suggestion: I think we should reset the matched counter here too, so it reflects that the FileStrategy node was removed and doing another pass discovers just one.

wait/walk_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/testcontainers/testcontainers-go"
+	"github.com/testcontainers/testcontainers-go/wait"
+)
+
+func TestWalk(t *testing.T) {
+	req := testcontainers.ContainerRequest{
+		WaitingFor: wait.ForAll(
```
mdelapenya, line 15: suggestion: let's add a test for having just one strategy as child in the multi, for covering the edge cases.

modules/cockroachdb/data/cluster_defaults.sql
```diff
@@ -0,0 +1,8 @@
+SET CLUSTER SETTING kv.range_merge.queue_interval = '50ms';
```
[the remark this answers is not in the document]
stevenh: Looks like there is one change since this advice was introduced in v21.2.17 (a setting was removed in v21.2.22), so looks like we're good. If this was needed we could look to introduce version conditionals.

modules/cockroachdb/options.go
```diff
@@ -1,69 +1,122 @@
...
 }
 
-// WithStoreSize sets the amount of available in-memory storage.
-// See https://www.cockroachlabs.com/docs/stable/cockroach-start#store
-func WithStoreSize(size string) Option {
-	return func(o *options) {
-		o.StoreSize = size
+// WithInitScripts adds the given scripts to those automatically run when the container starts.
+// These will be ignored if data exists in the `/cockroach/cockroach-data` directory within the container.
+func WithInitScripts(scripts ...string) testcontainers.CustomizeRequestOption {
```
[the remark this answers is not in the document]
stevenh: Readers provides more flexibility so switching to that would be a good idea. Users that want to read from a file can easily use `embed` + `bytes.NewReader` to achieve that.

wait/tls_test.go
stevenh: I saw that, but it was hiding errors e.g. [SelfSignedFromRequest](https://pkg.go.dev/github.com/mdelapenya/tlscert#SelfSignedFromRequest) returns nil on error, so followed the pattern in the standard library. If that got cleaned up we could change later.

wait/walk_test.go
stevenh: You're correct, but I wanted the test to act as an example of its use as well, hence using request.

wait/walk.go
stevenh: I wanted to allow the caller to customise the behaviour. This way round they could return an error or VisitRemove to remove the element without having to traverse it first. Added a comment to this effect.

wait/walk_test.go
stevenh: As we visit the root first to allow it to be customised, the root is also counted.

modules/cockroachdb/options.go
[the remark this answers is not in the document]
stevenh: Just came to implement this and we missed that you need a file name, which determines the order, so this would either require a custom type or fixed order and generate names, thoughts?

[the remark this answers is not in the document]
mdelapenya: We already have a custom type: ContainerFile, which holds a reader and a target file path:
```go
type ContainerFile struct {
	HostFilePath      string    // If Reader is present, HostFilePath is ignored
	Reader            io.Reader // If Reader is present, HostFilePath is ignored
	ContainerFilePath string
	FileMode          int64
}
```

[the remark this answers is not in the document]
stevenh: IMO that wouldn't add much value as you could just use that now. If we used `ContainerFile` for this the only benefit to having this method would be not having to set the destination directory.

modules/cockroachdb/cockroachdb.go
```diff
@@ -252,29 +254,57 @@ func connString(opts options, host string, port nat.Port) string {
...
-		return nil, err
-	}
+	// Walk the wait strategies to find the TLS strategy and either remove it or
+	// update the client certificate files to match the user and configure the
+	// container to use the TLS strategy.
+	if err := wait.Walk(&req.WaitingFor, func(strategy wait.Strategy) error {
+		if cert, ok := strategy.(*wait.TLSStrategy); ok {
+			if insecure {
+				// If insecure mode is enabled, the certificate strategy is removed.
+				return errors.Join(wait.VisitRemove, wait.VisitStop)
```
mdelapenya, line 293: I see now it pretty well, and I like it. Thanks!

follow-up: we probably need an example like this one in https://golang.testcontainers.org/features/wait/walk/

modules/cockroachdb/data/cluster_defaults.sql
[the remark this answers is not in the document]
mdelapenya: Indeed, although I'd not like to be in the place of maintaining multiple setups, but instead letting users do it in their code. I'm worried about coupling a certain configuration with a certain range of versions that is not documented anywhere. This comment applies to all modules, not only this one.

In any case, we could validate that the image is above a given semver value, also making the docs be very explicit about the `WithNoClusterDefaults` option.

mdelapenya (approved):
LGTM, thanks! Just a few comments, minor ones.

modules/cockroachdb/cockroachdb.go
stevenh: Yep planning to do a follow up PR for that.

stevenh:
@mdelapenya I've added clarification to the Run method about what version is supported. I tested all the way back to v22.2.0, which was the first version which introduced the init_success file creation once the cluster init has completed.

mdelapenya (approved):
LGTM, thanks!

PR #2490, opened 2024-04-19
Fix kafka internal docker connection

catinapoke:
## What does this PR do?
I fixed passing alias name in startup script when it's possible. 

## Why is it important?
You can't connect to kafka container from other container at the same docker network.  
Details: https://www.confluent.io/blog/kafka-listeners-explained/  

## Related issues

I have stumbled into issue with connecting from app docker container to kafka docker container at the same network. I tried many options including altering KAFKA_ADVERTISED_LISTENERS, but none of them worked. Finally, I found out that you overwrite KAFKA_ADVERTISED_LISTENERS in startup script.

Options I tried:  
1. Alter KAFKA_ADVERTISED_LISTENERS   
2. Change network type from bridge to host (it's impossible)  
3. Add new listeners  

## How to test this PR
I have tested it in my ProofOfConcept repo https://github.com/catinapoke/testcontainers-poc in kafka_test.go file
You can run it `go test -v ./tests/integration/kafka_test.go `

PR #2894, opened 2024-11-21
fix(kafka): Fix internal docker connection

mdelapenya:
## What does this PR do?

Follow-up of #2490 in order to be able to push commits and make progress.

## Why is it important?
Give importance to @catinapoke's work in #2490, keeping the commits (and authorship), and doing the toil tasks for them. This way we can push to this feature branch (the original PR is using the main branch, so we cannot update it)
 

## Related issues

- Supersedes #2490


modules/kafka/kafka.go
```diff
@@ -155,12 +215,43 @@ func copyStarterScript(ctx context.Context, c testcontainers.Container) error {
 	return nil
 }
 
-func WithClusterID(clusterID string) testcontainers.CustomizeRequestOption {
-	return func(req *testcontainers.GenericContainerRequest) error {
-		req.Env["CLUSTER_ID"] = clusterID
+func editEnvsForListeners(listeners []KafkaListener) map[string]string {
+	if len(listeners) == 0 {
+		// no change
+		return map[string]string{}
```
stevenh, line 221: suggestion: return nil as its just used as an iterator
```suggestion
		return nil
```

modules/kafka/kafka.go
```diff
@@ -155,12 +215,43 @@ func copyStarterScript(ctx context.Context, c testcontainers.Container) error {
...
+
+	// expect first listener has common network between kafka instances
+	envs["KAFKA_INTER_BROKER_LISTENER_NAME"] = listeners[0].Name
+
+	// expect small number of listeners, so joins is okay
+	for _, item := range listeners {
+		envs["KAFKA_LISTENERS"] = strings.Join(
+			[]string{
+				envs["KAFKA_LISTENERS"],
+				fmt.Sprintf("%s://0.0.0.0:%s", item.Name, item.Port),
```
stevenh, line 238: suggestion: use concat to simplify
```suggestion
				item.Name + "://0.0.0.0:" + item.Port,
```

modules/kafka/kafka_helpers_test.go
```diff
@@ -9,101 +9,171 @@ import (
...
+		if wantErr {
+			require.Error(t, err)
+		} else {
+			require.NoError(t, err)
+		}
 	}
+
+	t.Run("official/valid-version", func(t *testing.T) {
+		validateKRaftVersionFn(t, "confluentinc/confluent-local:7.5.0", false)
+	})
```
stevenh, line 69: suggestion: use direct to make it clearer, removing the function above and calling the methods directly.
```suggestion
	t.Run("official/valid-version", func(t *testing.T) {
		err := validateKRaftVersion("confluentinc/confluent-local:7.5.0")
		require.NoError(t, err)
	})
```

modules/kafka/kafka_helpers_test.go
```diff
@@ -9,101 +9,171 @@ import (
...
-			image:   "confluentinc/confluent-local:5.0.0",
-			wantErr: true,
-		},
-		{
-			name:    "Unofficial does not validate KRaft version",
-			image:   "my-kafka:1.0.0",
-			wantErr: false,
-		},
+func TestTrimValidateListeners(t *testing.T) {
+	validateFn := func(t *testing.T, listeners []KafkaListener, wantErr bool) {
```
stevenh, line 89: suggestion: as above, eliminate the func and embed directly to make it clearer as true / false are is ambiguous until you look at the function.

modules/kafka/kafka_test.go
```diff
@@ -74,6 +78,297 @@ func TestKafka_invalidVersion(t *testing.T) {
...
+
+		// test data
+		key      = "wow"
+		text_msg = "test-input-external"
+	)
+
+	Network, err := network.New(ctx)
+	if err != nil {
+		t.Fatal(err)
+	}
```
stevenh, line 100: suggestion: use require.NoError

modules/kafka/kafka_test.go
```diff
@@ -74,6 +78,297 @@ func TestKafka_invalidVersion(t *testing.T) {
...
+			{
+				Name: "BROKER",
+				Host: "kafka",
+				Port: "9092",
+			},
+		}),
+	)
+	// }
+	testcontainers.CleanupContainer(t, kafkaContainer)
+	require.NoError(t, err, "failed to start kafka container")
```
stevenh, line 117: suggestion: No need for the messages, more below

modules/kafka/kafka_test.go
```diff
@@ -74,6 +78,297 @@ func TestKafka_invalidVersion(t *testing.T) {
...
+	_, _, err = kcat.Exec(ctx, []string{"kcat", "-b", address, "-t", topic_out, "-P", "-l", tempfile})
+	require.NoError(t, err, "send message with kcat")
+
+	// External read
+	client, err := sarama.NewConsumerGroup(brokers, "groupName", config)
+	require.NoError(t, err, "create consumer group")
+
+	consumer, _, done, cancel := NewTestKafkaConsumer(t)
+
+	sCtx, _ := context.WithTimeout(context.Background(), time.Second)
```
stevenh, line 180: bug: as flagged by linter, need to cancel.

modules/kafka/kafka_test.go
```diff
@@ -74,6 +78,297 @@ func TestKafka_invalidVersion(t *testing.T) {
...
+	require.NoError(t, err)
+
+	_, _, err = kcat.Exec(ctx, []string{"kcat", "-b", address, "-t", topic_out, "-P", "-l", tempfile})
+	require.NoError(t, err, "send message with kcat")
+
+	// External read
+	client, err := sarama.NewConsumerGroup(brokers, "groupName", config)
+	require.NoError(t, err, "create consumer group")
+
+	consumer, _, done, cancel := NewTestKafkaConsumer(t)
```
stevenh, line 178: bug: should call cancel directly after, to ensure its called

stevenh (changes requested):
I wonder if we could use the same approach as see in the this https://github.com/testcontainers/testcontainers-go/pull/2883?

modules/kafka/kafka_test.go
mdelapenya: I'm having issues with this test, see https://github.com/testcontainers/testcontainers-go/pull/2490#issuecomment-2493002440

modules/kafka/kafka.go
```diff
@@ -36,6 +36,13 @@ echo '' > /etc/confluent/docker/ensure
 type KafkaContainer struct {
 	testcontainers.Container
 	ClusterID string
+	Listeners KafkaListener
```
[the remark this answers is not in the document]
mdelapenya: In fact, I'm realising this field is not used anywhere else. Do you think it's worth it to keep it (through the public function)?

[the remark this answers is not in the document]
stevenh: If it's not needed externally the. Just make it private

modules/kafka/kafka.go
```diff
@@ -121,32 +143,70 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
 	return c, nil
 }
 
+func trimValidateListeners(listeners []KafkaListener) error {
+	// Trim
+	for i := 0; i < len(listeners); i++ {
+		listeners[i].Name = strings.ToUpper(strings.Trim(listeners[i].Name, " "))
+		listeners[i].Host = strings.Trim(listeners[i].Host, " ")
+		listeners[i].Port = strings.Trim(listeners[i].Port, " ")
+	}
```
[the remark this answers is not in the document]
mdelapenya: Correct. I'm changing this to sanitise the listeners in the option, updating the validation tests

modules/kafka/kafka.go
mdelapenya: I think we should apply this values in the functional option. Wdyt?

stevenh: Yes, we should always try to have standalone functional options which act directly on the request if possible, makes things much cleaner and also allows further options to override if needed.

modules/kafka/kafka.go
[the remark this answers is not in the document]
mdelapenya: Not even assigned 🤔 so I'm removing it. If we see it's important to expose the listeners in the future, we can easily add it

mdelapenya:
@stevenh I addressed the majority of your comments: all except those related to the `TestKafka_networkConnectivity` test, which is still failing for me locally.

modules/kafka/kafka.go
```diff
@@ -121,32 +143,70 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
...
+	for _, item := range listeners {
+		if names[item.Name] {
+			return fmt.Errorf("duplicate of listener name: %s", item.Name)
+		}
+		names[item.Name] = true
+
+		if ports[item.Port] {
+			return fmt.Errorf("duplicate of listener port: %s", item.Port)
+		}
+		ports[item.Port] = true
```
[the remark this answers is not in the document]
mdelapenya: I'm not a kafka expert, so I'm not sure about that. @eddumelendez @catinapoke what are your thoughts on this?

modules/kafka/kafka.go
```diff
@@ -36,6 +36,13 @@ echo '' > /etc/confluent/docker/ensure
 type KafkaContainer struct {
 	testcontainers.Container
 	ClusterID string
+	Listeners KafkaListener
+}
+
+type KafkaListener struct {
```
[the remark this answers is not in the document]
mdelapenya: the `WithListeners` option needs it, so the client code needs to access it to provide valid values for the listeners.

stevenh (changes requested):

PR #2895, opened 2024-11-25
feat(wait): strategy walk

stevenh:
Add the Walk method which allows consumers to walk strategies and if desired update or remove them.

This is useful for modules which want to update strategy details based on user customisations performed by functional options passed to their run methods.


docs/features/wait/walk.md
```diff
@@ -0,0 +1,12 @@
+# Walk
```
mdelapenya, line 1: docs: we'll need to add it to the mkdocs.yaml file too. See https://github.com/testcontainers/testcontainers-go/pull/2896#discussion_r1858517337

docs/features/wait/walk.md
```diff
@@ -0,0 +1,12 @@
...
+
+Walk walks the strategies tree and calls the visit function for each node.
```
mdelapenya, line 3: question: I see the value in walking though the strategies, could you elaborate more about the use cases in the docs? 🙏

docs/features/wait/walk.md
```diff
@@ -2,11 +2,18 @@
 
 Walk walks the strategies tree and calls the visit function for each node.
 
+This allows modules to easily amend default wait strategies, updating or
+removing specific strategies based on requirements of functional options.
+
+For example removing a TLS strategy if a functional option enabled insecure mode
+or changing the location of the certificate based on the configured user.
+
```
mdelapenya, line 10: Fantastic! Thanks for making it clear 🙏

stevenh: Sorry you said TLS specifically, this is a FileStrategy removal example, we can add a TLS one when it's integrated into couchbasedb as it will be used there.

mdelapenya: Indeed, just mentioned the above example for 
> For example removing a TLS strategy if a functional option enabled insecure mode
or changing the location of the certificate based on the configured user.

Fine for me if we elaborate on this as a follow-up 👍

mdelapenya (approved):
LGTM, thanks!

PR #2896, opened 2024-11-25
feat(wait): tls strategy

stevenh:
Extract TLS certificate wait strategy into a dedicated wait type so it can be reused.

Use embed to simplify wait test loading of certs.


wait/tls_test.go
```diff
@@ -0,0 +1,146 @@
...
+		require.Equal(t, []tls.Certificate{cert}, got.Certificates)
+		require.True(t, pool.Equal(got.RootCAs))
+	})
+}
+
+func ExampleForTLSCert() {
+	ctx := context.Background()
+
+	// waitForTLSCert {
+	forCert := wait.ForTLSCert("/app/tls.pem", "/app/tls-key.pem").
```
mdelapenya, line 106: suggestion: add a comment with the origin of the certificates, which are set in the Dockerfile. Maybe interesting to add a code block in the docs site showing the entire Dockerfile

wait/tls.go
```diff
@@ -0,0 +1,167 @@
...
+
+// x509KeyPair is a pair of certificate and key files.
+type x509KeyPair struct {
+	certPEMFile string
+	keyPEMFile  string
+}
+
+// ForTLSCert returns a CertStrategy that will add a Certificate to the [tls.Config]
+// constructed from PEM formatted certificate key file pair in the container.
+func ForTLSCert(certPEMFile, keyPEMFile string) *TLSStrategy {
```
mdelapenya, line 37: question: do you think it would be interesting to provide a way to forget about generating certs and the library build them on the fly? Something like `wait.ForTLSTestCert`? The library would generate the file under the hood.

mdelapenya (commented):
LGTM, thanks! I added some comments, not blockers, but want to discuss them before accepting it.

Also, do you foresee any usage for this strategy in the existing modules? Elasticsearch, Cockroachdb, and more use TLS certs.

stevenh:
> LGTM, thanks! I added some comments, not blockers, but want to discuss them before accepting it.
> 
> Also, do you foresee any usage for this strategy in the existing modules? Elasticsearch, Cockroachdb, and more use TLS certs.

Great stuff, yes this is extracted from the [cockroachdb PR](https://github.com/testcontainers/testcontainers-go/pull/2883) but it could be used with any module where TLS is needed.

wait/tls.go
stevenh: Yep love that idea, something for a follow up PR.

mdelapenya (changes requested):
Last minute review of the docs site, and I found the new page is not discoverable.

Because this is a new page, we need to add it to the mkdocs.yml descriptor:

Please see https://deploy-preview-2896--testcontainers-go.netlify.app/features/wait/introduction/

This is not automated, although there is room for improving that.

docs/features/wait/introduction.md
```diff
@@ -15,6 +15,7 @@ Below you can find a list of the available wait strategies that you can use:
 - [Log](./log.md)
 - [Multi](./multi.md)
 - [SQL](./sql.md)
+- [TLS](./tls.md)
```
mdelapenya, line 18: Oh sorry, I was not clear, we also need to add it here: https://github.com/testcontainers/testcontainers-go/blob/main/mkdocs.yml#L56

mdelapenya (approved):
LGTM, only the mkdocs is missing, but other than that, it's ready

mdelapenya (approved):
LGTM, thanks!

PR #2899, opened 2024-11-28
fix: container binds syntax

zregvart:
## What does this PR do?

The container binds syntax is `host-src:container-dest[:options]` or `volume-name:container-dest[:options]` where `options` is a comma separated list of: `nocopy` `[ro|rw]`, `[z|Z]` or `[[r]shared|[r]slave|[r]private]`. See `HostConfig.Binds` of the Create a container request[1]

[1] https://docs.docker.com/reference/api/engine/version/v1.47/#tag/Container/operation/ContainerCreate

## Why is it important?

It is not possible to set some of these options (e.g. SELinux labels) otherwise.

## Related issues

-

### DIFF
--- a/container_test.go
+++ b/container_test.go
@@ -73,11 +73,33 @@ func Test_ContainerValidation(t *testing.T) {
 		},
 		{
 			Name:          "Invalid bind mount",
-			ExpectedError: "invalid bind mount: /data:/data:/data",
+			ExpectedError: "invalid bind mount: /data:/data:a:b",
 			ContainerRequest: testcontainers.ContainerRequest{
 				Image: "redis:latest",
 				HostConfigModifier: func(hc *container.HostConfig) {
-					hc.Binds = []string{"/data:/data:/data"}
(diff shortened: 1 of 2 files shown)


container_test.go
```diff
@@ -73,11 +73,33 @@ func Test_ContainerValidation(t *testing.T) {
...
 			ContainerRequest: testcontainers.ContainerRequest{
 				Image: "redis:latest",
 				HostConfigModifier: func(hc *container.HostConfig) {
-					hc.Binds = []string{"/data:/data:/data"}
+					hc.Binds = []string{"/data:/data:a:b"}
+				},
+			},
+		},
+		{
+			Name: "Can provide bind options",
```
mdelapenya, line 85: suggestion: we are slowly moving to using test names without spaces, something like "bind-options/provided" or similar, which allows running failed tests more easily.

mdelapenya (approved):
LGTM, thanks!

PR #2905, opened 2024-12-01
feat(nats): WithConfigFile - pass a configuration file to nats server

vyskocilm:
## What does this PR do?

Nats server can be started with a custom configuration file.

## Why is it important?

While a lot of options can be passed as a command line parameters, having an
ability to pass a custom configuration file helps a testing of more advanced
real world setups, like authn/authz.

Method `WithConfigFile` is provided by other modules like postgresql. The
difference is this version accepts `io.Reader`, so configuration file does not
need to be a real file stored on a host. This simplifies a testing setup.

## Related issues

N/A

## How to test this PR

This PR adds a simple test starting the server with a custom authorization
token and shows a client without a correct one can't connect to the server,
while the one with a right one can.

As a final validation the logs of the container are checked for `Using
configuration file: /etc/nats.conf` text.


modules/nats/options.go
```diff
@@ -49,3 +52,11 @@ func WithArgument(flag string, value string) CmdOption {
 		o.CmdArgs[flag] = value
 	}
 }
+
+// WithConfigFile pass io.Reader to the NATS container as /etc/nats.conf
+// Changes of a connectivity (listen address, or ports) may break a testcontainer
+func WithConfigFile(config io.Reader) CmdOption {
```
stevenh, line 58: suggestion: use a [ContainerCustomiser](https://pkg.go.dev/github.com/testcontainers/testcontainers-go#ContainerCustomizer) to avoid a two step process.

stevenh (changes requested):
Thanks for the PR, adding a config option is a good addition.

We're looking to use ContainerCustomizers where possible moving forward, looks like this would work for this case, so if we could make that switch it should simplify.

modules/nats/nats.go
```diff
@@ -58,6 +58,19 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
...
+		genericContainerReq.Cmd = append(genericContainerReq.Cmd, "-config", "/etc/nats.conf")
+		genericContainerReq.Files = append(
+			genericContainerReq.Files,
+			testcontainers.ContainerFile{
+				Reader:            settings.ConfigFile,
+				ContainerFilePath: "/etc/nats.conf",
+				FileMode:          0o644,
+			},
+		)
+	}
```
mdelapenya, line 72: As @stevenh mentioned, this block is the one that should go into the functional option.

Other than that, thanks for adding this, looks promising! 🙏

modules/nats/options.go
vyskocilm: https://github.com/testcontainers/testcontainers-go/pull/2905/commits/5da61731f5dfb3f77ab224f22ba36e3149307042

modules/nats/nats.go
vyskocilm: https://github.com/testcontainers/testcontainers-go/pull/2905/commits/5da61731f5dfb3f77ab224f22ba36e3149307042

modules/nats/options.go
```diff
@@ -16,12 +17,21 @@ func defaultOptions() options {
...
+var (
+	_ testcontainers.ContainerCustomizer = (*CmdOption)(nil)
+	_ testcontainers.ContainerCustomizer = (*ConfigFile)(nil)
+)
 
 // CmdOption is an option for the NATS container.
 type CmdOption func(opts *options)
 
+// ConfigFile optionally pass a configuration file into NATS container.
+type ConfigFile struct {
```
mdelapenya, line 31: I'm not sure if we need an extra type but instead leverage the `testcontainers.CustomiseRequestOption`. As an example, please take the postgres module: https://github.com/testcontainers/testcontainers-go/blob/main/modules/postgres/postgres.go#L66-L79, it will make the code simpler

Also, could you add the new option to the docs, at `docs/modules/nats.md`, right after the cmd-arguments entry 🙏 ? It will land into the [docs site for NATS](https://golang.testcontainers.org/modules/nats/#cmd-arguments)

vyskocilm: Rewritten code according postgres module: https://github.com/testcontainers/testcontainers-go/pull/2905/commits/a3636bbb3b0946ea20f2270d704a419faf20b7cb

Updated docs: https://github.com/testcontainers/testcontainers-go/pull/2905/commits/9a9225f150a516d4e49da0703bd7831ce19ae66a

Not sure about a proper since annotation, so I've omitted it from PR.

mdelapenya (approved):
LGTM, although I added two wording suggestions in the docs

modules/nats/options.go
mdelapenya: Oh you're right about the "since". Let me add a suggestion now

docs/modules/nats.md
```diff
@@ -75,6 +75,13 @@ These arguments are passed to the NATS server when it starts, as part of the com
 [Passing arguments](../../modules/nats/examples_test.go) inside_block:withArguments
 <!--/codeinclude-->
 
+#### Custom configuration file
+
```
mdelapenya, line 79:
```suggestion

- Not available until the next release of testcontainers-go <a href="https://github.com/testcontainers/testcontainers-go"><span class="tc-version">:material-tag: main</span></a>

```

modules/nats/nats_test.go
```diff
@@ -57,3 +59,52 @@ func TestNATS(t *testing.T) {
...
+listen: 0.0.0.0:4222
+authorization {
+    token: "s3cr3t"
+}
+`
+	ctx := context.Background()
+
+	//  createNATSContainer {
+	ctr, err := tcnats.Run(ctx, "nats:2.9", tcnats.WithConfigFile(strings.NewReader(natsConf)))
+	//  }
```
mdelapenya, line 74: Ups! sorry but last minute change: we use these markers for embedding code in the docs site. Unfortunately this is already in use, so I'd advocate for 1) removing it or 2) renaming and adding a code snippet in the module docs.

vyskocilm: I though those are folding markers for $EDITOR :-)

Replaced by regular comments: https://github.com/testcontainers/testcontainers-go/pull/2905/commits/554f949ca217ceb59479e4dca08a1c7b17cde7f1

mdelapenya (dismissed):
LGTM, thanks!

docs/modules/nats.md
```diff
@@ -75,6 +75,15 @@ These arguments are passed to the NATS server when it starts, as part of the com
...
+- Not available until the next release of testcontainers-go <a href="https://github.com/testcontainers/testcontainers-go"><span class="tc-version">:material-tag: main</span></a>
+
+It's possible to pass a custom config file to NATS container using `nats.WithConfigFile(strings.NewReader(config))`. The `io.Reader` is passed as a `-config /etc/nats.conf` arguments to an entrypoint.
```
stevenh, line 82: suggestion: clarify that a file is written to the container with the contents of the reader and not just passed directly, as that's a little confusing currently

modules/nats/nats_test.go
```diff
@@ -57,3 +59,45 @@ func TestNATS(t *testing.T) {
...
+	require.NoError(t, err)
+
+	uri, err := ctr.ConnectionString(ctx)
+	require.NoError(t, err)
+
+	// connect without a correct token must fail
+	mallory, err := nats.Connect(uri, nats.Name("Mallory"), nats.Token("secret"))
+	require.Error(t, err)
+	require.ErrorIs(t, err, nats.ErrAuthorization)
+	t.Cleanup(mallory.Close)
```
stevenh, line 83: suggestion: user EqualError to avoid the double test and make the Cleanup call first before it's a null check as currently it t will never be called

modules/nats/nats_test.go
```diff
@@ -57,3 +59,45 @@ func TestNATS(t *testing.T) {
...
+	// connect with a correct token must succeed
+	nc, err := nats.Connect(uri, nats.Name("API Token Test"), nats.Token("s3cr3t"))
+	require.NoError(t, err)
+	t.Cleanup(nc.Close)
+
+	// validate /etc/nats.conf mentioned in logs
+	const expected = "Using configuration file: /etc/nats.conf"
+	logs, err := ctr.Logs(ctx)
+	require.NoError(t, err)
+	sc := bufio.NewScanner(logs)
```
stevenh, line 94: A
suggestion: this looks like it will deadlock if the string is not found, add a timeout to avoid that.

modules/nats/options.go
```diff
@@ -49,3 +50,22 @@ func WithArgument(flag string, value string) CmdOption {
 		o.CmdArgs[flag] = value
 	}
 }
+
+// WithConfigFile pass io.Reader to the NATS container as /etc/nats.conf
```
stevenh, line 54: suggestion: as above clarify that reader is used to create a file not passed as the arg directly as that's confusing to the reader

stevenh (changes requested):
Just a few tweaks to the language and test

modules/nats/nats_test.go
vyskocilm: Used a direct string message, the `nats.ErrAuthorization` returned string with lower cases `nats: authorization violation`

```go
require.EqualError(t, err, "nats: Authorization Violation")
```

ff17ba124cb6cf9c72bfd04206630fd4afb6fb8a

modules/nats/nats_test.go
vyskocilm: used `time.After` in ff17ba124cb6cf9c72bfd04206630fd4afb6fb8a

modules/nats/options.go
vyskocilm: fixed in ff17ba124cb6cf9c72bfd04206630fd4afb6fb8a

docs/modules/nats.md
vyskocilm: fixed in ff17ba124cb6cf9c72bfd04206630fd4afb6fb8a

mdelapenya (approved):
LGTM, thanks!

PR #2914, opened 2024-12-10
chore(ci): add Github labels based on PR title

mdelapenya:
- **chore: add security as valid convention PR title**
- **chore: add labels based on PR title**

## What does this PR do?
It activates the autolabeler feature from the release-drafter Github Action, adding labels based on the PR title.

This is now possible thanks to the enforcement of conventional commits in the PR title.
 

## Why is it important?
It will simplify the burden of adding labels to PRs, making the creation of the release notes easier: as PRs will be automatically labelled, they will land into the right release notes section properly.

## Related issues

- Follows up #2911

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -29,22 +29,22 @@ categories:
 autolabeler:
   - label: 'breaking change'
     title:
-      - '/^[a-z]+(\(.+\))?!\:/i'
+      - '/^[a-z]+(\(.+\))?!\:/'
   - label: 'security'
     title:
-      - '/^security(\(.+\))?!?\:/i'
+      - '/^security(\(.+\))?!?\:/'
   - label: 'feature'
     title:
-      - '/^feat(\(.+\))?!?\:/i'
+      - '/^feat(\(.+\))?!?\:/'
   - label: 'bug'
     title:
-      - '/^(fix|bug)(\(.+\))?!?\:/i'
+      - '/^(fix|bug)(\(.+\))?!?\:/'
   - label: 'documentation'
     title:
-      - '/^docs(\(.+\))?!?\:/i'
+      - '/^docs(\(.+\))?!?\:/'
   - label: 'chore'
     title:
-      - '/^chore(\(.+\))?!?\:/i'
+      - '/^chore(\(.+\))?!?\:/'
   - label: 'dependencies'
     title:
-      - '/^deps(\(.+\))?!?\:/i'
+      - '/^deps(\(.+\))?!?\:/'
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -26,3 +26,25 @@ categories:
       - 'test flakiness'
   - title: 📦 Dependency updates
     label: 'dependencies'
+autolabeler:
+  - label: 'breaking change'
+    title:
+      - '/^break(\(.+\))?\:/i'
+  - label: 'security'
+    title:
+      - '/^security(\(.+\))?\:/i'
+  - label: 'feature'
+    title:
+      - '/^feat(\(.+\))?\:/i'
+  - label: 'bug'
+    title:
+      - '/^(fix|bug)(\(.+\))?\:/i'
+  - label: 'documentation'
+    title:
+      - '/^docs(\(.+\))?\:/i'
+  - label: 'chore'
+    title:
+      - '/^chore(\(.+\))?\:/i'
+  - label: 'dependencies'
+    title:
+      - '/^deps(\(.+\))?\:/i'
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -29,7 +29,7 @@ categories:
 autolabeler:
   - label: 'breaking change'
     title:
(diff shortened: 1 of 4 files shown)


.github/release-drafter.yml
```diff
@@ -26,3 +26,25 @@ categories:
       - 'test flakiness'
   - title: 📦 Dependency updates
     label: 'dependencies'
+autolabeler:
+  - label: 'breaking change'
+    title:
+      - '/^break(\(.+\))?\:/i'
```
stevenh, line 32: bug: The pattern for this should be looking for an [exclamation mark](https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with--to-draw-attention-to-breaking-change), I think you want:
```suggestion
      - '/^[a-z]+(\(.+\))?!\:/i'
```

.github/release-drafter.yml
```diff
@@ -26,3 +26,25 @@ categories:
...
+  - label: 'security'
+    title:
+      - '/^security(\(.+\))?\:/i'
```
stevenh, line 35: bug: because of the above all other patterns need to be updated, for example:
```suggestion
      - '/^security(\(.+\))?!?\:/i'
```

stevenh (changes requested):
Some issues with the patterns, to match conventional commit standard.

.github/workflows/conventions.yml
```diff
@@ -41,6 +41,7 @@ jobs:
             doesn't start with an uppercase character.
           types: |
             break
+            security
```
stevenh, line 44: suggestion: not done by this PR but break shouldn't be a type, thats done via exclamation mark, so we should remove this while here.

I also suggest we also add 'ci' (for continuous integration, like GitHub action changes), as a type as users aren't typically interested in those so useful to categorise them separately.

stevenh (changes requested):

.github/workflows/conventions.yml
mdelapenya: I usually add CI work under `chore(ci), chore(gha)`

.github/release-drafter.yml
mdelapenya: They also use `BREAKING CHANGE:` as convention, that's why I usually used `break` as our convention.

> BREAKING CHANGE: a commit that has a footer BREAKING CHANGE:, or appends a ! after the type/scope, introduces a breaking API change (correlating with [MAJOR](http://semver.org/#summary) in Semantic Versioning). A BREAKING CHANGE can be part of commits of any type.

.github/workflows/conventions.yml
stevenh: We can do that, will need documenting as `ci` could be what others, like me might naturally use but that should get caught by the validation so as long as we're consistent all good.

mdelapenya: Yeah, we need documenting that chore would be the umbrella category for build, ci, perf, test, etc, as we don't have separate sections in the changelog

.github/release-drafter.yml
stevenh: Yes but thats a footer, not a title item and its there to provide the context for the break which is typically used to describe the way to update. I would suggest sticking to the standard for `!` and ensuring that we also detail with the footer.

I'm not a fan of using a dedicated type, it's not compliant with the spec and as you loose the additional content so you don't know if it was a fix or feature etc that resulted in the breaking change.

mdelapenya: You're totally right, I did not read the `footer` word in the spec 🤦 

Let's make it 100% compliant with the spec for breaking changes then. Will update the regex then

.github/workflows/conventions.yml
mdelapenya: Question: how would you handle security? It's not covered by the conventional commits spec, and I always wanted to push security improvements to the changelog

stevenh: I think its fine if you want a separate type for that.

stevenh (changes requested):
Sorry missed a few

docs/contributing.md
```diff
@@ -29,12 +29,19 @@ Please just be sure to:
 * follow the style, naming and structure conventions of the rest of the project.
 * make commits atomic and easy to merge.
 * use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) for the PR title. This will help us to understand the nature of the changes, and to generate the changelog after all the commits in the PR are squashed.
+    * Please use the `break` type for breaking changes, as these categories are considered as `breaking change` in the changelog.
```
stevenh, line 32: suggestion: document the exclamation mark and BREAKING CHANGE: footer instead of `break` type.

stevenh (changes requested):
Just the docs which still mention `break` type instead of exclamation mark / BREAKING CHANGE: footer.

docs/contributing.md
mdelapenya: Forgot this! Will change it tomorrow, good catch

stevenh (approved):

PR #2920, opened 2024-12-15
feat: additional advertised listners/broker addresses for kafka

ghthor:
## What does this PR do?

1. Add additional kafka listeners to support more networking configurations
1. Add additional kafka advertised listeners for the above

## Why is it important?

1. kafka likes to know the exact address where it can be reached
1. the current ad. listeners returns `localhost:<random port>`
1. `localhost` is not a valid address to kafka when running within another docker container

This change is intended to support more complicated testing setups where the test code is running itself is another docker container, thus needs kafka to advertise a connection string other than `localhost`.

## Related issues

!!! TODO check for related issues

## How to test this PR

!!! TODO


mdelapenya:
Thanks @ghthor for your work here, I saw you linked to #2894, which is providing support for dynamically adding listeners to kafka. IIUC this PR is adding fixed listeners to it. What are the implications of having one or the other worlds for the kafka module here? In other words: if we merge this one, do we need the other one?

@eddumelendez could you take a look at this PR too?

ghthor:
@mdelapenya I think that most of the usecases for dynamic listeners would be provided by this one. The case that wouldn't be covered by this PR is if you need listeners that are a different protocol other than `PLAINTEXT`, IE TLS or whatever else kafka supports.

There is no reason that adding more default fixed listeners could not co-exist with the feature to add dynamic listeners. I don't know that I've found a reason to use dynamic listeners yet, and they add many elements of complication, as to declare them you need to know much information upfront or configure more stuff to make them work (hostnames / ports) etc.

The aim of this PR is to avoid as much boilerplate as possible to achieve Container <> Container support. I gave the user 2 options to achieve this aim.

1. Configure the non-kafka containers with an extra hostname `host.docker.internal` and use `BrokersByHostDockerInternal`
1. Configure both kafka and non-kafka containers to run in a new docker network so container name/id name resolution works and use `BrokersByContainer{Name,Id}`

You do either of those and you get Container <> Container kafka working AND we can assert all this is working within the kafka module tests because we're not dealing with user configurable listeners.

ghthor:
Will push up a fix for the lint failures.

ghthor:
@mdelapenya can you authorize another run of the CI test suite?

mdelapenya:
@ghthor could you run `make lint` from the kafka directory? It will give you more context about the lint error.

docs/modules/kafka.md
```diff
@@ -82,3 +91,34 @@ The `Brokers(ctx)` method returns the Kafka brokers as a string slice, containin
 <!--codeinclude-->
 [Get Kafka brokers](../../modules/kafka/kafka_test.go) inside_block:getBrokers
 <!--/codeinclude-->
+
+#### BrokersByHostDockerInternal
+
+The `BrokersByHostDockerInternal(ctx)` method returns the Kafka brokers as a
+string slice, containing the hostname `host.docker.internal` and a random port
+defined by Kafka's public port (`19092/tcp`).
```
stevenh, line 99: suggestion: clarify what the 19092 port is, it's not clear on reading if its an example of random port or is that re-enforcing the standard port

modules/kafka/examples_test.go
```diff
@@ -19,7 +27,7 @@ func ExampleRun() {
 	)
 	defer func() {
 		if err := testcontainers.TerminateContainer(kafkaContainer); err != nil {
-			log.Printf("failed to terminate container: %s", err)
+			log.Fatalf("failed to terminate container: %s", err)
```
stevenh, line 30: bug: this breaks clean up, please revert.

modules/kafka/examples_test.go
```diff
@@ -41,3 +49,337 @@ func ExampleRun() {
...
+	)
+	if err != nil {
+		log.Printf("failed to start container: %s", err)
+		return
+	}
+
+	// Clean up the container after
+	defer func() {
+		if err := kafkaContainer.Terminate(ctx); err != nil {
+			log.Fatalf("failed to terminate container: %s", err)
```
stevenh, line 68: bug: this should be `Printf`

modules/kafka/examples_test.go
```diff
@@ -41,3 +49,337 @@ func ExampleRun() {
...
+		}
+	}()
+
+	{
```
stevenh, line 72: suggestion: no need for wrapping braces which make this harder to read, more below.

docs/modules/kafka.md
```diff
@@ -82,3 +91,34 @@ The `Brokers(ctx)` method returns the Kafka brokers as a string slice, containin
...
+The `BrokersByContainerName(ctx)` method returns the Kafka brokers as a string
+slice, addressed by the container's name(`Ex: charming_dijkstra:19093`). This
+method is useful when you need to run additional containers that need to connect
+to Kafka.
+
+To use this broker address you should run all the containers inside a docker
+network.
+
+<!--codeinclude-->
+[Start Kafka inside a docker network](../../modules/kafka/examples_test.go) inside_block:getBrokersByContainerName_Kafka
```
stevenh, line 119: bug: looks like wrong name here says by docker network and then links to by container.

modules/kafka/examples_test.go
```diff
@@ -41,3 +49,337 @@ func ExampleRun() {
...
+	if err != nil {
+		log.Printf("failed to start container: %s", err)
+		return
+	}
+	// }
+
+	// Clean up the container after
+	defer func() {
+		if err := kafkaContainer.Terminate(ctx); err != nil {
+			log.Fatalf("failed to terminate container: %s", err)
```
stevenh, line 187: bug: should be `Printf`

modules/kafka/examples_test.go
```diff
@@ -41,3 +49,337 @@ func ExampleRun() {
...
+	)
+	if err != nil {
+		log.Printf("failed to start container: %s", err)
+		return
+	}
+
+	// Clean up the container after
+	defer func() {
+		if err := kafkaContainer.Terminate(ctx); err != nil {
+			log.Fatalf("failed to terminate container: %s", err)
```
stevenh, line 299: bug: should be `Printff`

modules/kafka/kafka.go
```diff
@@ -15,20 +15,37 @@ import (
 	"github.com/testcontainers/testcontainers-go/wait"
 )
 
-const publicPort = nat.Port("9093/tcp")
+const (
```
stevenh, line 18: suggest: these are expanding the API we have to maintain, lets keep them private.

modules/kafka/kafka.go
```diff
@@ -15,20 +15,37 @@ import (
...
+	ControllerPort = 9094
+)
+
 const (
 	starterScript = "/usr/sbin/testcontainers_start.sh"
 
 	// starterScript {
-	starterScriptContent = `#!/bin/bash
+	starterScriptContent = `
+#!/bin/bash
```
stevenh, line 38: bug: scripts shouldn't have a leading newline.

modules/kafka/kafka.go
```diff
@@ -46,14 +63,34 @@ func RunContainer(ctx context.Context, opts ...testcontainers.ContainerCustomize
 
 // Run creates an instance of the Kafka container type
 func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustomizer) (*KafkaContainer, error) {
+	listeners := []string{
+		fmt.Sprintf("LOCALHOST://0.0.0.0:%d", PublicLocalhostPort.Int()),
+		fmt.Sprintf("HOST_DOCKER_INTERNAL://0.0.0.0:%d", PublicDockerHostPort.Int()),
+		fmt.Sprintf("CONTAINER_NAME://0.0.0.0:%d", NetworkInternalContainerNamePort),
+		fmt.Sprintf("CONTAINER_ID://0.0.0.0:%d", NetworkInternalContainerIdPort),
+		fmt.Sprintf("BROKER://0.0.0.0:%d", BrokerToBrokerPort),
+		fmt.Sprintf("CONTROLLER://0.0.0.0:%d", ControllerPort),
```
stevenh, line 72: suggestion: looks like these could all be simplified to strings concatenation with a type change.

modules/kafka/kafka.go
```diff
@@ -46,14 +63,34 @@ func RunContainer(ctx context.Context, opts ...testcontainers.ContainerCustomize
 
 // Run creates an instance of the Kafka container type
 func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustomizer) (*KafkaContainer, error) {
+	listeners := []string{
```
stevenh, line 66: suggestion: looks like these are only ever consumed as a comma separated string, so define them as that here do avoid multiple joins.

modules/kafka/kafka.go
```diff
@@ -165,20 +217,61 @@ func WithClusterID(clusterID string) testcontainers.CustomizeRequestOption {
...
 	}
 
 	return []string{fmt.Sprintf("%s:%d", host, port.Int())}, nil
 }
 
+// BrokersByHostDockerInternal retrieves broker connection string suitable when
+// running 2 containers in the default docker network
+//
+// Example Output: host.docker.internal:<random_port>
+func (kc *KafkaContainer) BrokersByHostDockerInternal(ctx context.Context) ([]string, error) {
```
stevenh, line 240: question: why is this returning a slice when its hard coded to a single entry? More below

modules/kafka/kafka.go
```diff
@@ -165,20 +217,61 @@ func WithClusterID(clusterID string) testcontainers.CustomizeRequestOption {
...
+	port, err := kc.MappedPort(ctx, PublicDockerHostPort)
+	if err != nil {
+		return nil, err
+	}
+
+	return []string{fmt.Sprintf("%s:%d", "host.docker.internal", port.Int())}, nil
```
stevenh, line 246: suggestion: simplify with string concat, more below
```suggestion
	return []string{"host.docker.internal:" + port.Port()}, nil
```

modules/kafka/kafka_test.go
```diff
@@ -1,26 +1,42 @@
...
-	kafkaContainer, err := kafka.Run(ctx, "confluentinc/confluent-local:7.5.0", kafka.WithClusterID("kraftCluster"))
+	kafkaContainer, err := kafka.Run(ctx, "confluentinc/confluent-local:7.5.0", kafka.WithClusterID("kraftCluster"), network.WithNetwork(nil, net))
 	testcontainers.CleanupContainer(t, kafkaContainer)
 	require.NoError(t, err)
 
+	// Clean up the container after the test is complete
+	t.Cleanup(func() {
+		require.NoError(t, kafkaContainer.Terminate(ctx), "failed to terminate container: %v", err)
+	})
+
```
stevenh, line 39: bug: this isn't needed as that's what `CleanContainer` does.

modules/kafka/kafka_test.go
```diff
@@ -1,26 +1,42 @@
...
 func TestKafka(t *testing.T) {
-	topic := "some-topic"
+	const (
+		topic = "some-topic"
+		value = "kafka-message-value"
+	)
 
 	ctx := context.Background()
+	net, err := network.New(ctx)
+	require.NoError(t, err)
```
stevenh, line 29: bug: missing network clean up

stevenh (changes requested):
Thanks for the update, I've done an initial pass on this, theres a few bugs, suggestions and questions.

modules/kafka/kafka.go
ghthor: I made them public so the `kafka_test` package had access to them. It is a minor duplication without so I don't disagree with you here about moving them back to private and duplicating

modules/kafka/kafka.go
ghthor: Ah, good point. You cool with leaving the raw string with the leading newline but then running `strings.TrimLeft(script)` ?

modules/kafka/kafka.go
ghthor: I did it this way because I feel it makes it much much easier to read and confirm that the correct pairings of `${LISTENER_NAME} <> Port` are taking place compared to a big log string with multiple `fmt %` characters in it.

modules/kafka/kafka.go
ghthor: I think with string concat the ease of reading and verifying the LISTENER:// <> Port pairings will be maintained; I'll see what that looks like.

modules/kafka/kafka.go
stevenh: I wouldn't if you want this to be more readable I'd be tempted to move it to a file and include with embed.

modules/kafka/kafka.go
ghthor: Was just following the previous API for `Brokers()`. I suppose there is a future where maybe we do setup a multinode cluster and having the same API to return the list of brokers might be a good idea so the public API doesn't change if/when that happens.

modules/kafka/kafka.go
ghthor: Got it, we like string concat. :saluting_face:

modules/kafka/kafka.go
stevenh: To clarify I would keep them defined like you have, as you say its easier to read and maintain, but wrap with the `strings.Join` here so you done run that twice.

modules/kafka/kafka_test.go
ghthor: Excellent :champagne:

modules/kafka/kafka.go
stevenh: yep that's where I was going:
```go
	listeners := strings.Join([]string{
		"BROKER://0.0.0.0:"+brokerToBrokerPort,
		"CONTROLLER://0.0.0.0:"+controllerPort,
...
	})
```

modules/kafka/kafka.go
stevenh: Yer that's the pain with the separate test package name, but better as you say to have the slight duplication to keep it clean.

docs/modules/kafka.md
ghthor: I think this is correct, It's a 2 part explanation, I was trying to hightlight that this works best (maybe ONLY works) if you start the Kafka testcontainer inside a non-default docker network. So I broke this into 2 codeincludes, one focusing on starting the Kafka test container in a non-default network and the second part of starting another container in that same network.

modules/kafka/kafka_test.go
```diff
@@ -54,38 +75,141 @@ func TestKafka(t *testing.T) {
...
+		brokers, err := kafkaContainer.BrokersByContainerName(ctx)
+		require.NoError(t, err)
+
+		kcat, err := runKcatContainer(ctx, brokers, nil, []string{net.Name})
+		testcontainers.CleanupContainer(t, kcat)
+		require.NoError(t, err)
+
+		assertKcatReadMsg(ctx, t, kcat)
+	})
 }
```
ghthor, line 121: @stevenh Much happier with this after your suggestion to not use that giant closure here :saluting_face:

modules/kafka/kafka_test.go
```diff
@@ -75,17 +144,53 @@ func TestKafka_invalidVersion(t *testing.T) {
...
 
 	bs := testcontainers.RequireContainerExec(ctx, t, container, []string{"cat", "/usr/sbin/testcontainers_start.sh"})
 
-	require.Containsf(t, bs, brokerURL, "expected advertised listeners to contain %s, got %s", brokerURL, bs)
+	assert := func(listener string) {
+		t.Helper()
+		require.Containsf(t, bs, listener, "expected advertised listeners to contain %s, got %s", listener, bs)
+	}
+
+	mustBrokers := func(fn func(context.Context) ([]string, error)) string {
```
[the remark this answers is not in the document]
ghthor: Felt like this one benefited from being a internal closure, it's short, and requires to many parameters when lifted up to a function.

modules/kafka/kafka_test.go
```diff
@@ -54,38 +75,150 @@ func TestKafka(t *testing.T) {
...
+		kcat, err := runKcatContainer(ctx, brokers, nil, []string{net.Name})
+		testcontainers.CleanupContainer(t, kcat)
+		require.NoError(t, err)
+
+		l, err := kcat.Logs(ctx)
+		require.NoError(t, err)
+		defer l.Close()
+
+		assertKcatReadMsg(t, l)
+	})
```
ghthor, line 132: @stevenh Much happier with this after your suggestion to not use that giant closure here :saluting_face:

eddumelendez (commented):
The proposed idea is good; however, I lean towards prioritizing user flexibility. Introducing these changes as defaults could potentially break existing users who have already defined their own listeners. 

Additionally, there is a specific use case that this approach does not address: scenarios where tools like Toxiproxy are used in front of Kafka to simulate network conditions and test resiliency. This highlights the importance of maintaining configurability to accommodate diverse testing and deployment needs.

modules/kafka/kafka.go
```diff
@@ -141,12 +177,27 @@ func copyStarterScript(ctx context.Context, c testcontainers.Container) error {
...
-	scriptContent := fmt.Sprintf(starterScriptContent, host, port.Int(), hostname)
+	portDh, err := c.MappedPort(ctx, publicDockerHostPort)
+	if err != nil {
+		return fmt.Errorf("mapped port: %w", err)
+	}
+
+	// advertisedListeners {
+	advertisedListeners := strings.Join([]string{
+		fmt.Sprintf("LOCALHOST://%s:%d", host, portLh.Int()),
+		fmt.Sprintf("HOST_DOCKER_INTERNAL://%s:%d", "host.docker.internal", portDh.Int()),
```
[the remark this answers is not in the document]
ghthor: You can't use a broker address of `localhost` from another docker container, that's the entire point of this proposed addition.

https://github.com/testcontainers/testcontainers-go/pull/2920#issuecomment-2544753599

ghthor:
> The proposed idea is good; however, I lean towards prioritizing user flexibility. Introducing these changes as defaults could potentially break existing users who have already defined their own listeners.

Do you have an example of how to do that, as this change is proposed because that's not really possible to do.

> Additionally, there is a specific use case that this approach does not address: scenarios where tools like Toxiproxy are used in front of Kafka to simulate network conditions and test resiliency.

Do you have an example of what this looks like. I don't understand how adding a few more listeners and advertised listeners would break this.

@eddumelendez

ghthor:
I could see what it looks like to put all this behind functional options. It would differ from the other PR in that we're not going to allow you to configure any of them, just whether they are enabled or not.

This doesn't have a benefit that we could assert some things and produce errors if the API isn't used correctly.

Example being, if you want to use the `BrokersByContainer{Name,Id}` API you need to run the testcontainer in a docker network, not the default one. So if you enable that and don't pass in a network then we can produce a helpful error.

Using the `BrokersByHostDockerInternal` couldn't be verifed in the same way, as that requires configuring the OTHER container, But the other 2 methods could verify that; and actually we should probably produce an error for those 2 now cause we know whether kafka was started in the default network or not.

eddumelendez:
I have reviewed various use cases and examined bugs reported in other language implementations. Below is a summary of the key requirements we should address in the API:

1. Avoid Using host.docker.internal as a default
Using `host.docker.internal` as a default can create confusion with the equivalent feature provided by Docker Desktop ([reference](https://docs.docker.com/desktop/features/networking/#i-want-to-connect-from-a-container-to-a-service-on-the-host)). Moreover, it required additional configuration in the companion container, making it less convenient for users.

2. Flexible listener Configuration
Listeners can be configured using networking to be accessed via container name or container ID, as demonstrated in this PR. However, some users may prefer to utilize the container’s IP address. The API should accommodate these varied preferences, providing users with flexibility in how they configure and use it.

3. Proxy Configuration for Kafka Broker
As previously mentioned, a proxy can be configured on top of the Kafka broker. For this to function effectively, the proxy will need to register the network alias or the container IP to facilitate communication between containers.

PR #2922, opened 2024-12-17
chore: provide an internal tools module

mmorel-35:
## What does this PR do?

Provides an internal tools module that isolates tools in a dedicated folder (.tools) at the root of the folder instead of using local GOBIN

## Why is it important?

This helps sharing the same versions for tools between developers. It also allow to use dependencies managers to provide updates on those tools.

This is inspired by the work done on [opentelemetry-collector](https://github.com/open-telemetry/opentelemetry-collector/blob/main/Makefile.Common)


stevenh:
Is this impacted by the impending [tools addition](https://github.com/golang/go/issues/48429) to golang as [detailed here](https://cirw.in/blog/go-tools)?

PR #2923, opened 2024-12-17
feat(ollama): support calling the Ollama local process

stevenh:
Refactor local process handling for Ollama using a container implementation avoiding the wrapping methods.

This defaults to running the binary with an ephemeral port to avoid conflict.
This behaviour can be overridden my setting OLLAMA_HOST either in the parent  environment or in the values passed via WithUseLocal.

Improve API compatibility with:

- Multiplexed output streams
- State reporting
- Exec option processing
- WaitingFor customisation

Fix Container implementation:

- Port management
- Running checks
- Terminate processing
- Endpoint argument definition
- Add missing methods
- Consistent environment handling


modules/ollama/local.go
```diff
@@ -0,0 +1,724 @@
...
+	}
+
+	if err != nil {
+		return container, fmt.Errorf("start ollama: %w", err)
+	}
+
+	return container, nil
+}
+
+// validateRequest checks that req is valid for the local Ollama binary.
```
mdelapenya, line 123: suggestion: Can we include info about what represents an invalid request?

modules/ollama/local.go
```diff
@@ -0,0 +1,724 @@
...
+		c.mtx.Lock()
+		defer c.mtx.Unlock()
+
+		return c.exitErr
+	}
+}
+
+// Terminate implements testcontainers.Container interface for the local Ollama binary.
+// It stops the local Ollama process, removing the log file.
+func (c *localProcess) Terminate(ctx context.Context) error {
```
mdelapenya, line 525: todo: conflict with #2926, so depending on which one is merged first, we need to update it.

mdelapenya (dismissed):
This LGTM, waiting for you to mark is as ready. Great job with the process execution handling 🏆

modules/ollama/local.go
```diff
@@ -0,0 +1,724 @@
...
+		if _, ok := w.(*wait.HostPortStrategy); ok {
+			return wait.VisitRemove
+		}
+
+		return nil
+	}); err != nil {
+		return fmt.Errorf("walk strategies: %w", err)
+	}
+
+	logStrategy := wait.ForLog(localLogRegex).Submatch(c.extractLogDetails)
```
mdelapenya, line 686: Ok, now I see why you needed to add here the changes in the wait strategy. I was about to ask for a different PR for that, but it's fine for me to merge it here.

wait/log.go
```diff
@@ -24,6 +41,18 @@ type LogStrategy struct {
...
+	check func([]byte) error
+
+	// submatchCallback is a callback that will be called with the sub matches of the regexp.
+	submatchCallback func(pattern string, matches [][][]byte) error
+
+	// re is the optional compiled regexp.
+	re *regexp.Regexp
+
+	// log byte slice version of [LogStrategy.Log] used for count checks.
+	log []byte
```
mdelapenya, line 55: bug: we need to document these new fields, and relative methods, in the docs site.

modules/ollama/local.go
stevenh: My intent is to split that into a separate PR, if this looks good.

modules/ollama/local.go
stevenh: Yer just replied to that PR, I don't think it works in its current form as it breaks the interface separation.

wait/log.go
stevenh: Just the new method, internal fields shouldn't be documented.

modules/ollama/local.go
stevenh: Give me a thumbs up @mdelapenya if you like the approach and I'll create the separate wait PR with docs, which we can merge first.

stevenh: Went ahead and did the split you can find it [here](https://github.com/testcontainers/testcontainers-go/pull/2929), just waiting on tests to pass to take it out of draft.

wait/log.go
mdelapenya: Sorry, I meant the new exposed methods

modules/ollama/local.go
mdelapenya: We merged #2926 so I think we need to update this signature to match the interface

mdelapenya: Added it here: [9be4b71](https://github.com/testcontainers/testcontainers-go/pull/2923/commits/9be4b71acfa4c4b12ed46eaefb31b1f6776250eb)

I did not transfer the options from Terminate to the Stop call (stop duration), do you think we need to extract it from the options and override the default?

modules/ollama/local.go
```diff
@@ -548,36 +548,44 @@ func (c *localProcess) Stop(ctx context.Context, d *time.Duration) error {
 
 // Terminate implements testcontainers.Container interface for the local Ollama binary.
 // It stops the local Ollama process, removing the log file.
-func (c *localProcess) Terminate(ctx context.Context) error {
+func (c *localProcess) Terminate(ctx context.Context, opts ...testcontainers.TerminateOption) error {
+	options := testcontainers.NewTerminateOptions(ctx, opts...)
 	// First try to stop gracefully.
-	if err := c.Stop(ctx, &defaultStopTimeout); !c.isCleanupSafe(err) {
+	if err := c.Stop(options.Context(), options.StopTimeout()); !c.isCleanupSafe(err) {
```
mdelapenya, line 554: Neat! I forgot we had the getters for each terminate option. Cool!

modules/ollama/local.go
```diff
@@ -0,0 +1,758 @@
...
+	localLogVar     = "OLLAMA_LOGFILE"
+)
+
+var (
+	// Ensure localProcess implements the required interfaces.
+	_ testcontainers.Container           = (*localProcess)(nil)
+	_ testcontainers.ContainerCustomizer = (*localProcess)(nil)
+
+	// defaultStopTimeout is the default timeout for stopping the local Ollama process.
+	defaultStopTimeout = time.Second * 5
```
mdelapenya, line 47: bug: we don't need this, as the default value from the terminate options is applied thanks to the NewTerminateOptions call. The default is 10s then

stevenh: Nice catch.

mdelapenya (approved):
LGTM, great work with the refactor of the process Start/Wait, thanks for this!

Issue #2924, opened 2024-12-18
[Feature]: build log writer for container request


zenkovev:
### Problem

Now, there are two options for displaying logs of docker image build from Dockerfile: completely disable the output, however, it may contain useful information in case of problems, output to stderr, but the log may be large and clog up the output.

### Solution

To add the ability to specify an io.Writer for build log. I will attach a PR with a possible implementation below.

### Benefit

To solve this problem, add a flexible configuration for working with build logs.

### Alternatives

-

### Would you like to help contributing this feature?

Yes

zenkovev:
PR: https://github.com/testcontainers/testcontainers-go/pull/2925

PR #2925, opened 2024-12-18
feat!: build log writer for container request

zenkovev:
## What does this PR do?

It adds the ability to configure the output of logs of docker image build from Dockerfile.

## Why is it important?

Now, there are two options for displaying logs of docker image build from Dockerfile: completely disable the output, however, it may contain useful information in case of problems, output to stderr, but the log may be large and clog up the output.

## Related issues

- Closes https://github.com/testcontainers/testcontainers-go/issues/2924

## How to test this PR

I have added a test for changes.

### DIFF
--- a/container.go
+++ b/container.go
@@ -77,7 +77,8 @@ type ImageBuildInfo interface {
 	GetDockerfile() string                          // the relative path to the Dockerfile, including the file itself
 	GetRepo() string                                // get repo label for image
 	GetTag() string                                 // get tag label for image
-	ShouldPrintBuildLog() bool                      // allow build log to be printed to stdout
+	ShouldPrintBuildLog() bool                      // allow build log to be printed
+	GetBuildLogWriter() io.Writer                   // for output of build log if ShouldPrintBuildLog is true
 	ShouldBuildImage() bool                         // return true if the image needs to be built
 	GetBuildArgs() map[string]*string               // return the environment args used to build the from Dockerfile
 	GetAuthConfigs() map[string]registry.AuthConfig // Deprecated. Testcontainers will detect registry credentials automatically. Return the auth configs to be able to pull from an authenticated docker registry
@@ -93,6 +94,7 @@ type FromDockerfile struct {
 	Tag            string                         // the tag label for image, defaults to UUID
 	BuildArgs      map[string]*string             // enable user to pass build args to docker daemon
 	PrintBuildLog  bool                           // enable user to print build log
+	BuildLogWriter io.Writer                      // for output of build log if PrintBuildLog is true, defaults to os.Stderr
 	AuthConfigs    map[string]registry.AuthConfig // Deprecated. Testcontainers will detect registry credentials automatically. Enable auth configs to be able to pull from an authenticated docker registry
 	// KeepImage describes whether DockerContainer.Terminate should not delete the
 	// container image. Useful for images that are built from a Dockerfile and take a
@@ -412,6 +414,14 @@ func (c *ContainerRequest) ShouldKeepBuiltImage() bool {
 
 func (c *ContainerRequest) ShouldPrintBuildLog() bool {
 	return c.FromDockerfile.PrintBuildLog
+}
+
+func (c *ContainerRequest) GetBuildLogWriter() io.Writer {
+	if c.FromDockerfile.BuildLogWriter == nil {
+		c.FromDockerfile.BuildLogWriter = os.Stderr
+	}
+
+	return c.FromDockerfile.BuildLogWriter
 }
 
 // BuildOptions returns the image build options when building a Docker image from a Dockerfile.
--- a/container.go
+++ b/container.go
@@ -77,8 +77,7 @@ type ImageBuildInfo interface {
 	GetDockerfile() string                          // the relative path to the Dockerfile, including the file itself
 	GetRepo() string                                // get repo label for image
 	GetTag() string                                 // get tag label for image
-	ShouldPrintBuildLog() bool                      // allow build log to be printed
-	GetBuildLogWriter() io.Writer                   // for output of build log if ShouldPrintBuildLog is true
+	BuildLogWriter() io.Writer                      // for output of build log, use io.Discard to disable the output
 	ShouldBuildImage() bool                         // return true if the image needs to be built
 	GetBuildArgs() map[string]*string               // return the environment args used to build the from Dockerfile
 	GetAuthConfigs() map[string]registry.AuthConfig // Deprecated. Testcontainers will detect registry credentials automatically. Return the auth configs to be able to pull from an authenticated docker registry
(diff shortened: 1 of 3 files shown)


container.go
```diff
@@ -77,7 +77,8 @@ type ImageBuildInfo interface {
 	GetDockerfile() string                          // the relative path to the Dockerfile, including the file itself
 	GetRepo() string                                // get repo label for image
 	GetTag() string                                 // get tag label for image
-	ShouldPrintBuildLog() bool                      // allow build log to be printed to stdout
+	ShouldPrintBuildLog() bool                      // allow build log to be printed
+	GetBuildLogWriter() io.Writer                   // for output of build log if ShouldPrintBuildLog is true
```
mdelapenya, line 81: Given this change is a breaking change in the sense that anybody implementing this interface will be broken (very improbably, but true), I thikn we could merge both methods into one: `LogWriter() io.Writer`, so consumers can choose to print to stdout, nil or a custom one.

Wdyt?

zenkovev: Yes, I think it's really better to do this for the interface. I will make a commit with a fix.

container.go
```diff
@@ -410,8 +411,16 @@ func (c *ContainerRequest) ShouldKeepBuiltImage() bool {
 	return c.FromDockerfile.KeepImage
 }
 
-func (c *ContainerRequest) ShouldPrintBuildLog() bool {
-	return c.FromDockerfile.PrintBuildLog
+func (c *ContainerRequest) BuildLogWriter() io.Writer {
```
stevenh, line 414: suggestion: lets a doc string.

container.go
```diff
@@ -410,8 +411,16 @@ func (c *ContainerRequest) ShouldKeepBuiltImage() bool {
...
+	if c.FromDockerfile.BuildLogWriter == nil {
+		c.FromDockerfile.BuildLogWriter = os.Stderr
+	}
```
stevenh, line 417: question: should we try to maintain current behaviour so default to discard?

container.go
```diff
@@ -410,8 +411,16 @@ func (c *ContainerRequest) ShouldKeepBuiltImage() bool {
...
+
+	if c.FromDockerfile.PrintBuildLog {
```
stevenh, line 419: question: having two variable to control feel confusing, thoughts?

docker_test.go
```diff
@@ -705,6 +705,40 @@ func Test_BuildContainerFromDockerfileWithBuildLog(t *testing.T) {
...
+		ContainerRequest: req,
+		Started:          true,
+	}
+
+	c, err := GenericContainer(ctx, genContainerReq)
+	CleanupContainer(t, c)
+	require.NoError(t, err)
+
+	out, err := io.ReadAll(&buffer)
+	require.NoError(t, err)
```
stevenh, line 735: suggestion: no need for this on a buffer you can get the string from `Bytes()` or `String()` depending on the output format needed.

docker_test.go
```diff
@@ -705,6 +705,40 @@ func Test_BuildContainerFromDockerfileWithBuildLog(t *testing.T) {
...
+
+	temp := strings.Split(string(out), "\n")
+	require.NotEmpty(t, temp)
+	assert.Regexpf(t, `^Step\s*1/\d+\s*:\s*FROM alpine$`, temp[0], "Expected stdout first line to be %s. Got '%s'.", "Step 1/* : FROM alpine", temp[0])
```
stevenh, line 739: suggestion: use require instead of assert.

container.go
```diff
@@ -77,7 +77,7 @@ type ImageBuildInfo interface {
 	GetDockerfile() string                          // the relative path to the Dockerfile, including the file itself
 	GetRepo() string                                // get repo label for image
 	GetTag() string                                 // get tag label for image
-	ShouldPrintBuildLog() bool                      // allow build log to be printed to stdout
+	BuildLogWriter() io.Writer                      // for output of build log, use io.Discard to disable the output
```
stevenh, line 80: bug: as this is breaking change it will require ! and a BREAKING_CHANGE note, see: https://www.conventionalcommits.org/en/v1.0.0/

container.go
```diff
@@ -93,6 +93,7 @@ type FromDockerfile struct {
 	Tag            string                         // the tag label for image, defaults to UUID
 	BuildArgs      map[string]*string             // enable user to pass build args to docker daemon
 	PrintBuildLog  bool                           // enable user to print build log
+	BuildLogWriter io.Writer                      // for output of build log if PrintBuildLog is true, defaults to os.Stderr
```
stevenh, line 96: question: should the default be discard as it can become quite spammy for large builds?

container.go
```diff
@@ -410,8 +411,16 @@ func (c *ContainerRequest) ShouldKeepBuiltImage() bool {
...
+func (c *ContainerRequest) BuildLogWriter() io.Writer {
+	if c.FromDockerfile.BuildLogWriter == nil {
+		c.FromDockerfile.BuildLogWriter = os.Stderr
+	}
+
+	if c.FromDockerfile.PrintBuildLog {
+		return c.FromDockerfile.BuildLogWriter
+	} else {
+		return io.Discard
+	}
```
stevenh, line 423: suggestion: else isn't needed due to return.
```suggestion
	}

	return io.Discard
```

stevenh (changes requested):
Thanks for PR, have done a first pass with some questions and suggestions.

container.go
zenkovev: OK, I will add it

container.go
zenkovev: At the moment, the current behavior maintains because:
- PrintBuildLog is true and BuildLogWriter is default -> os.Stderr
- PrintBuildLog is false and BuildLogWriter is default -> io.Discard

However, it is possible to maintain compatibility in another way:
```go
if c.FromDockerfile.BuildLogWriter != nil {
	return c.FromDockerfile.BuildLogWriter
}
if c.FromDockerfile.PrintBuildLog {
	c.FromDockerfile.BuildLogWriter = os.Stderr
} else {
	c.FromDockerfile.BuildLogWriter = io.Discard
}
return c.FromDockerfile.BuildLogWriter
```

This way might be better because we can mark `PrintBuildLog` as deprecated field.

container.go
zenkovev: I answered in the previous question

docker_test.go
zenkovev: That's fair, I'll fix it

docker_test.go
zenkovev: This behavior is taken from a previous similar test, if correct it, probably need to correct both. It may be worth leaving it in its current form, because the output format is not the most stable thing

container.go
zenkovev: Fixed the name of the PR

container.go
zenkovev: Refers to the discussion about default as discard above

container.go
zenkovev: Yes, it can be fixed

zenkovev:
Thanks for the comments, I answered them, if there are no objections in this variant, I will make a new commit

docker_test.go
stevenh: Yer we're slowly moving to all require, no need to fix he other instances just don't want to introduce new uses of assert.

stevenh:
> Thanks for the comments, I answered them, if there are no objections in this variant, I will make a new commit

Go for it and will re-review once the changes are in.

docker_test.go
```diff
@@ -705,6 +705,37 @@ func Test_BuildContainerFromDockerfileWithBuildLog(t *testing.T) {
...
+		Started:          true,
+	}
+
+	c, err := GenericContainer(ctx, genContainerReq)
+	CleanupContainer(t, c)
+	require.NoError(t, err)
+
+	out := buffer.String()
+	temp := strings.Split(out, "\n")
+	require.NotEmpty(t, temp)
```
stevenh, line 735: nit: not required it would fail the regexp

stevenh (approved):
Looks good just one nit if you want to address, otherwise just let me know and we can get this closed out.

mdelapenya (approved):
LGTM, thanks both for your work during the review, great OSS colaboration 🙇

zenkovev:
Thanks for the review and quick feedback

Issue #2917, opened 2024-12-12
[Bug]: response from daemon: removal of container x is already in progress on calling Terminate()


moogacs:
### Testcontainers version

0.34.0

### Using the latest Testcontainers version?

Yes

### Host OS

linux 

### Host arch

any

### Go version

1.22

### Docker version

```shell
26.0.0 (any)
```

### Docker info

```shell
any
```

### What happened?

when using `Terminate()`  function see [here](https://github.com/testcontainers/testcontainers-go/blob/main/container.go#L53) , which i assume implemented [here](https://github.com/testcontainers/testcontainers-go/blob/main/docker.go#L306) because i am using norma docker container 

the container get terminated but still reports an error 
```
Error response from daemon: removal of container {xxxx} is already in progress
```

### Relevant log output

```shell

```

stevenh:
You can use [testcontainers.CleanupContainer](https://pkg.go.dev/github.com/testcontainers/testcontainers-go#CleanupContainer) to avoid that type cleanup race.

PR #2926, opened 2024-12-19
feat(termination)!: make container termination timeout configurable

moogacs:
## What does this PR do?

- extend `Terminate()` method with optional configuration for timeout configuration. 
- moved the `terminationOptions` under container to give the ability to make timeout, etc  configurable via a functional option.
- moved cleaning up under `Terminate()` function.

## Why is it important?

- `Terminate()` became broken with latest release version with latest release `0.34.0` `Terminate()` require handle for errors `"is already in progress"`
- this is CleanupContainer method signature and to use it in down streams that will require  many refactors because it expects `testing.TB`  
```
func CleanupContainer(tb testing.TB, ctr Container, options ...TerminateOption) {
```

so  with this change it will give the ability to configure the timeout and move the cleanup logic under termination 

## Related issues

- Closes https://github.com/testcontainers/testcontainers-go/issues/2917

a unit test added and shall be tested with all the existed pipelines


container.go
```diff
@@ -50,7 +50,7 @@ type Container interface {
 	Stop(context.Context, *time.Duration) error                                    // stop the container
 
 	// Terminate stops and removes the container and its image if it was built and not flagged as kept.
-	Terminate(ctx context.Context) error
+	Terminate(ctx context.Context, opts ...TerminateOption) error
```
mdelapenya, line 53: As this is breaking change, the PR title must be updated to `feat(termination)!: ...`

See https://golang.testcontainers.org/contributing/#submitting-your-changes

moogacs: thank you sure, gonna update it. i didn't think of it as breaking change, given the variadic function will make `opts ` optional

mdelapenya: Indeed, it could slip into a breaking change in two cases: somebody implementing their own container, i.e. modules, as you already spotted in the etcd module, and if somebody has the Terminate function assigned to a variable (see https://go.dev/play/p/5LsWTXgv_iU). Very rare thought

moogacs: gotcha, good point 👍🏼

docker.go
```diff
@@ -296,19 +297,33 @@ func (c *DockerContainer) Stop(ctx context.Context, timeout *time.Duration) erro
...
-func (c *DockerContainer) Terminate(ctx context.Context) error {
-	// ContainerRemove hardcodes stop timeout to 3 seconds which is too short
-	// to ensure that child containers are stopped so we manually call stop.
-	// TODO: make this configurable via a functional option.
-	timeout := 10 * time.Second
-	err := c.Stop(ctx, &timeout)
+//
+// Default: timeout is 10 seconds.
+func (c *DockerContainer) Terminate(ctx context.Context, opts ...TerminateOption) error {
+	if len(opts) == 0 {
```
stevenh, line 317: bug: this is fragile if other options are added.

stevenh (changes requested):
Thanks for PR, seems like a good idea, unfortunately this approach doesn't work with an interface design, see comments for details.

cleanup.go
```diff
@@ -54,41 +53,12 @@ func TerminateContainer(container Container, options ...TerminateOption) error {
...
-		ctx: context.Background(),
-	}
-
-	for _, opt := range options {
-		opt(c)
-	}
-
-	// TODO: Add a timeout when terminate supports it.
-	err := container.Terminate(c.ctx)
+	err := container.Terminate(context.Background(), options...)
```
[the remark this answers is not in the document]
moogacs: you mean like convert Terminate to accept `opts` only ?

cleanup.go
```diff
@@ -16,21 +15,21 @@ type terminateOptions struct {
 }
 
 // TerminateOption is a type that represents an option for terminating a container.
-type TerminateOption func(*terminateOptions)
+type TerminateOption func(*DockerContainer)
```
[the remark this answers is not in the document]
moogacs: yeah, i had similar feeling, but wasn't sure if shall i introduce a separate time, when i was opening this PR i tried to make as minimal breaking change but seems it has to be. open for suggestions 

i like the idea of `maintain the dedicated type but export it and the fields`

cleanup.go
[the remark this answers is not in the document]
stevenh: Ignore me I was confusing clean up and terminate, the former not having a context param.

cleanup.go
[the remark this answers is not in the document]
stevenh: I initially thought the dedicated type would work, but that wont ensure that all Container implementations support the options, which I'm not a fan of as a consumer would be left wondering why their options didn't work.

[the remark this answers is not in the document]
moogacs: the other thought i had before i come to this was introducing it in the `ContainerRequest`  and add getter in the `container` interface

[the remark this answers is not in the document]
stevenh: I thought about setters too, but I think that would make it much harder to discover for users.

I can't think of any perfect solution ATM, something mull over.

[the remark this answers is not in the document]
moogacs: i have made a change and revert it to make sure it uses `terminateOptions`  instead of the dependency on DockerContainer type. 

also added more test to show case the usage and the changes of configuration 

defaults of volumes is nil, timeout is 10   and context is passed already on `Terminate()` , lmk wdyt ?

modules/etcd/etcd.go
```diff
@@ -29,12 +29,12 @@ type EtcdContainer struct {
...
 // Terminate terminates the etcd container, its child nodes, and the network in which the cluster is running
 // to communicate between the nodes.
-func (c *EtcdContainer) Terminate(ctx context.Context) error {
+func (c *EtcdContainer) Terminate(ctx context.Context, opts ...testcontainers.TerminateOption) error {
 	var errs []error
 
 	// child nodes has no other children
 	for i, child := range c.childNodes {
-		if err := child.Terminate(ctx); err != nil {
+		if err := child.Terminate(ctx, opts...); err != nil {
```
stevenh, line 37: bug: not passed to the parent container below.

docker.go
```diff
@@ -296,19 +296,41 @@ func (c *DockerContainer) Stop(ctx context.Context, timeout *time.Duration) erro
...
-	// ContainerRemove hardcodes stop timeout to 3 seconds which is too short
-	// to ensure that child containers are stopped so we manually call stop.
-	// TODO: make this configurable via a functional option.
-	timeout := 10 * time.Second
-	err := c.Stop(ctx, &timeout)
+//
+// Default: timeout is 10 seconds.
+func (c *DockerContainer) Terminate(ctx context.Context, opts ...TerminateOption) error {
+	defaultTimeout := 10 * time.Second
+	defaultOptions := &terminateOptions{
```
stevenh, line 324: bug: for this to work for implementations outside of testcontainers package this would need to be exported alongside the fields.

docker.go
```diff
@@ -343,6 +365,25 @@ func (c *DockerContainer) Terminate(ctx context.Context) error {
 	c.sessionID = ""
 	c.isRunning = false
 
+	// Remove additional volumes if any.
```
stevenh, line 368: suggestion: can you add a TODO to simplify this when when we perform the client refactor please.

docker_test.go
```diff
@@ -279,6 +279,27 @@ func TestContainerStateAfterTermination(t *testing.T) {
 		require.Nil(t, state, "expected nil container inspect.")
 	})
 
+	t.Run("Nil State after termination with timeout", func(t *testing.T) {
```
stevenh, line 282: suggestion: we're standardising on short names with no spaces so a search from the failed sub test name can be successful.
```suggestion
	t.Run("termination-timeout", func(t *testing.T) {
```

docker_test.go
```diff
@@ -1087,6 +1108,56 @@ func TestContainerCreationWithVolumeAndFileWritingToIt(t *testing.T) {
...
+					ContainerFilePath: "/hello.sh",
+				},
+			},
+			Mounts:     Mounts(VolumeMount(volumeName, "/data")),
+			Cmd:        []string{"bash", "/hello.sh"},
+			WaitingFor: wait.ForLog("done"),
+		},
+		Started: true,
+	})
+	require.NoError(t, err)
```
stevenh, line 1137: bug: missing `ContainerCleanup` call before `NoError`.

docker_test.go
```diff
@@ -1087,6 +1108,56 @@ func TestContainerCreationWithVolumeAndFileWritingToIt(t *testing.T) {
 	require.NoError(t, err)
 }
 
+func TestContainerCreationWithVolumeCleaning(t *testing.T) {
+	absPath, err := filepath.Abs(filepath.Join(".", "testdata", "hello.sh"))
+	require.NoError(t, err)
```
stevenh, line 1113: Just so you're aware (no need to change) we've been using embed for this type recently to its easier to reuse.
```go
//go:embed testdata/hello.sh
var helloSh []byte
```
Then use:
```go
ContainerFile{
    Reader: bytes.NewReader(helloSh),
    ContainerFilePath: "/hello.sh",
    FileMode: 0o755,
}
```

docker_test.go
```diff
@@ -1087,6 +1108,56 @@ func TestContainerCreationWithVolumeAndFileWritingToIt(t *testing.T) {
...
+
+	// Create the container that writes into the mounted volume.
+	bashC, err := GenericContainer(ctx, GenericContainerRequest{
+		ProviderType: providerType,
+		ContainerRequest: ContainerRequest{
+			Image: "bash:5.2.26",
+			Files: []ContainerFile{
+				{
+					HostFilePath:      absPath,
+					ContainerFilePath: "/hello.sh",
```
stevenh, line 1128: bug: Missing `FileMode` field which will cause unpredictable behaviour.

docker_test.go
```diff
@@ -279,6 +279,27 @@ func TestContainerStateAfterTermination(t *testing.T) {
...
+		ctx := context.Background()
+		nginx, err := createContainerFn(ctx)
+		require.NoError(t, err)
+
+		err = nginx.Start(ctx)
+		require.NoError(t, err, "expected no error from container start.")
+
+		state, err := nginx.State(ctx)
+		require.NoError(t, err, "expected no error from container inspect.")
+		require.NotNil(t, state, "expected non nil container inspect.")
```
stevenh, line 292: nit: I wouldn't include this as we're not testing the behaviour of the `State` call, similarly with the `Nil` check after `require.Error`.

stevenh (changes requested):
Thanks for fleshing out what this looks like, I've done a review on how it stands and will discuss with @mdelapenya in the new year the pros and cons of functional options vs require args.

modules/etcd/etcd.go
moogacs: good catch, thanks

docker.go
```diff
@@ -296,19 +296,40 @@ func (c *DockerContainer) Stop(ctx context.Context, timeout *time.Duration) erro
...
+	defaultOptions := &TerminateOptions{
+		Timeout: &DefaultTimeout,
+		Context: ctx,
+	}
+
+	for _, opt := range opts {
+		opt(defaultOptions)
+	}
+
+	err := c.Stop(defaultOptions.Context, defaultOptions.Timeout)
```
stevenh, line 332: suggestion: use method as defined above
```suggestion
	err := c.Stop(options.Context(), options.StopTimeout())
```

docker.go
```diff
@@ -296,19 +296,40 @@ func (c *DockerContainer) Stop(ctx context.Context, timeout *time.Duration) erro
...
+// Default: timeout is 10 seconds.
+func (c *DockerContainer) Terminate(ctx context.Context, opts ...TerminateOption) error {
+	defaultOptions := &TerminateOptions{
+		Timeout: &DefaultTimeout,
+		Context: ctx,
+	}
+
+	for _, opt := range opts {
+		opt(defaultOptions)
+	}
```
stevenh, line 330:
```suggestion
	options := NewTimeoutOptions(ctx, opts...)
```

docker.go
```diff
@@ -343,6 +364,26 @@ func (c *DockerContainer) Terminate(ctx context.Context) error {
...
+
+	defer client.Close()
+
+	// Best effort to remove all volumes.
+	for _, volume := range defaultOptions.Volumes {
+		if errRemove := client.VolumeRemove(ctx, volume, true); errRemove != nil {
+			errs = append(errs, fmt.Errorf("volume remove %q: %w", volume, errRemove))
+		}
+	}
+
```
stevenh, line 386: suggest: leverage the Cleanup method.
```suggestion
	if err := options.Cleanup(ctx); err != nil {
		errs = append(errs, err)
	}
```

cleanup.go
```diff
@@ -2,35 +2,37 @@ package testcontainers
...
-	timeout *time.Duration
-	volumes []string
+// DefaultTimeout for termination
+var DefaultTimeout = 10 * time.Second
+
+// TerminateOptions is a type that holds the options for terminating a container.
+type TerminateOptions struct {
+	Context context.Context
+	Timeout *time.Duration
+	Volumes []string
```
stevenh, line 17: suggestion: use a constructor and methods to ensure consistent behaviour with minimum effort from different types.

```suggestion
// TerminateOptions is a type that holds the options for terminating a container.
type TerminateOptions struct {
        stopTimeout *time.Duration
        volumes []string
        ctx context.Context
}

// NewTerminateOptions returns a fully initialised TerminateOptions.
// Defaults: StopTimeout: 10 seconds.
func NewTerminateOptions(ctx context.Context, opts ...TerminateOption) *TerminateOptions {
        stopTimeout := 10 * time.Second
	options := &TerminateOptions{
		stopTimeout: &stopTimeout,
		ctx: ctx,
	}

	for _, opt := range opts {
		opt(options)
	}

	return options
}

// Context returns the context to use duration a Terminate.
func (o *TerminateOptions) Context() context.Context {
        return o.ctx
}

// StopTimeout returns the stop timeout to use duration a Terminate.
func (o *TerminateOptions) StopTimeout() time.Duration {
        return o.stopTimeout
}

// Cleanup performs any clean up needed.
func (o *TerminateOptions) Cleanup() error {
	// TODO: simplify this when when perform the client refactor.
	if len(o.volumes) == 0 {
		return nil
	}

	client, err := NewDockerClientWithOpts(o.ctx)
	if err != nil {
		return fmt.Errorf("docker client: %w", err)
	}
	defer client.Close()

	// Best effort to remove all volumes.
	var err []errors
	for _, volume := range o.volumes {
		if errRemove := client.VolumeRemove(o.ctx, volume, true); errRemove != nil {
			errs = append(errs, fmt.Errorf("volume remove %q: %w", volume, errRemove))
		}
	}

	return errors.Join(errs...)

**Review by stevenh — CHANGES_REQUESTED** [review]
Sorry to be a pain, but looking at what you've done with this, we should actually make the fields private with methods to avoid any ambiguity as to their use by other types, see suggestion in `docker.go`.

File: cleanup.go [hunk 17]
```diff
@@ -8,16 +8,63 @@ import (
...
+	stopTimeout *time.Duration
+	volumes     []string
 }
 
 // TerminateOption is a type that represents an option for terminating a container.
 type TerminateOption func(*terminateOptions)
 
+// NewTerminateOptions returns a fully initialised TerminateOptions.
+// Defaults: StopTimeout: 10 seconds.
+func NewTerminateOptions(ctx context.Context, opts ...TerminateOption) *terminateOptions {
```
stevenh, line 26: suggestion: The return type should be exported, as per my previous suggestion. Just keeping fields private ensures the behaviour is consistent.

docker.go
```diff
@@ -343,6 +356,10 @@ func (c *DockerContainer) Terminate(ctx context.Context) error {
 	c.sessionID = ""
 	c.isRunning = false
 
+	if err := options.Cleanup(); err != nil {
```
stevenh, line 359: bug: this `err` definition is actually shadowing the existing variable.
```suggestion
	if err = options.Cleanup(); err != nil {
```

docker.go
```diff
@@ -296,19 +296,32 @@ func (c *DockerContainer) Stop(ctx context.Context, timeout *time.Duration) erro
...
+	}
+}
+
+// WithTerminateVolumes is a functional option that sets the volumes for the container termination.
+func WithTerminateVolumes(volumes ...string) TerminateOption {
+	return func(opts *terminateOptions) {
+		opts.volumes = volumes
+	}
+}
+
```
stevenh, line 312: bug: These are already covered by `StopTimeout` and `RemoveVolumes` in `cleanup.go`.
```suggestion
```

cleanup.go
```diff
@@ -8,29 +8,76 @@ import (
 	"time"
 )
 
-// terminateOptions is a type that holds the options for terminating a container.
-type terminateOptions struct {
-	ctx     context.Context
-	timeout *time.Duration
-	volumes []string
+// DefaultTimeout for termination
+var DefaultTimeout = 10 * time.Second
```
stevenh, line 12: bug: sorry mentioned this previously but missed this in the last review, with it being exposed it can introduce race condition.
```suggestion
```

cleanup.go
```diff
@@ -8,29 +8,76 @@ import (
...
 
 // TerminateOption is a type that represents an option for terminating a container.
-type TerminateOption func(*terminateOptions)
+type TerminateOption func(*TerminateOptions)
+
+// NewTerminateOptions returns a fully initialised TerminateOptions.
+// Defaults: StopTimeout: 10 seconds.
+func NewTerminateOptions(ctx context.Context, opts ...TerminateOption) *TerminateOptions {
+	options := &TerminateOptions{
+		stopTimeout: &DefaultTimeout,
```
stevenh, line 28: suggestion: make local to fix the race condition mentioned above.
```suggestion
	timeout := time.Second * 10
	options := &TerminateOptions{
		stopTimeout: &timeout,
```

stevenh (changes requested):
Sorry missed a bug in the last review

stevenh (approved):
Thanks for all your work on this, looks good to me.

cleanup.go
```diff
@@ -8,29 +8,74 @@ import (
...
+		stopTimeout: &timeout,
+		ctx:         ctx,
+	}
+	for _, opt := range opts {
+		opt(options)
+	}
+	return options
+}
+
+// Context returns the context to use duration a Terminate.
```
mdelapenya, line 35: typo: 
```suggestion
// Context returns the context to use during a Terminate.
```

cleanup.go
```diff
@@ -8,29 +8,74 @@ import (
...
+func (o *TerminateOptions) Context() context.Context {
+	return o.ctx
+}
+
+// StopTimeout returns the stop timeout to use duration a Terminate.
```
mdelapenya, line 40: typo
```suggestion
// StopTimeout returns the stop timeout to use during a Terminate.
```

docker_test.go
mdelapenya: I think is not addressed yet. Could you add the mode?

cleanup.go
```diff
@@ -8,29 +8,74 @@ import (
...
+	volumes     []string
 }
 
 // TerminateOption is a type that represents an option for terminating a container.
-type TerminateOption func(*terminateOptions)
+type TerminateOption func(*TerminateOptions)
+
+// NewTerminateOptions returns a fully initialised TerminateOptions.
+// Defaults: StopTimeout: 10 seconds.
+func NewTerminateOptions(ctx context.Context, opts ...TerminateOption) *TerminateOptions {
```
mdelapenya, line 23: docs: can we make sure we include the new options in the docs site?

mdelapenya (commented):
I'm still on PTO, being back Jan 2nd, but had time between sweets and meals to take a quick look.

Just a few comments to be addressed, other than that, will approve once resolved.

Thank you both for your time and effort here

cleanup.go
moogacs: i've added doc, with consideration it will be released in next RC, lmk if it's not in the right place

stevenh (approved):

docs/features/common_functional_options.md
```diff
@@ -165,3 +165,41 @@ The above example is updating the predefined command of the image, **appending**
 
 !!!info
     This can't be used to replace the command, only to append options.
+
+#### NewTerminateOptions
```
mdelapenya, line 169: I'm sorry but this is not properly documented on our docs 😢 

This file is for the functional options that affect the `Run` function in the modules API.

I'd add this here: https://golang.testcontainers.org/features/garbage_collector/#terminate-function if possible.

moogacs: no worries, i was almost certain i added it in the writing place 😅, will do update it thank you

docs/features/garbage_collector.md
```diff
@@ -17,6 +17,45 @@ The primary method is to use the `Terminate(context.Context)` function that is
 available when a container is created. Use `defer` to ensure that it is called
 on test completion.
 
+#### NewTerminateOptions
+
+- Since testcontainers-go <a href="https://github.com/testcontainers/testcontainers-go/releases/tag/v0.35.0"><span class="tc-version">:material-tag: v0.35.0</span></a>
```
mdelapenya, line 22: Again not documented on our side, as the placeholder I'm suggesting will be replaced by the release script with the version we use (it'll be `0.35.0` but imagine we change our minds and decide to release `1.0.0`).

This is our bad, so I'm going to add it to the contributing docs asap 🙏. Please forgive me on this ping-pong.

```suggestion
- Not available until the next release of testcontainers-go <a href="https://github.com/testcontainers/testcontainers-go"><span class="tc-version">:material-tag: main</span></a>
```

docs/features/garbage_collector.md
```diff
@@ -17,6 +17,45 @@ The primary method is to use the `Terminate(context.Context)` function that is
 available when a container is created. Use `defer` to ensure that it is called
 on test completion.
```
mdelapenya, line 19:
```suggestion

The `Terminate` function can be customised with termination options to determine how a container is removed: termination timeout, and the ability to remove container volumes are supported at the moment. You can build the default options using the `testcontainers.NewTerminationOptions` function.

```

mdelapenya (approved):
LGTM, approving it although I added two docs changes.

Will merge this PR right after they are addressed and the CI passes 🤞 

Thank you for your work here 🙇

docs/features/garbage_collector.md
```diff
@@ -17,6 +17,47 @@ The primary method is to use the `Terminate(context.Context)` function that is
...
 
+The `Terminate` function can be customised with termination options to determine how a container is removed: termination timeout, and the ability to remove container volumes are supported at the moment. You can build the default options using the `testcontainers.NewTerminationOptions` function.
+
+#### NewTerminateOptions
+
+- Not available until the next release of testcontainers-go <a href="https://github.com/testcontainers/testcontainers-go"><span class="tc-version">:material-tag: main</span></a>
+
+If you want to attach option to container termination, you can use the `testcontainer.NewTerminateOptions(ctx context.Context, opts ...TerminateOption) *TerminateOptions` option, which receives a TerminateOption as parameter, creating custom termination options to be passed on the container termination.
```
mdelapenya, line 26:
```suggestion
If you want to attach option to container termination, you can use the `testcontainers.NewTerminateOptions(ctx context.Context, opts ...TerminateOption) *TerminateOptions` option, which receives a TerminateOption as parameter, creating custom termination options to be passed on the container termination.
```

mdelapenya (approved):

mdelapenya:
Merged, thank you @moogacs and @stevenh for your time reviewing this PR

container.go
ash2k, line 53: Hm, I wonder why we need a separate parameter for the timeout when a context is already provided to `Terminate()`?

As a user, I expect this to work fine. Does this code not do what is expected?

```go
ctx, _ := context.WithTimeout(context.Background(), time.Second*10)
c.Terminate()
```

If a `Duration` value is needed (vs reacting to a channel close), `context.Background().Deadline()` returns a deadline `Time`. It seems that there is no need for a new API.

PR #2929, opened 2024-12-20
feat(wait): log sub match callback

stevenh:
Add support for a sub match callback to wait.LogStrategy which allows containers to process the matched pattern storing details or otherwise validating them.

The callback can return a PermanentError if no more retries should be attempted.


mdelapenya (approved):
LGTM, thanks!

PR #2936, opened 2025-01-10
fix(localstack): more reliable legacy tag detection

NathanBaulch:
## What does this PR do?

Improves the internal `isLegacyMode` function to handle all tag conventions used by localstack.

## Why is it important?

I'm currently unable to use the much smaller `localstack/localstack:s3-latest` image tag because it's incorrectly detected as a legacy version.

## How to test this PR

Unit test included.

### DIFF
--- a/modules/localstack/localstack.go
+++ b/modules/localstack/localstack.go
@@ -21,14 +21,14 @@ const (
 	localstackHostEnvVar   = "LOCALSTACK_HOST"
 )
 
-var nonLegacyTags = []string{
+var recentVersionTags = []string{
 	"latest",
 	"s3",
 	"s3-latest",
 	"stable",
 }
 
-func isLegacyMode(image string) bool {
+func isMinimumVersion(image string, minVersion string) bool {
 	parts := strings.Split(image, ":")
 	version := parts[len(parts)-1]
 
@@ -36,30 +36,7 @@ func isLegacyMode(image string) bool {
 		version = version[0:pos]
 	}
 
-	if slices.Contains(nonLegacyTags, version) {
-		return false
-	}
-
-	if !strings.HasPrefix(version, "v") {
-		version = "v" + version
-	}
-
-	if semver.IsValid(version) {
-		return semver.Compare(version, "v0.11") < 0 // version < v0.11
-	}
-
-	return true
-}
-
-func isVersion2(image string) bool {
-	parts := strings.Split(image, ":")
-	version := parts[len(parts)-1]
-
-	if pos := strings.LastIndexByte(version, '-'); pos >= 0 {
-		version = version[0:pos]
-	}
-
-	if slices.Contains(nonLegacyTags, version) {
+	if slices.Contains(recentVersionTags, version) {
 		return true
 	}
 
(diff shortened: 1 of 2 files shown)


modules/localstack/localstack.go
```diff
@@ -24,7 +24,11 @@ func isLegacyMode(image string) bool {
 	parts := strings.Split(image, ":")
 	version := parts[len(parts)-1]
 
-	if version == "latest" {
+	if pos := strings.LastIndexByte(version, '-'); pos >= 0 {
+		version = version[0:pos]
+	}
+
+	if version == "latest" || version == "s3" || version == "s3-latest" || version == "stable" {
```
mdelapenya, line 31: suggestion: we could add an slice of valid non-legacy versions at the top of the file. Here we would just look up the used version to verify if it's legacy. Wdyt?

mdelapenya (commented):
Thanks for the contribution, added a comment regarding the non-legacy tags

modules/localstack/localstack.go
NathanBaulch: Are you suggesting every legacy version is simply hard coded in a slice? It looks like there are 31 versions prior to `v0.11`, plus the following unconventional tags that I have no idea about:

- `2021-11-15`
- `2022-01-27`
- `2022-01-28`
- `2022-07-25`
- `fix-perf`
- `jvm-2022-11-03`
- `m1-jre-fix`
- `multi`

NathanBaulch:
I just noticed the `isVersion2` function that should also be improved in similar ways.

modules/localstack/localstack.go
NathanBaulch: Here's a much simpler implementation that actually works (except for the junk "foo" test case).

```go
func isLegacyMode(image string) bool {
	return strings.Compare(image, "localstack/localstack:0.11") < 0
}
```

mdelapenya: > Are you suggesting every legacy version is simply hard coded in a slice? It looks like there are 31 versions prior to v0.11, plus the following unconventional tags that I have no idea aboutAre you suggesting every legacy version

No, just the ones in the if branch:
```
if version == "latest" || version == "s3" || version == "s3-latest" || version == "stable"
```

NathanBaulch: Sure, done. I've also fixed the `isVersion2` function along with unit tests.

mdelapenya (approved):
@stevenh anything to highlight here? This LGTM, thanks!

mdelapenya (approved):
LGTM, thanks!

Question (not a blocker at all): do you see we can take the chance to refactor both functions, isLegacyMode and isVersion2 as the code in there is practically the same? I'm approving the PR anyway, so will merge it if you don't consider it for this PR and prefer doing it in a follow-up.

Thanks again for your time 🙇

NathanBaulch:
> do you see we can take the chance to refactor both functions, isLegacyMode and isVersion2 as the code in there is practically the same?

Definitely. I thought about doing this but didn't want to mess with the code too much as a new contributor. See my latest change.

mdelapenya (approved):
LGTM, thanks for the refactor!

Issue #2859, opened 2024-10-30
[Bug]: `github.com/testcontainers/testcontainers-go.SkipIfProviderIsNotHealthy` panics on GHA `macos-latest` runners


RomainMuller:
### Testcontainers version

0.34.0

### Using the latest Testcontainers version?

Yes

### Host OS

macOS

### Host arch

x86, ARM64

### Go version

1.22 / 1.23

### Docker version

Docker is not available on `macos-latest` GitHub Actions runners.

### Docker info

Docker is not available on `macos-latest` GitHub Actions runners.

### What happened?

Attempting to skip tests on runners where Docker is not usable using `github.com/testcontainers/testcontainers-go.SkipIfProviderIsNotHealthy`; tests still fail on an unconditional `panic` on `macos-latest` runners in GitHub Actions.

### Relevant log output

    agent.go:76: Starting /Users/runner/hostedtoolcache/Python/3.13.0/arm64/bin/ddapm-test-agent on port 50048
    gen_test.go:18: Running setup
panic: rootless Docker not found [recovered]
	panic: rootless Docker not found

goroutine 53 [running]:
testing.tRunner.func1.2({0x102f964a0, 0x140001285d0})
	/Users/runner/hostedtoolcache/go/1.22.8/arm64/src/testing/testing.go:1631 +0x1c4
testing.tRunner.func1()
	/Users/runner/hostedtoolcache/go/1.22.8/arm64/src/testing/testing.go:1634 +0x33c
panic({0x102f964a0?, 0x140001285d0?})
	/Users/runner/hostedtoolcache/go/1.22.8/arm64/src/runtime/panic.go:770 +0x124
github.com/testcontainers/testcontainers-go/internal/core.MustExtractDockerHost.func1()
	/Users/runner/go/pkg/mod/github.com/testcontainers/testcontainers-go@v0.34.0/internal/core/docker_host.go:91 +0x78
sync.(*Once).doSlow(0x0?, 0x0?)
	/Users/runner/hostedtoolcache/go/1.22.8/arm64/src/sync/once.go:74 +0x100
sync.(*Once).Do(...)
	/Users/runner/hostedtoolcache/go/1.22.8/arm64/src/sync/once.go:65
github.com/testcontainers/testcontainers-go/internal/core.MustExtractDockerHost({0x103092ec0?, 0x1035f0f80?})
[...]

PR #2939, opened 2025-01-10
chore: recover from panic in `SkipIfProviderIsNotHealthy` test helper function

jasonyunicorn:
## What does this PR do?
Recovers from panic (in a `defer` statement) and skips the tests if necessary.

## Why is it important?
Attempting to skip tests on runners where Docker is not usable using `github.com/testcontainers/testcontainers-go.SkipIfProviderIsNotHealthy` is still failing unconditionally.

## Related issues
Closes [#2859](https://github.com/testcontainers/testcontainers-go/issues/2859)

## How to test this PR
Run the tests in CI/CD for `macos-latest` runners in GitHub Actions and observe if there any still any panics.


testing.go
```diff
@@ -20,6 +20,12 @@ var errAlreadyInProgress = regexp.MustCompile(`removal of container .* is alread
 // This is a function designed to be used in your test, when Docker is not mandatory for CI/CD.
 // In this way tests that depend on Testcontainers won't run if the provider is provisioned correctly.
 func SkipIfProviderIsNotHealthy(t *testing.T) {
+	defer func() {
+		if r := recover(); r != nil {
+			t.Skipf("Recovered from panic. Docker is not running. TestContainers can't perform is work without it")
```
mdelapenya, line 25: Typo: can you apply it to the rest of occurrences in the file 🙏 ?
```suggestion
			t.Skipf("Recovered from panic. Docker is not running. Testcontainers can't perform is work without it")
```

mdelapenya (approved):
LGTM, just suggesting fixing an existing typo.

Thanks!

testing.go
```diff
@@ -20,6 +20,12 @@ var errAlreadyInProgress = regexp.MustCompile(`removal of container .* is alread
 // This is a function designed to be used in your test, when Docker is not mandatory for CI/CD.
 // In this way tests that depend on Testcontainers won't run if the provider is provisioned correctly.
 func SkipIfProviderIsNotHealthy(t *testing.T) {
+	defer func() {
```
mdelapenya, line 23: I was not sure if t.Helper should go before the defer block, but it seems so.

Could you please update it, as golangci-lint suggests?

testing.go
stevenh: I would also include the value of `r` as there could be a few different cases and this would disambiguate them.

testing.go
jasonyunicorn: Whoops, sorry about that.  I didn't run `golangci-lint` in my haste.

Fixed.

mdelapenya (approved):
LGTM, thanks!

PR #2946, opened 2025-01-24
chore: enable context-as-argument from revive linter

mmorel-35:
## What does this PR do?

[revive](https://golangci-lint.run/usage/linters/#revive) is a linter for Go. Drop-in replacement of golint.

This PR list default rules and focus on applying context-as-argument rule.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -37,7 +37,6 @@ linters-settings:
     rules:
       - name: blank-imports
       - name: context-as-argument
-        disabled: true
         arguments:
           - allowTypesBefore: "*testing.T"
       - name: context-keys-type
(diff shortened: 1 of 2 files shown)


.golangci.yml
```diff
@@ -31,6 +30,51 @@ linters-settings:
       - standard
       - default
       - prefix(github.com/testcontainers)
+  nakedret:
+    max-func-lines: 0
+  revive:
```
stevenh, line 35: suggestion: This looks like the defaults, it's cleaner to override any settings we want to so we should be able to remove this.

.golangci.yml
```diff
@@ -9,13 +9,12 @@ linters:
     - nolintlint
     - nakedret
     - perfsprint
+    - revive
```
stevenh, line 12: thanks: the more linters the better :)

stevenh (changes requested):
Thanks for the PR, I think we can simplify by removing what looks like default config?

stevenh (approved):
Thanks for the clarification, LGTM

mdelapenya (approved):
LGTM, thanks!

PR #2948, opened 2025-01-27
chore: enable use-any from revive

mmorel-35:
#### What does this PR do?

enable [use-any](https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#use-any) rule from revive


mdelapenya (approved):
LGTM, thanks!

PR #2954, opened 2025-01-30
feat: added firebase module

xytis:
## What does this PR do?

This module allows running firebase emulator as a testcontainer.

It does currently depend on a forked docker image, residing under my ownership. There is no official firebase emulator docker image to my knowledge.

## Why is it important?

We use firebase, thought someone else might be using firebase too.


modules/firebase/examples_test.go
```diff
@@ -0,0 +1,41 @@
...
+func ExampleRunContainer() {
+	// runFirebaseContainer {
+	ctx := context.Background()
+
+	firebaseContainer, err := firebase.RunContainer(
+		ctx,
+		firebase.WithRoot(filepath.Join(BasePath(), "firebase")),
+	)
+	if err != nil {
+		log.Fatalf("failed to start container: %s", err)
```
stevenh, line 21: suggestion: please see the latest template for this which ensures that defer's are run by not calling `log.Fatal`, more below.

modules/firebase/examples_test.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/testcontainers/testcontainers-go/modules/firebase"
+)
+
+func ExampleRunContainer() {
+	// runFirebaseContainer {
+	ctx := context.Background()
+
+	firebaseContainer, err := firebase.RunContainer(
+		ctx,
+		firebase.WithRoot(filepath.Join(BasePath(), "firebase")),
```
stevenh, line 18: suggestion: just use "." here to avoid the need for `BasePath`.

modules/firebase/anchor_test.go
```diff
@@ -0,0 +1,13 @@
+package firebase_test
+
+import (
+	"path/filepath"
+	"runtime"
+)
+
+// BasePath returns the catalog of the module this function resides in
+func BasePath() string {
```
stevenh, line 9: suggestion: I believe this can be removed see the comment in `examples_test.go`

modules/firebase/firebase.go
```diff
@@ -0,0 +1,109 @@
...
+type FirebaseContainer struct {
+	testcontainers.Container
+}
+
+const defaultImageName = "ghcr.io/u-health/docker-firebase-emulator:13.29.2"
+
+// WithRoot sets the directory which is copied to the destination container as firebase root
+func WithRoot(rootPath string) testcontainers.CustomizeRequestOption {
+	return func(req *testcontainers.GenericContainerRequest) error {
+		if !strings.HasSuffix(rootPath, "/firebase") {
```
stevenh, line 25: question: Could you clarify why this requirement is needed?

modules/firebase/firebase.go
```diff
@@ -0,0 +1,109 @@
...
+
+// WithData names the data directory in firebase root
+func WithData(dataPath string) testcontainers.CustomizeRequestOption {
+	return func(req *testcontainers.GenericContainerRequest) error {
+		req.Env["DATA_DIRECTORY"] = dataPath
+		return nil
+	}
+}
+
+func cache(volumeName string, volumeOptions *mount.VolumeOptions) testcontainers.CustomizeRequestOption {
```
stevenh, line 45: suggestion: fold this into WithCache, no need for a separate method.

modules/firebase/firebase.go
```diff
@@ -0,0 +1,109 @@
...
+		Started:          true,
+	}
+
+	for _, opt := range opts {
+		if err := opt.Customize(&genericContainerReq); err != nil {
+			return nil, fmt.Errorf("customize: %w", err)
+		}
+	}
+
+	container, err := testcontainers.GenericContainer(ctx, genericContainerReq)
```
stevenh, line 103: bug: see the current template or other modules on how they handle this case, as container can still be not nil and hence required for clean up.

modules/firebase/firebase_test.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/stretchr/testify/assert"
+
+	"github.com/testcontainers/testcontainers-go/modules/firebase"
+)
+
+func TestFirebase(t *testing.T) {
+	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(3*time.Minute))
+
+	thing := fmt.Sprintf("%s/test", os.Getenv("PWD"))
+	fmt.Println(thing)
```
stevenh, line 20: question: looks like debug?

modules/firebase/firebase_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+	container, err := firebase.RunContainer(
+		ctx,
+		firebase.WithRoot(filepath.Join(BasePath(), "firebase")),
+		firebase.WithCache(),
+	)
+	if err != nil {
```
stevenh, line 27: bug: container can require clean up see other modules and template for how to handle this. We're also standarising on require for checks like this in tests.

modules/firebase/firebase_test.go
```diff
@@ -0,0 +1,47 @@
...
+		t.Fatal(err)
+	}
+
+	// Clean up the container after the test is complete
```
stevenh, line 31: bug: see comment above on how to use:
```go
	require.NoError(t, err)
```

modules/firebase/firebase_test.go
```diff
@@ -0,0 +1,47 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/testcontainers/testcontainers-go/modules/firebase"
+)
+
+func TestFirebase(t *testing.T) {
+	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(3*time.Minute))
```
stevenh, line 17: bug: cancel should be called in a clean up directly after this

modules/firebase/ports.go
```diff
@@ -0,0 +1,73 @@
...
+		return "", err
+	}
+	port, err := c.MappedPort(ctx, portName)
+	if err != nil {
+		return "", err
+	}
+	return fmt.Sprintf("%s:%s", host, port.Port()), nil
+}
+
+func (c *FirebaseContainer) UIConnectionString(ctx context.Context) (string, error) {
```
stevenh, line 35: question: seems like lots of additional API surface area instead of just exposing `ConnectionString`, thoughts?

If not then we'd need comments for all the methods.

modules/firebase/ports.go
```diff
@@ -0,0 +1,73 @@
...
+	DatabasePort  = "9000/tcp"
+	AuthPort      = "9099/tcp"
+	StoragePort   = "9199/tcp"
+	HostingPort   = "6000/tcp"
+)
+
+func (c *FirebaseContainer) connectionString(ctx context.Context, portName nat.Port) (string, error) {
+	host, err := c.Host(ctx)
+	if err != nil {
+		return "", err
```
stevenh, line 26: suggestion: wrap the error so the user knows the cause, more below.

stevenh (changes requested):
Thanks for the PR, I've done a initial pass. The majority of comments center around bringing it inline with the latest template for modules in terms of style and correct behaviour e.g. handling non nil container with error.

modules/firebase/firebase.go
xytis: This has been an issue a year ago. Mounting did not correctly work at that time, and if the source directory did not match the destination directory, mounting would nest the directory instead of mapping it directly.

I see that the issue has been fixed, removing.

modules/firebase/examples_test.go
xytis: Remade the whole PR using the new template. This code has been stale for a bit, we used it internally for over a year now :)

modules/firebase/examples_test.go
xytis: ✅

modules/firebase/ports.go
xytis: This one is a bit more tricky.

I would love to change all of these methods, and figure out a better way to configure the launched container.

Currently I ship a catalog which contains static predefined configuration for the emulator. These methods happen to match the port numbers defined in the said file.

But, in the tests that we use, we provide a completely different setup (because we don't want to boot the full emulation suite), and we have to match the port numbers in our config so that all of this works.

So a question from my side would be:
Is there a way to generate file based config during container setup? To my knowledge firebase emulator does not allow passing those values as environment variables.

I think, ideal use example would be this:
```go
	firebaseContainer, err := firebase.Run(ctx, 
	  "ghcr.io/u-health/docker-firebase-emulator:13.29.2",
	  firebase.WithUI(),
	  firebase.WithFirestore(),
	  firebase.WithCache(),
	)
```

And then, helper methods `XXXConnectionString` would actually return meaningful errors if the emulator part was not launched.

mdelapenya: > Is there a way to generate file based config during container setup? 

@xytis you can take a look at the redpanda module, where there is a Go template the module is using to build the configuration file.

xytis: After some testing on my side, I remembered why I shipped the config as a directory in the first place.

Default (and probably expected) use case for firebase projects, is to run `firebase init` and then use the scaffolded configuration to interact with the dependencies.

In the following commit, I've changed the behaviour of this module. It now attempts to understand the emulator configuration of your project and then boots the containerised emulator with that info.

At least in my use case, that makes sense, and when I launch the tests, I use the same project-wide config for all of them.

modules/firebase/firebase.go
xytis: Ah, my bad. The issue still exits. I wrote a failing test to exhibit the issue.

Is there something incorrect with the way I copy over the configuration?

Also, I don't think that volume mount would work in this case, because firebase emulator creates a lot of trash data in the root catalog, which in some cases can even make testing flaky (especially if using DATA_DIRECTORY as a fixture).

stevenh: I'll try to find some time to look. I had a some in progress work to fix up the copy behaviour which was odd.

If you have a test to exercise this that, would help alongside a description of the desired behaviour and what you're seeing instead.

modules/firebase/firebase.go
```diff
@@ -0,0 +1,165 @@
...
+				VolumeOptions: volumeOptions,
+			},
+			Target: cacheFilePath,
+		}
+		req.Mounts = append(req.Mounts, m)
+		return nil
+	}
+}
+
+func gatherPorts(config FirebaseConfig) ([]string, error) {
```
stevenh, line 70: question: could you describe what this is doing as it looks like its just getting the ports from UI field which wouldn't require reflection.

modules/firebase/firebase.go
```diff
@@ -0,0 +1,165 @@
...
+			return nil, fmt.Errorf("customize: %w", err)
+		}
+	}
+
+	// Check if user supplied root:
+	rootPathIdx := slices.IndexFunc(req.ContainerRequest.Files, func(file testcontainers.ContainerFile) bool {
+		return file.ContainerFilePath == rootFilePath
+	})
+	if rootPathIdx == -1 {
+		return nil, fmt.Errorf("firebase emulator is unable to boot without configuration root (obtained via 'firebase init')")
```
stevenh, line 131: suggestion: this seems to indicate there is a missing configuration to the container, could be be more explicit / concise about what the user did wrong?

modules/firebase/firebase.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+	defer cfg.Close()
+	bytes, err := io.ReadAll(cfg)
+	if err != nil {
+		return nil, fmt.Errorf("read firebase.json: %w", err)
+	}
+	var parsed FirebaseConfig
+	if err := json.Unmarshal(bytes, &parsed); err != nil {
+		return nil, fmt.Errorf("parse firebase.json: %w", err)
+	}
```
stevenh, line 147: suggestion: use streamed reading instead so its more memory efficent.

modules/firebase/firebase.go
```diff
@@ -0,0 +1,165 @@
...
+	expectedExposedPorts, err := gatherPorts(parsed)
+	if err != nil {
+		return nil, fmt.Errorf("issues with emulator config block: %w", err)
```
stevenh, line 150: suggestion: seems like this is just an invalid config
```suggestion
		return nil, fmt.Errorf("invalid config %q: %w", configPath, err)
```

modules/firebase/firebase_test.go
```diff
@@ -0,0 +1,74 @@
...
+		firebase.WithRoot(filepath.Join(".", "firebase")),
+	)
+	testcontainers.CleanupContainer(t, ctr)
+	require.NoError(t, err)
+
+	// perform assertions
+	// Ports are linked to the example config in firebase/firebase.json
+
+	firestoreUrl, err := ctr.ConnectionString(ctx, "8080/tcp")
+	assert.NoError(t, err)
```
stevenh, line 32: suggestion: use require instead of assert, we're moving to that as a standard for the project, more below

modules/firebase/firebase_test.go
```diff
@@ -0,0 +1,74 @@
...
+	assert.NotEmpty(t, firestoreUrl)
+	_ = os.Setenv("FIRESTORE_EMULATOR_HOST", firestoreUrl)
```
stevenh, line 34: bug: use `t.Setenv` which gets restored when the test returns.

modules/firebase/firebase_test.go
```diff
@@ -0,0 +1,74 @@
...
+	c, err := firestore.NewClient(ctx, firestore.DetectProjectID)
+	require.NoError(t, err)
+	defer c.Close()
+	w, err := c.Collection("example").Doc("one").Set(ctx, map[string]interface{}{
```
stevenh, line 38: suggestion: use modern `any`

modules/firebase/firebase_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+	// perform assertions
+	// Ports are linked to the example config in firebase/firebase.json
+
+	firestoreUrl, err := ctr.ConnectionString(ctx, "8080/tcp")
+	assert.NoError(t, err)
+	assert.NotEmpty(t, firestoreUrl)
+	_ = os.Setenv("FIRESTORE_EMULATOR_HOST", firestoreUrl)
+	c, err := firestore.NewClient(ctx, firestore.DetectProjectID)
+	require.NoError(t, err)
```
stevenh, line 36: suggestion: add a blank line after block of logic to make it easier to understand.

modules/firebase/firebase_test.go
```diff
@@ -0,0 +1,74 @@
...
+	defer c.Close()
+	w, err := c.Collection("example").Doc("one").Set(ctx, map[string]interface{}{
+		"foo": "bar",
+	})
+	require.NotNil(t, w)
+	require.NoError(t, err)
```
stevenh, line 42: bug: error check should always come first as it will provide the correct context to the user.

modules/firebase/firebase_test.go
```diff
@@ -0,0 +1,74 @@
...
+	snap, err := c.Collection("example").Doc("one").Get(ctx)
+	require.NoError(t, err)
+	var out map[string]string
+	err = snap.DataTo(&out)
+	require.NoError(t, err)
+	assert.Equal(t, "bar", out["foo"])
+}
+
+func TestFirebaseBadDirectory(t *testing.T) {
```
stevenh, line 51: question: Is this issue you were referencing as in you are expecting:
`./failure` to be copied to `/srv/firebase` and instead if ends up as `/srv/failure`?

modules/firebase/firebase_test.go
```diff
@@ -0,0 +1,74 @@
...
+	require.NoError(t, err)
+}
+
+func TestFirebaseRequiresRoot(t *testing.T) {
+	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(3*time.Minute))
+	defer cancel()
+
+	ctr, err := firebase.Run(ctx, "ghcr.io/u-health/docker-firebase-emulator:13.29.2")
+	testcontainers.CleanupContainer(t, ctr)
+	require.Error(t, err)
```
stevenh, line 72: suggestion: remove redundant require.Error.

sonar-project.properties
```diff
@@ -1,8 +1,8 @@
 # This file is autogenerated by the 'modulegen' tool.
-# GitHub organization linked to sonarcloud
+# Github organization linked to sonarcloud
```
stevenh, line 2: bug: reverting a newer fix

sonar-project.properties
```diff
@@ -1,8 +1,8 @@
...
 sonar.organization=testcontainers
 
-# Project key from sonarcloud dashboard for GitHub Action, otherwise pick a project key you like
+# Project key from sonarcloud dashboard for Github Action, otherwise pick a project key you like
```
stevenh, line 5: bug: reverting a newer fix

.github/workflows/ci.yml
```diff
@@ -134,13 +134,12 @@ jobs:
           # Disabling shallow clone is recommended for improving relevancy of reporting
           fetch-depth: 0
       
-      - uses: actions/download-artifact@fa0a91b85d4f404e444e00e005971372dc801d16 # v4.1.8
+      - uses: actions/download-artifact@9bc31d5ccc31df68ecc42ccf4149144866c47d8a # v3.0.2
```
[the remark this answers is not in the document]
xytis: oops, sorry about this. I used modulegen and just trusted it made the correct change.

modules/firebase/firebase.go
```diff
@@ -0,0 +1,165 @@
...
+
+const rootFilePath = "/srv/firebase"
+
+// WithRoot sets the directory which is copied to the destination container as firebase root
+func WithRoot(rootPath string) testcontainers.CustomizeRequestOption {
+	return func(req *testcontainers.GenericContainerRequest) error {
+		req.Files = append(req.Files, testcontainers.ContainerFile{
+			HostFilePath:      rootPath,
+			ContainerFilePath: rootFilePath,
+			FileMode:          0o775,
```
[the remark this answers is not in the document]
xytis: This is a directory that has to be copied over. It should work without this file mode, I added this as part of my testing.

modules/firebase/firebase.go
```diff
@@ -0,0 +1,165 @@
...
+
+		return nil
+	}
+}
+
+// WithData names the data directory (by default under firebase root), can be used as a way of setting up fixtures.
+// Usage of absolute path will imply that the user knows how to mount external directory into the container.
+func WithData(dataPath string) testcontainers.CustomizeRequestOption {
+	return func(req *testcontainers.GenericContainerRequest) error {
+		req.Env["DATA_DIRECTORY"] = dataPath
```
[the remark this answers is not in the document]
xytis: No, weirdly enough, the emulator accepts this as an environment variable and it has no relation to the config at all.

modules/firebase/firebase.go
xytis: It gathers the port numbers from config.emulators sub-block.

Instead of reflection here, I could demolish the structure definition of the `partialFirebaseConfig`, parse the json into `map[string]interface{}` and then extract matching paths by doing key comparisons (I need to get all values under the glob `config.emulators.*.port` and `config.emulators.*.websocketPort`).

However, I think that having the structure definition copied over from the typescript library is a bit more comprehensive, and if the emulator config schema changes, it will be easier to see what has changed from a structure definition.

modules/firebase/firebase.go
xytis: I am a bit conflicted about this error check in general.

Depending on the container, users might have a default root available, so that the emulator would launch even if no external root is mounted/copied.

I guess I am removing it, with the assumption that users can read container boot logs and see what is missing.

xytis: But, then - the port discovery process breaks... Unless I can inspect the container file system before launch.

modules/firebase/firebase.go
xytis: Could not find a similar example along other modules. Is adding a streaming json parser as a dependency really worth the several kilobytes of expected memory efficiency? (firebase.json is not infinite)

modules/firebase/firebase_test.go
xytis: Yes, exactly.

And if I attempt to specify the destination path as `/srv/firebase/` I get either `/srv/firebase/firebase` or `/srv/firebase/failure`, neither of which work for my use case.

I could attempt to modify the container launch script to accommodate for this, but `firebase` executable expects the configuration to be in the launch dir. That means, I would have to stray away from default execution method, and other containers would be less likely to work with this adapter.

stevenh: Thanks, I was looking for a good example of this in real world.

modules/firebase/config.go
```diff
@@ -0,0 +1,87 @@
+package firebase
+
+// Based on https://github.com/firebase/firebase-tools/blob/master/src/firebaseConfig.ts
+
+type EmulatorsConfig struct {
+	SingleProjectMode bool `json:"singleProjectMode,omitempty"`
+
+	Auth struct {
+		Host string `json:"host,omitempty"`
+		Port int    `json:"port,omitempty"`
```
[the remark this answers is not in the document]
xytis: This structure is parsed from JSON, and I can only do int64 or float64 by default. Since I am later just placing the given value into a string, I think it is fine to keep as is.

.github/workflows/ci.yml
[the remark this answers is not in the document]
stevenh: modulegen wouldn't have touched this, its likely just your branch is out of date and needed updating.

modules/firebase/config.go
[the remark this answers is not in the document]
stevenh: I'm not sure what you mean, here JSON will parse correctly if these are changed to uint16, the only difference is it will correctly error if the number is out of range e.g. less than zero or greater than 65,535.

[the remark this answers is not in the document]
xytis: Ah, my bad. I misunderstood the error I got after I changed the types.

Last commit changes this.

modules/firebase/config.go
```diff
@@ -0,0 +1,87 @@
+package firebase
+
+// Based on https://github.com/firebase/firebase-tools/blob/master/src/firebaseConfig.ts
+
+type EmulatorsConfig struct {
```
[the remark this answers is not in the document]
xytis: Made private (as it should've been).

mdelapenya (commented):
I added a few minor comments, we are in the good way.

Thanks for your contribution

.github/workflows/ci.yml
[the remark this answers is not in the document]
mdelapenya: We fixed the templates in #2956 so it should not happen anymore

docs/modules/firebase.md
```diff
@@ -0,0 +1,55 @@
...
+- `testcontainers.ContainerCustomizer`, a variadic argument for passing options.
+
+### Container Options
+
+When starting the Firebase container, you can pass options in a variadic way to configure it.
+
+#### Image
+
+If you need to set a different Firebase Docker image, you can set a valid Docker image as the second argument in the `Run` function.
+E.g. `Run(context.Background(), "ghcr.io/u-health/docker-firebase-emulator:13.29.2")`.
```
mdelapenya, line 49: Question: is there an Official docker image for the Google Cloud firebase emulator?

xytis: No. The emulator is shipped as an executable that is supposed to run on localhost.

mdelapenya: thanks for confirming, I'd have expected the google folks delivered that emulator. Let me think about this for a while, and I will come back here with more context about the used Docker image. 🙏

xytis: Probably this will take a while on firebase side. Mostly because it should be blocked by firebase emulator UI, which is not currently configurable to work with custom remotes. And packing emulator into docker implies that a non standard address+port is used.

mdelapenya: Oh I saw you're the maintainer of the image 🙇  Let me take a look at the image first, but then I think we can move on until Google decides to release their artifact.

Thanks for your work!

mdelapenya:
One question: reading the README in the docker image:

> Supports GRPC for firestore and pub/sub.

What are the differences with the existing Google Cloud emulators? We do have them under the gcloud module, and we currently plan to follow the same package structure as in the upcoming azure module (see #3008). Do you think this firebase has enough entity to be a separate module, or should it live alongside the other gcloud modules?

xytis:
Last time I checked, gcloud emulators did not provide the full suite of emulations. I will check this again and come back to you.

mdelapenya:
@xytis we merged the layout revamp in the gcloud module. Do you think we can move this there?

xytis:
Hey,

unfortunately not.

`gcloud emulators` do not include the following:
- firebase auth
- firebase storage (which is an access control layer on top of gcloud storage)
- firebase pubsub (again, a layer on top, with some helpers)

I am not certain about the rest.

mdelapenya:
> gcloud emulators do not include the following:

but we're going to use the `ghcr.io/u-health/docker-firebase-emulator:13.29.2` Docker image, so I'd like to know if it's "semantically" correct to locate this code under the gcloud module, so it's distributed with all GCP services, or if on the contrary, it makes more sense to have it as a separate module.

I do not have an opinion on this, so as the author of the image, please feel free to be prescriptive about it. 🙏

xytis:
Well, if I was searching for "firebase" emulators, I would not look under "gcloud". Mostly because the tooling which is provided by google, and the documentation is very different for these two products.

I think, having these two separate makes more sense as long as there are development differences between running `gcloud` and `firebase` executables.

mdelapenya:
LGTM then to continue as a separate module, thanks! 🙏 

I can help you out rebasing this PR resolving the conflicts before moving on with the review

xytis:
Please do, I am a bit occupied at the moment.

Thanks.

mdelapenya:
@xytis I cannot push to your branch because this PR was submitted from your main branch 😞 

I've pushed it to my origin so that you can check and pick up the commits: https://github.com/testcontainers/testcontainers-go/compare/main...mdelapenya:testcontainers-go:firebase-module?expand=1

I still have doubts about the `WithData` option, probably as a result of being this one my first interaction with Firebase. Could you help me out in understanding its motivation? I'd need some tests in the code to verify it works as expected.

xytis:
@mdelapenya merged your changes.

I was forced to use WithData because Firebase emulators expect a specific directory structure and files to exist in the "project root".

I toyed with providing structured config, but I am not using that in my own tests. It was easier to bootstrap a directory with a specific catalog structure, and I just clone that into the container.

Rough docs on this are here: https://firebase.google.com/docs/emulator-suite/install_and_configure

mdelapenya:
@xytis lint is complaining. Could you run `make lint` from the `firebase` module dir? 🙏

docs/modules/firebase.md
```diff
@@ -0,0 +1,55 @@
...
+
+{% include "../features/common_functional_options.md" %}
+
```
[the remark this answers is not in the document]
mdelapenya: @xytis let's not forget adding the options here 🙏

modules/firebase/go.mod
```diff
@@ -0,0 +1,83 @@
+module github.com/testcontainers/testcontainers-go/modules/firebase
+
+go 1.23.0
+
+require (
```
mdelapenya, line 5: suggestion: just before submitting the changes, run `go mod tidy` from the module dir to make sure all dependencies are in sync with the core. Else we'll receive an extra PR from dependabot 😅

xytis: `go mod tidy` returned no changes. Was this expected?

mdelapenya:
@xytis for some reason, the lint still fails:

``` 
  Error: modules/firebase/modules/firebase/firebase_test.go:13:1: File is not properly formatted (gci)
  
  ^
  Error: modules/firebase/modules/firebase/options.go:7:1: File is not properly formatted (gofumpt)
```

Once fixed, and the docs are updated (see https://github.com/testcontainers/testcontainers-go/pull/2954#discussion_r2054327097 and https://github.com/testcontainers/testcontainers-go/pull/2954#discussion_r2054338308), I think we are good to merge.

mdelapenya (dismissed):
I'm approving this PR already, although there are some comments to be addressed before we merge it:

- documentation adding the WithData options and the ConnectionString container method
- Steven's suggestion regarding wrapping errors.

Other than that, LGTM, thanks so much for your patience during the review, we do appreciate your hard work with Testcontainers and the Firebase emulator. 🙇

modules/firebase/ports.go
```diff
@@ -0,0 +1,20 @@
...
+	"context"
+	"fmt"
+
+	"github.com/docker/go-connections/nat"
+)
+
+func (c *Container) ConnectionString(ctx context.Context, portName nat.Port) (string, error) {
+	host, err := c.Host(ctx)
+	if err != nil {
+		return "", err
```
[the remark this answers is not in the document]
mdelapenya: @xytis friendly ping! 🙏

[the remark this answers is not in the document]
xytis: I'll get to this once I have some time :)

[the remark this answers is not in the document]
mdelapenya: I'm cutting a release today. Do you mind if, as this PR was sent from your main branch and I cannot contribute to it, I close this PR and submit another PR with your commits and my suggestions on top? Then I can incorporate this new module in today's release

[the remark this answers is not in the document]
xytis: Oh, sure. Do as you wish :)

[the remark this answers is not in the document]
xytis: Honestly, I am fine if you took the code and rebased into whatever location you need it to be :)

Just note that the issue which was blocking this module is still here:

I can not copy directories into container, if the source dir name and the dest dir name are not matching.

There is a test/example for that, but I placed an inverse condition for some reason.

I'll revert it now, so it would be clear why this is lagging for so long.

modules/firebase/ports.go
```diff
@@ -0,0 +1,20 @@
...
+	}
+	port, err := c.MappedPort(ctx, portName)
+	if err != nil {
+		return "", err
```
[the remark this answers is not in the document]
mdelapenya: Same here 🙏

modules/firebase/ports.go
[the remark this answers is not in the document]
mdelapenya: Indeed, that limitation exists at the moment. My question now would be: why would the users of the firebase module like to provide a wrong dir name? I can see two types of users:
1. developers of the firebase emulator, who would need different failure modes while developing it
2. developers using the emulator, who would not need that particular failure mode.

I guess developers in group 1 would love to have multiple root dirs with different use cases. As a _workaround_ I can think of a previous step that copies those dirs to a (t.TempDir() + "/firebase"), and then copy that temporary dir to the container. We can add _now_ an issue to testcontainers to resolve the issue as soon as possible and remove that workaround right after it's released.

Of course, as a firebase noob, I'm probably missing other concerns you may have clear. Please let me know what you think about this 🙏

modules/firebase/ports.go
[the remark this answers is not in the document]
xytis: This one is fixed too.

modules/firebase/ports.go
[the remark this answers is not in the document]
xytis: The only reason why someone would have issues with directory name is when you have different setups for different tests.

In that case, the workaround seems to be a good solution.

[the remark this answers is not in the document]
mdelapenya: Let's do this: let's the functional option do that automatically, storing the temp root dir in the container so that we add a new Terminate function that: 1) removes the temp dir, and 2) calls container.Terminate. Please honor the Terminate signature 

@stevenh @xytis thoughts?

modules/firebase/config.go
```diff
@@ -0,0 +1,87 @@
...
+		PostgresPort uint16 `json:"postgresPort,omitempty"`
+		DataDir      string `json:"dataDir,omitempty"`
+	} `json:"dataconnect,omitempty"`
+
+	Tasks struct {
+		Host string `json:"host,omitempty"`
+		Port uint16 `json:"port,omitempty"`
+	} `json:"tasks,omitempty"`
+}
+type partialFirebaseConfig struct {
```
stevenh, line 85: nit: add a blank line between types, suspect it will fail linting without.

modules/firebase/examples_test.go
```diff
@@ -0,0 +1,40 @@
...
+
+	"github.com/testcontainers/testcontainers-go"
+	"github.com/testcontainers/testcontainers-go/modules/firebase"
+)
+
+func ExampleRun() {
+	ctx := context.Background()
+
+	firebaseContainer, err := firebase.Run(ctx, "ghcr.io/u-health/docker-firebase-emulator:13.29.2",
+		firebase.WithRoot(filepath.Join("testdata", "firebase")),
```
stevenh, line 17: info: using literal "testdata/firebase" works fine on all platforms as even though windows uses `\` it also accepts `/`.

modules/firebase/firebase.go
```diff
@@ -0,0 +1,127 @@
...
+
+		hostF := emulator.FieldByName("Host")
+		portF := emulator.FieldByName("Port")
+		websocketPortF := emulator.FieldByName("WebsocketPort")
+
+		if hostF != (reflect.Value{}) && !hostF.IsZero() {
+			host := hostF.String()
+			if host != "0.0.0.0" {
+				return nil, fmt.Errorf("config specified %s emulator host on non public ip: %s", name, host)
+			}
```
stevenh, line 52: question: this seems to limit host to 0.0.0.0, if so why have the field? that said I'm not sure this check is needed, do you have an example of why it is?

modules/firebase/firebase.go
```diff
@@ -0,0 +1,127 @@
...
+		enabledF := emulator.FieldByName("Enabled")
+		if enabledF != (reflect.Value{}) && !enabledF.Bool() {
+			continue
+		}
+
+		hostF := emulator.FieldByName("Host")
+		portF := emulator.FieldByName("Port")
+		websocketPortF := emulator.FieldByName("WebsocketPort")
+
+		if hostF != (reflect.Value{}) && !hostF.IsZero() {
```
stevenh, line 48: suggestion: use IsValid to simplify test, more below
```suggestion
		if hostF.isValid() && !hostF.IsZero() {
```

modules/firebase/firebase.go
```diff
@@ -0,0 +1,127 @@
...
+func gatherPorts(config partialFirebaseConfig) ([]string, error) {
+	var ports []string
+
+	v := reflect.ValueOf(config.Emulators)
+	for i := 0; i < v.NumField(); i++ {
+		emulator := v.Field(i)
+		if emulator.Kind() != reflect.Struct {
+			continue
+		}
+		name := v.Type().Field(i).Name
```
stevenh, line 37: suggestion: extract type lookup out of the loop as its invariant.

modules/firebase/firebase.go
```diff
@@ -0,0 +1,127 @@
...
+
+	bytes, err := io.ReadAll(cfg)
+	if err != nil {
+		return nil, fmt.Errorf("read firebase.json: %w", err)
+	}
+
+	var parsed partialFirebaseConfig
+	if err := json.Unmarshal(bytes, &parsed); err != nil {
+		return nil, fmt.Errorf("parse firebase.json: %w", err)
+	}
```
stevenh, line 108: suggestion: use a stream reader to be more memory efficient, its a small benefit but good practice.

modules/firebase/firebase.go
```diff
@@ -0,0 +1,127 @@
...
+
+	expectedExposedPorts, err := gatherPorts(parsed)
+	if err != nil {
+		return nil, fmt.Errorf("gather ports: %w", err)
+	}
+	req.ExposedPorts = expectedExposedPorts
```
stevenh, line 114: question: should we wait for all ports instead of the fragile log wait?

modules/firebase/ports.go
```diff
@@ -0,0 +1,20 @@
...
+func (c *Container) ConnectionString(ctx context.Context, portName nat.Port) (string, error) {
+	host, err := c.Host(ctx)
+	if err != nil {
+		return "", fmt.Errorf("host: %w", err)
+	}
+	port, err := c.MappedPort(ctx, portName)
+	if err != nil {
+		return "", fmt.Errorf("mapped port: %w", err)
+	}
+	return fmt.Sprintf("%s:%s", host, port.Port()), nil
```
stevenh, line 19: nit: blank lines after error returns help users grok the code better
```suggestion
	if err != nil {
		return "", fmt.Errorf("host: %w", err)
	}

	port, err := c.MappedPort(ctx, portName)
	if err != nil {
		return "", fmt.Errorf("mapped port: %w", err)
	}

	return fmt.Sprintf("%s:%s", host, port.Port()), nil
```

modules/firebase/firebase_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+	ctr, err := firebase.Run(ctx, "ghcr.io/u-health/docker-firebase-emulator:13.29.2",
+		firebase.WithRoot(filepath.Join("testdata", "failure")),
+	)
+	// In this case, the file gets copied over at /srv/failure (instead of /srv/firebase)
+	// and this stops working.
+	// What would be a solution here? Previously I just added a check that the root must
+	// end in "/firebase"... I could do the same.
+	testcontainers.CleanupContainer(t, ctr)
+	require.NoError(t, err)
```
stevenh, line 64: bug: this should fail, as I understand it this is down to how our copy works or doesn't. I have an unfinished PR in this [area here](https://github.com/testcontainers/testcontainers-go/pull/2782).

I would change this to check for err as expected and mark the test as skip for now.

stevenh (changes requested):
Just a quick pass as I'm busy today.

xytis:
Ah, cool. I'll adapt to that container once I have some free time.

mdelapenya:
Fantastic! thanks for your time @xytis , we appreciate it!

PR #2960, opened 2025-02-01
chore!: enable var-naming from revive

mmorel-35:
#### What does this PR do?

enable [var-naming](https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#var-naming) rule from revive

This also moves output configuration from Makefile to golangci-lint config

#### BREAKING CHANGES
⚠️ There are naming changes for constants and functions

### DIFF
--- a/docs/modules/grafana-lgtm.md
+++ b/docs/modules/grafana-lgtm.md
@@ -67,7 +67,7 @@ The Grafana LGTM container exposes the following methods:
 
 - Since testcontainers-go <a href="https://github.com/testcontainers/testcontainers-go/releases/tag/v0.33.0"><span class="tc-version">:material-tag: v0.33.0</span></a>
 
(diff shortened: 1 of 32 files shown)


stevenh (changes requested):
Quite a few breaking changes here @mdelapenya due to public method renaming, thoughts?

mdelapenya (commented):
I'm tempted to block this PR until we release a v1. Almost all the modules will be affected, impacting lots of potential users of the library.

We already have some BC in the current release, some more in the previous one, but they were very narrowed. Here we are not offering a deprecation path for the renamed methods.

Wdyt?

mdelapenya:
Thanks for understanding, and your work!, @mmorel-35, it's a pleasure working with engineers like you.

If we can do that, first tackling the private fields, then we can move on with an another round of changes for the public API.

We can keep this PR as reference, in draft, and send different chunks. Thoughts?

PR #2962, opened 2025-02-03
fix(valkey): fix port race

stevenh:
Fix port race on start up of valkey containers by waiting for the port instead of the log entry.

### DIFF
--- a/modules/valkey/valkey.go
+++ b/modules/valkey/valkey.go
@@ -57,7 +57,7 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
 	req := testcontainers.ContainerRequest{
 		Image:        img,
 		ExposedPorts: []string{"6379/tcp"},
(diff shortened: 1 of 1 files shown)


modules/valkey/valkey.go
```diff
@@ -57,7 +57,7 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
 	req := testcontainers.ContainerRequest{
 		Image:        img,
 		ExposedPorts: []string{"6379/tcp"},
-		WaitingFor:   wait.ForLog("* Ready to accept connections"),
+		WaitingFor:   wait.ForListeningPort("6379/tcp"),
```
mdelapenya, line 60: question: do you think we could keep both conditions, the log entry and the listening port?

stevenh: That would only add more work, so keeping it simple would be my preference.

PR #2967, opened 2025-02-04
chore!: remove variadic arguments from nats ConnectionString

### LINKED ISSUE #2951: NATS: remove variadic argument for the connection string
I like this PR because it's pointing to bad designs on our side. E.g. this function should not receive the variadic argument at all, or we could check here if the NATS query string supports parameters that could be appended, demonstrating we have a bug here.

_Originally posted by @mdelapenya in https://github.com/testcontainers/testcontainers-go/pull/2949#discussion_r1931626901_

jasonyunicorn:
## What does this PR do?

Removes the (unnecessary) variadic arguments from `nats.ConnectionString` and `nats.MustConnectionString` since the arguments are not used.

## Why is it important?

Removing unused variadic arguments improves code clarity and prevents misleading usage expectations.

## Related issues

- Closes #2951


mdelapenya (commented):
LGTM, although there is a reference to the variation arguments in the docs/modules/nats.md file. Could you update that 🙏?

mdelapenya (approved):

PR #2974, opened 2025-02-05
fix(ci): use same condition for sonar steps

mdelapenya:
## What does this PR do?
It uses the same filter for the sonar steps in the CI workflow.

## Why is it important?
If artifact name is not properly generated, then the sonarcloud files won't be uploaded correctly, as the artifact name will be empty, causing a collision.

## Related issues

- Caused by #2876


mdelapenya:
Merging, as it will fix the ongoing PRs

PR #2978, opened 2025-02-06
chore: enable var-naming from revive (private vars only)

mmorel-35:
#### What does this PR do?

enable [var-naming](https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#var-naming) rule from revive

This also moves output configuration from Makefile to golangci-lint config

⚠️ Only private fields are fixed, the other are annotated with `//nolint:revive //FIXME` to be fixed later as agreed on https://github.com/testcontainers/testcontainers-go/pull/2960#issuecomment-2640515755

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -71,11 +71,18 @@ linters-settings:
       - name: use-any
       - name: var-declaration
       - name: var-naming
-        disabled: true
+        arguments:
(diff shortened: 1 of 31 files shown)


mdelapenya (approved):
LGTM, thanks for adding the `// FIXME` comment 🙏

PR #2979, opened 2025-02-07
feat!: log package for consistent output

KenxinKun:
## What does this PR do?

Implements solution for https://github.com/testcontainers/testcontainers-go/discussions/2977

## Why is it important?

Prepares the codebase for moving the logger to a new client, moving the interface and implementations behind `testcontainers/log`. It also silences some of the identified leaking logs in the `wait` package.

## How does it fix the issue?

- Moves the `Logger` interface and implementations to its own package `testcontainers/log`
- Adds methods to retrieve (`log.Default()`) or set (`log.SetDefault(logger)`) the default logger
- Re-uses the default logger in the `wait` package which was using the standard `log` package previously and leaking logs to STDOUT


logconsumer_test.go
```diff
@@ -523,10 +532,10 @@ func (l *bufLogger) String() string {
...
 	// Capture global logger.
 	logger := &bufLogger{}
-	Logger = logger
-	oldLogger := Logger
+	oldLogger := logging.Logger
+	SetLogger(logger)
 	t.Cleanup(func() {
-		Logger = oldLogger
+		logging.Logger = oldLogger
 	})
```
KenxinKun, line 539: Please check, it felt like the order of operations was wrong here (original first patched the logger, then it stored it as oldLogger, with the clean up not really changing anything?

container.go
```diff
@@ -469,7 +470,7 @@ func (c *ContainerRequest) BuildOptions() (types.ImageBuildOptions, error) {
 		}
 
 		if modifiedTag != tag {
-			Logger.Printf("✍🏼 Replacing image with %s. From: %s to %s\n", is.Description(), tag, modifiedTag)
+			log.Default().Printf("✍🏼 Replacing image with %s. From: %s to %s\n", is.Description(), tag, modifiedTag)
```
stevenh, line 473: suggestion: Expose a `Printf` in the `log` package so it can have cleaner consumers:
```suggestion
			log.Printf("✍🏼 Replacing image with %s. From: %s to %s\n", is.Description(), tag, modifiedTag)
```

More below

log/log.go
```diff
@@ -0,0 +1,88 @@
...
+	"strings"
+	"testing"
+
+	"github.com/docker/docker/client"
+)
+
+// Logger defines the Logger interface
+type Logger interface {
+	Printf(format string, v ...any)
+	Print(v ...any)
```
stevenh, line 16: question: does this need to include `Print`?

KenxinKun: Strictly no, it was sort of replacing some `Println` uses in `wait` but that can also be achieved with `Printf`. Happy to remove so the interface remains the same that it was, so it doesn't break any existing user implementations.

container.go
KenxinKun: Will do!

docker.go
```diff
@@ -817,7 +818,7 @@ func (c *DockerContainer) copyLogsTimeout(stdout, stderr io.Writer, options *con
 		// Timeout or client connection closed, retry.
 	default:
 		// Unexpected error, retry.
-		Logger.Printf("Unexpected error reading logs: %v", err)
+		log.Printf("Unexpected error reading logs: %v", err)
```
stevenh, line 821: bug: not new but this should use c.logger
```suggestion
		c.logger.Printf("Unexpected error reading logs: %v", err)
```

log/log.go
```diff
@@ -0,0 +1,71 @@
+package log
+
+import (
+	"log"
+	"os"
+	"strings"
+	"testing"
+)
+
+// Logger defines the Logger interface
```
stevenh: nit: could we make the doc comments end with a full stop, more below.

log/log.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func Printf(format string, v ...any) {
+	defaultLogger.Printf(format, v...)
+}
+
+// Validate our types implement the required interfaces.
+var (
+	_ Logger = (*log.Logger)(nil)
+)
```
stevenh: suggestion: can we move this to the top of the file add add one for each logger implementation

stevenh (changes requested):
Looking good, just a few little bits.

stevenh (dismissed):
Changes LGTM.

As this is a breaking change lets add the ! to title and footer as per https://www.conventionalcommits.org/en/v1.0.0/

As it is a fix, I would suggestion we capture the what in the title and then in expand with how / why in the description.

wait/host_port_test.go
```diff
@@ -454,11 +455,15 @@ func TestHostPortStrategySucceedsGivenShellIsNotInstalled(t *testing.T) {
...
 		WithPollInterval(100 * time.Millisecond)
 
-	oldWriter := log.Default().Writer()
-	var buf bytes.Buffer
-	log.Default().SetOutput(&buf)
+	oldLogger := log.Default()
+
+	buf := &bytes.Buffer{}
+	logger := standardlog.New(buf, "test", standardlog.LstdFlags)
+	logger.SetOutput(buf)
```
stevenh, line 462: suggestion: you can just declare and no need to set the output that was done by the New, more below.
```suggestion
	var buf bytes.Buffer
	logger := standardlog.New(&buf, "test", standardlog.LstdFlags)
```

stevenh (changes requested):
Just a little fix for log capture in the wait tests.

wait/host_port_test.go
KenxinKun: Is this right? the `buf` variable is already a pointer, I don't think we need the extra `&`?

wait/host_port.go
```diff
@@ -161,10 +162,10 @@ func (hp *HostPortStrategy) WaitUntilReady(ctx context.Context, target StrategyT
 	if err = internalCheck(ctx, internalPort, target); err != nil {
 		switch {
 		case errors.Is(err, errShellNotExecutable):
-			log.Println("Shell not executable in container, only external port validated")
+			log.Printf("Shell not executable in container, only external port validated")
```
mdelapenya, line 165: question: do we want to remove the `\n` here?

wait/host_port_test.go
```diff
@@ -15,6 +15,7 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/testcontainers/testcontainers-go/exec"
+	"github.com/testcontainers/testcontainers-go/log"
```
mdelapenya, line 18: question: I usually alias the testcontainers packages with the `tc` prefix instead: `tclog "github.com/.../log"`, so the stdlib `log` would remain the same.

BUT I'm not sure if this idiomatic and should be done exactly the other way around (as you suggested: library packages go first).

What is the idiomatic way of doing this?

mdelapenya (commented):
Added a few comments, no blockers

wait/host_port_test.go
KenxinKun: Just pushed the fix

modules/neo4j/neo4j.go
```diff
@@ -73,7 +74,7 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
 
 	genericContainerReq := testcontainers.GenericContainerRequest{
 		ContainerRequest: request,
-		Logger:           testcontainers.Logger,
+		Logger:           log.Default(),
```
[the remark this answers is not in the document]
stevenh: That would be a good, removes a bunch of boiler plate.

wait/host_port.go
stevenh: Println isn't part of the interface and since log inserts a new line anyway it works.

wait/host_port_test.go
stevenh: I only do it if there is a conflict.

mdelapenya: Indeed, just in case of conflicts, as in this file. Else, keep it without the alias

KenxinKun: fix pushed :)

stevenh (changes requested):
just a typo and the request from @mdelapenya about the default logger.

modules/neo4j/config.go
```diff
@@ -93,7 +94,7 @@ func WithNeo4jSettings(settings map[string]string) testcontainers.CustomizeReque
 
 // WithLogger sets a custom logger to be used by the container
 // Consider calling this before other "With functions" as these may generate logs
-func WithLogger(logger testcontainers.Logging) testcontainers.CustomizeRequestOption {
+func WithLogger(logger log.Logger) testcontainers.CustomizeRequestOption {
```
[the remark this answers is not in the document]
stevenh: Yer mark as deprecated and remove any uses.

stevenh (dismissed):

mdelapenya (approved):
LGTM, thank you so much for your work here. 🙇 

@KenxinKun @stevenh I added df2324e to avoid having you working more in this. It's basically adding the new `log` package to one missing markdown file.

We are merging this PR when the CI passes 🚀 

Cheers!

mdelapenya:
BTW @KenxinKun @stevenh wdyt if we change the title of this PR to communicate better we are refactoring the log infra? Something like `feat(log)!: add log package to stop logging debug messages to stdout`, but open to any suggestion 🙏

stevenh:
> BTW @KenxinKun @stevenh wdyt if we change the title of this PR to communicate better we are refactoring the log infra? Something like `feat(log)!: add log package to stop logging debug messages to stdout`, but open to any suggestion 🙏

As this is a new package, I'd go with:
feat!: log package for consistent output

And then expand on the body.

The reason to not include log in the scope is it doesn't exist yet (updates would), and for description if wraps in github its too long, so trying to be concise 😄

KenxinKun:
@mdelapenya @stevenh I synced the fork and update this branch you may need to kick CI again, happy with suggestions above

PR #2984, opened 2025-02-15
fix(mongodb): replica set initialization & connection handling

ttruongatl:
## What does this PR do?

This PR addresses critical issues in the MongoDB container module regarding replica set initialization, host IP detection, and connection string formation. In addition, it expands test coverage by verifying change stream functionality when a replica set is configured.

1. **Replica Set Initialization Reliability:**  
   The previous implementation initialized the replica set using the container’s internal IP and a hardcoded port (27017). This approach caused issues when custom host port mappings were applied (e.g., mapping container port 27017 to host port 52120) because the advertised address was incorrect. Furthermore, the old code did not reliably wait for MongoDB to be fully ready before issuing the `rs.initiate` command, leading to intermittent failures that went undetected by tests.  
   _This PR introduces a `waitForMongoReady` helper, which waits for MongoDB’s readiness (using a ping command) before retrieving the host and mapped port and issuing the replica set initialization command with the correct values._

2. **Accurate Host IP Detection:**  
   Previously, using Unix sockets defaulted to "localhost," which often broke replica set functionality since that address isn't externally reachable. This PR introduces a helper, getLocalNonLoopbackIP, to obtain a non-loopback IPv4 address—and if that fails, it falls back to "host.docker.internal"—ensuring the connection string advertises an externally accessible IP.

3. **Enhanced Connection String Formation:**  
   The `ConnectionString` method has been updated to dynamically build the connection URI. When a replica set is configured, it now appends the appropriate query parameter (e.g., `?replicaSet=rs0`) so that clients can connect properly. Previously, the tests did not verify this because the connection string did not accurately reflect the replica set configuration.

4. **Expanded Testing with Change Streams:**  
   The test suite has been extended to include change stream tests for containers configured with a replica set. If the connection string contains a replica set parameter, the tests now create a change stream on a collection, insert a document, and verify that the change event reflects an insert operation. This ensures that the replica set functionality is not only initialized correctly but also operational in terms of change stream processing.

## Why is it important?

Without these enhancements, the replica set configuration might not work correctly—especially in CI or Docker host environments where custom port mappings are common. The shortcomings of the old code meant that, even though tests passed, the replica set functionality (and thus features like change streams) was not truly exercised. This PR improves reliability and test coverage, ensuring that integration tests can now fully validate replica set behavior.

## How to test this PR:

```sh
git clone git@github.com:ttruongatl/testcontainers-go.git
cd modules/mongodb
make test
```


modules/mongodb/mongodb.go
```diff
@@ -125,10 +125,19 @@ func (c *MongoDBContainer) ConnectionString(ctx context.Context) (string, error)
 	if err != nil {
 		return "", err
 	}
+
```
stevenh, line 128: suggestion: this could fail depending on the values, lets change it to use url / net functions to ensure the value is valid, for example:
```go
	u := url.URL{
		Scheme: "mongodb",
		Host:   net.JoinHostPort(host, port),
	}

	if c.username != "" || c.password != "" {
		u.User = url.UserPassword(c.username, c.password)
	}

	if c.replicaSet != "" {
	        v := url.Values{}
		v.Add("replicaSet", c.replicaSet)
	        u.RawQuery = v.Encode()
	}

        return  u.String(), nil
```

modules/mongodb/mongodb.go
```diff
@@ -176,17 +185,27 @@ func initiateReplicaSet(req *testcontainers.GenericContainerRequest, cli mongoCl
 		req.LifecycleHooks, testcontainers.ContainerLifecycleHooks{
 			PostStarts: []testcontainers.ContainerHook{
 				func(ctx context.Context, c testcontainers.Container) error {
-					ip, err := c.ContainerIP(ctx)
+					// Wait for MongoDB to be ready
+					if err := waitForMongoReady(ctx, c, cli); err != nil {
+						return fmt.Errorf("failed to wait for MongoDB ready: %w", err)
```
stevenh, line 190: suggestion: "failed" prefix causes hard to read errors when wrapped. General rule is to describe what failed and then report the error for example:
```suggestion
						return fmt.Errorf("wait for mongo: %w", err)
```

More below.

modules/mongodb/mongodb_test.go
```diff
@@ -13,7 +17,48 @@ import (
 	"github.com/testcontainers/testcontainers-go/modules/mongodb"
 )
 
+func getLocalNonLoopbackIP() (string, error) {
+	interfaces, err := net.Interfaces()
+	if err != nil {
+		return "", err
```
stevenh, line 23: suggestion: wrap the error.

modules/mongodb/mongodb_test.go
```diff
@@ -13,7 +17,48 @@ import (
 	"github.com/testcontainers/testcontainers-go/modules/mongodb"
 )
 
+func getLocalNonLoopbackIP() (string, error) {
```
stevenh, line 20: suggestion: remove get prefix as it's idiomatic in go to not use get / set prefix unless required.

modules/mongodb/mongodb_test.go
```diff
@@ -13,7 +17,48 @@ import (
...
+		if err != nil {
+			continue // try next interface
+		}
+		for _, addr := range addrs {
+			var ip net.IP
+			switch v := addr.(type) {
+			case *net.IPNet:
+				ip = v.IP
+			case *net.IPAddr:
+				ip = v.IP
```
stevenh, line 40: suggestion: add default case which continues, simplifying the test below.

modules/mongodb/mongodb_test.go
```diff
@@ -125,18 +170,53 @@ func TestMongoDB(t *testing.T) {
...
+			}
 		})
 	}
 }
+
+// hasReplicaSet checks if the connection string includes a replicaSet query parameter.
+func hasReplicaSet(connStr string) (bool, error) {
+	u, err := url.Parse(connStr)
+	if err != nil {
+		return false, err
```
stevenh, line 217: suggestion: wrap the error.

modules/mongodb/mongodb_test.go
```diff
@@ -125,18 +170,53 @@ func TestMongoDB(t *testing.T) {
...
+	}
+	q := u.Query()
+	_, ok := q["replicaSet"]
```
stevenh, line 220: question: would a blank replica set also be no replica set?

stevenh (changes requested):
Thanks for the PR, here's some initial feedback.

modules/mongodb/mongodb_test.go
ttruongatl: Good question! With the updated code, a blank replica set (e.g., replicaSet=) will be considered as no replica set. The function now checks for a non-empty value associated with the replicaSet parameter.

modules/mongodb/mongodb.go
ttruongatl: Updated the code by following the suggestion.

modules/mongodb/mongodb.go
ttruongatl: Removed "failed" prefix in wrapped errors.

modules/mongodb/mongodb_test.go
ttruongatl: Updated.

modules/mongodb/mongodb_test.go
ttruongatl: Updated.

modules/mongodb/mongodb_test.go
```diff
@@ -125,18 +173,63 @@ func TestMongoDB(t *testing.T) {
...
+	if !ok {
+		return false, nil
+	}
+	// Check if any of the replica set values are non-empty.
+	for _, value := range replicaSetValues {
+		if value != "" {
+			return true, nil // Found a non-empty replicaSet value
+		}
+	}
+	return false, nil // replicaSet parameter exists, but all values are empty
```
stevenh, line 234: suggestion: while query params can be duplicated replicaSet can't so this can't so this can be simplifed
```suggestion
	return q.Get("replicaSet") == ""
```

stevenh (changes requested):
Thanks for the updates, just done another quick pass.

modules/mongodb/mongodb.go
```diff
@@ -125,10 +127,23 @@ func (c *MongoDBContainer) ConnectionString(ctx context.Context) (string, error)
...
 	if c.username != "" && c.password != "" {
-		return fmt.Sprintf("mongodb://%s:%s@%s:%s", c.username, c.password, host, port.Port()), nil
+		u.User = url.UserPassword(c.username, c.password)
+	}
+
+	if c.replicaSet != "" {
+		q := url.Values{}
+		q.Add("replicaSet", c.replicaSet)
+		u.RawQuery = q.Encode()
+		u.Path = "/"
```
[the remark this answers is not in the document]
ttruongatl: - I investigated the issue and found that when the replica set is specified, omitting the slash results in a URI like `mongodb://192.168.86.62:61879?replicaSet=rs`, which triggers the error **“must have a / before the query ?”**. 
- Adding u.Path = "/" changes the URI to mongodb://192.168.86.62:61879/?replicaSet=rs, which is valid. 
- In scenarios without a replica set, the absence of the slash doesn’t cause any problems, but including it universally doesn’t hurt.
- Therefore, I’ve updated the code to always set the path to “/” (as your suggestion) to ensure the connection string is valid in all cases.

mdelapenya (commented):
I just left a few non-blocking comments. Will approve once discussed, thanks!

modules/mongodb/mongodb_test.go
```diff
@@ -13,7 +17,50 @@ import (
 	"github.com/testcontainers/testcontainers-go/modules/mongodb"
 )
 
+func localNonLoopbackIP() (string, error) {
+	interfaces, err := net.Interfaces()
+	if err != nil {
```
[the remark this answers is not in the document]
ttruongatl: Sure, I can make the change. I’m wondering if there’s a configuration in `golangci-lint` that can automatically detect coding style issues.

modules/mongodb/mongodb_test.go
```diff
@@ -13,7 +17,50 @@ import (
...
+			return ip.String(), nil
+		}
+	}
+	return "", errors.New("no non-loopback IPv4 address found")
+}
+
 func TestMongoDB(t *testing.T) {
+	host, err := localNonLoopbackIP()
+	if err != nil {
+		host = "host.docker.internal"
```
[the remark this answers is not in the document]
ttruongatl: From the document:  
[Docker Networking - Connecting from a container to a service on the host](https://docs.docker.com/desktop/features/networking/#i-want-to-connect-from-a-container-to-a-service-on-the-host),  

> "The host has a changing IP address, or none if you have no network access. We recommend that you connect to the special DNS name `host.docker.internal`, which resolves to the internal IP address used by the host."

I understand that `host.docker.internal` serves as a reliable fallback since it resolves to the host's internal IP address. However, I'm open to any insights or suggestions from experts in the repo regarding its behavior across different container runtimes.

modules/mongodb/mongodb_test.go
```diff
@@ -13,7 +17,50 @@ import (
 	"github.com/testcontainers/testcontainers-go/modules/mongodb"
 )
 
+func localNonLoopbackIP() (string, error) {
```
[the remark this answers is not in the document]
mdelapenya: I saw this code being not specific to mongo, but thinking about it twice, keeping it internal to the module is fine until we have another use case. Thanks for the clarification

stevenh (changes requested):
Sorry a little miss-understanding on a previous suggestion, should be good to go after this.

stevenh (approved):
Looks good to me

stevenh:
@ttruongatl looks like the file needs a format, likely just a spacing change from one of my suggestions.

ttruongatl:
Fixed the lint check issues.

stevenh (dismissed):

modules/mongodb/mongodb_test.go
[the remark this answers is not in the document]
mdelapenya: Thanks for checking. I asked @eddumelendez about this, and he already mentioned the concerns we could face adding this code, but for the kafka module: https://github.com/testcontainers/testcontainers-go/pull/2920#issuecomment-2581330637.

With that in mind, I think we should avoid using `host.docker.internal` for the reasons explained there.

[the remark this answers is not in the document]
ttruongatl: Thank you for helping me understand the concerns, @mdelapenya. I understand that `host.docker.internal` isn't recommended based on the Kafka module discussion (#2920). Since it's Docker-specific, what would you suggest as a more compatible approach across different container runtimes? Let me know your thoughts!

modules/mongodb/mongodb.go
```diff
@@ -176,17 +191,27 @@ func initiateReplicaSet(req *testcontainers.GenericContainerRequest, cli mongoCl
...
+						return fmt.Errorf("get mapped port: %w", err)
 					}
 
 					cmd := cli.eval(
-						"rs.initiate({ _id: '%s', members: [ { _id: 0, host: '%s:27017' } ] })",
+						"rs.initiate({ _id: '%s', members: [ { _id: 0, host: '%s:%s' } ] })",
 						replSetName,
-						ip,
+						host,
+						mappedPort.Port(),
```
mdelapenya, line 213: question: why do we need the mapped port here? The command mongoCli's eval command executes **inside the container**, so the listening port must be 27017 instead of the exposed port in the host.

I tried locally using the old cli command, and removing the code in the tests that calculated the network interfaces, setting the `TESTCONTAINERS_HOST_OVERRIDE` env var, and all the tests pass again.

mdelapenya: @ttruongatl I still see this error when trying to push to your branch:

```
fatal: The upstream branch of your current branch does not match
the name of your current branch.  To push to the upstream branch
on the remote, use

    git push upstream HEAD:refs/pull/2984/head

To push to the branch of the same name on the remote, use

    git push upstream HEAD

To avoid automatically configuring an upstream branch when its name
won't match the local branch, see option 'simple' of branch.autoSetupMerge
in 'git help config'.
```

I'm attaching a PATCH file so you can add it to your PR.

[0001-chore-simplify-tests.patch](https://github.com/user-attachments/files/19433212/0001-chore-simplify-tests.patch)

Cheers!

modules/mongodb/mongodb.go
```diff
@@ -208,3 +222,7 @@ func withAuthReplicaset(
 		return nil
 	}
 }
+
+func waitForMongoReady(ctx context.Context, c testcontainers.Container, cli mongoCli) error {
```
mdelapenya, line 226: BTW, I noticed that this function is unused now. Could you please remove it, or verify it's not needed? 🙏

mdelapenya (approved):
LGTM, thanks!

PR #2988, opened 2025-02-17
feat: add WithScripts and GetSQLCmdPath functions

MattiasMTS:
## What does this PR do?
Add some helper functions to the MS SQL Server module to easier bootstrap some ddl sql scripts etc. Also, helper function to get the path to the `sqlcmd` tool, see more [here](https://hub.docker.com/r/microsoft/mssql-server).

## Why is it important?
It will improve developer experience and follow the same conventions that exist for other modules, e.g. `mysql`. 

## How to test this PR
Run the `make test` in the folder and it should run. I am using `podman`, so I modify the containerprovider during that time but should work as is if you're using docker


mdelapenya (commented):
I left some comments that need to be addressed before accepting the PR.

It's also important to add the new option to the docs at `./docs/modules/mssql.md`, that will land into our docs here: https://golang.testcontainers.org/modules/mssql/

Other than that, the PR looks good, we just need to talk about my comments.

Thanks for your work and time here 🙇

MattiasMTS:
>It's also important to add the new option to the docs at ./docs/modules/mssql.md, that will land into our docs here: https://golang.testcontainers.org/modules/mssql/

Sure, I was a bit unsure on adding that myself or if it was generated. As for the structure, ok if I follow the previous conventions or do I need to add e.g. the `tag` of upcoming release this feature will be included in?

modules/mssql/mssql_test.go
```diff
@@ -128,3 +128,98 @@ func TestMSSQLServerWithInvalidPassword(t *testing.T) {
...
+		got = append(got, p)
+	}
+
+	require.EqualValues(t, want, got)
+}
+
+// tests that the correct path to the sqlcmd executable is returned based on the image tag
+func TestGetSQLCmdPath(t *testing.T) {
+	tests := []struct {
+		name  string
```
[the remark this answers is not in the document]
MattiasMTS: Alright, sure I can rename. I usually follow this guide https://github.com/uber-go/guide/blob/master/style.md#test-tables from uber on how this should be named. But nice with the copy&paste part, I'll take that with me 😄

modules/mssql/mssql.go
```diff
@@ -41,6 +43,46 @@ func WithPassword(password string) testcontainers.CustomizeRequestOption {
...
 
+// WithScripts adds SQL scripts to be executed after the container is ready.
+// The scripts are executed in the order they are provided using
+// the sqlcmd utility.
+func WithScripts(scripts ...string) testcontainers.CustomizeRequestOption {
+	return func(req *testcontainers.GenericContainerRequest) error {
+		password := defaultPassword
+		if req.Env["MSSQL_SA_PASSWORD"] != "" {
+			password = req.Env["MSSQL_SA_PASSWORD"]
+		}
```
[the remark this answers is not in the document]
MattiasMTS: nice catch, very much a possible bug caught. I'll refactor it

modules/mssql/mssql.go
```diff
@@ -41,6 +43,45 @@ func WithPassword(password string) testcontainers.CustomizeRequestOption {
...
+// The scripts are executed in the order they are provided using sqlcmd tool.
+func WithScripts(scripts ...string) testcontainers.CustomizeRequestOption {
+	return func(req *testcontainers.GenericContainerRequest) error {
+		hooks := make([]testcontainers.ContainerHook, 0, len(scripts))
+		for _, script := range scripts {
+			hook := func(ctx context.Context, c testcontainers.Container) error {
+				password := defaultPassword
+				if req.Env["MSSQL_SA_PASSWORD"] != "" {
+					password = req.Env["MSSQL_SA_PASSWORD"]
+				}
```
stevenh, line 56: suggestion: as this is loop invariant, extract it out.

modules/mssql/mssql.go
```diff
@@ -41,6 +43,45 @@ func WithPassword(password string) testcontainers.CustomizeRequestOption {
 	}
 }
 
+// WithScripts adds SQL scripts to be executed after the container is ready.
+// The scripts are executed in the order they are provided using sqlcmd tool.
+func WithScripts(scripts ...string) testcontainers.CustomizeRequestOption {
```
stevenh, line 48: question: in other models we've been moving towards `io.Reader` as the source for things like this, as it provides more flexibility. Would that work here?

modules/mssql/mssql.go
```diff
@@ -41,6 +43,45 @@ func WithPassword(password string) testcontainers.CustomizeRequestOption {
...
+				cmd := testcontainers.NewRawCommand([]string{
+					GetSQLCmdPath(req.Image),
+					"-S", "localhost",
+					"-U", defaultUsername,
+					"-P", password,
+					"-No",
+					"-i", targetPath,
+				})
+				_, _, err := c.Exec(ctx, cmd.AsCommand(), cmd.Options()...)
+				return err
```
stevenh, line 72: suggestion: wrap the error so its clear about the cause
```suggestion
				if _, _, err := c.Exec(ctx, cmd.AsCommand(), cmd.Options()...); err != nil {
					return fmt.Errorf("exec script %q: %w", script, err)
				}
				return nil
```

modules/mssql/mssql.go
```diff
@@ -41,6 +43,45 @@ func WithPassword(password string) testcontainers.CustomizeRequestOption {
...
+		for _, script := range scripts {
+			hook := func(ctx context.Context, c testcontainers.Container) error {
+				password := defaultPassword
+				if req.Env["MSSQL_SA_PASSWORD"] != "" {
+					password = req.Env["MSSQL_SA_PASSWORD"]
+				}
+
+				targetPath := "/tmp/" + filepath.Base(script)
+				if err := c.CopyFileToContainer(ctx, script, targetPath, 0o644); err != nil {
+					return err
```
stevenh, line 60: suggestion: wrap the error to provide context.

modules/mssql/mssql.go
```diff
@@ -99,3 +140,25 @@ func (c *MSSQLServerContainer) ConnectionString(ctx context.Context, args ...str
 
 	return connStr, nil
 }
+
+// GetSQLCmdPath helper function to return the
+// sqlcmd path based on the image version
+func GetSQLCmdPath(image string) string {
```
stevenh, line 146: suggestion: keep this private unless there's a good reason not to, as it reduces the maintainable API surface area.

modules/mssql/mssql.go
```diff
@@ -99,3 +140,25 @@ func (c *MSSQLServerContainer) ConnectionString(ctx context.Context, args ...str
...
+	const sqlCmd, sqlCmdOld = "/opt/mssql-tools18/bin/sqlcmd", "/opt/mssql-tools/bin/sqlcmd"
```
stevenh, line 147: suggestion: lets add some comments here about what it's trying to achieve as it's pretty hard to follow the logic, some examples would be good, as well as tests for this function.

modules/mssql/mssql_test.go
```diff
@@ -128,3 +128,125 @@ func TestMSSQLServerWithInvalidPassword(t *testing.T) {
 	testcontainers.CleanupContainer(t, ctr)
 	require.NoError(t, err)
 }
+
+// tests that a container can be created with a DDL script
+func TestMSSQLServerWithScriptsDDL(t *testing.T) {
+	const password = "MyCustom@Passw0rd"
+	seedFile := filepath.Join("testdata", "seed.sql")
+
+	tests := []struct {
```
stevenh, line 137: suggestion: use t.Run instead of test table as that allows specific tests to be run from editors and also reduces the amount of code.

```golang
t.Run("password-then-scripts", func(t *testing.T) {
      testWithScripts(t, mssql.WithScripts(seedFile), mssql.WithPassword(password))
}

t.Run("scripts-then-password", func(t *testing.T) {
      testWithScripts(t, mssql.WithScripts(seedFile), mssql.WithPassword(password))
}

...

// testWithScripts tests the behaviour of WithScripts
func testScripts(t *testing.T, options ...testcontainer.ContainerCustomiser) {
     t.Helper()

    // Content for t.Run from the for loop goes here.
}
```

modules/mssql/mssql_test.go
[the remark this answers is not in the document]
stevenh: We're actually moving away from test tables as they require more code and don't allow individual tests to be run directly from editors like VS Code.

stevenh (changes requested):
Thanks for the updates, here's some more feedback based on the latest version

modules/mssql/mssql.go
MattiasMTS: that's what I asked in the description. What's your preference here rather than saying we move it to internal 😋

modules/mssql/mssql.go
MattiasMTS: What's unclear what this function aims to?

- we have docstrings explaining what it does
- the function itself is small and contains reasonable naming of variables etc
- we have tests already showing how this is used

Do you want an example in the docstring? I appreciate the review but please try to make them more expressive what you want to do instead of just making them 'nit'

modules/mssql/mssql.go
MattiasMTS: sure it might work, but I see still `string` being used in other modules. You decide...

modules/mssql/mssql.go
MattiasMTS: this error is prob explaining itself, no need to do so.

modules/mssql/mssql.go
MattiasMTS: if we want it private, where should it be placed best in order to add unit test for it too?

mdelapenya:
> or do I need to add e.g. the `tag` of upcoming release this feature will be included in?

Yes, please use the following block:

```md
```

modules/mssql/mssql.go
mdelapenya: @stevenh is right. We are trying to make these kind of options more portable thanks to readers. String path are already there, although it's on our plans to start converting all of them into readers, or at least offer the two choices. So for new options like this one, I'd go with Readers first, and if you want to add paths, then add it making sure they call the reader version.

MattiasMTS: alright, didn't know this was the preferred way. I'll refactor it

modules/mssql/mssql.go
mdelapenya: @MattiasMTS we try to follow conventional comments, https://conventionalcomments.org/, with the sole goal of avoiding unclear messages from our side. But we sometimes commit mistakes and are not that clear. Sorry if it caused confusion, our intention is always to remove ambiguity following those conventions, so we'll try to be more explicit/expressive next time.

Our goal as maintainers is to create an healthy OSS community, not only in the code we produce/accept but also in the interactions we participate in.

modules/mssql/mssql.go
MattiasMTS: refactored.

modules/mssql/mssql.go
stevenh: To clarify my comment was about what it's actually achieving internally (so not a doc comment) looking at this and not knowing how `mssql` images are named I found it hard to follow the logic, so not what it's trying to do but how it's doing it, if we need to look into a logic error in the future.

So what I was looking for was examples container image XYZ -> mssql-tools18 (new style) but ABC -> mssql-tools (old style)?

That being said, I just thought, why not add both to PATH environment variable and then exec an unpathed `sqlcmd`.

modules/mssql/mssql.go
stevenh: Sorry @MattiasMTS I didn't spot that question, if there's no need for others to consume then I would make it private.

If the current test set is using a different module name so can't access I would just add another file which can with an `_internal` suffix.

That said, see comment about using PATH environment instead to eliminate the need for it.

modules/mssql/mssql.go
MattiasMTS: sure, domain knowledge is tricky. I added in the description of the PR link to the docs on when they did this change etc.

We can sure mix with `PATH` etc but I'd prefer simplicity over that as of now. Given, the only way you're prob going to interact with the tool is by e.g. setting a variable from this function and then re-use that in the code. I'd buy that we should set it to `PATH` if we'd attach to the container etc and do stuff.

All and all, this is mainly an utility function that facilitates `WithScripts` function. Happy to put it as not an exposed function but then I need to understand where it is best to put it in order to add unit test for it (without scope creeping the scope of this PR) 😄

MattiasMTS: I'll make it internal and do as per your suggestion above @stevenh .

MattiasMTS: do you still want that test to be tdt or not?

modules/mssql/mssql.go
mdelapenya: I think the message `This is used internally by the `WithScripts` option but can also be used directly if needed.` from https://github.com/testcontainers/testcontainers-go/pull/2988/files#diff-1261868bc467efdc41a1d4c94027d9befaa46eb533242a0800737fdd182289a9R72 makes it clear that it must be a private helper in the module. Let's make it private, and make it public when/if needed

MattiasMTS:
Well, it's as they say
image
thanks for the comments @mdelapenya  and @stevenh. Should be ready for somewhat of a final review.

Do you want to extend the scope of `WithScripts` to support other types of executable, e.g. `*.sh` etc. I see that is quite typical based on previous modules.

stevenh (changes requested):
Thanks for the quick updates, think we're getting close.

Did you see the replies to other threads about using PATH instead of parsing custom image?

modules/mssql/mssql_test.go
```diff
@@ -128,3 +129,122 @@ func TestMSSQLServerWithInvalidPassword(t *testing.T) {
 	testcontainers.CleanupContainer(t, ctr)
 	require.NoError(t, err)
 }
+
+// tests that a container can be created with a DDL script
+func TestMSSQLServerWithScriptsDDL(t *testing.T) {
+	const password = "MyCustom@Passw0rd"
+	seedFilePath := filepath.Join("testdata", "seed.sql")
```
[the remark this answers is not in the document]
MattiasMTS: that is nifty, learned something new today. Thanks!

modules/mssql/mssql_test.go
```diff
@@ -128,3 +129,76 @@ func TestMSSQLServerWithInvalidPassword(t *testing.T) {
...
+			var p Pizza
+			err := rows.Scan(&p.ID, &p.ToppingName, &p.Deliciousness)
+			require.NoError(t, err)
+			got = append(got, p)
+		}
+
+		require.EqualValues(t, want, got)
+	}
+
+	ctx := context.Background()
```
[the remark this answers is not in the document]
MattiasMTS: couldn't make this work, could you elaborate a bit more?

MattiasMTS:
Renamed it to `WithInitSQL` and also updated the code to use `PATH` as you suggested, which reduced a bunch of code which is nice.

modules/mssql/mssql_test.go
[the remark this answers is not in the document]
MattiasMTS: ah, solved by bumping my nixpkg to go_1_24, was still on the old one.

stevenh (dismissed):
Thanks for all your work on this, hope you managed to take away some learnings as I did.

stevenh:
@MattiasMTS could you update the PR title and description to match the code please

stevenh:
@mdelapenya my suggestion to include `t.Context()` fails because its 1.24 we should consider bumping to 1.24 instead of 1.23 to take advantage of that.

In the mean time @MattiasMTS feel free to change that back to `context.Background()` so we don't delay this PR any further.

MattiasMTS:
Done. I also added a test of an old image to verify the `PATH` of sqlcmd works backwards compatible.

mdelapenya:
> we should consider bumping to 1.24 instead of 1.23 to take advantage of that.

@stevenh we usually develop the library following Go's release policy: https://golang.testcontainers.org/system_requirements/#go-version, which means we always develop the library in the lower supported version (1.23 at the moment)

PR #2992, opened 2025-02-20
feat: add ScyllaDB module

mdelapenya:
## What does this PR do?
This PR adds ScyllaDB as a module, written on top of #2919 (it was submitted from the main branch, so I cannot add commits on top, that's why I'm submitting a new PR).

## Why is it important?
New modules, cool!

## Related issues

- Supersedes and closes #2919


mdelapenya:
@danielhe4rt I took your original branch and added the commits from the review process we did on #2919 on top of yours.

I'd appreciate a final review from you here 🙏

modules/scylladb/scylladb.go
```diff
@@ -0,0 +1,159 @@
...
+			"--alternator-write-isolation": "always",
+		})
+
+		return nil
+	}
+}
+
+// WithCustomCommands sets custom commands with  values for the ScyllaDB container.
+// This is an option to overwrite the default commands with a custom one.
+// See more [here](https://opensource.docs.scylladb.com/stable/operating-scylla/procedures/tips/best-practices-scylla-on-docker.html)
```
stevenh, line 68: bug: this invalid formatting, see https://tip.golang.org/doc/comment links section

modules/scylladb/scylladb.go
```diff
@@ -0,0 +1,159 @@
...
+		setCommandFlag(req, map[string]string{
+			"--alternator-port":            alternatorPortStr,
+			"--alternator-write-isolation": "always",
+		})
+
+		return nil
+	}
+}
+
+// WithCustomCommands sets custom commands with  values for the ScyllaDB container.
```
stevenh, line 66: suggestion: remove the double space
```suggestion
// WithCustomCommands sets custom commands with values for the ScyllaDB container.
```

modules/scylladb/scylladb.go
```diff
@@ -0,0 +1,159 @@
...
+// This is an option to overwrite the default commands with a custom one.
+// See more [here](https://opensource.docs.scylladb.com/stable/operating-scylla/procedures/tips/best-practices-scylla-on-docker.html)
+func WithCustomCommands(cmds map[string]string) testcontainers.CustomizeRequestOption {
```
stevenh, line 69: suggestion: just pass a variadic and do the processing in setCommandFlag as that will also allow singles to be passed e.g. `--my-flag`

modules/scylladb/scylladb.go
```diff
@@ -0,0 +1,159 @@
...
+// and obtaining the host and exposed port from the container
+// Eg: "host:port" -> "localhost:9042" -> "localhost:19042" -> "localhost:8000"
+func (c Container) ConnectionHost(ctx context.Context, port uint16) (string, error) {
+	host, err := c.Host(ctx)
+	if err != nil {
+		return "", fmt.Errorf("host: %w", err)
+	}
+
+	portStr := strconv.FormatInt(int64(port), 10)
+	containerPort, err := c.MappedPort(ctx, nat.Port(portStr))
```
stevenh, line 86: bug: nat.Port expects to be port/proto suggest using NewPort with a protocol.

modules/scylladb/scylladb.go
```diff
@@ -0,0 +1,159 @@
...
+	return func(req *testcontainers.GenericContainerRequest) error {
+		setCommandFlag(req, cmds)
+		return nil
+	}
+}
+
+// ConnectionHost returns the host and port of the Scylladb container with the given port
+// and obtaining the host and exposed port from the container
+// Eg: "host:port" -> "localhost:9042" -> "localhost:19042" -> "localhost:8000"
+func (c Container) ConnectionHost(ctx context.Context, port uint16) (string, error) {
```
stevenh, line 79: question: as the ports can vary would be better to provide the distinct methods for each type of port so the user doesn't have to keep track of them?

modules/scylladb/scylladb.go
```diff
@@ -0,0 +1,159 @@
...
+		return "", fmt.Errorf("host: %w", err)
+	}
+
+	portStr := strconv.FormatInt(int64(port), 10)
+	containerPort, err := c.MappedPort(ctx, nat.Port(portStr))
+	if err != nil {
+		return "", fmt.Errorf("mapped port: %w", err)
+	}
+
+	return host + ":" + containerPort.Port(), nil
```
stevenh, line 91: bug: use `net.JoinHostPort`

modules/scylladb/scylladb.go
```diff
@@ -0,0 +1,159 @@
...
+		if !ok {
+			cmds = append(cmds, cmd)
+		} else {
+			cmds = append(cmds, fmt.Sprintf("%s=%s", seps[0], val))
+			delete(flag, seps[0])
+		}
+	}
+
+	for key, val := range flag {
+		cmds = append(cmds, fmt.Sprintf("%s=%s", key, val))
```
stevenh, line 155: suggestion: no need for sprintf here just add
```suggestion
		cmds = append(cmds, key+"="+val)
```

modules/scylladb/scylladb.go
```diff
@@ -0,0 +1,159 @@
...
+	cmds := make([]string, 0, len(req.Cmd)+len(flag))
+
+	for _, cmd := range req.Cmd {
+		seps := strings.SplitN(cmd, "=", 1)
+		val, ok := flag[seps[0]]
+
+		if !ok {
+			cmds = append(cmds, cmd)
+		} else {
+			cmds = append(cmds, fmt.Sprintf("%s=%s", seps[0], val))
```
stevenh, line 149: suggestion: no need for sprintf
```suggestion
			cmds = append(cmds, seps[0]+"="+val)
```

modules/scylladb/scylladb_test.go
```diff
@@ -0,0 +1,195 @@
...
+func TestScyllaWithConfig(t *testing.T) {
+	ctx := context.Background()
+
+	ctr, err := scylladb.Run(ctx,
+		"scylladb/scylla:6.2",
+		scylladb.WithConfig(bytes.NewReader(scyllaYaml)),
+		scylladb.WithShardAwareness(),
+	)
+	require.NoError(t, err)
+	testcontainers.CleanupContainer(t, ctr)
```
stevenh, line 76: bug: order is reversed.

modules/scylladb/scylladb_test.go
```diff
@@ -0,0 +1,195 @@
...
+
+	alternatorPort := uint16(8000)
+
+	t.Run("test-with-alternator", func(t *testing.T) {
+		ctr, err := scylladb.Run(ctx,
+			"scylladb/scylla:6.2.2",
+			scylladb.WithAlternator(alternatorPort),
+		)
+		require.NoError(t, err)
+		testcontainers.CleanupContainer(t, ctr)
```
stevenh, line 124: bug: order is reversed.

modules/scylladb/scylladb_test.go
```diff
@@ -0,0 +1,195 @@
...
+		require.NoError(t, err)
+		requireCreateTable(t, client)
+	})
+
+	t.Run("test-without-alternator", func(t *testing.T) {
+		ctr, err := scylladb.Run(ctx,
+			"scylladb/scylla:6.2",
+		)
+		require.NoError(t, err)
+		testcontainers.CleanupContainer(t, ctr)
```
stevenh, line 136: bug: order it reversed.

modules/scylladb/scylladb_test.go
```diff
@@ -0,0 +1,195 @@
...
+	}, nil
+}
+
+func getDynamoAlternatorClient(t *testing.T, c *scylladb.Container, port uint16) (*dynamodb.Client, error) {
+	t.Helper()
+	var errs []error
+
+	hostPort, err := c.ConnectionHost(context.Background(), port)
+	if err != nil {
+		errs = append(errs, fmt.Errorf("get connection string: %w", err))
```
stevenh, line 159: suggestion: describe the method
```suggestion
		errs = append(errs, fmt.Errorf("connection host: %w", err))
```

modules/scylladb/scylladb_test.go
```diff
@@ -0,0 +1,195 @@
...
+		URI: url.URL{Host: r.HostPort, Scheme: "http"},
+	}, nil
+}
+
+func getDynamoAlternatorClient(t *testing.T, c *scylladb.Container, port uint16) (*dynamodb.Client, error) {
+	t.Helper()
+	var errs []error
+
+	hostPort, err := c.ConnectionHost(context.Background(), port)
+	if err != nil {
```
stevenh, line 158: question: is there any reason not to just fail the test here so use (more below)?
```go
require.NoError(t, err)
```

modules/scylladb/scylladb_test.go
```diff
@@ -0,0 +1,195 @@
...
+	HostPort string
+}
+
+func (r *scyllaAlternatorResolver) ResolveEndpoint(_ context.Context, _ dynamodb.EndpointParameters) (smithyendpoints.Endpoint, error) {
+	return smithyendpoints.Endpoint{
+		URI: url.URL{Host: r.HostPort, Scheme: "http"},
+	}, nil
+}
+
+func getDynamoAlternatorClient(t *testing.T, c *scylladb.Container, port uint16) (*dynamodb.Client, error) {
```
stevenh, line 153: suggestion: add a doc comment.

modules/scylladb/scylladb_test.go
```diff
@@ -0,0 +1,195 @@
...
+		)
+		require.NoError(t, err)
+		testcontainers.CleanupContainer(t, ctr)
+
+		_, err = getDynamoAlternatorClient(t, ctr, alternatorPort)
+		require.Error(t, err)
+	})
+}
+
+type scyllaAlternatorResolver struct {
```
stevenh, line 143: suggestion: add a doc comment

stevenh (changes requested):
Done an initial pass, skipping examples as depending on the results they might need changing.

modules/scylladb/scylladb_test.go
mdelapenya: Indeed, this caused me confusion too, as I saw assertions performed even after an error. I think we can simply check the errors, and do not execute code after them

mdelapenya: There is a test that uses the alternator return value and check that it's an error

modules/scylladb/scylladb.go
mdelapenya: I like the idea of simplifying the usage of both ports. I'm adding both methods

mdelapenya:
@stevenh I think I addressed all your comments 🙏

modules/scylladb/scylladb.go
```diff
@@ -0,0 +1,214 @@
...
+
+	"github.com/docker/go-connections/nat"
+
+	"github.com/testcontainers/testcontainers-go"
+	"github.com/testcontainers/testcontainers-go/wait"
+)
+
+const (
+	port           = "9042/tcp"
+	shardAwarePort = "19042/tcp"
```
mdelapenya, line 19: @danielhe4rt in the Java implementation, the alternator port is a constant set to `8000`. In Go, we allow customising this port. We'd like to align both implementations, so I'd like you to help us understand the rationale behind making it fixed Vs dynamic. 🙏

mdelapenya: @eddumelendez I noticed the SSL implementation in Java is hardcoding the paths for the certificates and trustStore: https://github.com/testcontainers/testcontainers-java/blob/5af66d700e942474afabd8e550bf089196fdb5f9/modules/scylladb/src/main/java/org/testcontainers/scylladb/ScyllaDBContainer.java#L77-L83 But those values are set in a test file, the scylladb config: https://github.com/testcontainers/testcontainers-java/blob/5af66d700e942474afabd8e550bf089196fdb5f9/modules/scylladb/src/test/resources/scylla-test-ssl/scylla.yaml#L463-L468

If users have different paths set in their test files, then they cannot update them in the code. Is this a potential bug?

mdelapenya:
@eddumelendez regarding the SSL example, I'd make progress merging the module as soon it's ready, and contribute the SSL support as a new feature. Wdyt?

modules/scylladb/scylladb.go
```diff
@@ -0,0 +1,214 @@
...
+			flagsMap[before] = after
+		} else {
+			flagsMap[flag] = ""
+		}
+	}
+
+	return func(req *testcontainers.GenericContainerRequest) error {
+		if errInvalidFlag != nil {
+			return errInvalidFlag
+		}
```
stevenh, line 95: suggestion: merge the two blocks to simplify the logic
```suggestion
	return func(req *testcontainers.GenericContainerRequest) error {
		flagsMap := make(map[string]string, len(flags))
		for _, flag := range flags {
			if !strings.HasPrefix(flag, "--") && !strings.HasPrefix(flag, "-") {
				return fmt.Errorf("invalid flag: %s", flag)
			}

			before, after, found := strings.Cut(flag, "=")
			if found {
				flagsMap[before] = after
			} else {
				flagsMap[flag] = ""
			}
		}
```

modules/scylladb/scylladb.go
```diff
@@ -0,0 +1,214 @@
...
+		Image:        img,
+		ExposedPorts: []string{port},
+		Cmd: []string{
+			"--developer-mode=1",
+			"--overprovisioned=1",
+			"--smp=1",
+			"--memory=512M",
+		},
+		WaitingFor: wait.ForAll(
+			wait.ForLog(".*initialization completed..*").AsRegexp(),
```
stevenh, line 146: question: should this be removed, as log checks are fragile and I would expect the other checks to do the job?

modules/scylladb/scylladb_test.go
```diff
@@ -0,0 +1,288 @@
...
+			},
+		},
+		BillingMode: types.BillingModePayPerRequest,
+	})
+	require.NoError(t, err)
+}
+
+func TestWithCustomCommands(t *testing.T) {
+	t.Run("invalid-flag", func(t *testing.T) {
+		t.Parallel()
```
stevenh, line 202: question: looks like there's quite a bit of memory in use if these all run in parallel, could that be counter productive

stevenh (commented):
Looking good, a few questions and one suggestion

modules/scylladb/scylladb.go
mdelapenya: I kept it because it was added by the author (devrel at scylladb), and it's also present in the Java impl.

I agree that new modules should not rely on the log if they provide a more robust strategy, which is the case. I'm removing it as you suggest

modules/scylladb/scylladb_test.go
mdelapenya: Indeed, the container request is big in memory. I added the parallel execution because they are unit tests that happens really fast even without parallel execution, so I guess it's fine to remove that

modules/scylladb/scylladb.go
mdelapenya: @danielhe4rt @eddumelendez I'm taking the decision of pinning the alternator port to the 8000 default port, not allowing its customisation. This is how the Java implementation is done, so I'm aligning with that.

PLMK if you have any concern about it, specially if you prefer it to be configurable (as it was in this current state)

eddumelendez: About the port, there is no specific reason. IIRC I used 8000 because is the one used in scylladb docs. Regarding to the certificates and truststore, I think there is one issue with it now because we are forcing format instead to be open.

eddumelendez (approved):

PR #2994, opened 2025-02-23
chore: enable empty-block from revive

mmorel-35:
#### What does this PR do?

enable [empty-block](https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#empty-block) rule from revive


mdelapenya (approved):
LGTM, thanks!

PR #2995, opened 2025-02-24
chore!: extract image-related code to the top-level image package

mdelapenya:
- **chore: extract dockerignore to the image package**
- **chore: extract image list to the image package**
- **chore: extract save images to the image package**
- **chore: extract image substitutors to the image package**
- **chore: extract image build to the image package**

## What does this PR do?
This PR refactors the code to extract the docker image related code to a top-level package: `image`.

This package will be responsible of:
- pulling images
- saving images as tar files
- parsing dockerignore files for image build
- listing images
- calculating image substitutors
- building images

The related tests have been moved to the package, but there are "e2e" tests (they are actually using a ContainerRequest) that have been kept in the root package of the library.

## Why is it important?
Separation of concerns, reducing the API surface in the root package.


stevenh (changes requested):
I like the idea of a smaller package with more focused responsibilities.

I've done a quick pass to capture some observations, however I wonder for if the current pattern of passing in a type which implements an interface hides the important elements making it harder to understand the logic and creating a hard dependency on external types such as `types.ImageBuildOptions`.

Thinking from first principles for a minute, would it be better if we made the image follow the our new best practices. For example have Build method take the required arguments + functional options for example:
```go
func Build(ctx context.Context, context io.Reader, options ... Option) (string, error) {
...
}
```

That way we can avoid hard dependences for consumers on external types and provide a cleaner more maintainable package going forward, thoughts?

PR #2996, opened 2025-02-25
feat: client proof of concept

stevenh:
This is a client proof of concept which contains basic config support alongside an example method BuildImage.


internal/client/client.go
```diff
@@ -0,0 +1,118 @@
...
+	client *client.Client
+	cfg    *config
+	err    error
+}
+
+// ClientOption is a type that represents an option for configuring a client.
+type ClientOption func(*Client) error
+
+// Logger returns a client option that sets the logger for the client.
+func Logger(log slog.Logger) ClientOption {
```
mdelapenya, line 46: question: are we going to use `WithXXX` options or `XXX` (like here)?

PR #2997, opened 2025-02-26
chore: readd dependabot, including a way to refresh the project files for all the modules

mdelapenya:
- **chore: bring back dependabot generation**
- **chore: replace assert with require**
- **feat: add a modulegen command to refresh the project files**
- **chore: exclude compose from the mkdocs nav entries**
- **fix: remove extra comma after refresh**
- **chore: readd dependabot updates**
- **fix: lint**
- **chore: wrap errors in dependabot files**
- **docs: document new types and functions**
- **chore: simplify types merging generator and refresher**
- **chore: include modulegen in dependabot**
- **chore: check modulegen is included in the dependabot updates**
- **docs: document new command**
- **chore: exclude dependabot changes from the build pipeline**

## What does this PR do?
This PR enables dependabot for all the modules in the project so that they receive updates when needed.

It also adds the automation for creating dependabot updates for the new modules added with the code generation tool, and a new `refresh` command to refresh all the project files:
- dependabot.yml
- mkdocs.yml
- sonar.properties
- vscode project file

As a consequence, the unit tests for the modulegen module has been updated (simplified?) making them not affecting the current workspace: for that, the project and module files will be copied to a temp dir, and the test context will happen there, not adding dirty to the current workspace.

Finally, the dependabot file has been excluded from the automation to detect changed, so that a change on it does not include all the modules in the CI pipeline.

## Why is it important?
Given the CI pipeline is now more efficient, only running the modified modules, now it's possible to handle dependabot updates not consuming all the GHA runners at a time. Instead just 4 runners will be used for testing the changes: detect + lint + testx2 + end, which is way faster than before, where the entire pipeline had to be built (~100 workers), adding flakiness.


scripts/changed-modules.sh
```diff
@@ -74,7 +74,7 @@ readonly ROOT_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)
 readonly excluded_modules=(".devcontainer" ".vscode" "docs")
 
 # define an array of files that won't be included in the list
-readonly excluded_files=("mkdocs.yml" "sonar-project.properties")
+readonly excluded_files=("mkdocs.yml" ".github/dependabot.yml" "sonar-project.properties")
```
mdelapenya, line 77: Because this file is updated, this PR will run all the modules, which makes total sense.

mdelapenya:
@mmorel-35 as contributor of the modulegen, would you like to take a look at this? 🙏

modulegen/internal/dependabot/main.go
```diff
@@ -0,0 +1,69 @@
...
+// AddModule update dependabot with the new module
+func (g Generator) AddModule(ctx context.Context, tcModule context.TestcontainersModule) error {
+	configFile := ctx.DependabotConfigFile()
+
+	config, err := readConfig(configFile)
+	if err != nil {
+		return fmt.Errorf("read config: %w", err)
+	}
+
+	if err := config.addUpdate("/" + tcModule.ParentDir() + "/" + tcModule.Lower()); err != nil {
```
stevenh, line 21: question: should this be filepath.Join?

modulegen/internal/dependabot/writer.go
```diff
@@ -0,0 +1,23 @@
...
+func writeConfig(configFile string, config *Config) error {
+	err := os.MkdirAll(filepath.Dir(configFile), 0o755)
+	if err != nil {
+		return fmt.Errorf("create directory: %w", err)
+	}
+	data, err := yaml.Marshal(config)
+	if err != nil {
+		return fmt.Errorf("marshal config: %w", err)
+	}
+	header := "# This file is autogenerated by the 'modulegen' tool.\n# Please look at the generator code when updating it.\n"
```
stevenh, line 20: suggestion: use the standard header format detailed [here](https://github.com/golang/go/issues/13560#issuecomment-288457920)

modulegen/internal/main.go
```diff
@@ -85,8 +125,17 @@ func GenerateFiles(ctx context.Context, tcModule context.TestcontainersModule) e
 		sonar.Generator{},  // update sonar-project.properties
 	}
 
+	examples, err := ctx.GetExamples()
+	if err != nil {
+		return err
```
stevenh, line 130: suggestion: wrap the error

modulegen/internal/main.go
```diff
@@ -85,8 +125,17 @@ func GenerateFiles(ctx context.Context, tcModule context.TestcontainersModule) e
...
+	}
+	modules, err := ctx.GetModules()
+	if err != nil {
+		return err
```
stevenh, line 134: suggestion: wrap the error

modulegen/internal/main.go
```diff
@@ -85,8 +125,17 @@ func GenerateFiles(ctx context.Context, tcModule context.TestcontainersModule) e
...
+	}
+
 	for _, generator := range projectGenerators {
-		err := generator.Generate(ctx)
+		err := generator.Generate(ctx, examples, modules)
```
stevenh, line 138: suggestion: wrap the error while here

modulegen/internal/mkdocs/main.go
```diff
@@ -31,16 +34,48 @@ func (g Generator) AddModule(ctx context.Context, tcModule context.Testcontainer
...
+// excluding compose as it has its own page in the docs.
+func (g Generator) Generate(ctx context.Context, examples []string, modules []string) error {
+	configFile := ctx.MkdocsConfigFile()
+	config, err := ReadConfig(configFile)
+	if err != nil {
+		return fmt.Errorf("read config: %w", err)
+	}
+
+	for _, module := range modules {
+		if module == "compose" {
```
stevenh, line 53: suggestion: document why we're skipping the compose module

stevenh (changes requested):
Loving all the additional docs and general clean ups. Just a few suggestions.

modulegen/internal/dependabot/main.go
mdelapenya: Indeed, I'm changing it now, although the first element must be the `filepath.Separator`(see dependabot.yaml, the directories for an update)

modulegen/internal/dependabot/writer.go
mdelapenya: But the generated file in this case it not a Go file, but the dependabot.yml descriptor. We should use the proper comment character.

modulegen/internal/mkdocs/main.go
```diff
@@ -26,8 +26,8 @@ func (g Generator) AddModule(ctx context.Context, tcModule context.Testcontainer
 	if err != nil {
 		return fmt.Errorf("generate md file: %w", err)
 	}
-	moduleMd := tcModule.ParentDir() + "/" + tcModule.Lower() + ".md"
-	indexMd := tcModule.ParentDir() + "/index.md"
+	moduleMd := filepath.Join(tcModule.ParentDir(), tcModule.Lower()) + ".md"
```
stevenh, line 29: suggestion: add the extension inside the join, for consistency with other.s
```suggestion
	moduleMd := filepath.Join(tcModule.ParentDir(), tcModule.Lower() + ".md")
```

modulegen/internal/vscode/types.go
```diff
@@ -19,26 +20,26 @@ func newConfig(examples []string, modules []string) *Config {
 	config := Config{
 		Folders: []Folder{
 			{
-				Path: "../modulegen",
+				Path: filepath.Join("..", "modulegen"),
```
stevenh, line 23: suggestion: I really wouldn't make these changes, all OSs handle / separated relative paths and it just makes it harder to grok.

modulegen/internal/dependabot/main.go
```diff
@@ -0,0 +1,70 @@
...
+func (g Generator) Generate(ctx context.Context, examples []string, modules []string) error {
+	configFile := ctx.DependabotConfigFile()
+
+	config, err := readConfig(configFile)
+	if err != nil {
+		return fmt.Errorf("read config: %w", err)
+	}
+
+	for _, example := range examples {
+		if err := config.addUpdate(filepath.Join(string(filepath.Separator), "examples", example)); err != nil {
```
stevenh, line 39: question: this looks like its creating "/examples" or "\examples" which would break in windows if used what's the intent here?

modulegen/dependabot_test.go
```diff
@@ -0,0 +1,89 @@
...
+)
+
+// isWindows returns if the current OS is Windows. For that it checks the GOOS environment variable or the runtime.GOOS constant.
+func isWindows() bool {
+	return os.Getenv("GOOS") == "windows" || runtime.GOOS == "windows"
+}
+
+func TestGetDependabotConfigFile(t *testing.T) {
+	if isWindows() {
+		t.Skip("skipping on windows because the dependabot config file uses the '/' separator")
```
stevenh, line 24: question: not sure I understand, what's the issue here?

I'm concerned that if the tests don't work the the generation wouldn't work either.

stevenh (commented):
Some questions before I continue on review.

modulegen/internal/dependabot/main.go
mdelapenya: Mmm thinking carefully, dependabot file will always exist under our GitHub workers, so no need to run tests with different file separators, as they will always be linux

modulegen/dependabot_test.go
mdelapenya: I'm reverting it, so that the tests will check the unix-separator. My intention was to not run the tests for our files, which are sanity checks for the format of the current state of the dependabot file.

mdelapenya: Test passing even on Windows 😉

modulegen/internal/dependabot/writer.go
stevenh: There might be an added benefit, would need to check, but editors and github understood this header when I last checked, although that might just be for go files as you mentioned.

stevenh (approved):
LGTM

PR #3007, opened 2025-03-05
chore(ci): use new GH action for SonarQube

mdelapenya:
- **choe(ci): use new sonarqube GH action**
- **revert: for testing on CI**

## What does this PR do?
It uses the new SonarQube GH action, as the old one was deprecated. It generates the sonar properties on the fly, removing the need of maintaining a properties file with all the modules. As a result, the code generation tool for the sonar properties is removed in this PR.

Also, the project in Sonar will be a monorepo, having as project key:
- core: `testcontainers_testcontainers-go` 
- modules: `testcontainers_testcontainers-go_modules_postgres`

and as project name:
- core: `testcontainers-go` 
- modules: `testcontainers-go-modules-postgres`

As a consequence, the sonar step will be executed per module, updating each project of the parent project with the given changes.

## Why is it important?
Sonar is failing since March 4th 2025. Besides that, simplifying the maintenance of modules and collecting sonar metrics per module, not per monorepo.


mdelapenya:
I'm going to merge this one and explore how it goes in the Sonarcloud instance.

PR #3009, opened 2025-03-07
feat(gcloud): add option to run firestore in datastore mode

Zarux:
## What does this PR do?

This PR adds an option to the gcloud module applicable only to `RunFirestore` that runs the emulator in `Firestore in Datastore mode`. 

It also changes the base example to point to the non-depracated function in order to create a new example with datastore mode

## Why is it important?

Datastore in GCP is now `Firestore in Datastore mode`, which means the datastore emulator behaves slightly different to what one would expect as it emulates the legacy datastore product.

In the gcloud cloud-sdk version 465.0.0 a parameter was added to run the firestore emulator in datastore mode which should behave more like the product in GCP.

## How to test this PR
The simplest way to test this would be to run the firestore emulator with the `gcloud.WithDatastoreMode()` and connect using the datastore package.
There is an example test case added in this PR as well that does this.


Zarux:
> Thanks for submitting this improving. I took a quick look and initially it LGTM.
> 
> I want to share that after #3008, I'm considering following the same pattern of one submodule per package, deprecating all the existing GCP containers. As contributor of this GCP module, does it reasonable to you?

My view is quite limited as this is my first contribution to this module :smile:  But at first glance it looks reasonable, yes :+1:

eddumelendez:
more general approach would be to allow adding more flags to the main command. See https://github.com/testcontainers/testcontainers-java/pull/10067 and https://github.com/testcontainers/testcontainers-node/pull/926

mdelapenya:
> more general approach would be to allow adding more flags to the main command. See [testcontainers/testcontainers-java#10067](https://github.com/testcontainers/testcontainers-java/pull/10067) and [testcontainers/testcontainers-node#926](https://github.com/testcontainers/testcontainers-node/pull/926)

We'd need to separate the containers in different modules, as with the current approach, all of them are sharing the flags config.

Zarux:
@mdelapenya Updated the pr to match the new structure. 

Had to change firestore/options slightly to allow non-shared options using embedding to allow extending it instead of just aliasing.
Let me know if you want to do it a different way

mdelapenya (commented):
LGTM, just one thing is missing: updating the `docs/modules/gcloud.md` file with the new option 🙏

docs/modules/gcloud.md
```diff
@@ -176,6 +176,12 @@ In example: `Run(context.Background(), "gcr.io/google.com/cloudsdktool/cloud-sdk
 
 {% include "./gcloud-shared.md" %}
 
+### Datastore mode
+
+Using the `WithDatastoreMode` option will run the Firestore emulator using `Firestore In Datastore` mode allowing you to use Datastore APIs and clients towards the Firestore emulator.
+
+Requires `cloud-sdk:465.0.0` or higher
```
mdelapenya, line 183: question: do you think we need to add a validation of the docker image in this case, or is it enough to let it fail?

mdelapenya (approved):
LGTM, just added a question regarding the Docker image version, but not a blocker, so I'll merge it once addressed.

Thanks for your hard work with testcontainers-go!

docs/modules/gcloud.md
Zarux: In my opinion it's enough to let it fail. The version requirement is well documented and 465 has been out for a while.

mdelapenya: Thanks for confirming, I'll let the CI to finish and then merge it

mdelapenya:
Oh wait! I forgot to mention that we added new functional options in the core, specifically one to contribute to the CMD: https://golang.testcontainers.org/modules/gcloud/#withcmdargs

Last minute change: could you please update the new option to internally call `testcontainers.WithCmdArgs` 🙏 ? This will also address @eddumelendez's concerns here: https://github.com/testcontainers/testcontainers-go/pull/3009#issuecomment-2741402542

Zarux:
> could you please update the new option to internally call testcontainers.WithCmdArgs 🙏 ?

I don't think that will work without a refactor 
1. The request command is defined after the options are applied. Changing this would require changing the behaviour of default options
2. The command running is not `gcloud`, but `bash` so any arguments added will not be added to the gcloud command, but the bash command and will not be applied correctly. This is why the extra options are just appended strings to the gcloud command

mdelapenya:
> > could you please update the new option to internally call testcontainers.WithCmdArgs 🙏 ?
> 
> I don't think that will work without a refactor
> 
> 1. The request command is defined after the options are applied. Changing this would require changing the behaviour of default options
> 2. The command running is not `gcloud`, but `bash` so any arguments added will not be added to the gcloud command, but the bash command. Thus they will not be applied correctly. This is why the extra options are just appended strings to the gcloud command

Ok, let's do this. Because this PR is in good shape, let's make progress and merge, and think about that refactor as a follow up. Will do a last pass of the review, and eventually merge, thank you so much for your contribution!

mdelapenya (approved):

docs/modules/gcloud.md
```diff
@@ -176,6 +176,12 @@ In example: `Run(context.Background(), "gcr.io/google.com/cloudsdktool/cloud-sdk
 
 {% include "./gcloud-shared.md" %}
 
+### Datastore mode
```
mdelapenya, line 179:
```suggestion
### Datastore mode

- Not available until the next release of testcontainers-go <a href="https://github.com/testcontainers/testcontainers-go"><span class="tc-version">:material-tag: main</span></a>

```

docs/modules/gcloud.md
```diff
@@ -176,6 +176,15 @@ In example: `Run(context.Background(), "gcr.io/google.com/cloudsdktool/cloud-sdk
...
+
+- Not available until the next release of testcontainers-go <a href="https://github.com/testcontainers/testcontainers-go"><span class="tc-version">:material-tag: main</span></a>
+
```
mdelapenya, line 182:
```suggestion
```

PR #3010, opened 2025-03-08
docs: correct container variable

tacerus:
## What does this PR do?

Repair a code example.

## Why is it important?

Because working examples make people happy.

## Related issues

n/a

n/a

n/a


mdelapenya (approved):
Indeed, we usually try to embed code snippets directly from the source code, to about what you wisely points out, and this one seems it's not using this approach, so thank you very much for fixing the docs. 

LGTM

PR #3012, opened 2025-03-10
docs: refine texts on how to set the module image

mdelapenya:
- **fix: remove outdated text from the module's docs template**
- **docs: fix the texts to set the container image in modules**
- **docs: use In example instead of E.g. in the image option**

## What does this PR do?
This PR refines two things:

1. the module template for module docs: removing the outdated message for the old `RunContainer` function, which is not generated anymore.
2. Change the text for the Image selection, as it was using the old message for using a different image, when in fact, it's the only way to specify the image.

## Why is it important?
Consistency in the docs, and also not generating outdated texts.


stevenh (approved):
LGTM

PR #3032, opened 2025-03-12
fix: skip nil strategies in wait.ForAll

mdelapenya:
## What does this PR do?
This PR skips nil wait strategies in case some downstream consumer uses the wait.ForAll without checking the validity of all the straegies in the list.

A clear example is building a list of strategies depending on certain conditions, and using forAll multiple times.

## Why is it important?
Avoid runtime panics


wait/all.go
```diff
@@ -62,6 +62,13 @@ func (ms *MultiStrategy) WaitUntilReady(ctx context.Context, target StrategyTarg
 	}
 
 	for _, strategy := range ms.Strategies {
+		if strategy == nil {
```
stevenh, line 65: bug: this won't detect a nil type which implements Strategy see https://go.dev/play/p/yz46T3tUXOR

wait/all_test.go
```diff
@@ -124,3 +124,27 @@ func TestMultiStrategy_WaitUntilReady(t *testing.T) {
 		})
 	}
 }
+
+func TestMultiStrategy_handleNils(t *testing.T) {
```
stevenh, line 128: bug: missing test for nil type which implements Strategy.

stevenh (changes requested):

stevenh (changes requested):
Looking good just the lint failures to fix.

stevenh (approved):

PR #3034, opened 2025-03-14
fix: pass secret to the reusable workflow

mdelapenya:
## What does this PR do?
Passes the secret to the reusable workflow.

## Why is it important?
GH actions needs to explicitly pass the secrets to reusable workflows, else the value is empty, causing the sonar step to fail, breaking the build.

## Related issues

- Completes #3007


mdelapenya:
merging, as tests are not affected

PR #3035, opened 2025-03-14
fix: use empty dir for the core

mdelapenya:
Use the empty path for the core library


mdelapenya (commented):
I created the project using the REST APIs: https://community.sonarsource.com/t/monorepo-how-to-add-subprojects-automatically-to-the-parent-project-when-a-build-is-triggered/136819/3

We must create the new project on Sonarcloud when a new project is created.

PR #3039, opened 2025-03-17
chore(ci): manage Sonar projects from Github actions

mdelapenya:
## What does this PR do?
This PR adds a shell script that uses the Sonarcloud APIs to manage the projects on it. With the script, we are able to:

- create one project
- delete one project
- update the main branch for the project (for some reason it creates it with master as main branch)
- createAll the projects (not the core)
- deleteAll the projects (not the core)

The script needs a SONAR_TOKEN env var, which in the CI will be provided as a secret.

The PR also adds three GH workflows to create, delete and perform bulk operations, using an input form with the module name for creating/deleting workflows, and for action to perform in the build one.

## Why is it important?
When a new module is created, we need to recreate it on Sonarcloud, and this is tricky to perform when a new module is added. In the meantime, when/if a new-module-detector is in place, we are doing this manually, so maintainers can go the GH actions and trigger a build for the given module.


mdelapenya:
Merging to enable the creation of project in Sonarcloud

PR #3044, opened 2025-03-18
feat: support for mounting images

mdelapenya:
## What does this PR do?
This PR updates the code for mounting images, creating a new mount type for Images, and adding the types and methods needed to implement it.

A new optional Validator interface has been added, in order to verify that the subpaths from the Image Mount are not absolute, which is a requirement from the Docker spec.

Unit tests have been added to cover the different use cases:
- creation of image mounts
- preparation of the mounts
- validation of the image mounts

A functional options to be consumed by modules has been added, so they can simply mount the file system of another image with that option. An example for the Ollama module has been added too: a first container loads an LLM, the container is committed to a different image, and a second container mounts that image, resulting in having the LLM already loaded.

It's IMPORTANT to notice that the image mount capabilities are only available if the underlying container runtime supports it. Docker v28 does it.

## Why is it important?
Consume new features from the Docker engine!

- 

## Follow-ups
I do not like the code for mounting volumes and tmpfs, as it's hard to follow, and in my opinion there are too many abstractions. So I'd advocate for starting a discussion on how to do a hard refactor for that part.


mounts.go
```diff
@@ -35,6 +40,9 @@ type ContainerMountSource interface {
 	// Type determines the final mount type
 	// possible options are limited by the Docker API
 	Type() MountType
+
+	// Validate validates the source of the mount
+	Validate() error
```
mdelapenya, line 45: I was forced to add this to have early feedback when preparing the mounts for a container

docker_mounts.go
```diff
@@ -85,6 +100,40 @@ func (s DockerTmpfsMountSource) GetTmpfsOptions() *mount.TmpfsOptions {
 	return s.TmpfsOptions
 }
 
+// DockerImageMountSource is a mount source for an image
+type DockerImageMountSource struct {
```
eddumelendez, line 104: what about just `ImageMountSource`?

mdelapenya: I used this for keeping consistency with the existing types. As mentioned in the follow-ups section:

> I do not like the code for mounting volumes and tmpfs, as it's hard to follow, and in my opinion there are too many abstractions. So I'd advocate for starting a discussion on how to do a hard refactor for that part.

So I'm open to start a rewrite of that layer

mounts.go
stevenh: suggestion: make is an optional interface so we don't have no-op implementations all over.

This would mean we get the following in `preCreateContainerHook`:
```go
	var mountErrors []error
	for _, m := range req.Mounts {
		if v, ok := m.source.(Validater); ok {
			if err := m.Source.Validate(); err != nil {
				mountErrors = append(mountErrors, err)
			}
		}
	}

	if len(mountErrors) > 0 {
		return errors.Join(mountErrors...)
	}
```

options_test.go
```diff
@@ -238,3 +238,52 @@ func TestWithHostPortAccess(t *testing.T) {
 		})
 	}
 }
+
+func TestWithImageMount(t *testing.T) {
```
stevenh, line 242: question: does this need a docker version check which calls https://pkg.go.dev/testing#T.Skip if its not high enough?

docker_mounts.go
stevenh: I'd go with the simple naming too.

stevenh: Agreed there must be a simpler abstraction for this seems there's multiple layers just adding complexity.

docker_mounts.go
```diff
@@ -85,6 +100,40 @@ func (s DockerTmpfsMountSource) GetTmpfsOptions() *mount.TmpfsOptions {
...
+	// ImageName is the image name
+	ImageName string
```
stevenh, line 106: question: do these need to be public so should we keep them private and use a NewXXX method to ensure the type is constructed properly?

stevenh (changes requested):

options_test.go
mdelapenya: Done in [a143f15](https://github.com/testcontainers/testcontainers-go/pull/3044/commits/a143f15bf34d0b3e0d8c31562799cb655f02695f)

mdelapenya:
@stevenh @eddumelendez I acknowledge that the names used in this PR could be better, but they are consistent with the existing types. As mentioned [above](https://github.com/testcontainers/testcontainers-go/pull/3044#discussion_r2006886661) I'm open to start a full-rewrite of this part, from scratch.

In the meantime, I think this is good to go for the next release.

mdelapenya:
I'm merging this one, as interesting to have it for the next release.

PR #3051, opened 2025-03-22
feat: allow overriding the session ID

### LINKED ISSUE #2804: [Feature]: Support reuse + reaper keeping a container alive as long as more tests run within a time window
### Problem

I run tests that require a mysql, over and over again. While I'm iterating, I'd like to leverage `reuse` to only pay that boot price on the first test, and then subsequent tests can reuse the same mysql (which is close to a 6x improvement in speed).

_However_, I also want the reaper to clean up this container once I'm done with my testing "session".

I'm imagining a config like `reuse: true, reuseTimeout: 5m`. Which makes it so as long as I run a test within 5m of the last test, the reaper leaves the container alone, but once 5m elapses without triggering a test, the reaper cleans everything up.

I thought this might be possible today with a combination of `reuse: true` and `TESTCONTAINERS_RYUK_RECONNECTION_TIMEOUT=5m`, but that still cleans up my mysql after 5m (and also spawns a new reaper for every run, which quickly pile up):

```
CONTAINER ID  IMAGE                                                  COMMAND     CREATED             STATUS             PORTS                                                                   NAMES
b551cb564516  docker.io/testcontainers/ryuk:0.8.1                    /bin/ryuk   2 minutes ago       Up 2 minutes       0.0.0.0:39033->8080/tcp, 8080/tcp                                       reaper_42d84025974814469f367f10224b4657f7ad888e2a5d827e76703f06292c39ec
5a7876b38a26  docker.io/mysql:8.0.36                                 mysqld      2 minutes ago       Up 2 minutes       0.0.0.0:42559->3306/tcp, 0.0.0.0:44633->33060/tcp, 3306/tcp, 33060/tcp  tests-mysql
c93fdf315582  docker.io/testcontainers/ryuk:0.8.1                    /bin/ryuk   2 minutes ago       Up 2 minutes       0.0.0.0:34995->8080/tcp, 8080/tcp                                       reaper_3ccf995c484a246a36990a92f85c3fc17ffb3912f867ba70b029aaf323f48547
e63820a62ce3  docker.io/testcontainers/ryuk:0.8.1                    /bin/ryuk   2 minutes ago       Up 2 minutes       0.0.0.0:44359->8080/tcp, 8080/tcp                                       reaper_fe49fa57fcc7031f3d1e9dbb6f305a2f4b392d1b10921d42b6a0df56e4970667
56d082f1fbbd  docker.io/testcontainers/ryuk:0.8.1                    /bin/ryuk   About a minute ago  Up About a minute  0.0.0.0:38461->8080/tcp, 8080/tcp                                       reaper_669ec11b7226f1c446fddd1a0e3f1aadfd1627687ed6adf3b5c846e66ecbe276
9922c527850c  docker.io/testcontainers/ryuk:0.8.1                    /bin/ryuk   About a minute ago  Up About a minute  0.0.0.0:34041->8080/tcp, 8080/tcp                                       reaper_74976de06285397df7d0edffbdd2e5fd284cf833f5b3de025705fbcad5e3608b
c76bcd09a892  docker.io/testcontainers/ryuk:0.8.1                    /bin/ryuk   52 seconds ago      Up 52 seconds      0.0.0.0:32785->8080/tcp, 8080/tcp                                       reaper_6e46bea7477b04a07464880892df435bdbd55753551da0d80b51d8917f4ab094
df476617ebdf  docker.io/testcontainers/ryuk:0.8.1                    /bin/ryuk   42 seconds ago      Up 42 seconds      0.0.0.0:39403->8080/tcp, 8080/tcp                                       reaper_24fdf990fc0d0ce6e34d030f619611233f7b7e714e623e1b6c635f7a85fdee4a
fc398aaabe82  docker.io/testcontainers/ryuk:0.8.1                    /bin/ryuk   27 seconds ago      Up 27 seconds      0.0.0.0:34295->8080/tcp, 8080/tcp                                       reaper_f9596c3bfefde5404c227e99286798894109b8d3b1de2fe207580a1baf340ac6
8ac0c88b5298  docker.io/testcontainers/ryuk:0.8.1                    /bin/ryuk   18 seconds ago      Up 18 seconds      0.0.0.0:42809->8080/tcp, 8080/tcp                                       reaper_8a5edab68adc986d01ff49f0774b013c3933ee8854bda319df02d26dfa233ded
64edfcfa3ec0  docker.io/testcontainers/ryuk:0.8.1                    /bin/ryuk   8 seconds ago       Up 9 seconds       0.0.0.0:35387->8080/tcp, 8080/tcp                                       reaper_3d76e1e0b630828fad795b4fcc603edaf11179da4d56148248f89862e92139d1



### Solution

Add a rolling reaper cleanup window

### Benefit

Users can reuse containers, but not have to worry about manually cleaning them up

### Alternatives

Manually delete reusable containers

Yes

### PR DESCRIPTION
**Pull request by jcmfernandes**
## What does this PR do?
Makes the session ID configurable, allowing for long-running sessions.

## Why is it important?
Speaking of my use-case, Apache Pulsar takes a long time to boot, so I wanted testcontainers to boot it up, eventually take it down, but allow me to use the same container across several runs. This wasn't the case, since I was getting a new reaper running on every test run. I wanted a rolling-window behavior, hence this change.

## Related issues

- Closes #2804

## Follow-ups

Update the documentation.

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -1010,7 +1010,7 @@ func (c *DockerContainer) connectReaper(ctx context.Context) error {
 		return nil
 	}
 
-	reaper, err := spawner.reaper(context.WithValue(ctx, core.DockerHostContextKey, c.provider.host), core.SessionID(), c.provider)
+	reaper, err := spawner.reaper(context.WithValue(ctx, core.DockerHostContextKey, c.provider.host), c.provider.config.SessionID, c.provider)
 	if err != nil {
 		return fmt.Errorf("reaper: %w", err)
 	}
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -52,6 +52,13 @@ type Config struct {
 	// Environment variable: TESTCONTAINERS_HUB_IMAGE_NAME_PREFIX
 	HubImageNamePrefix string `properties:"hub.image.name.prefix,default="`
 
+	// SessionID is the ID of the testing session.
+	// Setting this value will preclude runs from creating more than one reaper. Therefore,
+	// changes to ryuk settings past its creation will be ignored.
+	//
+	// Environment variable: TESTCONTAINERS_SESSION_ID
+	SessionID string `properties:"session.id,default="`
+
 	// RyukDisabled is a flag to enable or disable the Garbage Collector.
 	// Setting this to true will prevent testcontainers from automatically cleaning up
 	// resources, which is particularly important in tests which timeout as they
@@ -119,6 +126,11 @@ func read() Config {
 		hubImageNamePrefix := os.Getenv("TESTCONTAINERS_HUB_IMAGE_NAME_PREFIX")
 		if hubImageNamePrefix != "" {
 			config.HubImageNamePrefix = hubImageNamePrefix
+		}
+
+		sessionID := os.Getenv("TESTCONTAINERS_SESSION_ID")
+		if sessionID != "" {
+			config.SessionID = sessionID
 		}
 
 		ryukPrivilegedEnv := os.Getenv("TESTCONTAINERS_RYUK_CONTAINER_PRIVILEGED")
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/magiconair/properties"
+
 	"github.com/testcontainers/testcontainers-go/internal/core/bootstrap"
 )
 
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/magiconair/properties"
+	"github.com/testcontainers/testcontainers-go/internal/core/bootstrap"
 )
 
 const ReaperDefaultImage = "testcontainers/ryuk:0.14.0"
@@ -131,6 +132,8 @@ func read() Config {
 		sessionID := os.Getenv("TESTCONTAINERS_SESSION_ID")
 		if sessionID != "" {
 			config.SessionID = sessionID
+		} else if config.SessionID == "" {
+			config.SessionID = bootstrap.SessionID()
 		}
 
 		ryukPrivilegedEnv := os.Getenv("TESTCONTAINERS_RYUK_CONTAINER_PRIVILEGED")
--- a/internal/core/bootstrap.go
+++ b/internal/core/bootstrap.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/google/uuid"
 	"github.com/shirou/gopsutil/v4/process"
+
 	"github.com/testcontainers/testcontainers-go/internal/config"
 )
 
--- a/internal/core/bootstrap.go
+++ b/internal/core/bootstrap.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/google/uuid"
 	"github.com/shirou/gopsutil/v4/process"
+	"github.com/testcontainers/testcontainers-go/internal/config"
 )
 
 // sessionID returns a unique session ID for the current test session. Because each Go package
@@ -45,6 +46,11 @@ var processID string
 const sessionIDPlaceholder = "testcontainers-go:%d:%d"
 
 func init() {
+	cfg := config.Read()
+	if cfg.SessionID != "" {
+		sessionID = cfg.SessionID
+	}
+
 	processID = uuid.New().String()
 
 	parentPid := os.Getppid()
@@ -57,7 +63,9 @@ func init() {
 
 	processes, err := process.Processes()
 	if err != nil {
-		sessionID = uuid.New().String()
+		if sessionID == "" {
+			sessionID = uuid.New().String()
+		}
 		projectPath = fallbackCwd
 		return
 	}
@@ -75,7 +83,9 @@ func init() {
 
 		t, err := p.CreateTime()
 		if err != nil {
-			sessionID = uuid.New().String()
+			if sessionID == "" {
+				sessionID = uuid.New().String()
+			}
 			return
 		}
 
@@ -83,14 +93,16 @@ func init() {
 		break
 	}
 
-	hasher := sha256.New()
-	_, err = fmt.Fprintf(hasher, sessionIDPlaceholder, parentPid, createTime)
-	if err != nil {
-		sessionID = uuid.New().String()
-		return
-	}
+	if sessionID == "" {
+		hasher := sha256.New()
+		_, err = fmt.Fprintf(hasher, sessionIDPlaceholder, parentPid, createTime)
+		if err != nil {
+			sessionID = uuid.New().String()
+			return
+		}
 
-	sessionID = hex.EncodeToString(hasher.Sum(nil))
+		sessionID = hex.EncodeToString(hasher.Sum(nil))
+	}
 }
 
 func ProcessID() string {
--- a/container.go
+++ b/container.go
@@ -22,6 +22,7 @@ import (
 	"github.com/moby/patternmatcher/ignorefile"
 
 	tcexec "github.com/testcontainers/testcontainers-go/exec"
+	"github.com/testcontainers/testcontainers-go/internal/config"
 	"github.com/testcontainers/testcontainers-go/internal/core"
 	"github.com/testcontainers/testcontainers-go/log"
 	"github.com/testcontainers/testcontainers-go/wait"
@@ -176,7 +177,7 @@ func (c *ContainerRequest) sessionID() string {
 		return sessionID
 	}
 
-	return core.SessionID()
+	return config.Read().SessionID
 }
 
 // containerOptions functional options for a container
--- a/docker_client.go
+++ b/docker_client.go
@@ -9,7 +9,9 @@ import (
 	"github.com/moby/moby/client"
 
 	"github.com/testcontainers/testcontainers-go/internal"
+	"github.com/testcontainers/testcontainers-go/internal/config"
 	"github.com/testcontainers/testcontainers-go/internal/core"
+	"github.com/testcontainers/testcontainers-go/internal/core/bootstrap"
 	"github.com/testcontainers/testcontainers-go/log"
 )
 
@@ -17,6 +19,8 @@ import (
 // It implements the SystemAPIClient interface in order to cache the docker info and reuse it.
 type DockerClient struct {
 	*client.Client // client is embedded into our own client
+
+	config config.Config
 }
 
 var (
@@ -85,8 +89,8 @@ func (c *DockerClient) Info(ctx context.Context, options client.InfoOptions) (cl
 		internal.Version,
 		host,
 		core.MustExtractDockerSocket(ctx),
-		core.SessionID(),
-		core.ProcessID(),
+		c.config.SessionID,
+		bootstrap.ProcessID(),
 	)
 
 	return dockerInfo, nil
@@ -123,8 +127,11 @@ func NewDockerClientWithOpts(ctx context.Context, opt ...client.Opt) (*DockerCli
 		return nil, err
 	}
 
+	tcConfig := config.Read()
+
 	tcClient := DockerClient{
 		Client: dockerClient,
+		config: tcConfig,
 	}
 
 	if _, err = tcClient.Info(ctx, client.InfoOptions{}); err != nil {
--- a/docker_client.go
+++ b/docker_client.go
@@ -127,11 +127,9 @@ func NewDockerClientWithOpts(ctx context.Context, opt ...client.Opt) (*DockerCli
 		return nil, err
 	}
 
-	tcConfig := config.Read()
-
 	tcClient := DockerClient{
 		Client: dockerClient,
-		config: tcConfig,
+		config: config.Read(),
 	}
 
 	if _, err = tcClient.Info(ctx, client.InfoOptions{}); err != nil {
--- a/generic.go
+++ b/generic.go
@@ -7,6 +7,7 @@ import (
 	"maps"
 	"sync"
 
+	"github.com/testcontainers/testcontainers-go/internal/config"
 	"github.com/testcontainers/testcontainers-go/internal/core"
 	"github.com/testcontainers/testcontainers-go/log"
 )
@@ -100,7 +101,7 @@ type GenericProvider interface {
 // reaper is enabled, otherwise this is excluded to prevent resources being
 // incorrectly reaped.
 func GenericLabels() map[string]string {
-	return core.DefaultLabels(core.SessionID())
+	return core.DefaultLabels(config.Read().SessionID)
 }
 
 // AddGenericLabels adds the generic labels to target.
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -569,7 +569,7 @@ func TestReadTCConfig(t *testing.T) {
 				},
 			},
 			{
-				"With Ryuk disabled using an env var and properties. Env var wins",
+				"With Session ID set using an env var and properties. Env var wins",
 				`session.id=bar`,
 				map[string]string{
 					"TESTCONTAINERS_SESSION_ID": "foo",
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+
 	"github.com/testcontainers/testcontainers-go/internal/core/bootstrap"
 )
 
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"github.com/testcontainers/testcontainers-go/internal/core/bootstrap"
 )
 
 const (
@@ -43,6 +44,7 @@ func TestReadConfig(t *testing.T) {
 		config := Read()
 
 		expected := Config{
+			SessionID:    bootstrap.SessionID(),
 			RyukDisabled: true,
 			Host:         "", // docker socket is empty at the properties file
 		}
@@ -67,7 +69,9 @@ func TestReadTCConfig(t *testing.T) {
 
 		config := read()
 
-		expected := Config{}
+		expected := Config{
+			SessionID: bootstrap.SessionID(),
+		}
 
 		assert.Equal(t, expected, config)
 	})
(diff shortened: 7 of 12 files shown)

### CODE REVIEW

File: docker.go [hunk 1]
```diff
@@ -1454,15 +1454,8 @@ func (p *DockerProvider) RunContainer(ctx context.Context, req ContainerRequest)
...
-func (p *DockerProvider) Config() TestcontainersConfig {
-	return TestcontainersConfig{
-		Host:           p.config.Host,
-		TLSVerify:      p.config.TLSVerify,
-		CertPath:       p.config.CertPath,
-		RyukDisabled:   p.config.RyukDisabled,
-		RyukPrivileged: p.config.RyukPrivileged,
-		Config:         p.config,
-	}
+func (p *DockerProvider) Config() config.Config {
```
mdelapenya: bug: removing this method will cause a breaking change for current users. If we really want to do this, we'd need to mark the PR as a BC (adding `feat:!` as PR title prefix). We can bypass that with a deprecation and a new method. More below

config.go
```diff
@@ -1,29 +0,0 @@
-package testcontainers
-
-import (
-	"github.com/testcontainers/testcontainers-go/internal/config"
-)
-
-// TestcontainersConfig represents the configuration for Testcontainers
-type TestcontainersConfig struct {
```
mdelapenya, line 8: bug: removing this will cause a breaking change, see below.

kiview:
Hey @jcmfernandes, I am actually surprised your setup doesn't work with the current [Test Session Semantics](https://golang.testcontainers.org/features/test_session_semantics/).

Also, your change doesn't take the parent PID into account anymore, so it would be a breaking change for downstream integrations, that rely on the current session ID semantics.

jcmfernandes:
> Hey @jcmfernandes, I am actually surprised your setup doesn't work with the current [Test Session Semantics](https://golang.testcontainers.org/features/test_session_semantics/).
> 
> Also, your change doesn't take the parent PID into account anymore, so it would be a breaking change for downstream integrations, that rely on the current session ID semantics.

Thanks for getting back to me.

Without these changes, every `go test ./...` run create a new instance of the reaper, and I'm not alone (#2804).

As stated in the PR description:
>Regarding point 3, I must admit that I don't understand the advantages of the hashing logic I ended up deleting. I'm happy to revert it, though.

I'm happy to revert those changes, but so far, setting the session ID myself (through an env var) has been the only solution that worked for me.

kiview:
> Without these changes, every go test ./... run create a new instance of the reaper

Alright, that is how it is supposed to work.
But what it sounds to me from what you are trying to achieve, it sounds you need [Reusable Containers](https://golang.testcontainers.org/features/creating_container/#reusable-container).

Either way, the semantics around sessions IDs and cleanup behavior are fairly involved. I'd suggest to not proceed with this work without specifically syncing with @mdelapenya and @stevenh on the design.

jcmfernandes:
> > Without these changes, every go test ./... run create a new instance of the reaper
> 
> Alright, that is how it is supposed to work. But what it sounds to me from what you are trying to achieve, it sounds you need [Reusable Containers](https://golang.testcontainers.org/features/creating_container/#reusable-container).
> 
> Either way, the semantics around sessions IDs and cleanup behavior are fairly involved. I'd suggest to not proceed with this work without specifically syncing with @mdelapenya and @stevenh on the design.

Thanks for the pointers. @mdelapenya and @stevenh, your input is very much welcome.

>But what it sounds to me from what you are trying to achieve, it sounds you need [Reusable Containers](https://golang.testcontainers.org/features/creating_container/#reusable-container).

I am using a reusable container, and the container was always reused as expected; no issues there. The problem is that the reaper isn't reused because the session ID keeps changing.

jcmfernandes:
>Also, your change doesn't take the parent PID into account anymore, so it would be a breaking change for downstream integrations, that rely on the current session ID semantics.

I pushed a new commit that brings back the logic I had previously removed. This shouldn't break things for anyone anymore.

stevenh:
Hi @jcmfernandes can you clarify to me the problem you're trying to solve?

Your description seems to indicate to me you want to share containers between tests due to the slow startup time?

jcmfernandes:
> Hi @jcmfernandes can you clarify to me the problem you're trying to solve?
> 
> Your description seems to indicate to me you want to share containers between tests due to the slow startup time?

Hi @stevenh! Correct. In other words, I want:
1. Reusable containers (all good!)
2. Reusable reapers for reusable containers (**not** good)

Copying from #2804, this is what I see when I enable reusable containers:

> ```
> CONTAINER ID  IMAGE                                                  COMMAND     CREATED             STATUS             PORTS                                                                   NAMES
> b551cb564516  docker.io/testcontainers/ryuk:0.8.1                    /bin/ryuk   2 minutes ago       Up 2 minutes       0.0.0.0:39033->8080/tcp, 8080/tcp                                       reaper_42d84025974814469f367f10224b4657f7ad888e2a5d827e76703f06292c39ec
> 5a7876b38a26  docker.io/mysql:8.0.36                                 mysqld      2 minutes ago       Up 2 minutes       0.0.0.0:42559->3306/tcp, 0.0.0.0:44633->33060/tcp, 3306/tcp, 33060/tcp  tests-mysql
> c93fdf315582  docker.io/testcontainers/ryuk:0.8.1                    /bin/ryuk   2 minutes ago       Up 2 minutes       0.0.0.0:34995->8080/tcp, 8080/tcp                                       reaper_3ccf995c484a246a36990a92f85c3fc17ffb3912f867ba70b029aaf323f48547
> e63820a62ce3  docker.io/testcontainers/ryuk:0.8.1                    /bin/ryuk   2 minutes ago       Up 2 minutes       0.0.0.0:44359->8080/tcp, 8080/tcp                                       reaper_fe49fa57fcc7031f3d1e9dbb6f305a2f4b392d1b10921d42b6a0df56e4970667
> 56d082f1fbbd  docker.io/testcontainers/ryuk:0.8.1                    /bin/ryuk   About a minute ago  Up About a minute  0.0.0.0:38461->8080/tcp, 8080/tcp                                       reaper_669ec11b7226f1c446fddd1a0e3f1aadfd1627687ed6adf3b5c846e66ecbe276
> 9922c527850c  docker.io/testcontainers/ryuk:0.8.1                    /bin/ryuk   About a minute ago  Up About a minute  0.0.0.0:34041->8080/tcp, 8080/tcp                                       reaper_74976de06285397df7d0edffbdd2e5fd284cf833f5b3de025705fbcad5e3608b
> c76bcd09a892  docker.io/testcontainers/ryuk:0.8.1                    /bin/ryuk   52 seconds ago      Up 52 seconds      0.0.0.0:32785->8080/tcp, 8080/tcp                                       reaper_6e46bea7477b04a07464880892df435bdbd55753551da0d80b51d8917f4ab094
> df476617ebdf  docker.io/testcontainers/ryuk:0.8.1                    /bin/ryuk   42 seconds ago      Up 42 seconds      0.0.0.0:39403->8080/tcp, 8080/tcp                                       reaper_24fdf990fc0d0ce6e34d030f619611233f7b7e714e623e1b6c635f7a85fdee4a
> fc398aaabe82  docker.io/testcontainers/ryuk:0.8.1                    /bin/ryuk   27 seconds ago      Up 27 seconds      0.0.0.0:34295->8080/tcp, 8080/tcp                                       reaper_f9596c3bfefde5404c227e99286798894109b8d3b1de2fe207580a1baf340ac6
> 8ac0c88b5298  docker.io/testcontainers/ryuk:0.8.1                    /bin/ryuk   18 seconds ago      Up 18 seconds      0.0.0.0:42809->8080/tcp, 8080/tcp                                       reaper_8a5edab68adc986d01ff49f0774b013c3933ee8854bda319df02d26dfa233ded
> 64edfcfa3ec0  docker.io/testcontainers/ryuk:0.8.1                    /bin/ryuk   8 seconds ago       Up 9 seconds       0.0.0.0:35387->8080/tcp, 8080/tcp                                       reaper_3d76e1e0b630828fad795b4fcc603edaf11179da4d56148248f89862e92139d1
> ```

I want to see **one** reaper for the reusable container. As a side effect, having one reaper creates a “rolling window” behavior, where the reaper resets its timeout upon a new connection to the container, which is precisely what I would like to see: keep the container up and running between test runs and terminate it after the defined period of inactivity.

Am I making sense?

kiview:
Oh yes, that makes absolutely sense and I believe this is more of a bug and something that Testcontainers should already handle correctly, rather than having a user to specify a session id manually.

stevenh:
Thanks for the confirmation @jcmfernandes.

To facilitate this are you intending to set a long `ReconnectionTimeout` for all containers? I ask as if so I think thats going to prevent clean up of all other containers too, which I'm not sure is your intention?

jcmfernandes:
@stevenh yes, because I want to keep the testcontainer running for the entirety of my work session.

I usually start a watcher that runs tests automatically as I save changes. I want the testcontainer to boot on the first test run and to stay around for, say, one hour, waiting for new connections. Furthermore, I then want new connections to reset that timeout. Example.

1. Run a test that starts the testcontainer at 9:05 AM with a reconnection timeout of 1 hour. Ryuk must wait for new connections until 10:05 AM, and GC the testcontainer if that doesn't happen.
2. Run another test at 9:35 AM --> I now want ryuk to wait until 10:35 AM for new connections.

This is precisely the behavior I get by manually setting the session ID and `RYUK_RECONNECTION_TIMEOUT` to 1 hour. Now, I'm happy to discuss ways to make the session ID stable and auto-generated.

stevenh:
Thanks again for the clarification, the challenge I see with that is not only does it impact the long lived container but all containers for example:

You run a test which starts the long lived DB container as well a short lived test container. The desired behaviour is the DB container stays running between tests, but the test container itself exits after the test is run.

However in the case where the test fails and the short lived contain didn't exit as expected, with a standard ryuk setup this wouldn't be an issue as the test container would be quickly cleaned up, but with `RYUK_RECONNECTION_TIMEOUT` set to one hour no clean up will happen until that hour expires.

Does this make sense?

To handle this situation properly I think we'd need more fine grained control of when specific containers are cleaned up.

jcmfernandes:
Thanks for the prompt reply, @stevenh.

That's a fair concern — it seems that testcontainers is missing some kind of "group" concept to which certain GC options apply. Currently, that's a "session", and this PR won't change it. That is, I don't see how this PR makes things worse regarding your concern, and it solves an actual reported issue.

mdelapenya (commented):
Hi @jcmfernandes, thanks for opening this PR, and following-up after @kiview's comments. I'm returning today after summer break, working on the open PRs/issues in the repo.

I'm not against the changes in this PR, although it introduces some breaking changes (BCs) after removing public API functions. See my comments for that. We could avoid it offering a deprecation path when/if possible, although we sometimes introduce BCs if needed.

I'd rename the PR to `feat: allow overriding the sessionID` and would focus on that, only. The added value if we do this is we would also enable Bazel folks to contribute their own env var for the session ID at build time. Nevertheless, we must check how the reaper behaves in that scenario. As discussed here, it's by design that the reaper is spawned just once per test session (a shell), but it's true that the design could be inconsistent with reusable containers used in multiple shells (different session IDs). Sharing state, even in tests, is a complex thing to maintain, so we probably need a more robust design for both features: reusable containers and the reaper, and it could need dedicated work on that direction.

@kiview @stevenh wdyt?

docker.go
jcmfernandes: Right, this isn't internal-only. I'll split it into two PRs. Thanks for the feedback!

jcmfernandes:
@mdelapenya thanks for your input. I modified this PR, making it focus solely on allowing devs to set the session ID, and I opened a separate PR #3269 where I delete `TestcontainersConfig`.

internal/core/bootstrap.go
```diff
@@ -8,6 +8,7 @@ import (
 
 	"github.com/google/uuid"
 	"github.com/shirou/gopsutil/v4/process"
+	"github.com/testcontainers/testcontainers-go/internal/config"
```
mdelapenya: bug: linter is complaining about this line. Could you run `make lint` to address that? 🙏

jcmfernandes: Done!

internal/core/bootstrap.go
```diff
@@ -45,6 +47,11 @@ var processID string
 const sessionIDPlaceholder = "testcontainers-go:%d:%d"
 
 func init() {
+	cfg := config.Read()
```
stevenh: @mdelapenya could calling `config.Read()` in `init` cause undesired behaviour due to how early it called?

I'm thinking this would prevent things like changing the environment variables in a test from having the desired effect as an example.

internal/core/bootstrap.go
```diff
@@ -57,7 +64,9 @@ func init() {
 
 	processes, err := process.Processes()
 	if err != nil {
-		sessionID = uuid.New().String()
+		if sessionID == "" {
+			sessionID = uuid.New().String()
+		}
```
stevenh: suggestion: put this in a `defer` check in an else block to the `cfg.SessionID != ""` less to maintain that way.

stevenh (changes requested):
I'm concerned calling `config.Read` in `init` could have some unintended side effects which are impossible to address, something we'll need to confirm.

internal/core/bootstrap.go
jcmfernandes: I'm no longer pushing this change.

internal/core/bootstrap.go
jcmfernandes: Thanks for the feedback, @stevenh. I understand your concern: reading the config in an `init()` function felt wrong.

I change the code so that `config.Read` falls back to the session in `core/bootstrap` (I moved `bootstrap.go` to a new package to avoid import cycle between `internal/config` and `internal/core`). Please let me know what you think.

internal/config/config.go
```diff
@@ -121,6 +130,13 @@ func read() Config {
 			config.HubImageNamePrefix = hubImageNamePrefix
 		}
 
+		sessionID := os.Getenv("TESTCONTAINERS_SESSION_ID")
+		if sessionID != "" {
+			config.SessionID = sessionID
+		} else if config.SessionID == "" {
+			config.SessionID = bootstrap.SessionID()
+		}
```
mdelapenya, line 138: suggestion: is this more correct to set the default value to the bootstrap one?
```suggestion
		config.SessionID = bootstrap.SessionID()
		sessionID := os.Getenv("TESTCONTAINERS_SESSION_ID")
		if sessionID != "" {
			config.SessionID = sessionID
		}
```

mdelapenya (dismissed):
LGTM! I'm approving it even though I added a comment that would be good to add.

Once addressed, I think we can re-run the CI and merge.

Thanks!

PS: Many times, we are forced to do multitasking, so please sorry for taking too long to address this.

internal/config/config.go
jcmfernandes: The "else if" comparison is indeed irrelevant. Fixing that. Thank you!

jcmfernandes:
> LGTM! I'm approving it even though I added a comment that would be good to add.
> 
> Once addressed, I think we can re-run the CI and merge.
> 
> Thanks!
> 
> PS: Many times, we are forced to do multitasking, so please sorry for taking too long to address this.

Nothing to apologize for! Thanks, everyone, for the thorough review. :raised_hands:

mdelapenya:
Some tests are failing on CI, can you check?

jcmfernandes:
> Some tests are failing on CI, can you check?

After all, [that "else if" statement](https://github.com/testcontainers/testcontainers-go/pull/3051#discussion_r2745643205) is necessary, as the AI bot correctly [pointed out](https://github.com/testcontainers/testcontainers-go/pull/3051#discussion_r2745676269). I deleted the commit.

internal/config/config.go
mdelapenya: @jcmfernandes is this updated? I think the else if is still there 🙈

jcmfernandes: @mdelapenya the "if else" really is necessary because we would rather not ignore values set in `.testcontainers.properties`. The priority is:

1. Env var
2. Property
3. `bootstrap.SessionID()`

Your suggestion causes us to ignore the property.

jcmfernandes:
@mdelapenya mind giving this a final look? Thank you so much.

mdelapenya:
@jcmfernandes can you rebase from current main 🙏 Fine if you push-force, as the changeset's size is reduced

BTW I'm cutting a new release soon, and this work will land into it (sorry for the late review 😞 )

jcmfernandes:
Rebased @mdelapenya, thank you! :raised_hands: 

I opened #3837 as it was raised the [LLM code review](https://github.com/testcontainers/testcontainers-go/pull/3051#pullrequestreview-4856522740) added to this branch and it's a genuine bug.

mdelapenya:
@jcmfernandes for some reason I cannot update your PR branches. Could you update them yourself? I already fixed the lint error for the azurite module

jcmfernandes:
> @jcmfernandes for some reason I cannot update your PR branches. Could you update them yourself? I already fixed the lint error for the azurite module

Rebased on main.

mdelapenya:
CI fails, could you take a look?

jcmfernandes:
> CI fails, could you take a look?

Done. It was something wonky that stem from the review but truly is necessary.

internal/config/config.go
```diff
@@ -121,6 +130,13 @@ func read() Config {
 			config.HubImageNamePrefix = hubImageNamePrefix
 		}
 
+		sessionID := os.Getenv("TESTCONTAINERS_SESSION_ID")
+		if sessionID != "" {
+			config.SessionID = sessionID
+		} else if config.SessionID == "" {
```
mdelapenya, line 136: question: if this function is called just once , who could have set the session ID value here? Could this be a sympton the tests are kind of setting a state and this elseIf path is exactly for that?

jcmfernandes: Clarifying: are you refering to https://github.com/testcontainers/testcontainers-go/actions/runs/31088981575/job/92592530979?pr=3051#step:8:5482 ?

If so, env vars take priority over the values included in the properties file - that is, we **first** load `config` with values from the properties file and only then we **override** them with values coming from env vars. Therefore, if we remove the `config.SessionID == ""` check, then we **always** override the value coming from the properties file when the env var isn't set.

Does this make sense?

mdelapenya: Oh I see, it's the properties the one that could have set it, sorry, Now I get it, thanks for the clarification

mdelapenya (approved):
LGTM, thanks

Issue #2470, opened 2024-04-07
[Enhancement]: Reuse flag to start exited or paused containers 


FredrikMorstad:
### Proposal

As of now using the reuse flag on container who's not running(but is created) will cause a the program to ping and wait for the container to be running for 15 minutes `DefaultMaxElapsedTime` before exiting(with ryku disabled). My proposal is to restart the container if it's not running as I think it's a natural extension of the reuse name. As stetted in the documentation `With Reuse option you can reuse an existing container.` . I interpret existing covering both running containers as well as exited or paused containers, so I think this will be an enhancement for the reuse flag to restart a paused or exited container. 

### Why is this useful?
In my case I have an interface providing the necessary functions for running a local authentication emulator, I want to keep using the same container rather than recreating the container as this image rarely changes. By extending the reuse capability we can avoid the need for calling the docker api to start the container before actually using testcontainers. 

Here is the code for some context:
```go
func (testauth *TestAuth) Start() error {
	containerPort := "9099"

	// emulator specific endpoint used to check if emulator is ready
	firebaseConfigURI := "/emulator/v1/projects/demo-project/config"

	request := testcontainers.ContainerRequest{
		Image:        "our_ghcr_image",
		ExposedPorts: []string{containerPort},
		WaitingFor:   wait.NewHTTPStrategy(firebaseConfigURI),
	}

	if testauth.reuse {
		request.Name = "our_testcontainer_name"
	}
	
	// Must add code here to check if container of name `req.Name` is paused or exited and possible restart the container for this code to work 
	
	container, err := testcontainers.GenericContainer(testauth.containerCtx, testcontainers.GenericContainerRequest{
		ContainerRequest: request,
		Started:          true,
		Reuse:            testauth.reuse,
	})

	if err != nil {
		return err
	}

	// ... some other code 
}
```

### Some extra info
I have ryku disabled as I had a problem with it and haven't gotten around to investigate the problem. So this may not be a problem if the garbage collector cleans up in the background so when `reuse` is enabled it's almost always needing to create a new container. However if this is the case I still belive that this feature will improve the reuse functionality

If it turns out to be something we want to do I can create the PR as I have already gotten the bear minimum version to work

mdelapenya:
Hey @FredrikMorstad I'd like to know more about your setup to understand better your use case, more specifically the number of packages using that "singleton" container.

[...]

PR #3062, opened 2025-03-26
fix: handle stopped containers more gracefully when reuse is enabled

georgespalding:
## What does this PR do?

Fixes a bug that causes testcontainers to hang for a really long time waiting on a stopped container when reuse is enabled.

The fix in short is:

When a container is being created with reuse enabled:
- Look among all containers, even the ones not running.
  - If a container with the expected name is found:
    - If it is not running:
      - Attempt to start it.
        - If the start fails, return an error

The alternative cases above should work same as before 

## Why is it important?

The way the library works today, if a reused container exists, but is not running, it freezes for a long time, which is a really bad developer experience.

This fix is not watertight (ses the `docker pause` scenario below), but instantly tells you that something is wrong, giving fast feedback.

## Related issues

- Relates https://github.com/testcontainers/testcontainers-go/issues/2470

## How to test this PR

Create a little program that runs a testcontainer:
```golang
package main

import (
	"context"
	"fmt"
	"github.com/testcontainers/testcontainers-go/modules/postgres"
)

func main() {
	true := true
	containerName := "my-postgres"
	container, err := postgres.Run(context.Background(),
		"postgres:14-bookworm",
		postgres.BasicWaitStrategies(),
		testcontainers.CustomizeRequestOption(func(req *testcontainers.GenericContainerRequest) error {
			req.Name = containerName
			req.Reuse = true
			return nil
		}),
	)
	if err != nil {
		fmt.Println(fmt.Sprintf("run postgres testcontainer: %s", err.Error()))
	} else {
		fmt.Println(fmt.Sprintf("postgres testcontainer is running: %t", container.Container.IsRunning()))
	}
}
```

1. `docker ps` -> container is not there
2. `go run ./` -> prints `postgres testcontainer is running: true`
3. `docker ps` -> container _is_ running
4. `docker stop my-postgres`
5. `go run ./` -> prints `postgres testcontainer is running: true`
6. `docker pause my-postgres`
7. `go run ./` -> prints `run postgres testcontainer: generic container: create container: start container my-postgres in state paused: container start: Error response from daemon: cannot start a paused container, try unpause instead`


docker.go
```diff
@@ -1369,6 +1369,16 @@ func (p *DockerProvider) ReuseOrCreateContainer(ctx context.Context, req Contain
 		lifecycleHooks:    []ContainerLifecycleHooks{combineContainerHooks(defaultHooks, req.LifecycleHooks)},
 	}
 
+	if p.config.RyukDisabled {
+		// ryuk disabled, ensure the container we are about to reuse is started
+		if !dc.IsRunning() {
```
georgespalding, line 1374: Bug: `dc.IsRunning` always returns false, that struct was initialized right above... :facepalm: 
```suggestion
		if c.State == "running" {
```

docker.go
```diff
@@ -1369,6 +1369,17 @@ func (p *DockerProvider) ReuseOrCreateContainer(ctx context.Context, req Contain
 		lifecycleHooks:    []ContainerLifecycleHooks{combineContainerHooks(defaultHooks, req.LifecycleHooks)},
 	}
 
+	if p.config.RyukDisabled {
+		// ryuk disabled, ensure the container we are about to reuse is running
+		if c.State != "running" {
+			// found container to reuse, but it is not in state running
+			fmt.Println("Attempting to start: " + req.Name + " in state: " + c.State)
```
stevenh, line 1376: suggestion: removing this debugging.

stevenh (changes requested):

docker.go
georgespalding: Committed by mistake when checking that the `c.State` check was doing what I expected...
Removed now.

georgespalding: @stevenh I've removed the debugging as you suggested. Would you be up for doing a second pass?

mdelapenya:
@georgespalding thanks for submitting this PR.  I'm trying to reproduce it locally with this test:

```go
func TestWithReuse(t *testing.T) {
	containerName := "my-postgres"

	opts := []testcontainers.ContainerCustomizer{
		postgres.BasicWaitStrategies(),
		testcontainers.CustomizeRequestOption(func(req *testcontainers.GenericContainerRequest) error {
			req.Name = containerName
			req.Reuse = true
			return nil
		}),
	}

	ctr, err := postgres.Run(context.Background(), "postgres:14-bookworm", opts...)
	if err != nil {
		t.Logf("run postgres testcontainer: %s", err.Error())
	} else {
		t.Logf("postgres testcontainer is running: %t", ctr.IsRunning())
	}

	require.NoError(t, err)
	require.NotNil(t, ctr)

	err = ctr.Stop(context.Background(), nil)
	require.NoError(t, err)

	ctr1, err := postgres.Run(context.Background(), "postgres:14-bookworm", opts...)
	if err != nil {
		t.Logf("run postgres testcontainer: %s", err.Error())
	} else {
		t.Logf("postgres testcontainer is running: %t", ctr1.IsRunning())
	}

	require.NoError(t, err)
	require.NotNil(t, ctr1)

	cli, err := testcontainers.NewDockerClientWithOpts(context.Background())
	require.NoError(t, err)
	require.NotNil(t, cli)

	err = cli.ContainerPause(context.Background(), ctr1.GetContainerID())
	require.NoError(t, err)

	ctr2, err := postgres.Run(context.Background(), "postgres:14-bookworm", opts...)
	if err != nil {
		t.Logf("run postgres testcontainer: %s", err.Error())
	} else {
		t.Logf("postgres testcontainer is running: %t", ctr2.IsRunning())
	}

	require.NoError(t, err)
	require.NotNil(t, ctr2)
}
```

This test fails for `ctr2` to be run, and this is because it needs to be `unpaused` first, which is the error message I see in the description of the PR.

I tested it against `main` and your code with the same results, although I spotted a different bug for reused containers: the code to skip starting a container if Ryuk was disabled I think is incorrect, and should apply always, checking if the container is in the unpaused state too (cc/ @stevenh)

docker.go:L1367
```diff
-	if p.config.RyukDisabled {
-		// ryuk disabled, ensure the container we are about to reuse is running
-		if c.State != "running" {
-			// found container to reuse, but it is not in state running
-			if err := dc.Start(ctx); err != nil {
-				return dc, fmt.Errorf("start container %s in state %s: %w", req.Name, c.State, err)
-			}
-		}
-	}
+	// Ensure the container we are about to reuse is running,
+	// but not if it is paused.
+	if c.State != "running" && c.State != "paused" {
+		// found container to reuse, but it is not in state running
+		if err := dc.Start(ctx); err != nil {
+			return dc, fmt.Errorf("start container %s in state %s: %w", req.Name, c.State, err)
+		}
+	}
```

OTOH, testcontainers-go does not provide native support for Pausing/Unpausing containers, unless using the underlying Docker client we expose (see my test).

Said that, I'm not sure how to proceed with this PR

mdelapenya:
@georgespalding I added a commit with the related fix that I found

docker.go
```diff
@@ -1364,13 +1364,17 @@ func (p *DockerProvider) ReuseOrCreateContainer(ctx context.Context, req Contain
...
 	}
 
-	if p.config.RyukDisabled {
-		// ryuk disabled, ensure the container we are about to reuse is running
-		if c.State != "running" {
-			// found container to reuse, but it is not in state running
-			if err := dc.Start(ctx); err != nil {
-				return dc, fmt.Errorf("start container %s in state %s: %w", req.Name, c.State, err)
-			}
+	// If a container was stopped programmatically, and Ryuk is enabled, we want to
```
stevenh, line 1367: question: the comment details Ryuk state but there's no check for this, should there be?

docker.go
```diff
@@ -1364,13 +1364,17 @@ func (p *DockerProvider) ReuseOrCreateContainer(ctx context.Context, req Contain
...
+	// ensure the container is running again, but only if it is not paused, as it's
+	// not possible to start a paused container. The Docker Engine returns the
+	// "cannot start a paused container, try unpause instead" error.
+	switch c.State {
+	case "running", "paused":
```
stevenh, line 1372: question: should we unpause it or error?

reuse_test.go
```diff
@@ -0,0 +1,39 @@
...
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/testcontainers/testcontainers-go"
+)
+
+func TestGenericContainer_stop_start_withReuse(t *testing.T) {
+	containerName := "my-postgres"
```
stevenh, line 13: nit: change to my-nginx to avoid confusion given the image used.

stevenh (commented):
A few clarifying questions

reuse_test.go
mdelapenya: Oh yes, I started with postgres, but reduced the test size. Changing it asap

docker.go
mdelapenya: Mmm I added it because the original message included it 🤔 But you're right, and readers don't care about Ryuk's state in this case. Removing that asap

docker.go
mdelapenya: I added the //TODO message

stevenh: Would it be better to just error for paused, indicating it's not supported, that way it will fail fast?

mdelapenya: Yeah, I like the idea, thanks!

PR #3063, opened 2025-03-28
chore(gcloud): deprecate old gcp containers, creating subpackages for them

mdelapenya:
- **chore: use new version (v0.36.0) in modules and examples**
- **chore: prepare for next minor development cycle (0.37.0)**
- **chore: migrate BigQuery to its own package**
- **chore: migrate BigTable to its own package**
- **chore: migrate Datastore to its own package**
- **chore: migrate Firestore to its own package**
- **chore: migrate Pubsub to its own package**
- **chore: migrate Spanner to its own package**
- **chore: use shared options**

## What does this PR do?
Following the new Azure module, this PR creates subpackages inside the `gcloud` module, so that each sspecialised container is able to offer a different set of functional options to customise it.

As a result, gcloud exports new packages to interact with the different services of the emulator: bigquery, bigtable, datastore, firestore, pubsub and spanner.

The old containers have been deprecated, so we recommend moving to the new ones as soon as possible.

Because all the services used a shared ProjectID configuration, we created an internal shared package to handle it, creating package and variable aliases in all the service packages, so they are able to reuse the same code, maintaining their package namespace.

The docs has been updated, and the lint and tests pass.

## Why is it important?
Improve the maintainability of each service in an independent manner. I.e. the bigquery service uses a WithDataYAML option that must not be used in the other services.

As an example, #3009 adds a new option to the firestore container, which could be added to its own package with ease. 

## Related issues

- Relates #3009 (cc/ @Zarux)


modules/gcloud/bigtable.go
```diff
@@ -7,12 +7,13 @@ import (
 	"github.com/testcontainers/testcontainers-go/wait"
 )
 
-// Deprecated: use RunBigTable instead
+// Deprecated: use [RunBigTable] instead
```
stevenh, line 10: question: this looks like a deprecation chain, could this point to the ultimate desired function?

modules/gcloud/datastore.go
```diff
@@ -7,12 +7,13 @@ import (
 	"github.com/testcontainers/testcontainers-go/wait"
 )
 
-// Deprecated: use RunDatastore instead
+// Deprecated: use [RunDatastore] instead
```
stevenh, line 10: question: as above deprecation chain?

modules/gcloud/firestore.go
```diff
@@ -7,12 +7,13 @@ import (
 	"github.com/testcontainers/testcontainers-go/wait"
 )
 
-// Deprecated: use RunFirestore instead
+// Deprecated: use [RunFirestore] instead
```
stevenh, line 10: question: deprecation chain?

stevenh (commented):
Looking good, just question on the deprecation chain.

modules/gcloud/bigtable.go
mdelapenya: Yeah, I was thinking about that, if pointing to the new function or keep it as is. I'll directly link to the new ones, everywhere

mdelapenya:
@stevenh I think this is ready to go 🚀

stevenh (approved):
LGTM :shipit:

PR #3070, opened 2025-03-31
feat: add more functional options to the modules API

mdelapenya:
- **chore: deprecate request fields that can be modified by the existing modifiers**
- **feat: add the missing functional options for the container request**
- **chore: refactor option tests to follow the test helper function pattern**
- **docs: document new functional options**
- **chore: do not use postgres in functional options docs**
- **chore(localstack): use new customisers instead of the override**
- **chore: apply deprecations**

## What does this PR do?
This PR adds the missing functional options to the modules API:

- `testcontainers.WithExposedPort(ports ...string)`
- `testcontainers.WithEntrypoint(entrypoint ...string)`
- `testcontainers.WithCmd(cmd ...string)`
- `testcontainers.WithLabels(labels map[string]string)`
- `testcontainers.WithFiles(files ...testcontainers.ContainerFile)`
- `testcontainers.WithMounts(mounts ...testcontainers.ContainerMount)`
- `testcontainers.WithTmpfs(tmpfs map[string]string)`

At the same time, it deprecates some fields of the ContainerRequest that won't be used anymore, but instead should be configured using the proper ConfigModifier:

- Hostname: use the `req.ConfigModifier` instead
- WorkingDir: use the `req.ConfigModifier` instead
- Privileged: use the `req.HostConfigModifier` instead
- User: use the `req.ConfigModifier` instead
- ShmSize: use the `req.HostConfigModifier` instead

The usage of those fields has been updated across the library, and the default config modifiers has been updated to support passing those values in the case the client code stil uses the deprecated fields.

All options has been added to the docs.

## Why is it important?
Keep the basic creation of containers simple, delegating advanced features to the modifiers. Also adding all supported fields as functional options makes the modules API feature complete in terms of customisation, making the override container request option not needed for common use cases.

## Follow-ups
I'm thinking about creating a `GenericContainer` new module, which is able to consume all the options as any other module, therefore, the core library would just be the foundation for intearacting with the docker engine, and the modules would be the real meat for Testcontainers users.


options.go
```diff
@@ -334,3 +334,69 @@ func WithWaitStrategyAndDeadline(deadline time.Duration, strategies ...wait.Stra
 		return nil
 	}
 }
+
```
mdelapenya, line 337: In case there is a discussion about variadic Vs single arguments in the new options, I left a comment here: https://github.com/testcontainers/testcontainers-go/pull/3071#discussion_r2022567464

docker_test.go
```diff
@@ -1758,12 +1764,16 @@ func TestContainerRunningCheckingStatusCode(t *testing.T) {
 }
 
 func TestContainerWithUserID(t *testing.T) {
+	expectedUserID := "60125"
```
stevenh, line 1767: suggestion: use a const

docs/features/common_functional_options.md
```diff
@@ -32,7 +128,7 @@ postgres, err = postgresModule.Run(ctx, "postgres:15-alpine", testcontainers.Wit
 If you need to access a port that is already running in the host, you can use `testcontainers.WithHostPortAccess` for example:
 
 ```golang
-postgres, err = postgresModule.Run(ctx, "postgres:15-alpine", testcontainers.WithHostPortAccess(8080))
+ctr, err = aModule.Run(ctx, "image:tag", testcontainers.WithHostPortAccess(8080))
```
stevenh, line 131: suggestion: use an idiomatic name, as I read that and thought it was a bug vs a placeholder.
More below

options.go
```diff
@@ -41,6 +41,15 @@ func CustomizeRequest(src GenericContainerRequest) CustomizeRequestOption {
 	}
 }
 
+// BuildFromDockerfile allows to build a container from a Dockerfile
+func BuildFromDockerfile(ctx context.Context, df FromDockerfile) CustomizeRequestOption {
```
stevenh, line 45: suggestion: remove ctx as it's not used.

stevenh (changes requested):
Like the consistency improvement, just a few suggestions

docker_test.go
```diff
@@ -1779,7 +1789,7 @@ func TestContainerWithUserID(t *testing.T) {
 	b, err := io.ReadAll(r)
 	require.NoError(t, err)
 	actual := regexp.MustCompile(`\D+`).ReplaceAllString(string(b), "")
-	assert.Equal(t, req.User, actual)
+	assert.Equal(t, expectedUserID, actual)
```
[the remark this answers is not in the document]
mdelapenya: I sometimes hate Cursor when it suggests too much code that overlaps the block you're interested in updating

options.go
mdelapenya: I thought I had removed it 🤦

options.go
```diff
@@ -334,3 +343,69 @@ func WithWaitStrategyAndDeadline(deadline time.Duration, strategies ...wait.Stra
...
+// WithEntrypoint appends the entrypoint to the entrypoint for a container
+func WithEntrypoint(entrypoint ...string) CustomizeRequestOption {
+	return func(req *GenericContainerRequest) error {
+		req.Entrypoint = append(req.Entrypoint, entrypoint...)
+		return nil
+	}
+}
+
+// WithExposedPorts appends the ports to the exposed ports for a container
+func WithExposedPorts(ports ...string) CustomizeRequestOption {
```
mdelapenya, line 356: question: do we want to pass strings, or nat.Port? :thinking:

stevenh: my initial reaction as string, but nat.Port might have it's benefits. The key for me is usability, can I do:
```go
testcontainers.WithExposedPorts("8080")
```

stevenh (changes requested):
Looking good, concerned about a few which are appending when I would expect to replace due to how they work.

options.go
```diff
@@ -334,3 +343,69 @@ func WithWaitStrategyAndDeadline(deadline time.Duration, strategies ...wait.Stra
...
 	}
 }
+
+// WithEntrypoint appends the entrypoint to the entrypoint for a container
+func WithEntrypoint(entrypoint ...string) CustomizeRequestOption {
+	return func(req *GenericContainerRequest) error {
+		req.Entrypoint = append(req.Entrypoint, entrypoint...)
+		return nil
+	}
+}
```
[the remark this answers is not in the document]
mdelapenya: We are never overriding, but "appending". PTAL at the existing WithWaitStrategy, that uses ForAll to not lose the initial value.

I.e. a module defining its own entrypoint, would see it being overridden by this function if we replace instead of append.

options.go
mdelapenya: Yes, because nat.Port is declared as a new string type:

```go
// Port is a string containing port number and protocol in the format "80/tcp"
type Port string
```

stevenh: Yer it works for literals but not for variables unless defined correctly, is that an issue?
Example: https://go.dev/play/p/nfjRMSr94-G

mdelapenya: let use string then

stevenh (approved):
LGTM

PR #3071, opened 2025-03-31
feat: add socat container

mdelapenya:
- **chore: create socat module from modulegen**
- **chore: implement socat**
- **chore: bump tc-go**
- **chore: mod tidy**
- **chore: fix examples**

## What does this PR do?
It adds a SOCAT container, a utility container that provides TCP port forwarding and network tunneling between services, enabling transparent communication between containers and networks.

This is particularly useful in testing scenarios where you need to simulate network connections or provide transparent access to services running in different containers.

## Why is it important?
It can be used alongside other modules as a helper for network communication

## Related issues

- Relates https://github.com/testcontainers/testcontainers-node/pull/964
- Relates https://github.com/testcontainers/testcontainers-python/pull/795

## Other concerns

Other languages are implementing the socat container as part of the core, and we are discussing about it here too. The socat module in this PR just uses the stdlib plus the `github.com/docker/go-connections/nat` pacakge, so all the pulled packages would already be there when using tc-go.

I understand the benefits for having it in the core library, although with the current project layout it would be more difficult to add it directly in the core, without a package to enclose the functionality: we wouldn't use the same APIs for running it as in the modules, with the `socat.Run` function. If in the core, we should:

1. Create a socat package, which is basically the same as having it in the module
2. Add it to the core but exposed as testcontainers.RunSocat which would lose a bit of the semantics provided by the package namespace plus the Run function (as in the rest of the modules)


modules/socat/options.go
```diff
@@ -0,0 +1,76 @@
...
+	return Target{
+		exposedPort:  exposedPort,
+		internalPort: internalPort,
+		host:         host,
+	}
+}
+
+// WithTargets sets the targets for the socat container.
+// The host of each target must be without the port, as it is internally mapped to the exposed port.
+func WithTargets(targets ...Target) Option {
```
stevenh, line 59: question: would it be better to use multiple WithTarget aor WithTargetInternalPort?

modules/socat/socat.go
```diff
@@ -0,0 +1,83 @@
...
+func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustomizer) (*Container, error) {
+	req := testcontainers.GenericContainerRequest{
+		ContainerRequest: testcontainers.ContainerRequest{
+			Image:      img,
+			Entrypoint: []string{"/bin/sh"},
+		},
+		Started: true,
+	}
+
+	// 2. Gather all config options (defaults and then apply provided options)
```
stevenh, line 32: suggestion: remove the 2. given there's no 1.

modules/socat/options.go
```diff
@@ -0,0 +1,76 @@
...
+	return options{
+		Targets: map[int]Target{},
+	}
+}
+
+// Compiler check to ensure that Option implements the testcontainers.ContainerCustomizer interface.
+var _ testcontainers.ContainerCustomizer = (Option)(nil)
+
+// Option is an option for the Redpanda container.
+type Option func(*options)
```
stevenh, line 26: suggestion: return an error to future proof

modules/socat/socat.go
```diff
@@ -0,0 +1,83 @@
...
+		targetURLs[k] = targetURL
+	}
+
+	c.targetURLs = targetURLs
+
+	return c, nil
+}
+
+// TargetURL returns the URL for the target's exposed port, nil if the port is not mapped
+func (c *Container) TargetURL(t Target) *url.URL {
```
stevenh, line 81: suggestion: just pass in the port to simplify and only provide the needed info.

testing.go
```diff
@@ -53,6 +53,21 @@ func SkipIfDockerDesktop(t *testing.T, ctx context.Context) {
 	}
 }
 
+// SkipIfNotDockerDesktop is a utility function capable of skipping tests
+// if tests are not run using Docker Desktop.
+func SkipIfNotDockerDesktop(t *testing.T, ctx context.Context) {
```
stevenh, line 58: kudos: nice little helper.

stevenh (changes requested):
Couple of questions but no blockers.

modules/socat/options.go
mdelapenya: this is one of the dilemmas I recently have: allow variadic arguments to do the job, or pass the same option multiple times... 

I do not have a strong opinion. The variadic argument is more concise Vs the explicitness of the multi-usage of the option.

What are your thoughts on this?

testing.go
mdelapenya: I was rebasing -i to remove one commit and I thought I had removed this block, used for a test including the new Docker Model Runner... but indeed, it's fine to have it

modules/socat/options.go
mdelapenya: On the other hand, I'm looking at #3070, and for those structured types, I see no issue in passing them as variadic arguments instead 🤔 

I'm fine with keeping the single argument here for the targets, as it's very scoped to the socat container. But for the other PR, I see the value in adding them as they are passed as collections

mdelapenya:
@stevenh I think this is ready to go 🚀

.github/dependabot.yml
```diff
@@ -63,6 +63,7 @@ updates:
         - /modules/redpanda
         - /modules/registry
         - /modules/scylladb
+        - /modules/socat
```
eddumelendez, line 66: not a blocker, but IMO adding it to `core` will be great as a utility container implementation.

eddumelendez (commented):
implementation LGTM, just wonder if should be a module or keep it in `core`

.github/dependabot.yml
mdelapenya: @eddumelendez could you elaborate a bit? I started with that in mind but considered the modules are easier to consume, dependencies wise. But I'm not opposed to sing it to them core

mdelapenya: In this case, the socat module just uses the stdlib so all the pulled packages would already be there when using tc-go.

I understand the benefits for having it in the core, although with the current project layout would be more difficult to add it directly in the core, without a package to enclose the functionality. I.e. not using the same APIs for running it with the `socat.Run` function. If in the core, we should:
1. Create a socat package, which is basically the same as the module
2. Add it to the core but exposed as `testcontainers.RunSocat` which would lose a bit of the semantics above

PR #3072, opened 2025-03-31
feat(influxdb): support for influxdbv2

### LINKED ISSUE #3066: [Feature]: Support for InfluxDB v2
### Problem

As far as I could tell from the documentation, only InfluxDB v1 is supported.
Influxdb V2 could be deployed using `influxdb.Run(context.Background(), "influxdb:2.7.8))`, but the container won't be set up properly as the environment variables for Docker differ from the v1.

### Solution

Adding support for InfluxDB V2 with updated environment variables. 
Also, it would make sense to add two more functions: 
- `GetConnectionURL` that return a connection URL for the influxdb client
- `GetInfluxClient` that returns an InfluxDB V2 compatible API client

### Benefit

Support for InfluxDB V2. Ease of use for the functions.

### Alternatives

Influxdb V2 could be deployed using `influxdb.Run(context.Background(), "influxdb:2.7.8))`, but the container won't be set up properly as the environment variables for Docker differ from the v1.

Yes

xBlaz3kx:
Feature

## What does this PR do?

I added a `WithV2Env` option to the InfluxDB module to setup the V2 instance correctly according to their confguration guidelines. Added tests for changes and updated docs/examples for setting up a V2 instance.

## Why is it important?

Lack of documentation on how to setup V2 instances and prettified the workaround with a neater solution (custom options with configuration for V2).

## Related issues

- Closes #3066

```
	ctx := context.Background()

	username := "username"
	password := "password"
	org := "org"
	bucket := "bucket"
	authEnabled := true
	token := "influxdbv2token"

	influxdbContainer, err := influxdb.Run(ctx,
		"influxdb:2.7.11",
		influxdb.WithV2Env(influxdb.InfluxDBV2Config{
			Username:    &username,
			Password:    &password,
			Org:         org,
			Bucket:      bucket,
			Token:       &token,
			AuthEnabled: &authEnabled,
		}),
	)
	defer func() {
		if err := testcontainers.TerminateContainer(influxdbContainer); err != nil {
			log.Printf("failed to terminate container: %s", err)
		}
	}()
	if err != nil {
		log.Printf("failed to start container: %s", err)
		return
	}
	// }

	state, err := influxdbContainer.State(ctx)
	if err != nil {
		log.Printf("failed to get container state: %s", err)
		return
	}

	fmt.Println(state.Running)
	// Output:
	// true

	// Query the InfluxDB API to verify the setup
	url, err := influxdbContainer.ConnectionUrl(ctx)
	if err != nil {
		log.Printf("failed to get host: %s", err)
		return
	}

	// Initialize a new InfluxDB client
	client := influxdb2.NewClientWithOptions(url, token, influxdb2.DefaultOptions())
	defer client.Close()

	// Get the bucket
	influxBucket, err := client.BucketsAPI().FindBucketByName(ctx, bucket)
	if err != nil {
		log.Printf("failed to get bucket: %s", err)
		return
	}

	log.Printf("Got bucket: %s", influxBucket.Name)
```


modules/influxdb/influxdb.go
```diff
@@ -118,6 +118,62 @@ func WithConfigFile(configFile string) testcontainers.CustomizeRequestOption {
...
+	PasswordFile *string // File containing the passwords (e.g., /run/secrets/password)
+	TokenFile    *string // File containing the token (e.g., /run/secrets/token)
+	Org          string  // Organization name
+	Bucket       string  // Bucket name
+	Retention    *string // Retention policy (e.g., "30d" for 30 days)
+	Token        *string // Admin token
+	AuthEnabled  *bool   // Enable authentication (true/false); defaults to false.
+}
+
+func WithV2Env(config InfluxDBV2Config) testcontainers.CustomizeRequestOption {
```
mdelapenya, line 134: bug: missing docs for the exposed function

modules/influxdb/influxdb.go
```diff
@@ -118,6 +118,62 @@ func WithConfigFile(configFile string) testcontainers.CustomizeRequestOption {
 	}
 }
 
+type InfluxDBV2Config struct {
+	Username     *string // Username for the initial user
+	Password     *string // Password for the initial user
+	UsernameFile *string // File containing the usernames (e.g., /run/secrets/username)
```
stevenh, line 124: question: this says usernames is that a typo?

Same for passwords

modules/influxdb/influxdb.go
```diff
@@ -118,6 +118,62 @@ func WithConfigFile(configFile string) testcontainers.CustomizeRequestOption {
 	}
 }
 
+type InfluxDBV2Config struct {
+	Username     *string // Username for the initial user
```
stevenh, line 122: suggestion: make the types without a pointer to make their use easier. Zero value can be used to check if they need setting still.

modules/influxdb/influxdb.go
stevenh, line 134: @mdelapenya should this be a seperate module so it keeps the dependencies clean, or is that just too much duplication?

stevenh, line 134: suggestion: add a doc comment.

@mdelapenya we should enable the linter for this.

modules/influxdb/influxdb_test.go
```diff
@@ -44,6 +45,54 @@ func TestV2Container(t *testing.T) {
...
+			Bucket:      bucket,
+			Token:       &token,
+			AuthEnabled: &authEnabled,
+		}),
+	)
+	testcontainers.CleanupContainer(t, influxdbContainer)
+
+	state, err := influxdbContainer.State(ctx)
+	require.NoError(t, err)
+	assert.True(t, state.Running)
```
stevenh, line 72: suggestion: use assert for all tests for consistency, more below.

stevenh (changes requested):
Thanks for PR, a few suggestions and clarifying question

modules/influxdb/influxdb.go
xBlaz3kx: Afaik he suggested to just create an option that would override the default env. I dont think there is any benefit in separating the modules at this moment.

modules/influxdb/influxdb_test.go
xBlaz3kx: Well other tests use `require` everywhere, but I would argue it really depends when to use `require` and `assert`. `Require` will abort the tests, so I used it when checking for critical errors (as further testing makes no sense)

modules/influxdb/influxdb.go
xBlaz3kx: But it can happen that user provided a default value unintentionally. With pointers, we can guarantee the user wanted to change the value. Also this way we would distinct the **required** vs **optional** values.

modules/influxdb/influxdb.go
xBlaz3kx: Yeah its a typo. [This](https://docs.influxdata.com/influxdb/v2/install/use-docker-compose/) is the reference for the config, in case you're wondering where the naming came from.

modules/influxdb/influxdb.go
xBlaz3kx: Other functions in the module also don't have docs. Should I add them in scope of this PR?

stevenh: Feel free, in the goal of improving code quality and usability all exported should have comments. Personally I even tend to do internal ones :)

modules/influxdb/influxdb_test.go
stevenh: The problem with assert is used wrong you get bunch of additional noise, thats the reason why we're standarising on require for all tests. This has the benefit of reducing the cognitive load on review, as you don't have to consider all the nuances of the content of each test.

xBlaz3kx: Sure, I can use `require` instead of `assert` to keep things consistent.

modules/influxdb/influxdb.go
mdelapenya: This is a tough topic, as we do not have a support matrix for a module version corresponding a technology version.

I'd go with a best effort to support as much as possible without compromising the maintainability. If adding the option for V2 is enough and clear for users, I'd go with that.

modules/influxdb/influxdb.go
```diff
@@ -118,6 +119,65 @@ func WithConfigFile(configFile string) testcontainers.CustomizeRequestOption {
 	}
 }
 
+// InfluxDBV2Config contains the configuration for InfluxDB v2.
+type InfluxDBV2Config struct {
```
mdelapenya, line 123: suggestion: let's not include the package name into the type to avoid redundancy in calls (i.e. `influxdb.ConfigV2` Vs `influxdb.InfluxDBV2Config`)
```suggestion
type ConfigV2 struct {
```

modules/influxdb/influxdb_test.go
```diff
@@ -44,6 +45,200 @@ func TestV2Container(t *testing.T) {
...
+
+			opt := influxdb.WithV2Env(test.influxConfig)
+			_ = opt(genericReq)
+
+			assert.Equal(t, test.expectedEnvMap, genericReq.Env)
+		})
+	}
+}
+
+func TestV2Container_WithOptions(t *testing.T) {
```
mdelapenya, line 194:
```suggestion
func TestRun_v2WithOptions(t *testing.T) {
```

modules/influxdb/influxdb_test.go
```diff
@@ -7,6 +7,7 @@ import (
 	"testing"
 	"time"
 
+	influxdb2 "github.com/influxdata/influxdb-client-go/v2"
```
mdelapenya, line 10: suggestion: for consistency with the v1 import
```suggestion
	influxclient2 "github.com/influxdata/influxdb-client-go/v2"
```

modules/influxdb/influxdb_test.go
```diff
@@ -44,6 +45,200 @@ func TestV2Container(t *testing.T) {
 	require.Truef(t, state.Running, "InfluxDB container is not running")
 }
 
+func TestWithV2Env(t *testing.T) {
+	username, password := "root", "password"
+	token := "token"
+	retention := "30d"
+	usernameFile, passwordFile, tokenFile := "username.txt", "password.txt", "token.txt"
+
+	tests := []struct {
```
mdelapenya, line 54: suggestion: we are adoption a different pattern for testing multiple values in tests. Instead of defining a test-table, we are adopting using a test helper function receiving the parameters to be tests (the code block inside the for loop of the table), and one `t.Run` subtest per entry in the table. This allows us to select one subtests to be run (in the IDE or in console, filtering go test with the subtest name). We need then subtests names without spaces: "with-required-only")

modules/influxdb/examples_test.go
```diff
@@ -41,3 +43,69 @@ func ExampleRun() {
...
+	defer client.Close()
+
+	// Get the bucket
+	influxBucket, err := client.BucketsAPI().FindBucketByName(ctx, bucket)
+	if err != nil {
+		log.Printf("failed to get bucket: %s", err)
+		return
+	}
+
+	log.Printf("Got bucket: %s", influxBucket.Name)
```
mdelapenya, line 106: bug: this is not honouring the testable example format, causing the tests to fail https://github.com/testcontainers/testcontainers-go/actions/runs/14207921652/job/39809564404#step:7:437
```suggestion
	fmt.Println(influxBucket.Name)
```

modules/influxdb/examples_test.go
```diff
@@ -41,3 +43,69 @@ func ExampleRun() {
...
+
+	// Output:
+	// true
+	// Got bucket: bucket
```
mdelapenya, line 110: suggestion: simpler
```suggestion
	// bucket
```

mdelapenya (commented):
Left a few comments. I do not have a strong opinion on the Nil Vs zero values discussion, as I see the value in separating required Vs optional values using pointers.

We are usually setting zero values, in general, without complains for our users, so I'd say we are fine not using pointers, checking the zero value was not set (booleans raising hands atm 😄 )

modules/influxdb/influxdb_test.go
xBlaz3kx: I mean sure, I can fix the naming, but I'd leave it as table test. Should be picked up by the IDE regardless. Also I saw there are multiple instances of table tests in the repository.

mdelapenya: > there are multiple instances of table tests in the repository

yes, and we are adopting the different pattern for new code (obviously not documented 😞 )

In any case, it's a recommendation, not a blocker, although we'd prefer adding new code in that format if possible

docs/modules/influxdb.md
```diff
@@ -54,7 +57,7 @@ Use the second argument in the `Run` function to set a valid Docker image.
 In example: `Run(context.Background(), "influxdb:1.8.0")`.
 
 !!!info
-    Note that `influxdb:latest` will get you a version 2 image which is not supported by this module.
+    Note that `influxdb:latest` will pull a version 2 image.
```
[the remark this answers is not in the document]
xBlaz3kx: Could you clarify what "Not available until the new release" should do? Any examples? I've updated the docs in [a7980d8](https://github.com/testcontainers/testcontainers-go/pull/3072/commits/a7980d82f5d5d25ddc7529620914147affdaf7c4), but not sure if this is what was requested.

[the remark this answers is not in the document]
mdelapenya: Yes, sorry: we add this line right after each new option: https://github.com/testcontainers/testcontainers-go/blob/322a05707b5895eac7e475582453e05f7f9e5789/docs/modules/socat.md?plain=1#L32. The release process will automatically replace it at release time.

modules/influxdb/influxdb.go
stevenh: I'm not sure I follow, could you explain how a user can provide a value unintentionally?

Using pointers makes it much harder to use as can't use direct values, at least not without a generics helper.

Something simple like specifying a username and password, which could be:
```go
Withv2Env(ConfigV2{Username: "myuser", Password: "mypasswd"})
```
becomes:
```go
username := "myuser"
password := "mypasswd"
Withv2Env(ConfigV2{Username: &username, Password: &password})
```

modules/influxdb/influxdb_test.go
stevenh: Here's an [example in ollama](https://github.com/testcontainers/testcontainers-go/blob/main/modules/ollama/local_test.go#L52)

As a direct example from this:
```go
	t.Run("with-required-only", func(t *testing.T) {
		genericReq := &testcontainers.GenericContainerRequest{
			ContainerRequest: testcontainers.ContainerRequest{}
		}
		opt := influxdb.WithV2Env(
			influxdb.ConfigV2{
				Org:    "Org",
				Bucket: "Bucket",
			},
		}
		err := opt.Customise(&genericReq)
		require.NoError(t, err)

		require.Equal(t, map[string]string{
			"DOCKER_INFLUXDB_INIT_MODE":         "setup",
			"DOCKER_INFLUXDB_INIT_ORG":          "Org",
			"DOCKER_INFLUXDB_INIT_BUCKET":       "Bucket",
			"DOCKER_INFLUXDB_INIT_AUTH_ENABLED": "false",
		}, request.Env}
	}

	t.Run("with-username-and-password", func(t *testing.T) {
		influxConfig: influxdb.ConfigV2{
			Org:      "Org",
			Bucket:   "Bucket",
			Username: &username,
			Password: &password,
		},
		err := opt.Customise(&genericReq)
		require.NoError(t, err)

		require.Equal(t, map[string]string{
			"DOCKER_INFLUXDB_INIT_MODE":         "setup",
			"DOCKER_INFLUXDB_INIT_ORG":          "Org",
			"DOCKER_INFLUXDB_INIT_BUCKET":       "Bucket",
			"DOCKER_INFLUXDB_INIT_AUTH_ENABLED": "false",
			"DOCKER_INFLUXDB_INIT_USERNAME":     username,
			"DOCKER_INFLUXDB_INIT_PASSWORD":     password,
		}, request.Env}
	}
```

modules/influxdb/influxdb.go
stevenh: Oh and there's the other issue that the values could be altered between pass in the run, which can result in unexpected behaviour.

modules/influxdb/influxdb.go
```diff
@@ -120,6 +121,65 @@ func WithConfigFile(configFile string) testcontainers.CustomizeRequestOption {
 	}
 }
 
+// ConfigV2 contains the configuration for InfluxDB v2.
+type ConfigV2 struct {
```
stevenh, line 125: suggestion: thinking about this some more we shouldn't use a big struct like this as it's the anti pattern of functional options. Instead you should have multiple options each which serves a single purpose. For example:

```go
func withV2Base(req *testcontainers.GenericContainerRequest, org, bucket string) {
	req.Env["DOCKER_INFLUXDB_INIT_ORG"] = org
	req.Env["DOCKER_INFLUXDB_INIT_BUCKET"] = bucket
	req.Env["DOCKER_INFLUXDB_INIT_MODE"] = "setup" // Always setup, we wont be migrating from v1 to v2
}

func secretsPath(path string) string {
	return dockerSecretPath + "/" + path
}

func WithV2Auth(org, bucket, username, password string) testcontainers.CustomizeRequestOption {
		withV2Base(req, org, bucket)
		req.Env["DOCKER_INFLUXDB_INIT_USERNAME"] = username
		req.Env["DOCKER_INFLUXDB_INIT_PASSWORD"] = password
		// TODO: check for conflicting options e.g. WithV2SecretsAuth

		return nil
	}
}

func WithV2SecretsAuth(org, bucket, usernameFile, passwordFile string) testcontainers.CustomizeRequestOption {
		withV2Base(req, org, bucket)
		req.Env["DOCKER_INFLUXDB_INIT_USERNAME_FILE"] = secretsPath(usernameFile)
		req.Env["DOCKER_INFLUXDB_INIT_PASSWORD_FILE"] = secretsPath(passwordFile)
		// TODO: check for conflicting options e.g. WithV2Auth

		return nil
	}
}

func WithV2Retention(retention time.duration) testcontainers.CustomizeRequestOption {
		req.Env["DOCKER_INFLUXDB_INIT_RETENTION"] = retention.String()

		return nil
	}
}

func WithV2AdminToken(token string) testcontainers.CustomizeRequestOption {
		req.Env["DOCKER_INFLUXDB_INIT_ADMIN_TOKEN"] = token
		// TODO: check for conflicting options e.g. WithV2SecretsAdminToken

		return nil
	}
}

func WithV2SecretsAdminToken(tokenFile string) testcontainers.CustomizeRequestOption {
		req.Env["DOCKER_INFLUXDB_INIT_ADMIN_TOKEN_FILE"] = secretsPath(tokenFile)
		// TODO: check for conflicting options e.g. WithV2AdminToken

		return nil
	}
}
```

Not sure if this is the right split, but you get the idea. This makes it easier to use and more idiomatic.

stevenh (changes requested):
Updated with clarification of previous asks, hopefully this makes sense.

Having thought about this more, the big config struct is an antipattern to functional options, sorry for not spotting this earlier, see comments for details.

modules/influxdb/influxdb.go
```diff
@@ -120,6 +121,65 @@ func WithConfigFile(configFile string) testcontainers.CustomizeRequestOption {
...
+		if config.Password != nil {
+			req.Env["DOCKER_INFLUXDB_INIT_PASSWORD"] = *config.Password
+		}
+
+		if config.Token != nil {
+			req.Env["DOCKER_INFLUXDB_INIT_ADMIN_TOKEN"] = *config.Token
+		}
+
+		if config.AuthEnabled != nil {
+			req.Env["DOCKER_INFLUXDB_INIT_AUTH_ENABLED"] = strconv.FormatBool(*config.AuthEnabled)
```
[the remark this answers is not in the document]
xBlaz3kx: In the [specs](https://docs.influxdata.com/influxdb/v2/install/use-docker-compose/) and also going through their setup procedure in the Dockerfile.

modules/influxdb/influxdb.go
mdelapenya: Oh you're right @stevenh thanks for pointing it out, and sorry @xBlaz3kx for my initial recommendation. I had an online conversation with @stevenh and I now get it more correctly.

His recommendation for the granular options is the way to go. Thanks!

modules/influxdb/influxdb.go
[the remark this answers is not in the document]
stevenh: that's where I looked and there's no mention of `DOCKER_INFLUXDB_INIT_AUTH_ENABLED` also no mention in [their repo ](https://github.com/search?q=repo%3Ainfluxdata%2Finfluxdata-docker%20DOCKER_INFLUXDB_INIT_AUTH_ENABLED&type=code) am I missing something?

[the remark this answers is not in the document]
xBlaz3kx: Hm, odd, I was sure it was there. Maybe I was referencing an older version or it was removed, unsure.

modules/influxdb/influxdb.go
```diff
@@ -120,6 +122,142 @@ func WithConfigFile(configFile string) testcontainers.CustomizeRequestOption {
...
+
+		return nil
+	}
+}
+
+const dockerSecretPath = "/run/secrets"
+
+func secretsPath(path string) string {
+	return dockerSecretPath + "/" + path
+}
```
mdelapenya, line 157: question: this means that a file must be created in the container, right? How are we setting it? I see the env var initialisation but I'm not able yet to check where the file is copied into the container. Is that assumption correct?

mdelapenya: Never mind, I realised this is the Docker secrets path. Candidate to live in the core for a follow-up

mdelapenya:
@xBlaz3kx I added a commit on top of yours, if you approve, I'm merging this PR, thanks!

PR #3073, opened 2025-03-31
fix(dind): use docker image load

acouvreur:
Loading images with docker image import does not preserve metatdata such as entrypoint, cmd, labels and tags. This makes the loaded image unusable as is.

This replaces the command with docker image load which fully preserves the metadata.

## How to test this PR

Using the previous implementation will have the new test fail because the entrypoint and cmd are not preserved.


mdelapenya (approved):
LGTM, thanks for noticing this 🙏

PR #3075, opened 2025-04-01
chore(deps): replace `golang.org/x/exp/slices` with stdlib

Juneezee:
## What does this PR do?

Replace `golang.org/x/exp/slices` with `slices`

## Why is it important?

The experimental functions in `golang.org/x/exp/slices` are now available in the standard library in Go 1.21.

Reference: https://go.dev/doc/go1.21#slices


mdelapenya (approved):
LGTM, thanks!

stevenh (approved):
Thanks for this

PR #3082, opened 2025-04-02
chore: bump golangci-lint to v2

mdelapenya:
## What does this PR do?
This PR updates the golangci-lint versino to v2, migrating the existing configuration file with this command:

```shell
golangci-lint migrate -c .golangci.yml
```

Before that, I had to update my local version of golangci-lint to the new release, using the `make lint` command with the updated version and path to the package (updated in this PR too).

Finally, I ran the lint to all modules, fixing the following errors:

- add `staticcheck` to the existing exclusions 
- do not start errors with uppercase
- do not end errors with punctuation
- replace require.Equal with require.Empty where needed
- skip embedded type in method calls. I.e.: a.EmbeddedType.Foo => a.Foo
- deprecate the existing errors in the wait package that are not starting with Err. New ones appear as replacement.

## Why is it important?
golangci-lint v2 came, and if we want to continue using the tool, we need to update our toolchain.


mdelapenya:
@mmorel-35 as a golangci-lint advocate, I'd appreciate your review here 🙏

PR #3091, opened 2025-04-08
feat: create an internal sdk for docker contexts and docker configs

mdelapenya:
- **feat: create a seed for an sdk in the internal package**
- **chore: wrap errors**
- **chore: use new dockerconfig package instead of cpuguy83's**
- **chore: run mod tidy**

## What does this PR do?
This PR adds two packages into a new internal docker package:

- context: locates and loads docker contexts from the docker config path.
- config: locates and reads docker config from the docker config path.

The config package embedes the cpuyguy83's `dockercfg` package, to have control on the config and auth types.

## Why is it important?
Both new packages could represent the seed for an eventual docker-sdk-go package, that would naturally emerge from this internal set of packages.

They are internal in order to iterate faster, and not break clients of testcontainers-go.

## Related issues

- #2810 would benefit from this new context package, so it must be updated to use it instead.


mdelapenya:
@stevenh I think this PR would be the foundation for the internal SDK in which the work for the new docker client could land.

Can we prioritise this PR if possible?

internal/docker/context/context.go
```diff
@@ -0,0 +1,109 @@
...
+	// metadataDir is the name of the directory containing the metadata
+	metadataDir = "meta"
+)
+
+// ErrDockerHostNotSet is the error returned when the Docker host is not set in the Docker context
+var ErrDockerHostNotSet = internal.ErrDockerHostNotSet
+
+// getContextFromEnv returns the context name from the environment variables.
+func getContextFromEnv() string {
+	if os.Getenv(EnvOverrideHost) != "" {
```
stevenh, line 46: bug: this looks like it should be an == ""?

internal/docker/context/context.go
```diff
@@ -0,0 +1,109 @@
...
+		return DefaultContextName
+	}
+
+	if ctxName := os.Getenv(EnvOverrideContext); ctxName != "" {
+		return ctxName
+	}
+
+	return ""
```
stevenh, line 54: suggestion: simplify to only one read so it also closes the race condition.

internal/docker/context/context.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// Current returns the current context name, based on
+// environment variables and the cli configuration file. It does not
+// validate if the given context exists or if it's valid.
+//
+// If the current context is not found, it returns the default context name.
```
stevenh, line 61: suggestion: eliminate conflict between statement that it doesn't validate and the not found check.

internal/docker/context/context.go
```diff
@@ -0,0 +1,109 @@
...
+func Current() (string, error) {
```
stevenh, line 62: question: is current the best name? It seems odd as all functions are typically point in time, possibly Name or something similar.

stevenh (changes requested):

internal/docker/context/context.go
mdelapenya: I've seen current and active as vamos in the Docker docs, although current seems part of the response when setting a context:

```shell 
docker context use docker-test
docker-test
Current context is now "docker-test"
```

internal/docker/context/context.go
mdelapenya: https://github.com/docker/cli/blob/61c6818f0b0c3caf80685b319be22f8babd11586/cli/command/cli.go#L449-L467

This is the original code from docker/cli, so it must work as is

internal/docker/context/context.go
mdelapenya: we are reading two different variables, is that what you meant?

internal/docker/context/context.go
mdelapenya: Also, the docker-config file contains this entry: `"currentContext": "value",`

Please check out with `cat ~/.docker/config.json`

internal/docker/context/context.go
stevenh: Interesting as that specifically breaks context if you specify a host. This is documented in the comment for `CurrentContext`. We should replicate that so folks know the history if we want to keep that behaviour as it's quite unexpected.

internal/docker/context/context.go
stevenh: It's not wrong, and as you said matches the use elsewhere, I just found it odd / superfluous.

PR #3092, opened 2025-04-08
feat: add toxiproxy module

mdelapenya:
- **chore: initial toxiproxy module**
- **chore: add examples and tests for toxiproxy**
- **docs: update container types in code examples**
- **docs: add examples**
- **docs: document exposed variables**

## What does this PR do?
This PR converts the existing example into a tc-go module.

It adds a functional option to modify the number of exposed ports in the range used by toxiproxy to add proxies.

This PR, intentionally, does not couple the implementation of the module with any toxiproxy Go client. On the other hand,
it uses `github.com/Shopify/toxiproxy/v2` for the tests and examples.

Finally, it removes the existing example module, converting that code into the testable examples of the new module.

## Why is it important?
More modules!


docs/modules/scylladb.md
```diff
@@ -30,7 +30,7 @@ go get github.com/testcontainers/testcontainers-go/modules/scylladb
 The ScyllaDB module exposes one entrypoint function to create the ScyllaDB container, and this function receives three parameters:
 
 ```golang
-func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustomizer) (*ScyllaDBContainer, error)
+func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustomizer) (*Container, error)
```
mdelapenya, line 33: docs: leftovers from previous merges

docs/modules/socat.md
```diff
@@ -34,7 +34,7 @@ go get github.com/testcontainers/testcontainers-go/modules/socat
 The Socat module exposes one entrypoint function to create the Socat container, and this function receives three parameters:
 
 ```golang
-func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustomizer) (*SocatContainer, error)
+func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustomizer) (*Container, error)
```
mdelapenya, line 37: docs: leftovers from previous merges

modules/toxiproxy/toxiproxy.go
```diff
@@ -0,0 +1,83 @@
...
+
+const (
+	// ControlPort is the port of the Toxiproxy control API
+	ControlPort = "8474/tcp"
+
+	// firstProxiedPort is the first port of the range of ports that will be proxied
+	firstProxiedPort = 8666
+
+	// defaultPortRange is the default range of ports that will be proxied
+	defaultPortRange = 31
```
eddumelendez, line 20: I've been wondering if we should have a default or just let the users define the container ports they want to open. Meanwhile debugging and performing `docker ps`, all 31 ports will be displayed and maybe just a couple are used, which is troublesome.

So, my proposal is adding a new fn that would simplify the code below:

```go
toBeMergeRequest := testcontainers.GenericContainerRequest{
		ExposedPorts: []string{"8666/tcp"},
	},
}

toxiproxyContainer, err := tctoxiproxy.Run(
	ctx,
	"ghcr.io/shopify/toxiproxy:2.12.0",
	testcontainers.CustomizeRequest(toBeMergeRequest),
	network.WithNetwork([]string{"toxiproxy"}, nw),
)
```

I couldn't find `testcontainers.WithExposedPort`, so, maybe good idea to have it?

This is different to what tc-java does but I think would provide better experience in general. WDYT?

mdelapenya: That new option is added in #3070

mdelapenya: Regarding the ability to customize the default range, we added a functional option for it, see options.go

mdelapenya: Ahh I see your point now: instead of defining a fixed number of random ports, let the user add a proxy connection (host and port). That would allow users have just those ports that are required. WDYT?

eddumelendez: Only adding exposed ports to the container.

If you need to define a [config file](https://github.com/eddumelendez/testcontainers-samples/blob/main/spring-boot-r2dbc-postgresql-chaos/src/test/resources/toxiproxy.json) for toxiproxy then check this code to [configure toxiproxy with config file](https://github.com/eddumelendez/testcontainers-samples/blob/main/spring-boot-r2dbc-postgresql-chaos/src/test/java/com/example/springbootr2dbcpostgresqlchaos/SpringBootR2dbcPostgresqlChaosConfigCliApplicationTests.java#L40-L42). I think this will be nice to add to tc-go as well.

mdelapenya: we would need to parse the config file looking for proxies to expose their ports, right?

eddumelendez: not exactly what I meant. Passing the config file with new `toxiproxy.WithConfigFile(file)` and setting the right command would be enough IMO.

mdelapenya:
@eddumelendez I added the ability to add proxies and generate the config on the fly. Could you PTAL?

modules/toxiproxy/options_test.go
```diff
@@ -0,0 +1,182 @@
...
+	testcontainers.CleanupContainer(t, redisContainer)
+	require.NoError(t, err)
+
+	toxiproxyContainer, err := Run(
+		ctx,
+		"ghcr.io/shopify/toxiproxy:2.12.0",
+		// the config file defines a proxy named "redis"
+		WithConfigFile(strings.NewReader(configContent)),
+		// this proxy will be added to the existing proxies
+		WithProxy("redis2", "redis2:6379"),
```
eddumelendez, line 133: the documentation states that `WithProxy` uses ports starting with 8666. However, it depends on the order of those functions, for the given example, in order to access redis the mapped port is the one with 8666 and for redis2 is 8667. If the order changes, then redis uses 8667 and redis2 8666. I see how convenient this is but it is adding another option to create proxies.

Existing options to create proxies:
* [Using Toxiproxy CLI](https://github.com/eddumelendez/testcontainers-samples/blob/main/spring-boot-r2dbc-postgresql-chaos/src/test/java/com/example/springbootr2dbcpostgresqlchaos/SpringBootR2dbcPostgresqlChaosCliApplicationTests.java#L44)
* Using Toxiproxy Client
* Using Config file.

On top of that, this is assuming we want to expose all the proxies ports. There could be cases where you just want to introduce latency among two containers.

docs/modules/toxiproxy.md
```diff
@@ -0,0 +1,140 @@
...
+```golang
+func WithProxy(name string, upstream string) Option
+```
+
+If this option is used in combination with the `WithConfigFile` option, the proxy defined in this option
+is added to the proxies defined in the config file.
+
+!!!info
+    If you add proxies in a programmatic manner using the Toxiproxy client, then you need to manually
+    add exposed ports in the Toxiproxy container.
```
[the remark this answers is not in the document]
mdelapenya: https://deploy-preview-3092--testcontainers-go.netlify.app/modules/toxiproxy/#__tabbed_5_1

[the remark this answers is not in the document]
eddumelendez: can we include how proxyPort is created?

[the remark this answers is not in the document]
mdelapenya: There is a specific section with examples: https://deploy-preview-3092--testcontainers-go.netlify.app/modules/toxiproxy/#__tabbed_6_1

I can link to it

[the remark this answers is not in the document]
mdelapenya: I added a new code snippet for it

modules/toxiproxy/options_test.go
eddumelendez: > We could explicitly add the ability to expose ports in the WithProxy option.

As I said in the original comment I really see how convenient is this fn. However, it also open the question to what else should be supported. For example, can we support the `enabled` feature as well? what about toxic definitions? I'd prefer to delegate this to existing clients for maintenance purposes.

> OTOH, do you see a real concern by exposing all proxied ports, even they are not in use?

It's just not nice to have all those displaying in the terminal while debugging. This is based on real experience :D and share it based on tc-java. This is not an issue anymore in the PR because it is opening ports as they are declared.

> The docs say that the ports start in 8666, but we can extend the docs to explicitly say that the orders of the ports determines the exposed port, starting from 8666.

At the beginning, I thought the config file was created first and then will add the entry defined in `WithProxy`. I had to debug it in order to understand and check if the order impacts, wish I was assuming will not bus just wanted to double check.

mdelapenya: With the current choose, is possible to add proxies with both a config file, as requested, and a single proxy, using the same concept: get all of them and build a config file to be added to the container. 

Then users can list the existing proxies with the go client and do whatever they need with it, even disabling it. IIRC there is an example about this. 

Regarding ports, this code is but defining a range, so the huge list of exposed ports is not a problem anymore, because just the proxied ports will be exposed. 

Do you think the docs are not descriptive enough? Let's improve then to make them more clear

eddumelendez (dismissed):

modules/toxiproxy/config.go
```diff
@@ -0,0 +1,78 @@
...
+	Upstream     string `json:"upstream"`
+	upstreamIP   string
+	upstreamPort int
+	Enabled      bool `json:"enabled"`
+}
+
+// sanitize is a helper function that returns another proxy
+// in which all the fields have been extracted from the
+// string representation of the upstream and listen fields.
+func (p proxy) sanitize() (proxy, error) {
```
stevenh, line 24: suggestion: act on a pointer so you can just return an error.
```suggestion
func (p *proxy) sanitize() error {
```

modules/toxiproxy/config.go
```diff
@@ -0,0 +1,78 @@
...
+		upstreamIP:   upstreamIP,
+		upstreamPort: upstreamPort,
+		Enabled:      p.Enabled,
+	}
+
+	return pp, nil
+}
+
+// newProxy creates a new proxy configuration with default values
+func newProxy(name string, upstream string) (proxy, error) {
```
stevenh, line 62: suggestion: return a pointer so you can return nil on error, it's also idiomatic to return a pointer from new functions.

modules/toxiproxy/config.go
```diff
@@ -0,0 +1,78 @@
...
+// proxy represents a single proxy configuration in the toxiproxy config file
+type proxy struct {
+	Name         string `json:"name"`
+	Listen       string `json:"listen"`
+	listenIP     string
+	listenPort   int
+	Upstream     string `json:"upstream"`
+	upstreamIP   string
+	upstreamPort int
+	Enabled      bool `json:"enabled"`
```
stevenh, line 18: suggestion: group public and private fields so it's easier to tell them appart.

stevenh (changes requested):
A quick pass with some idiomatic go suggestions.

mdelapenya:
@stevenh comments addressed. Thank you for your thorough review 🙇

mdelapenya:
I'm merging this. Let's open an issue if we have last minute questions on it, thanks!

Issue #2542, opened 2024-05-18
[Bug]: Postgres Testcontainers for Golang: WithInitScripts with multiple scripts does not work


DiplomateProgrammer:
### Testcontainers version

0.31.0

### Using the latest Testcontainers version?

Yes

### Host OS

Windows

### Host arch

x86

### Go version

1.22.3

### Docker version

```shell
Client:
 Cloud integration: v1.0.35+desktop.13
 Version:           26.1.1
 API version:       1.45
 Go version:        go1.21.9
 Git commit:        4cf5afa
 Built:             Tue Apr 30 11:48:43 2024
 OS/Arch:           windows/amd64
 Context:           default

Server: Docker Desktop 4.30.0 (149282)
 Engine:
  Version:          26.1.1
  API version:      1.45 (minimum version 1.24)
  Go version:       go1.21.9
  Git commit:       ac2de55
  Built:            Tue Apr 30 11:48:28 2024
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.31
  GitCommit:        e377cd56a71523140ca6ae87e30244719194a521
 runc:
  Version:          1.1.12
  GitCommit:        v1.1.12-0-g51d5e94
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### Docker info

```shell
Client:
 Version:    26.1.1
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.14.0-desktop.1
    Path:     C:\Program Files\Docker\cli-plugins\docker-buildx.exe
  compose: Docker Compose (Docker Inc.)
    Version:  v2.27.0-desktop.2
    Path:     C:\Program Files\Docker\cli-plugins\docker-compose.exe
  debug: Get a shell into any image or container (Docker Inc.)
    Version:  0.0.29
    Path:     C:\Program Files\Docker\cli-plugins\docker-debug.exe
  dev: Docker Dev Environments (Docker Inc.)
    Version:  v0.1.2
    Path:     C:\Program Files\Docker\cli-plugins\docker-dev.exe
  extension: Manages Docker extensions (Docker Inc.)
    Version:  v0.2.23
    Path:     C:\Program Files\Docker\cli-plugins\docker-extension.exe
  feedback: Provide feedback, right in your terminal! (Docker Inc.)
    Version:  v1.0.4
[...]

PR #3093, opened 2025-04-08
feat(postgres): add WithOrderedInitScripts for Postgres testcontainers

### LINKED ISSUE #3090: [Enhancement]: Postgres WithInitScripts files should be run in the order they are being passed
### Proposal

Hi! Thanks for the great work here.

I ran into a problem while trying to write some tests using testcontainers, specifically the Postgres module.

The files that are passed to `WithInitScripts` are copied inside the container, and I assume - at least that's what my tests are showing - that they are then run in alphanumerical order. I believe this is surprising behavior, I expect the init scripts to be run in the order I passed them to testcontainers.

Someone ran into this behavior before and raised an issue : https://github.com/testcontainers/testcontainers-go/issues/2542, then they probably figured it out and closed it. I think this is proof that this behavior is surprising, or at least not documented.

What I suggest is that the files get copied inside the container and prefixed with their ordering the list, so that for example : 

`.WithInitScripts("file.sql", "second.sql", "final.sql")`
-> (Current behavior is that file.sql runs first, final.sql runs second and second runs last.) :x:
-> The files should get copied inside the container as `00-file.sql`, `01-second.sql`, `02-final.sql`. That way they get run in the right order.

# Considerations and workaround

I'm assuming this is a breaking change because it changes behavior. Users relying on the ordering by filename will run into issues when updating the package. 

It can also be that the current behavior is expected and wanted, in which case this should at least be documented.

As a workaround, users can implement the logic described above themselves (which is what I'll be doing in the short term)

Thoughts?

AlisCode:
Type of change: Enhancement

## What does this PR do?

As suggested in the discussion, this adds a brand new option to add init scripts to your Postgres testcontainer, and run them in the order that the user provides them to the function.

It also further documents `WithInitScripts` to clarify in what order the scripts are executed. 

## Why is it important?

The files that are passed to `WithInitScripts` are copied inside the container, then run in alphanumerical order. I believe this is surprising behavior, as I expect the init scripts to be run in the order I passed them to testcontainers.

Someone ran into this behavior before and raised an issue : https://github.com/testcontainers/testcontainers-go/issues/2542

## Related issues

- Closes #3090

## How to test this PR

Try using the newly-added function `WithOrderedInitScripts` for the postgres container.
Any setup using `WithInitScripts` should work just like before.

### DIFF
--- a/modules/postgres/postgres.go
+++ b/modules/postgres/postgres.go
@@ -95,13 +95,12 @@ func WithDatabase(dbName string) testcontainers.CustomizeRequestOption {
 }
 
 // WithInitScripts sets the init scripts to be run when the container starts.
-// These init scripts will be executed in sorted name order as defined by the current locale, which defaults to en_US.utf8.
+// These init scripts will be executed in sorted name order as defined by the container's current locale, which defaults to en_US.utf8.
 // If you need to run your scripts in a specific order, consider using `WithOrderedInitScripts` instead.
 func WithInitScripts(scripts ...string) testcontainers.CustomizeRequestOption {
 	return func(req *testcontainers.GenericContainerRequest) error {
 		for _, script := range scripts {
-			filename := filepath.Base(script)
-			appendInitScript(req, script, filename)
+			appendInitScript(req, script, filepath.Base(script))
 		}
 		return nil
 	}
@@ -112,8 +111,7 @@ func WithInitScripts(scripts ...string) testcontainers.CustomizeRequestOption {
 func WithOrderedInitScripts(scripts ...string) testcontainers.CustomizeRequestOption {
 	return func(req *testcontainers.GenericContainerRequest) error {
 		for idx, script := range scripts {
-			filename := filepath.Base(script)
-			containerFilePath := fmt.Sprintf("%06d-%s", idx, filename)
+			containerFilePath := fmt.Sprintf("%03d-%s", idx, filepath.Base(script))
 			appendInitScript(req, script, containerFilePath)
 		}
 		return nil
--- a/modules/postgres/postgres.go
+++ b/modules/postgres/postgres.go
@@ -94,22 +94,40 @@ func WithDatabase(dbName string) testcontainers.CustomizeRequestOption {
 	}
 }
 
-// WithInitScripts sets the init scripts to be run when the container starts
+// WithInitScripts sets the init scripts to be run when the container starts.
+// These init scripts will be executed in sorted name order as defined by the current locale, which defaults to en_US.utf8.
+// If you need to run your scripts in a specific order, consider using `WithOrderedInitScripts` instead.
 func WithInitScripts(scripts ...string) testcontainers.CustomizeRequestOption {
 	return func(req *testcontainers.GenericContainerRequest) error {
-		initScripts := []testcontainers.ContainerFile{}
 		for _, script := range scripts {
-			cf := testcontainers.ContainerFile{
-				HostFilePath:      script,
-				ContainerFilePath: "/docker-entrypoint-initdb.d/" + filepath.Base(script),
-				FileMode:          0o755,
-			}
-			initScripts = append(initScripts, cf)
+			filename := filepath.Base(script)
+			appendInitScript(req, script, filename)
 		}
-		req.Files = append(req.Files, initScripts...)
-
 		return nil
 	}
+}
+
+// WithOrderedInitScripts sets the init scripts to be run when the container starts.
+// The scripts will be run in the order that they are provided in this function.
+func WithOrderedInitScripts(scripts ...string) testcontainers.CustomizeRequestOption {
+	return func(req *testcontainers.GenericContainerRequest) error {
+		for idx, script := range scripts {
+			filename := filepath.Base(script)
+			containerFilePath := fmt.Sprintf("%06d-%s", idx, filename)
+			appendInitScript(req, script, containerFilePath)
+		}
+		return nil
+	}
+}
+
+// Adds an initialization script to the Postgres container
+func appendInitScript(req *testcontainers.GenericContainerRequest, hostFilePath string, containerFilePath string) {
+	cf := testcontainers.ContainerFile{
+		HostFilePath:      hostFilePath,
+		ContainerFilePath: "/docker-entrypoint-initdb.d/" + containerFilePath,
+		FileMode:          0o755,
+	}
+	req.Files = append(req.Files, cf)
 }
 
 // WithPassword sets the initial password of the user to be created when the container starts
--- a/modules/postgres/postgres.go
+++ b/modules/postgres/postgres.go
@@ -99,10 +99,12 @@ func WithDatabase(dbName string) testcontainers.CustomizeRequestOption {
 // If you need to run your scripts in a specific order, consider using `WithOrderedInitScripts` instead.
 func WithInitScripts(scripts ...string) testcontainers.CustomizeRequestOption {
 	return func(req *testcontainers.GenericContainerRequest) error {
+		containerFiles := []testcontainers.ContainerFile{}
 		for _, script := range scripts {
-			appendInitScript(req, script, filepath.Base(script))
+			initScript := buildInitScriptContainerFile(script, filepath.Base(script))
+			containerFiles = append(containerFiles, initScript)
 		}
-		return nil
+		return testcontainers.WithFiles(containerFiles...)(req)
 	}
 }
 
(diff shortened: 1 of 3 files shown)


mdelapenya:
A first look and this looks good. Could you please add tests verifying the new behavior?

AlisCode:
I sure can, would you mind pointing me to a test that I can take inspiration from ?

mdelapenya:
Sure, please take as look here: https://github.com/testcontainers/testcontainers-go/blob/322a05707b5895eac7e475582453e05f7f9e5789/modules/postgres/postgres_test.go#L262

AlisCode:
Hi! I've added a test that checks that both scripts are executed in the order they are provided. If the behavior changes or becomes inconsistent somehow, and insert-user.sql gets executed first, then the test should break.

modules/postgres/postgres_test.go
```diff
@@ -288,6 +288,44 @@ func TestWithInitScript(t *testing.T) {
...
 
+func TestWithOrderedInitScript(t *testing.T) {
+	ctx := context.Background()
+
+	ctr, err := postgres.Run(ctx,
+		"postgres:15.2-alpine",
+		// Executes first the init-user-db shell-script, then the insert-user SQL script
+		postgres.WithOrderedInitScripts(
+			filepath.Join("testdata", "init-user-db.sh"),
+			filepath.Join("testdata", "insert-user.sql"),
```
mdelapenya, line 299: suggestion: because `init-user-db` < `insert-user`, it will be applied first even without the changes in the PR.

I'd simply make sure the logic in this PR is satisfy, probably checking the files are renamed in the target dir of the container 🤔  Thoughts?

AlisCode: I'm happy to do that, but I couldnt find an easy way to list the files that are inside the container after it's created.

What do you think of just renaming `insert-user.sql` to `do-insert-user.sql` ?

mdelapenya: That would work, just making sure the insert file has a name that goes first (i.e. `aaaa-insert-user.sql` for clearness) and if the database is there it means the files where processed correctly.

For checking the content of the directory, you could `ctr.Exec(ctx, []string{"ls -l /path/to/files"})`, get&read the reader output, and verify the names included the new prefixes

AlisCode: Done. Thanks for the guidance :)

modules/postgres/postgres.go
```diff
@@ -94,24 +94,42 @@ func WithDatabase(dbName string) testcontainers.CustomizeRequestOption {
 	}
 }
 
-// WithInitScripts sets the init scripts to be run when the container starts
+// WithInitScripts sets the init scripts to be run when the container starts.
+// These init scripts will be executed in sorted name order as defined by the current locale, which defaults to en_US.utf8.
```
mdelapenya, line 98:
```suggestion
// These init scripts will be executed in sorted name order as defined by the container's current locale, which defaults to en_US.utf8.
```

modules/postgres/postgres.go
```diff
@@ -94,24 +94,42 @@ func WithDatabase(dbName string) testcontainers.CustomizeRequestOption {
...
 	return func(req *testcontainers.GenericContainerRequest) error {
-		initScripts := []testcontainers.ContainerFile{}
 		for _, script := range scripts {
-			cf := testcontainers.ContainerFile{
-				HostFilePath:      script,
-				ContainerFilePath: "/docker-entrypoint-initdb.d/" + filepath.Base(script),
-				FileMode:          0o755,
-			}
-			initScripts = append(initScripts, cf)
+			filename := filepath.Base(script)
```
mdelapenya, line 103: suggestion: inline the variable if it's not used anymore

modules/postgres/postgres.go
```diff
@@ -94,24 +94,42 @@ func WithDatabase(dbName string) testcontainers.CustomizeRequestOption {
...
+		return nil
+	}
+}
 
+// WithOrderedInitScripts sets the init scripts to be run when the container starts.
+// The scripts will be run in the order that they are provided in this function.
+func WithOrderedInitScripts(scripts ...string) testcontainers.CustomizeRequestOption {
+	return func(req *testcontainers.GenericContainerRequest) error {
+		for idx, script := range scripts {
+			filename := filepath.Base(script)
```
mdelapenya, line 115: suggestion: inline the variable if it's not used anymore

modules/postgres/postgres.go
```diff
@@ -94,24 +94,42 @@ func WithDatabase(dbName string) testcontainers.CustomizeRequestOption {
...
+			containerFilePath := fmt.Sprintf("%06d-%s", idx, filename)
```
mdelapenya, line 116: nit, not a blocker: let's use a smaller number. Not sure if we need thousands of files 😅. Please update the tests too if agreed 
```suggestion
			containerFilePath := fmt.Sprintf("%03d-%s", idx, filename)
```

mdelapenya (commented):
LGTM, just added some minor comments

Thanks!

AlisCode:
@mdelapenya I've taken the time to address all those comments :) 

Enjoy!

modules/postgres/postgres.go
mdelapenya: I guess we need to update the tests here https://github.com/testcontainers/testcontainers-go/pull/3093/commits/8c32f0f2e0d24d731d3bea1f864b844bf4f98abc#diff-929bacfd366ab1235318667cb165ce78b7ed08b9c56a0d48f5a29ef6b4379b00R324

AlisCode: Oops. Sorry for the noise :)

modules/postgres/postgres.go
```diff
@@ -94,24 +94,40 @@ func WithDatabase(dbName string) testcontainers.CustomizeRequestOption {
...
+		for idx, script := range scripts {
+			containerFilePath := fmt.Sprintf("%03d-%s", idx, filepath.Base(script))
+			appendInitScript(req, script, containerFilePath)
+		}
 		return nil
 	}
 }
 
+// Adds an initialization script to the Postgres container
+func appendInitScript(req *testcontainers.GenericContainerRequest, hostFilePath string, containerFilePath string) {
```
mdelapenya, line 122: BTW, we just merged an option to append container files, I think we can simplify this code here, just calling the WithFiles option: https://golang.testcontainers.org/modules/postgres/#withfiles

Thoughts?

AlisCode: Addressed

modules/postgres/postgres.go
```diff
@@ -94,21 +94,40 @@ func WithDatabase(dbName string) testcontainers.CustomizeRequestOption {
...
+			containerFilePath := fmt.Sprintf("%03d-%s", idx, filepath.Base(script))
+			initScript := buildInitScriptContainerFile(script, containerFilePath)
+			containerFiles = append(containerFiles, initScript)
+		}
+		return testcontainers.WithFiles(containerFiles...)(req)
+	}
+}
+
+// Adds the file located at hostFilePath as an init script that postgres will launch
+func buildInitScriptContainerFile(hostFilePath string, containerFilePath string) testcontainers.ContainerFile {
```
mdelapenya, line 126: question: do we need this simple function that just returns the struct? I'd consider removing it as it does not do anything special inside. But I'll let you decide

mdelapenya (commented):
LGTM; although I'd like to have the new option documented at `docs/modules/postgres.md`. Please do not forget adding the `Not available...` marker 🙏  (https://github.com/testcontainers/testcontainers-go/blob/main/docs/modules/arangodb.md?plain=1#L62)

modules/postgres/postgres.go
AlisCode: You asked to share some code between `WithOrderedInitScripts` and `WithInitScripts`, so this function does just that. The only alternative I can think of is duplicating the `ContainerFile` declaration in the 2 functions, so I believe this is better?

mdelapenya: Yes, I asked that before having the new WithFiles options, which now removes the duplication. I'm fine having it as a function, although with the current implementation I see it as a pass-through function. In any case, it's not a blocker to me.

Once the docs are added, I'm happy to merge this 😊

mdelapenya:
@AlisCode friendly ping 🙏

mdelapenya:
@AlisCode given this PR is sent from your main branch, I cannot contribute to it with the requested changes.

I'm pushing a new branch with all your commits, adding mine on top, so that you appear as contributor.

PR #3106, opened 2025-04-15
feat: add Docker Model Runner module

mdelapenya:
- **chore: initial commit for the Docker Model Runner module**
- **chore: add missing module to the VSCode settings**
- **chore: add example on how to pull a model**
- **feat: wait for DMR to be ready**
- **chore: verify the creation of the model returns 200 OK**
- **feat: support for listing the local models**
- **chore: make sure socat is installed in the socat module**
- **feat: support for getting a local model**
- **chore: skip tests for if not running on Docker Desktop**
- **chore: add testable examples**
- **fix: use proper model response type**
- **chore: add example with openai-go**
- **chore: add example with langchaingo**

## What does this PR do?
Adds a module for the Docker Model Runner, including examples with langchaingo and openai-go.

It also exposes Pull, Get, List methods on the container to pull models, get a model and list all the available models in the DMR.

## Why is it important?
New module for DMR


docs/modules/dockermodelrunner.md
```diff
@@ -0,0 +1,100 @@
...
+- `testcontainers.ContainerCustomizer`, a variadic argument for passing options.
+
+### Container Options
+
+When starting the Docker Model Runner container, you can pass options in a variadic way to configure it.
+
+#### Image
+
+Use the second argument in the `Run` function to set a valid Docker image.
+In example: `Run(context.Background(), "alpine/socat:1.8.0.1")`.
```
kiview, line 46: **suggestion (non-blocking)**: I think we need to explain here, that it needs to be a `socat` image.

docs/modules/dockermodelrunner.md
```diff
@@ -0,0 +1,100 @@
...
+
+{% include "../features/common_functional_options.md" %}
+
+#### WithModel
+
+- Not available until the next release of testcontainers-go <a href="https://github.com/testcontainers/testcontainers-go"><span class="tc-version">:material-tag: main</span></a>
+
+Use the `WithModel` option to set the model to pull when the container is started.
```
kiview, line 54:
```suggestion
Use the `WithModel` option to set the model to pull when the container is started.
Please be aware, that only Models as OCI Artifacts are compatible with Docker Model Runner.
```

modules/dockermodelrunner/dockermodelrunner.go
```diff
@@ -0,0 +1,129 @@
...
+		if err != nil {
+			return c, fmt.Errorf("pull model: %w", err)
+		}
+	}
+
+	return c, nil
+}
+
+// GetModel returns a model that is already pulled using the Docker Model Runner format.
+func (c *Container) GetModel(ctx context.Context, namespace string, name string) (types.ModelResponse, error) {
```
kiview, line 81: **question**: Shouldn't that be `InspectModel`, to map against the CLI and against other APIs?

modules/dockermodelrunner/dockermodelrunner.go
```diff
@@ -0,0 +1,129 @@
...
+
+// Container represents the DockerModelRunner container type used in the module
+type Container struct {
+	*socat.Container
+	model   string
+	baseURL string
+}
+
+// Run creates an instance of the DockerModelRunner container type
+func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustomizer) (*Container, error) {
```
kiview, line 31: **suggestion**: Provide a default value for `img` (which Socat image to provide seems like a very involved implementation detail). 

We don't have a `asCompatibleSubstitute` check in tc-go, right? Else it would make sense to check for this against the provided image name, but if we don't do it in tc-go right now, ignore this suggestion.

kiview (approved):

modules/dockermodelrunner/dockermodelrunner.go
mdelapenya: We decided not using default images, that's why the APIs are forcing the pass of the image. What the module does, is to check if the `socat` binary is installed. Else, the module fails

kiview: I am still in favor of a default image for this module, since the role of the image is a different one here (utility, rather than the actual content of the module). But if it is awkward do it with the current API design, we can leave it.

mdelapenya: Because there is no interface (yet) to satisfy the Run function contract, we can define the Run function for the DMR module in the way we need it

docs/modules/dockermodelrunner.md
mdelapenya: Done, thanks!

modules/dockermodelrunner/dockermodelrunner.go
```diff
@@ -0,0 +1,134 @@
...
+	if err != nil {
+		return models, fmt.Errorf("list models: %w", err)
+	}
+
+	return models, nil
+}
+
+// PullModel pulls a model from the Docker Model Runner
+func (c *Container) PullModel(ctx context.Context, model string) error {
+	log.Default().Printf("🙏 Pulling model %s. Please be patient, no progress bar yet!", model)
```
stevenh, line 114: suggestion: Definitely agree a progress bar option would be good the amount of time needed

modules/dockermodelrunner/dockermodelrunner_test.go
```diff
@@ -0,0 +1,96 @@
...
+	})
+
+	t.Run("success/with-image", func(t *testing.T) {
+		ctr, err := dockermodelrunner.RunWithImage(ctx, socat.DefaultImage)
+		testcontainers.CleanupContainer(t, ctr)
+		require.NoError(t, err)
+	})
+
+	t.Run("failure/with-image", func(t *testing.T) {
+		_, err := dockermodelrunner.RunWithImage(ctx, "alpine:latest")
```
stevenh, line 39: bug: while its expecting an error it should handle not getting one, so call CleanupContainer.

modules/dockermodelrunner/main_test.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"context"
+	"log"
+	"os"
+	"testing"
+
+	"github.com/testcontainers/testcontainers-go"
+)
+
+func TestMain(m *testing.M) {
```
stevenh, line 12: suggestion: this won't show the tests as skipped, use a helper include a once if needed.

modules/dockermodelrunner/sdk/client/client.go
```diff
@@ -0,0 +1,23 @@
+package client
+
+// openAIEndpointSuffix is the suffix for the OpenAI endpoint
+const openAIEndpointSuffix = "/engines/v1"
+
+// Client is the client for the Docker Model Runner
+type Client struct {
+	BaseURL        string
```
stevenh, line 8: suggestion: make private.

modules/dockermodelrunner/sdk/client/create.go
```diff
@@ -0,0 +1,42 @@
...
+	"io"
+	"net/http"
+	"strings"
+)
+
+// CreateModel creates a model
+func (c *Client) CreateModel(ctx context.Context, fqmn string) ([]byte, error) {
+	var bytes []byte
+
+	payload := fmt.Sprintf(`{"from": "%s"}`, fqmn)
```
stevenh, line 15: suggestion: use %q

modules/dockermodelrunner/sdk/client/create.go
```diff
@@ -0,0 +1,42 @@
...
+	reqURL := c.BaseURL + "/models/create"
+
+	httpClient := http.Client{}
```
stevenh, line 18: suggestion: use the default client if we don't need anything special

modules/dockermodelrunner/sdk/client/create.go
```diff
@@ -0,0 +1,42 @@
...
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, reqURL, strings.NewReader(payload))
+	if err != nil {
+		return bytes, fmt.Errorf("new post request (%s): %w", reqURL, err)
```
stevenh, line 22: suggestion: return nil instead of bytes to avoid confusion, more below.

modules/dockermodelrunner/sdk/client/create.go
```diff
@@ -0,0 +1,42 @@
...
+	if resp.StatusCode != http.StatusOK {
+		return bytes, fmt.Errorf("status code: %d", resp.StatusCode)
+	}
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return bytes, fmt.Errorf("read all: %w", err)
+	}
+	bytes = body
+	return bytes, nil
```
stevenh, line 41: question: do we really want to force this to read into memory?

modules/dockermodelrunner/sdk/client/inspect.go
```diff
@@ -0,0 +1,43 @@
...
+	"net/http"
+
+	"github.com/testcontainers/testcontainers-go/modules/dockermodelrunner/sdk/types"
+)
+
+// InspectModel returns a model by namespace and name
+func (c *Client) InspectModel(ctx context.Context, namespace string, name string) (types.ModelResponse, error) {
+	var model types.ModelResponse
+
+	httpClient := http.Client{}
```
stevenh, line 17: suggestion: use the default client

modules/dockermodelrunner/sdk/client/inspect.go
```diff
@@ -0,0 +1,43 @@
...
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+
+	"github.com/testcontainers/testcontainers-go/modules/dockermodelrunner/sdk/types"
+)
+
+// InspectModel returns a model by namespace and name
+func (c *Client) InspectModel(ctx context.Context, namespace string, name string) (types.ModelResponse, error) {
```
stevenh, line 14: suggestion: return a pointer so it can be nil on error.

modules/dockermodelrunner/sdk/client/inspect.go
```diff
@@ -0,0 +1,43 @@
...
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return model, fmt.Errorf("read all: %w", err)
+	}
+
+	err = json.Unmarshal(body, &model)
+	if err != nil {
+		return model, fmt.Errorf("json unmarshal: %w", err)
+	}
```
stevenh, line 40: suggestion: parse on the fly instead of reading into memory.

modules/dockermodelrunner/sdk/client/list.go
```diff
@@ -0,0 +1,43 @@
...
+	"net/http"
+
+	"github.com/testcontainers/testcontainers-go/modules/dockermodelrunner/sdk/types"
+)
+
+// ListModels lists all models
+func (c *Client) ListModels(ctx context.Context) ([]types.ModelResponse, error) {
+	var models []types.ModelResponse
+
+	httpClient := http.Client{}
```
stevenh, line 17: suggestion: same comments as per InspectModel

modules/dockermodelrunner/sdk/types/types.go
```diff
@@ -0,0 +1,16 @@
...
+	ID string `json:"id"`
+
+	// Tags are the tags of the model
+	Tags []string `json:"tags"`
+
+	// Config is the config of the model
+	Config map[string]string `json:"config"`
+
+	// Created is the creation time of the model
+	Created int64 `json:"created"`
```
stevenh, line 15: question: can this be a time.Time so its more idiomatic?

stevenh (changes requested):

modules/dockermodelrunner/dockermodelrunner.go
```diff
@@ -0,0 +1,134 @@
...
+
+// Container represents the DockerModelRunner container type used in the module
+type Container struct {
+	*socat.Container
+	model   string
+	baseURL string
+}
+
+// Run creates an instance of the DockerModelRunner container type.
+func Run(ctx context.Context, opts ...testcontainers.ContainerCustomizer) (*Container, error) {
```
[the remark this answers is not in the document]
mdelapenya: Me neither! I'd have liked to use this API since the beginning too

modules/dockermodelrunner/dockermodelrunner.go
```diff
@@ -0,0 +1,134 @@
...
+		c = &Container{Container: socatCtr, model: settings.model}
+	}
+
+	if err != nil {
+		return c, fmt.Errorf("socat run: %w", err)
+	}
+
+	c.baseURL = socatCtr.TargetURL(modelRunnerPort).String()
+
+	if settings.model != "" {
```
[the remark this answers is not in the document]
mdelapenya: @kiview do you thing the withModel option is mandatory? @stevenh's concerns are really valid in regards of usability, even more without having a progress bar that informs the user what's going on.

modules/dockermodelrunner/dockermodelrunner.go
mdelapenya: @kiview I followed your example. Is it it, or the backend not providing the progress bar?

modules/dockermodelrunner/main_test.go
mdelapenya: The issue is with the testable examples, that cannot use t.Skip. I can provide a helper function that checks that, but then if not on Docker Desktop, I need to printout the expected outputs in that early code branch.

Not against this, though

modules/dockermodelrunner/sdk/client/create.go
mdelapenya: do we need special characters in the model name?

modules/dockermodelrunner/sdk/client/create.go
mdelapenya: This was part of another code example, removing the bytes part

modules/dockermodelrunner/sdk/types/types.go
mdelapenya: > inspect model: decode json: Time.UnmarshalJSON: input is not a JSON string

😢 

Probably worthy to mention to @kiview and team

modules/dockermodelrunner/dockermodelrunner.go
```diff
@@ -0,0 +1,132 @@
...
+func (c *Container) ListModels(ctx context.Context) ([]types.ModelResponse, error) {
+	dmrClient := client.NewClient(c.baseURL)
+
+	models, err := dmrClient.ListModels(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("list models: %w", err)
+	}
+
+	return models, nil
+}
```
eddumelendez, line 108: tc modules must not be coupled to client SDKs

modules/dockermodelrunner/sdk/client/client.go
```diff
@@ -0,0 +1,23 @@
+package client
+
+// openAIEndpointSuffix is the suffix for the OpenAI endpoint
+const openAIEndpointSuffix = "/engines/v1"
```
eddumelendez, line 4: it must be only `/engines`. `/v1` belongs to OpenAI compatible API.

modules/dockermodelrunner/dockermodelrunner.go
stevenh: Could you clarify the intent of that guidance? 

Are you suggesting that all callers have to take an endpoint from testcontainers create a client and then make the ListModels call?

Also does the fact this is a provided client not an external implementation have any bearing?

mdelapenya: The drmClient is an http client provided by the module. We could, as follow up, allow client code to customize the client, but with the current implementation, the client code cannot change it, on purpose.

modules/dockermodelrunner/sdk/client/client.go
mdelapenya: In any case, the const is explicitly mentioning OpenAI, so users would have it clear in the name and in the comments.

modules/dockermodelrunner/docker_desktop_test.go
```diff
@@ -0,0 +1,44 @@
...
+	info, err := cli.Info(context.Background())
+	if err != nil {
+		return false, fmt.Errorf("failed to get docker info: %w", err)
+	}
+
+	if info.OperatingSystem == "Docker Desktop" {
+		return true, nil
+	}
+
+	l.Printf("Skipping because Docker Desktop is not running")
```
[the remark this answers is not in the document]
mdelapenya: Yeah, but not in the testable example

modules/dockermodelrunner/sdk/client/client.go
mdelapenya: We provided two examples for two different libraries: langchaingo and openai-go, and in both, simply passing the container method with the OpenAIEndpoint is enough to create the chat experience. You can find them [here](https://github.com/testcontainers/testcontainers-go/pull/3106/files/c2a096a9860c3e4c6c551396aa254e96f5d26824#diff-a54804014f6b2844a842877658de61d82d041cfaffe119e70a6bf3bd6f79c440R327) and [here](https://github.com/testcontainers/testcontainers-go/pull/3106/files/c2a096a9860c3e4c6c551396aa254e96f5d26824#diff-a54804014f6b2844a842877658de61d82d041cfaffe119e70a6bf3bd6f79c440R259)

stevenh (changes requested):
A few questions and a suggestion to make ModelResponse a more idiomatic.

modules/dockermodelrunner/dockermodelrunner.go
```diff
@@ -0,0 +1,132 @@
...
+		wait.ForListeningPort("80/tcp"),
+		wait.ForHTTP("/").WithPort("80/tcp").WithStatusCodeMatcher(func(status int) bool {
+			return status == http.StatusOK
+		}).WithResponseMatcher(func(body io.Reader) bool {
+			bodyBytes, err := io.ReadAll(body)
+			if err != nil {
+				return false
+			}
+			return strings.Contains(string(bodyBytes), "The service is running")
+		}),
```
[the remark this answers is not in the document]
mdelapenya: Probably yes, removing

modules/dockermodelrunner/dockermodelrunner.go
```diff
@@ -0,0 +1,132 @@
...
+			if err := apply(&settings); err != nil {
+				return nil, err
+			}
+		}
+	}
+
+	// Add socat options, which are applied to the socat container.
+	opts = append(opts, testcontainers.WithWaitStrategy(
+		wait.ForListeningPort("80/tcp"),
+		wait.ForHTTP("/").WithPort("80/tcp").WithStatusCodeMatcher(func(status int) bool {
```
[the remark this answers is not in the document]
mdelapenya: For same reason as below, will remove it, thanks

modules/dockermodelrunner/internal/sdk/client/create.go
```diff
@@ -0,0 +1,38 @@
...
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("status code: %d", resp.StatusCode)
+	}
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("read all: %w", err)
+	}
+
+	return body, nil
```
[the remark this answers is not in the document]
mdelapenya: Removing it as this was part of demo code that landed here

modules/dockermodelrunner/internal/sdk/client/pull.go
```diff
@@ -0,0 +1,37 @@
...
+func (c *Client) PullModel(ctx context.Context, fqmn string) error {
+	payload := fmt.Sprintf(`{"from": %q}`, fqmn)
+	reqURL := c.baseURL + "/models/create"
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, reqURL, strings.NewReader(payload))
+	if err != nil {
+		return fmt.Errorf("new post request (%s): %w", reqURL, err)
+	}
+
+	log.Default().Printf("🙏 Pulling model %s. Please be patient, no progress bar yet!", fqmn)
```
stevenh, line 21: question: do we want to add a TODO to remove these log lines when we have a better solution?

modules/dockermodelrunner/internal/sdk/client/pull.go
```diff
@@ -0,0 +1,37 @@
...
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return fmt.Errorf("http post: %w", err)
+	}
+	defer resp.Body.Close()
+
+	// The Docker Model Runner returns a 200 status code for a successful pulls
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("status code: %d", resp.StatusCode)
```
stevenh, line 31: suggestion: as this isn't a wrap I would do something more descriptive for the user
```suggestion
		return fmt.Errorf("unexpected status code: %d", resp.StatusCode)
```

modules/dockermodelrunner/internal/sdk/types/types_test.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+func TestUnmarshalJSONError(t *testing.T) {
+	// Invalid JSON with malformed created timestamp
+	invalidJSON := `{
+        "id": "model123",
+        "tags": ["tag1", "tag2"],
+        "config": {"key1": "value1"},
+        "created": "not-a-number"
+    }`
```
stevenh, line 40: nit: strange indentation

stevenh (changes requested):
Looking good, just a few bits.

modules/dockermodelrunner/internal/sdk/client/pull.go
mdelapenya: Working on a progress bar atm, using https://github.com/schollz/progressbar/ as reference

mdelapenya: I added a progress bar, so this is not valid anymore

modules/dockermodelrunner/internal/sdk/client/pull.go
```diff
@@ -0,0 +1,69 @@
...
+type pullOptions struct {
+	// Progress is an optional writer to show download progress
+	Progress ProgressWriter
+}
+
+// PullOption is a function that configures PullOptions
+type PullOption func(*pullOptions)
+
+// PullModel creates a model in the Docker Model Runner, by pulling the model from Docker Hub.
+func (c *Client) PullModel(ctx context.Context, fqmn string, opts ...PullOption) error {
```
stevenh, line 22: suggestion: I'm guessing this is fully qualified model name, but it's not 100% obvious so I would use a more obvious name even if it is longer.

modules/dockermodelrunner/internal/sdk/client/pull.go
```diff
@@ -0,0 +1,69 @@
...
+)
+
+// pullOptions contains options for pulling a model
+type pullOptions struct {
+	// Progress is an optional writer to show download progress
+	Progress ProgressWriter
+}
+
+// PullOption is a function that configures PullOptions
+type PullOption func(*pullOptions)
```
stevenh, line 19: question: could we ever have one which returned an error? If so do that now so we don't have to break the API later down the line.

stevenh (changes requested):
Nice changes, here's a few bits.

modules/dockermodelrunner/internal/sdk/client/pull.go
mdelapenya: Oh indeed! we need that

modules/dockermodelrunner/internal/sdk/client/pull.go
```diff
@@ -0,0 +1,69 @@
...
+			options.Progress.SetTotal(-1)
+		}
+		reader = io.TeeReader(resp.Body, options.Progress)
+	}
+
+	// Read the response
+	_, err = io.Copy(io.Discard, reader)
+	if err != nil {
+		return fmt.Errorf("copy response: %w", err)
+	}
```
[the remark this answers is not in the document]
mdelapenya: Sorry but I don't understand: we are sending a POST request to the Docker Model Runner (DMR), which streams the model data in the response body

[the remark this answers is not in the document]
stevenh: Followed up in slack, but it doesn't seem like the content of the model is streamed, just a chunked encoding detailing the current download state of the model, so unless I'm missing something I don't think this will work.

stevenh (approved):
LGTM

mdelapenya:
Merging when the CI passes, thanks for your as-always-helpful review @stevenh 🙇 

image

PR #3115, opened 2025-04-17
feat(redis): add TLS support

mdelapenya:
- **chore: move existint options to its own file**
- **feat(redis): add TLS support**

## What does this PR do?

Adds TLS support for Redis, using https://github.com/mdelapenya/tlscert to generate certs on the fly.

The module will use a temporary dir in the OS to create the TLS certs, passing their bytes to the container. Finally, the module will store a TLS config for client code to build a secure client using that config.

## Why is it important?
More advanced scenarios for Redis.

## Follow-ups
Do we want users to contribute their own TLS certificates, or is it fine if the module generates them on the fly?


modules/redis/options.go
```diff
@@ -0,0 +1,141 @@
...
+// Customize is a NOOP. It's defined to satisfy the testcontainers.ContainerCustomizer interface.
+func (o Option) Customize(*testcontainers.GenericContainerRequest) error {
+	// NOOP to satisfy interface.
+	return nil
+}
+
+// WithTLS sets the TLS configuration for the redis container, setting
+// the port to listen on for TLS connections, whether to use a secure URL,
+// and whether to disable MTLS (mutual TLS).
+func WithTLS(tlsPort string, secureURL bool, mtlsDisabled bool) Option {
```
mdelapenya, line 35: Question: do we want three separate options or just this one?

modules/redis/examples_test.go
```diff
@@ -6,19 +6,21 @@ import (
 	"log"
 	"path/filepath"
 
+	"github.com/go-redis/redis/v8"
```
gaby, line 9: Current is v9

modules/redis/go.mod
```diff
@@ -5,8 +5,10 @@ go 1.23.0
 toolchain go1.23.6
 
 require (
+	github.com/docker/go-connections v0.5.0
 	github.com/go-redis/redis/v8 v8.11.5
```
gaby, line 9: We should prob bump go-redis to v9 while we are at it

modules/redis/redis.go
```diff
@@ -40,7 +64,12 @@ func (c *RedisContainer) ConnectionString(ctx context.Context) (string, error) {
 		return "", err
 	}
 
-	uri := fmt.Sprintf("redis://%s:%s", hostIP, mappedPort.Port())
+	prefix := "redis://"
```
mdelapenya, line 67: suggestion: 
```suggestion
	schemme := "redis://"
```

modules/redis/options.go
mdelapenya: Let's be opinionated and not allow passing a port, let the module use a default one.

Same for the the schema, and the mtls, so we could have just WithTLS(), because advance users can always replace the CMD with WithCmd or WithEntrypoint

modules/redis/redis.go
```diff
@@ -55,90 +84,101 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
...
+		// Create a temporary directory to store the TLS certificates.
+		tmpDir := os.TempDir()
+
+		// Generate TLS certificates in the fly and add them to the container before it starts.
+		// Update the CMD to use the TLS certificates.
+		caCert, clientCert, serverCert := createTLSCerts(tmpDir)
+
+		// Update the CMD to use the TLS certificates.
+		cmds := []string{
+			"--port", "6379",
```
mdelapenya, line 123: suggestion: remove this argument because it's already in the image.
```suggestion
```

modules/redis/redis.go
```diff
@@ -55,90 +84,101 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
...
 		}
 
-		// prepend the command to run the redis server with the config file, which must be the first argument of the redis server process
-		if req.Cmd[0] == redisServerProcess {
-			// just insert the config file, then the rest of the args
-			req.Cmd = append([]string{redisServerProcess, defaultConfigFile}, req.Cmd[1:]...)
-		} else if req.Cmd[0] != redisServerProcess {
-			// prepend the redis server and the config file, then the rest of the args
-			req.Cmd = append([]string{redisServerProcess, defaultConfigFile}, req.Cmd...)
+		tcOpts = append(tcOpts, testcontainers.WithCmd(cmds...)) // Replace the default CMD with the TLS certificates.
```
mdelapenya, line 135:
```suggestion
		tcOpts = append(tcOpts, testcontainers.WithCmdArgs(cmds...)) // Append the default CMD with the TLS certificates.
```

modules/redis/options.go
```diff
@@ -0,0 +1,141 @@
...
+		o.tlsPort = tlsPort
+		o.withSecureURL = secureURL
+		o.withMTLSDisabled = mtlsDisabled
+		return nil
+	}
+}
+
+// WithConfigFile sets the config file to be used for the redis container, and sets the command to run the redis server
+// using the passed config file
+func WithConfigFile(configFile string) testcontainers.CustomizeRequestOption {
```
stevenh, line 41: suggestion: pass in an io.Reader which makes it easier to use with the embed module.

modules/redis/options.go
```diff
@@ -0,0 +1,141 @@
...
+	const defaultConfigFile = "/usr/local/redis.conf"
+
+	return func(req *testcontainers.GenericContainerRequest) error {
+		cf := testcontainers.ContainerFile{
+			HostFilePath:      configFile,
+			ContainerFilePath: defaultConfigFile,
+			FileMode:          0o755,
```
stevenh, line 48: suggestion: config shouldn't need to be executable.

modules/redis/options.go
```diff
@@ -0,0 +1,136 @@
...
+	return func(o *options) error {
+		o.tlsEnabled = true
+		return nil
+	}
+}
+
+// WithConfigFile sets the config file to be used for the redis container, and sets the command to run the redis server
+// using the passed config file
+func WithConfigFile(configFile string) testcontainers.CustomizeRequestOption {
+	const defaultConfigFile = "/usr/local/redis.conf"
```
stevenh, line 42: suggestion: either raise to file level to increase visibility and reuse opportunities or just use literal below.
Raising it to file level is consistent with other options.

modules/redis/options.go
```diff
@@ -0,0 +1,136 @@
...
+// save the dataset every N seconds if there are at least M changes in the dataset.
+// This method allows Redis to benefit from copy-on-write semantics.
+// See https://redis.io/docs/management/persistence/#snapshotting for more information.
+func WithSnapshotting(seconds int, changedKeys int) testcontainers.CustomizeRequestOption {
+	if changedKeys < 1 {
+		changedKeys = 1
+	}
+	if seconds < 1 {
+		seconds = 1
+	}
```
stevenh, line 90: suggestion: return an error so the user knowns their options are invalid.

modules/redis/redis.go
```diff
@@ -40,7 +76,7 @@ func (c *RedisContainer) ConnectionString(ctx context.Context) (string, error) {
 		return "", err
 	}
 
-	uri := fmt.Sprintf("redis://%s:%s", hostIP, mappedPort.Port())
+	uri := fmt.Sprintf("rediss://%s:%s", hostIP, mappedPort.Port())
```
stevenh, line 79: question: I would expect this to only work with tls enabled?

modules/redis/redis.go
```diff
@@ -54,91 +90,97 @@ func RunContainer(ctx context.Context, opts ...testcontainers.ContainerCustomize
...
-		WaitingFor:   wait.ForLog("* Ready to accept connections"),
+		ExposedPorts: []string{redisPort},
 	}
 
 	genericContainerReq := testcontainers.GenericContainerRequest{
 		ContainerRequest: req,
 		Started:          true,
 	}
 
+	settings := options{}
```
stevenh, line 101: suggestion: just declare, which is idiomatic
```suggestion
	var settings options
```

modules/redis/redis.go
```diff
@@ -54,91 +90,97 @@ func RunContainer(ctx context.Context, opts ...testcontainers.ContainerCustomize
...
-	if container != nil {
-		c = &RedisContainer{Container: container}
-	}
+	tcOpts := []testcontainers.ContainerCustomizer{}
 
-	if err != nil {
-		return c, fmt.Errorf("generic container: %w", err)
+	waitStrategies := []wait.Strategy{
+		wait.ForListeningPort(redisPort).WithStartupTimeout(time.Second * 10),
+		wait.ForLog("* Ready to accept connections"),
```
stevenh, line 114: question: do we need the ForLog?

modules/redis/redis.go
```diff
@@ -54,91 +90,97 @@ func RunContainer(ctx context.Context, opts ...testcontainers.ContainerCustomize
...
 	}
 
-	return c, nil
-}
+	if settings.tlsEnabled {
```
stevenh, line 117: suggestion: I believe it's recommended to disable the standard port when TLS is enabled [see here](https://redis.io/docs/latest/operate/oss_and_stack/management/security/encryption/).

modules/redis/redis_test.go
```diff
@@ -95,19 +95,64 @@ func TestRedisWithSnapshotting(t *testing.T) {
 	assertSetsGets(t, ctx, redisContainer, 10)
 }
 
+func TestRedisWithTLS(t *testing.T) {
+	ctx := context.Background()
+
+	t.Run("mtls-disabled", func(t *testing.T) {
```
stevenh, line 101: suggestion: remove duplication now these are the same option.

modules/redis/redis.go
```diff
@@ -54,91 +90,97 @@ func RunContainer(ctx context.Context, opts ...testcontainers.ContainerCustomize
...
+		// wait for the TLS port to be available
+		waitStrategies = append(waitStrategies, wait.ForListeningPort(nat.Port(tlsPort)).WithStartupTimeout(time.Second*10))
 
-// WithConfigFile sets the config file to be used for the redis container, and sets the command to run the redis server
-// using the passed config file
-func WithConfigFile(configFile string) testcontainers.CustomizeRequestOption {
-	const defaultConfigFile = "/usr/local/redis.conf"
+		// Create a temporary directory to store the TLS certificates.
+		tmpDir := os.TempDir()
```
stevenh, line 122: suggestion: move into createTLSCerts as it doesn't seem to be needed here.

modules/redis/redis.go
```diff
@@ -27,10 +40,33 @@ const (
...
+// It uses the default 6379 port.
 func (c *RedisContainer) ConnectionString(ctx context.Context) (string, error) {
-	mappedPort, err := c.MappedPort(ctx, "6379/tcp")
+	return c.connectionString(ctx, nat.Port(redisPort))
+}
+
+// ConnectionStringTLS returns the connection string for the Redis container using TLS.
+// It returns an error if TLS is not enabled, else the TLS port defined in the options
+// is used to build the connection string.
+func (c *RedisContainer) ConnectionStringTLS(ctx context.Context) (string, error) {
```
stevenh, line 55: suggestion: eliminate as its either TLS or not see comment about [port = 0 when tls is enabled](https://raw.githubusercontent.com/redis/redis/unstable/redis.conf).

docs/modules/redis.md
```diff
@@ -71,16 +71,44 @@ By default Redis saves snapshots of the dataset on disk, in a binary file called
 
 In the case you have a custom config file for Redis, it's possible to copy that file into the container before it's started. E.g. `WithConfigFile(filepath.Join("testdata", "redis7.conf"))`.
 
+#### WithTLS
+
+In the case you want to enable TLS for the Redis container, you can use the `WithTLS` option. E.g. `WithTLS("6380", true, false)`.
```
stevenh, line 76: bug: using the old syntax

stevenh (changes requested):

modules/redis/options.go
mdelapenya: Would this represent a breaking change in the existing behavior? The snapshotting option was present before this changes. It's unfortunate that moving the code to a separate file (like I did in this PR) does not honor the history of the file.

modules/redis/options.go
mdelapenya: Same here regarding BCs: this function already existed prior to this PR

modules/redis/options.go
stevenh: Hmm, yes it looked new when I was reviewing. As a user I would still expect an error vs doing something magical to fix. If this does break someones code it should be a simple fix. Happy if this is split out from this PR as a separate fix though.

modules/redis/redis.go
mdelapenya: I'd keep it for consistency with previous behaviour.

We could have an issue to revisit all modules and simplify the wait strategies removing the for log

docs/modules/redis.md
```diff
@@ -71,24 +71,33 @@ By default Redis saves snapshots of the dataset on disk, in a binary file called
...
-[Image for Redis-Stack](../../modules/redis/redis_test.go) inside_block:redisStackImage
-<!--/codeinclude-->
+This method returns the TLS configuration for the Redis container, nil if TLS is not enabled.
 
 <!--codeinclude-->
-[Image for Redis-Stack Server](../../modules/redis/redis_test.go) inside_block:redisStackServerImage
+[Get TLS config](../../modules/redis/redis_test.go) inside_block:tlsConfig
 <!--/codeinclude-->
+
+In the above example, the optins are used to configure a Redis client with TLS enabled.
```
mdelapenya, line 103:
```suggestion
In the above example, the options are used to configure a Redis client with TLS enabled.
```

docs/modules/redis.md
```diff
@@ -71,24 +71,33 @@ By default Redis saves snapshots of the dataset on disk, in a binary file called
...
 
-This method returns the connection string to connect to the Redis container, using the default `6379` port.
+This method returns the connection string to connect to the Redis container, using the default `6379` port, and `redis` schema.
 
 <!--codeinclude-->
 [Get connection string](../../modules/redis/redis_test.go) inside_block:connectionString
 <!--/codeinclude-->
 
-### Redis variants
+If the container is started with TLS enabled, the connection string will be `rediss://host:port`, using the `rediss` schema.
```
mdelapenya, line 93:
```suggestion
If the container is started with TLS enabled, the connection string is `rediss://host:port`, using the `rediss` schema.
```

modules/redis/redis.go
stevenh: `t.TmpDir()` has the benefit of automatic clean up, are any files created already removed?

stevenh (changes requested):
Just some comment formatting

modules/redis/redis.go
mdelapenya: No, will add it right now

modules/redis/options.go
```diff
@@ -30,6 +32,8 @@ func (o Option) Customize(*testcontainers.GenericContainerRequest) error {
 
 // WithTLS sets the TLS configuration for the redis container, setting
 // the 6380/tcp port to listen on for TLS connections and using a secure URL (rediss://).
+// It also creates a temporary directory to store the TLS certificates, which is removed
+// when the container is terminated.
```
stevenh, line 36: suggestion: I would remove, its exposing internal workings the user shouldn't care about.

modules/redis/options.go
```diff
@@ -0,0 +1,131 @@
...
+	return testcontainers.WithCmdArgs("--save", strconv.Itoa(seconds), strconv.Itoa(changedKeys))
+}
+
+// createTLSCerts creates a CA certificate, a client certificate and a Redis certificate,
+// storing them in the given temporary directory.
+func createTLSCerts() (*tlscert.Certificate, *tlscert.Certificate, *tlscert.Certificate, error) {
+	tmpDir, err := os.MkdirTemp(os.TempDir(), "testcontainers-redis-tls")
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("mkdir temp dir: %w", err)
+	}
```
stevenh, line 95: suggestion: don't set ParentDir so it just returns struct as it looks like we never use the files anyway?

modules/redis/options.go
```diff
@@ -0,0 +1,131 @@
...
+	// ips is the extra list of IPs to include in the certificates.
+	// It's used to allow the client and Redis certificates to be used in the same host
+	// when the tests are run using a remote docker daemon.
+	ips := []net.IP{net.ParseIP("127.0.0.1")}
+
+	// Generate CA certificate
+	caCert := tlscert.SelfSignedFromRequest(tlscert.Request{
```
stevenh, line 102: bug: need to test for nil which indicates an error, more below.

We should also raise an issue to update SelfSignedFromRequest to return an explicit error, so we can provide that to the caller.

modules/redis/redis.go
```diff
@@ -27,10 +32,25 @@ const (
 
 type RedisContainer struct {
 	testcontainers.Container
+	settings options
+}
+
+// Terminate terminates the Redis container, removing the TLS certificates that were created in the fly.
```
stevenh, line 38: suggestion: remove if the above suggestion to no set ParentDir is valid as it won't be need.

stevenh (changes requested):
Sorry to be a pain, but looks like this can be simplified quite a bit.

modules/redis/options.go
mdelapenya: Sorry but I don't understand: we use the files to build the TLSConfig that is returned by the container, so any client can be configured. Is that what you mean?

modules/redis/options.go
mdelapenya: I submitted a PR to add Err functions to tlscert: https://github.com/mdelapenya/tlscert/pull/1

When merged, I can release `v0.2.0` and directly use it here

modules/redis/options.go
mdelapenya: let me change this in a follow-up refactor, as it's not related to TLS. I moved the options to its own file that's why it appears as new code

modules/redis/options.go
stevenh: I can't see the files being used, only returned values.

The library seems to support generation without writing them to disk. If the user wanted them written to disk that could be done as save or export method.

Did I miss where the files themselves are being used?

mdelapenya: Oh you mean the file representation, not the cert bytes, right? We are probably not using it, you're right and we could simply discard the creation, as the TLS config is just used in-memory

stevenh: yes, exactly no need to create on disk so no need for temp dir etc.

modules/redis/options.go
```diff
@@ -0,0 +1,129 @@
...
+	if seconds < 1 {
+		seconds = 1
+	}
+
+	return testcontainers.WithCmdArgs("--save", strconv.Itoa(seconds), strconv.Itoa(changedKeys))
+}
+
+// createTLSCerts creates a CA certificate, a client certificate and a Redis certificate,
+// storing them in the given temporary directory.
+func createTLSCerts() (*tlscert.Certificate, *tlscert.Certificate, *tlscert.Certificate, error) {
```
stevenh, line 87: suggestion: as all three types are the same I would name the return values so the consumer knows what to expect without having to read the function. Also as the consumer only needs the bytes you could just return that from this function.

mdelapenya: they need the bytes and access to the TLSConfig, that's why I returned the entire type.

Will update the signature to name the returned elements 👍

stevenh: thanks for clarifying, missed that.

stevenh (approved):
LGTM

PR #3117, opened 2025-04-23
chore(ci): run core tests on Testcontainers Cloud

mdelapenya:
## What does this PR do?
It runs the tests on Testcontainers Cloud, only when the core module is added to the build.

For that it extends the existing reusable workflow to accept one boolean argument, and if set to true, install the Testcontainers Cloud Client before running the tests.

## Why is it important?
Dogfood Testcontainers Cloud.

### DIFF
--- a/.github/workflows/ci-test-go.yml
+++ b/.github/workflows/ci-test-go.yml
@@ -18,6 +18,11 @@ on:
         type: string
         default: "."
         description: "The directory where the Go project is located."
+      testcontainers-cloud:
+        required: false
+        type: boolean
+        default: false
+        description: "Run the tests on Testcontainers Cloud"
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-test-go.yml
```diff
@@ -81,6 +86,13 @@ jobs:
             echo "No dependencies script found at $SCRIPT_PATH - skipping installation"
           fi
 
+      # Setup Testcontainers Cloud Client right before your Testcontainers tests
+      - name: Setup Docker Cloud Client
```
mdelapenya, line 90:
```suggestion
      - name: Setup Testcontainers Cloud Client
```

PR #3125, opened 2025-04-25
fix(dockermodelrunner): wait for the model to be pulled

mdelapenya:
## What does this PR do?
It adds a backoff-retry strategy after pulling a model in order to verify the model was actually pulled.

The retry uses the passed context to determine the exit.

## Why is it important?
When the HTTP connection is closed, the pull cancels, but the code continues its execution, resulting in a not pulled model.

## Related issues

- Caused by #3106

### DIFF
--- a/modules/dockermodelrunner/internal/sdk/client/pull.go
+++ b/modules/dockermodelrunner/internal/sdk/client/pull.go
@@ -20,12 +20,9 @@ func (c *Client) PullModel(ctx context.Context, fullyQualifiedModelName string) 
 		return fmt.Errorf("new post request (%s): %w", reqURL, err)
 	}
 
-	log.Default().Printf("🙏 Pulling model %s. Please be patient", fullyQualifiedModelName)
+	logger := log.Default()
 
-	// Check context before making request
-	if err := ctx.Err(); err != nil {
-		return fmt.Errorf("context done before request: %w", err)
-	}
+	logger.Printf("🙏 Pulling model %s. Please be patient", fullyQualifiedModelName)
 
 	resp, err := http.DefaultClient.Do(req)
 	if err != nil {
@@ -44,25 +41,16 @@ func (c *Client) PullModel(ctx context.Context, fullyQualifiedModelName string) 
 	}
 
 	scanner := bufio.NewScanner(resp.Body)
-	done := make(chan error, 1)
-
-	go func() {
-		for scanner.Scan() {
-			log.Default().Printf(scanner.Text())
-		}
-		done <- scanner.Err()
-	}()
-
-	select {
-	case <-ctx.Done():
-		return fmt.Errorf("context done: %w", ctx.Err())
-	case err := <-done:
-		if err != nil {
-			return fmt.Errorf("scan error: %w", err)
-		}
+	// TODO: use a progressbar instead of multiple line output.
+	for scanner.Scan() {
+		logger.Printf(scanner.Text())
 	}
 
-	log.Default().Printf("✅ Model %s pulled successfully!", fullyQualifiedModelName)
+	if err := scanner.Err(); err != nil {
+		return fmt.Errorf("scan error: %w", err)
+	}
+
+	logger.Printf("✅ Model %s pulled successfully!", fullyQualifiedModelName)
 
 	return nil
 }
--- a/modules/dockermodelrunner/internal/sdk/client/pull.go
+++ b/modules/dockermodelrunner/internal/sdk/client/pull.go
@@ -1,14 +1,11 @@
 package client
 
 import (
+	"bufio"
 	"context"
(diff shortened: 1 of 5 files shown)


modules/dockermodelrunner/internal/sdk/client/pull.go
```diff
@@ -31,6 +36,35 @@ func (c *Client) PullModel(ctx context.Context, fullyQualifiedModelName string)
 		return fmt.Errorf("unexpected status code: %d", resp.StatusCode)
 	}
 
+	namespace := strings.Split(fullyQualifiedModelName, "/")[0]
```
stevenh, line 39: suggestion: I think we just need to read the body to completion, no need for the retry which the caller can handle if desired.

So an io.Copy(io.Discard, resp.Body) should be all that's needed

stevenh (changes requested):
I think this can be simplified, see comments

modules/dockermodelrunner/internal/sdk/client/pull.go
mdelapenya: The cool thing of this is the logs providing progress

stevenh: you could do that by outputting each line received if needed.

modules/dockermodelrunner/internal/sdk/client/pull.go
```diff
@@ -18,19 +20,48 @@ func (c *Client) PullModel(ctx context.Context, fullyQualifiedModelName string)
...
+	}()
+
+	select {
+	case <-ctx.Done():
+		return fmt.Errorf("context done: %w", ctx.Err())
+	case err := <-done:
+		if err != nil {
+			return fmt.Errorf("scan error: %w", err)
+		}
+	}
```
stevenh, line 63: suggestion: no need for the go routine here as scanner will return the context error if needed from the its `Read` call to `resp.Body`.
```suggestion
	scanner := bufio.NewScanner(resp.Body)

	// TODO: use a progressbar instead of multiple line output.
	logger := log.Default()
	for scanner.Scan() {
		logger.Printf(scanner.Text())
	}
	
	if err := scanner.Err(); err != nil {
		return fmt.Errorf("scan error: %w", err)
	}
```

modules/dockermodelrunner/internal/sdk/client/pull.go
```diff
@@ -18,19 +20,48 @@ func (c *Client) PullModel(ctx context.Context, fullyQualifiedModelName string)
 		return fmt.Errorf("new post request (%s): %w", reqURL, err)
 	}
 
-	log.Default().Printf("🙏 Pulling model %s. Please be patient, no progress bar yet!", fullyQualifiedModelName)
+	log.Default().Printf("🙏 Pulling model %s. Please be patient", fullyQualifiedModelName)
```
stevenh, line 23: suggestion: create a logger var so we don't have to run the Default function multiple times, see example in the for loop suggestion below for an example.

stevenh (changes requested):

modules/dockermodelrunner/internal/sdk/client/pull.go
```diff
@@ -18,20 +20,37 @@ func (c *Client) PullModel(ctx context.Context, fullyQualifiedModelName string)
...
 	if err != nil {
 		return fmt.Errorf("http post: %w", err)
 	}
 	defer resp.Body.Close()
 
+	// Check context after getting response
+	if err := ctx.Err(); err != nil {
+		return fmt.Errorf("context done after response: %w", err)
+	}
+
```
stevenh, line 37: suggestion: not sure if this was missed, but this isn't needed, the read from resp.Body by scanner will detect if needed.

stevenh (changes requested):
Looks like one comment was missed, then good to go.

modules/dockermodelrunner/internal/sdk/client/pull.go
```diff
@@ -18,19 +20,48 @@ func (c *Client) PullModel(ctx context.Context, fullyQualifiedModelName string)
 		return fmt.Errorf("new post request (%s): %w", reqURL, err)
 	}
 
-	log.Default().Printf("🙏 Pulling model %s. Please be patient, no progress bar yet!", fullyQualifiedModelName)
+	log.Default().Printf("🙏 Pulling model %s. Please be patient", fullyQualifiedModelName)
+
+	// Check context before making request
+	if err := ctx.Err(); err != nil {
+		return fmt.Errorf("context done before request: %w", err)
+	}
```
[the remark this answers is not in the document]
mdelapenya: I only removed the first check, now it's updated

stevenh (approved):
LGTM

PR #3126, opened 2025-04-25
docs(redpanda): update sasl authentication option to use scram sha 256

peczenyj:
Highlight that the sasl authentication uses scram sha 256 (I spent some time trying to connect with 512 until I realize my mistake)

## What does this PR do?

Update go doc expanding some existing information

## Why is it important?

I think we should be explicit about sasl uses scram sha 256

## Related issues

N/A


mdelapenya (dismissed):
LGTM thanks!

Could you also update the docs at `./docs/modules/redpanda.md`? 🙏

peczenyj:
Done

mdelapenya (approved):
I kicked off the CI, once passed (it will 😄 ) we are good to merge, thanks for your work here with the documentation, they are pretty important to us!

LGTM!

PR #3131, opened 2025-04-30
feat(valkey): add TLS support for Valkey

mdelapenya:
- **feat(valkey): add TLS support for Valkey**
- **docs(redis): update missing version markers**

## What does this PR do?
It copies what we added for Redis in #3115, but for Valkey.

It also adds the versions when the new TLS options were added to the Redis docs.

## Why is it important?
Keep parity between similar modules.


modules/valkey/options.go
```diff
@@ -0,0 +1,82 @@
...
+var _ testcontainers.ContainerCustomizer = (Option)(nil)
+
+// Option is an option for the Redpanda container.
+type Option func(*options) error
+
+// Customize is a NOOP. It's defined to satisfy the testcontainers.ContainerCustomizer interface.
+func (o Option) Customize(*testcontainers.GenericContainerRequest) error {
+	// NOOP to satisfy interface.
+	return nil
+}
```
stevenh, line 28: suggestion: not needed for this PR but as this is a very common pattern we should create a noop customiser we can just embed to avoid this repeating boilerplate.

stevenh (approved):
LGTM

modules/valkey/options.go
mdelapenya: Makes sense! Thanks for the suggestion!

PR #3132, opened 2025-04-30
feat(memcached): add memcached module

mdelapenya:
## What does this PR do?
It adds the Memcached module, adding testable examples using `bradfitz/gomemcache` as client.

## Why is it important?
New module!


modules/memcached/memcached_test.go
```diff
@@ -0,0 +1,21 @@
...
+)
+
+func TestRun(t *testing.T) {
+	ctx := context.Background()
+
+	ctr, err := memcached.Run(ctx, "memcached:1.6-alpine")
+	testcontainers.CleanupContainer(t, ctr)
+	require.NoError(t, err)
+
+	// perform assertions
```
stevenh, line 20: question: was this left as a placeholder for more work?

stevenh (approved):
Just one question

modules/memcached/memcached_test.go
mdelapenya: It's part of the template. Will remove it in a follow, up , also cleaning it up from the modules where it does not apply anymore

Issue #3124, opened 2025-04-25
[Enhancement]: Add TLS/SSL Support to Cassandra


MitulShah1:
### Proposal

Currently, the Cassandra testcontainer module doesn't provide native support for TLS/SSL connections. This enhancement proposes implementing TLS support to make the module more useful for testing secure connections and configurations.

**Use Case**

When testing applications that connect to Cassandra via secure TLS connections, developers need to set up a test environment that mimics production security settings. Adding TLS support to the testcontainer would allow developers to:

1. Test client applications against a TLS-enabled Cassandra instance
2. Verify connection security settings
3. Test client certificate authentication
4. Ensure connection strings and configurations work correctly with encryption

mdelapenya:
@MitulShah1 would you like to contribute it? Please follow the recently added Redis impl so that it matches the current pattern for on-the-fly generation of the certs 🙏

MitulShah1:
Thanks @mdelapenya , Yes love to contribute, please assign to me

mdelapenya:
Closed as PR was merged

MitulShah1:
@mdelapenya  when is next release plan? i see no new release since 0.40.0, i want to implement this in https://github.com/gofiber/storage/issues/1688

PR #3151, opened 2025-05-02
Add ssl option cassandra

MitulShah1:
#3124 

## Related issues

-


modules/cassandra/cassandra.go
```diff
@@ -66,15 +73,72 @@ func WithInitScripts(scripts ...string) testcontainers.CustomizeRequestOption {
...
+			certFile := testcontainers.ContainerFile{
+				HostFilePath:      sslOpts.CertPath,
+				ContainerFilePath: "/etc/cassandra/conf/cassandra.crt",
+				FileMode:          0o644,
+			}
+			req.Files = append(req.Files, certFile)
+		}
+
+		// Mark that SSL is enabled for later use
+		req.Labels = mergeMap(req.Labels, map[string]string{"testcontainers.cassandra.ssl": "true"})
```
mdelapenya, line 131: question: I guess this is needed to notify the Run function about the SSL addition, right? I'd choose a different pattern for this, creating an `options` type as in the redis module. Then, you can simply generate the SSL certs on-the-fly as in that module.

mdelapenya (changes requested):
Looking good, although I added some comments about using the same approach as in the Redis module. Could you please take a look?

Cheers!

MitulShah1:
> Looking good, although I added some comments about using the same approach as in the Redis module. Could you please take a look?
> 
> Cheers!

I made changes, can you please check again?

modules/cassandra/cassandra.go
```diff
@@ -94,21 +153,35 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
...
+	}
+
+	tcOpts = append(tcOpts, testcontainers.WithExposedPorts(exposePort...))
+	tcOpts = append(tcOpts, testcontainers.WithWaitStrategy(waitStrategies...))
+
+	// Append the customizers passed to the Run function.
+	tcOpts = append(tcOpts, opts...)
+
+	// Apply the testcontainers customizers.
+	for _, opt := range tcOpts {
```
stevenh, line 184: suggestion: simplify by passing genericContainerReq into setupTLS and having it do the work, instead of double handling the values.

You could even pass it in to the options so you don't even need the IsTLSEnabled boolean and associated check.

```go
type Option func(*testcontainers.GenericContainerRequest, *Options) error
```

stevenh (changes requested):
Thanks for the quick updates, here's some suggestion which focus on simplification and reducing API surface area.

modules/cassandra/options.go
```diff
@@ -0,0 +1,85 @@
...
+// WithSSL enables SSL/TLS support on the Cassandra container
+func WithSSL() Option {
+	return func(o *Options) error {
+		o.IsTLSEnabled = true
+		return nil
+	}
+}
+
+// GenerateJKSKeystore generates a JKS keystore with a self-signed cert using keytool, and extracts the public cert for Go client trust.
+func GenerateJKSKeystore() (keystorePath, certPath string, err error) {
```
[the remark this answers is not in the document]
MitulShah1: @stevenh  making this function private, i need to change in options_test.go, its ok?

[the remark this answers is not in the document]
stevenh: Either that or make a seperate test file.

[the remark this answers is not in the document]
MitulShah1: options_test.go is created by me only but i cannot access that function as testcases are depend on it, do you want me to make this function private and remove options_test.go?

modules/cassandra/cassandra.go
```diff
@@ -14,28 +16,35 @@ import (
 )
 
 const (
-	port = nat.Port("9042/tcp")
+	port       = nat.Port("9042/tcp")
+	securePort = nat.Port("9142/tcp") // Common port for SSL/TLS connections
```
stevenh, line 20: question: do we need to two ports or can it just be one which is either SSL or not?

modules/cassandra/cassandra.go
```diff
@@ -94,30 +99,44 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
...
 	}
 
+	// Set up wait strategies
+	waitStrategies := []wait.Strategy{
+		wait.ForListeningPort(port),
+		wait.ForExec([]string{"cqlsh", "-e", "SELECT bootstrapped FROM system.local"}).WithResponseMatcher(func(body io.Reader) bool {
+			data, _ := io.ReadAll(body)
+			return strings.Contains(string(data), "COMPLETED")
+		}).WithStartupTimeout(1 * time.Minute),
+	}
```
stevenh, line 124: @mdelapenya looks like we have an issue with WithWaitStrategy, it sets vs appending to WaitFor preventing easy initialisation here.

modules/cassandra/cassandra.go
```diff
@@ -94,30 +99,44 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
...
+
+	// Add TLS wait strategy if TLS config exists
+	if settings.tlsConfig != nil {
+		waitStrategies = append(waitStrategies, wait.ForListeningPort(securePort).WithStartupTimeout(1*time.Minute))
+	}
+
+	// Apply wait strategy using the correct method
+	if err := testcontainers.WithWaitStrategy(wait.ForAll(waitStrategies...)).Customize(&genericContainerReq); err != nil {
+		return nil, err
```
stevenh, line 133: suggestion: wrap the error here
```suggestion
		return nil, fmt.Errorf("with wait strategy: %w", err)
```

modules/cassandra/cassandra.go
```diff
@@ -126,3 +145,11 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
 
 	return c, nil
 }
+
+// TLSConfig returns the TLS configuration for the Cassandra container, nil if TLS is not enabled.
+func (c *CassandraContainer) TLSConfig() *tls.Config {
```
stevenh, line 150: @mdelapenya quite a common pattern in modules, we should consider returning an error instead.

modules/cassandra/options.go
```diff
@@ -0,0 +1,127 @@
...
+type Options struct {
+	tlsConfig *TLSConfig
+}
+
+// Option is an option for the Cassandra container.
+type Option func(*testcontainers.GenericContainerRequest, *Options) error
+
+// Customize implements the testcontainers.ContainerCustomizer interface
+func (o Option) Customize(req *testcontainers.GenericContainerRequest) error {
+	return o(req, &Options{})
```
stevenh, line 32: bug: while this will run there options won't be accessible, so should still be just a noop.

We're looking to refactor how this works in a future version to avoid this, but that will be a bit.

modules/cassandra/cassandra.go
```diff
@@ -94,30 +99,44 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
...
-		),
+		ExposedPorts: []string{string(port)},
 	}
 
 	genericContainerReq := testcontainers.GenericContainerRequest{
 		ContainerRequest: req,
 		Started:          true,
 	}
 
+	var settings Options
```
stevenh, line 110: bug: settings is never set, you still need to check if you have an Option in the for loop and process it to ensure it set correctly.

If your tests are passing they shouldn't so might need a fix there too.

modules/cassandra/options.go
```diff
@@ -0,0 +1,127 @@
...
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+
+	"github.com/testcontainers/testcontainers-go"
+)
+
+// TLSConfig represents the TLS configuration for Cassandra
+type TLSConfig struct {
```
stevenh, line 16: question: are the paths needed? If not we can just return `*tls.Config`. I suspect the answer is dependent on eliminating the need for `keytool`

stevenh (changes requested):
A few bugs introduced here I think, the big question is eliminating the use of external dependency keytool

modules/cassandra/cassandra.go
mdelapenya: You mean the pointer to the config and the error, right? I think we changed the tlscert library precisely for that, so good to me to change it

mdelapenya: Oh I see, the recently added methods for Redis and Valkey. For Valkey, we are in the same release cycle, we can change it without BC, for Redis, it's a BC

modules/cassandra/options.go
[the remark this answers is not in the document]
MitulShah1: It seems you're suggesting we either minimize SSL-related dependencies or simply support SSL with minimal setup. I always try to follow the existing repository codebase and coding style as much possible.

Initially, I tried using only [tlscert](https://pkg.go.dev/github.com/mdelapenya/tlscert), but it has some limitations. If we choose to use tlscert, we can eliminate the keytool dependency, but we'd then need to rely on openssl. 

So, we have two options:

Use keytool as we currently do.

Follow an approach similar to [Cassandra's Go driver](https://github.com/apache/cassandra-gocql-driver/tree/trunk/testdata/pki), where certificates are added in a testdata folder and used for SSL testing, as seen in the official Go driver.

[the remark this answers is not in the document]
stevenh: tlscert would be the preferred option as it's what a lot of the other modules use.

If it needs changes that should be quick and easy as that module is written by @mdelapenya one of this repositories maintainers.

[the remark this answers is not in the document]
MitulShah1: i know but it can handle this cassandra cert generation? @mdelapenya ? i already tried that let me know if ok to use openssl then i will make change to tlscert.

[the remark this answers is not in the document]
mdelapenya: Yes, let's use it for consistency. If there is a missing feature in that library, let's open an issue in there and work on fix and release it

Cheers!

mdelapenya:
@MitulShah1 did you have the chance to take a look a the comments here?

MitulShah1:
> @MitulShah1 did you have the chance to take a look a the comments here?

Sorry lil busy, i will check and try to finish this week.

gaby:
@mdelapenya @MitulShah1 any progress on this?

MitulShah1:
> @mdelapenya @MitulShah1 any progress on this?

Hey @gaby @mdelapenya  sorry for delay but i tried with simple tlc certy which @stevenh  mentioned with @mdelapenya  library seems its not working with cassandra. so just stuck on this point only.

mdelapenya:
@gaby @MitulShah1 thanks for the ping. I'll resolve the conflicts locally and will try to check the TLS issues mentioned above. 

@MitulShah1 can you elaborate on those issues with tlscert?

MitulShah1:
@mdelapenya @stevenh thanks for patience, i've fixed all issues please have a look again and let me know if any changes :)

modules/cassandra/cassandra_test.go
```diff
@@ -117,3 +117,54 @@ func TestCassandraWithInitScripts(t *testing.T) {
...
+	ctx := context.Background()
+
+	// withTLS {
+	ctr, err := cassandra.Run(ctx, "cassandra:4.1.3", cassandra.WithTLS())
+	// }
+	testcontainers.CleanupContainer(t, ctr)
+	require.NoError(t, err)
+
+	// Verify TLS config is available
+	tlsConfig := ctr.TLSConfig()
```
mdelapenya: suggestion: could you add a check in the non-TLS tests that this config is nil?

modules/cassandra/options.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+// WithTLS enables TLS/SSL on the Cassandra container.
+// When enabled, the container will:
+//   - Generate self-signed certificates
+//   - Configure Cassandra to use client encryption
+//   - Expose the SSL port (9142)
+//
+// Use TLSConfig() on the returned container to get the *tls.Config for client connections.
+func WithTLS() Option {
```
mdelapenya, line 42: bug: let's document the new option in the `docs/modules/cassandra.md` file. Please remember to use the marker as described in the contributing docs: https://golang.testcontainers.org/contributing/#adding-functional-options

It could be of interest to embed a code snippet linking to `// withTLS {` (see https://golang.testcontainers.org/contributing/#adding-code-snippets)

modules/cassandra/cassandra.go
```diff
@@ -12,20 +15,34 @@ import (
...
 func (c *CassandraContainer) ConnectionHost(ctx context.Context) (string, error) {
+	if c.settings.tlsEnabled {
+		return c.PortEndpoint(ctx, sslPort, "")
+	}
 	return c.PortEndpoint(ctx, port, "")
 }
 
+// TLSConfig returns the TLS configuration for secure client connections.
+// Returns nil if TLS is not enabled on the container.
+func (c *CassandraContainer) TLSConfig() *tls.Config {
```
mdelapenya: bug: let's document the new container method in the `docs/modules/cassandra.md` file. Please remember to use the marker as described in the contributing docs: https://golang.testcontainers.org/contributing/#adding-functional-options

It could be of interest to embed a code snippet linking to `// TLSConfig {` (see https://golang.testcontainers.org/contributing/#adding-code-snippets)

mdelapenya (approved):
LGTM! Thanks @MitulShah1 for your work and sorry for the late review. I just added a few comments regarding the missing docs, which are needed. But I'm already approving the PR waiting for the docs to be in place.

Cheers!

modules/cassandra/cassandra.go
mdelapenya: BTW checking the rest of the comments, we agreed in having this method to return the pointer to tls.Config and an error. Let's change it

docs/modules/cassandra.md
```diff
@@ -70,6 +70,23 @@ In the case you have a custom config file for Cassandra, it's possible to copy t
...
+When enabled, the container will:
+- Generate self-signed certificates automatically
+- Configure Cassandra to use client encryption
+- Expose the SSL port (9142)
+
+Use the `TLSConfig()` method on the returned container to get the `*tls.Config` for client connections. The method returns an error if TLS was not enabled via `WithTLS()`.
+
+<!--codeinclude-->
+[Using TLS option](../../modules/cassandra/cassandra_test.go) inside_block:withTLS
+<!--/codeinclude-->
```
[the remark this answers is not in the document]
mdelapenya: This PR adds the WithTLS option

docs/modules/cassandra.md
```diff
@@ -70,6 +70,23 @@ In the case you have a custom config file for Cassandra, it's possible to copy t
...
+- Not available until the next release <a href="https://github.com/testcontainers/testcontainers-go"><span class="tc-version">:material-tag: main</span></a>
+
+If you need to enable TLS/SSL encryption for client connections, you can use the `cassandra.WithTLS()` option. 
+
+When enabled, the container will:
+- Generate self-signed certificates automatically
+- Configure Cassandra to use client encryption
+- Expose the SSL port (9142)
+
+Use the `TLSConfig()` method on the returned container to get the `*tls.Config` for client connections. The method returns an error if TLS was not enabled via `WithTLS()`.
```
mdelapenya, line 84: suggestion: add the entry for the `TLSConfig` container method under https://github.com/testcontainers/testcontainers-go/pull/3151/commits/50579f95192097f5e413b4cc99f09ba968f15bb7#diff-b27ccdbcf3e3815b7df9845a916f9fec349d71ad4f2baa18ce80edc5dd84bf8fR92, which is the dedicated section for them.

PR #3156, opened 2025-05-05
feat: add more functional options for customising containers

mdelapenya:
## What does this PR do?
- **feat: add new option for AlwaysPull image**
- **feat: add new option for Lifecycle Hooks image**
- **feat: add new option for log consumer config**
- **feat: new option for attaching to networks by name**
- **feat: add two more options (no-start and name)**
- **feat: add one more option (image platform)**

## Why is it important?
More options to fully customise a container, from the modules side, or the Run function.

### DIFF
--- a/network/network.go
+++ b/network/network.go
@@ -2,6 +2,7 @@ package network
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/docker/docker/api/types/network"
@@ -141,23 +142,32 @@ func WithNetwork(aliases []string, nw *testcontainers.DockerNetwork) testcontain
 }
 
 // WithNetworkName attachs a container to an already existing network, by its name.
-// Finally, if the network is not "bridge", it sets the network alias on that network to the given alias.
-// This is because network-scoped alias is supported only for containers in user defined networks.
+// If the network is not "bridge", it sets the network alias on that network
+// to the given alias, else, it returns an error. This is because network-scoped alias
+// is supported only for containers in user defined networks.
 func WithNetworkName(aliases []string, networkName string) testcontainers.CustomizeRequestOption {
 	return func(req *testcontainers.GenericContainerRequest) error {
+		if networkName == "bridge" {
+			return errors.New("network-scoped aliases are supported only for containers in user defined networks")
+		}
+
 		// attaching to the network because it was created with success or it already existed.
 		req.Networks = append(req.Networks, networkName)
-
-		// network-scoped alias is supported only for containers in user defined networks
-		if networkName == "bridge" {
-			return nil
-		}
 
 		if req.NetworkAliases == nil {
 			req.NetworkAliases = make(map[string][]string)
 		}
 		req.NetworkAliases[networkName] = aliases
 
+		return nil
+	}
+}
+
+// WithBridgeNetwork attachs a container to the "bridge" network.
+// There is no need to set the network alias, as it is not supported for the "bridge" network.
+func WithBridgeNetwork() testcontainers.CustomizeRequestOption {
+	return func(req *testcontainers.GenericContainerRequest) error {
+		req.Networks = append(req.Networks, "bridge")
 		return nil
 	}
 }
--- a/network/network.go
+++ b/network/network.go
@@ -137,11 +137,21 @@ func WithIPAM(ipam *network.IPAM) CustomizeNetworkOption {
 // WithNetwork reuses an already existing network, attaching the container to it.
 // Finally it sets the network alias on that network to the given alias.
 func WithNetwork(aliases []string, nw *testcontainers.DockerNetwork) testcontainers.CustomizeRequestOption {
-	return func(req *testcontainers.GenericContainerRequest) error {
-		networkName := nw.Name
+	return WithNetworkName(aliases, nw.Name)
+}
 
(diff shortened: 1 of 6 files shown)


network/network.go
```diff
@@ -137,12 +137,22 @@ func WithIPAM(ipam *network.IPAM) CustomizeNetworkOption {
...
+// Finally, if the network is not "bridge", it sets the network alias on that network to the given alias.
+// This is because network-scoped alias is supported only for containers in user defined networks.
+func WithNetworkName(aliases []string, networkName string) testcontainers.CustomizeRequestOption {
+	return func(req *testcontainers.GenericContainerRequest) error {
 		// attaching to the network because it was created with success or it already existed.
 		req.Networks = append(req.Networks, networkName)
 
+		// network-scoped alias is supported only for containers in user defined networks
+		if networkName == "bridge" {
+			return nil
```
stevenh, line 153: question: Should this return an error?

options.go
```diff
@@ -106,14 +106,33 @@ func WithHostPortAccess(ports ...int) CustomizeRequestOption {
 	}
 }
 
+// WithName will set the name of the container.
+func WithName(containerName string) CustomizeRequestOption {
+	return func(req *GenericContainerRequest) error {
+		if containerName == "" {
+			return errors.New("container name must be provided")
+		}
```
stevenh, line 114: question: is this check needed? I ask as this may be used to unset a name, which this prevents. If it's kept we should document it in the doc comment.

If removed we'd need to add that to WithReuseByName

options.go
```diff
@@ -106,14 +106,33 @@ func WithHostPortAccess(ports ...int) CustomizeRequestOption {
...
+	}
+}
+
+// WithNoStart will prevent the container from being started after creation.
+func WithNoStart() CustomizeRequestOption {
+	return func(req *GenericContainerRequest) error {
+		req.Started = false
+		return nil
+	}
+}
```
stevenh, line 126: suggestion: pass in start so it can he used to set and unset started?

options.go
```diff
@@ -255,6 +274,17 @@ func WithLogConsumers(consumer ...LogConsumer) CustomizeRequestOption {
 	}
 }
 
+// WithLogConsumerConfig sets the log consumer config for a container.
+// Beware that this option completely replaces the existing log consumer config,
+// including the log consumers and the log production options,
+// so it should be used with care.
+func WithLogConsumerConfig(config *LogConsumerConfig) CustomizeRequestOption {
```
stevenh, line 281: question: didn't we deprecate logconsumers?

options.go
```diff
@@ -376,6 +406,22 @@ func WithImageMount(source string, subpath string, target ContainerMountTarget)
 	}
 }
 
+// WithAlwaysPull will pull the image before starting the container
+func WithAlwaysPull() CustomizeRequestOption {
```
stevenh, line 410: suggestion: pass in the parameter so can set and unset, without two options. Also match the field name for consistency?

stevenh (changes requested):
A few questions, all in the same vain

options.go
mdelapenya: I considered it,  although I see the Started state as the default, so I won't Don't want users to explicitly set it

stevenh: Unfortunately while the intention might be for started to be the default, as this is a bool and the default value is false it actually isn't so it's totally possible to need to set it. If it was NoStart allowing the zero value this would make more sense. Another example of why keeping fields private provides implementation flexibility.

Not a blocker but something to consider

mdelapenya: > Another example of why keeping fields private provides implementation flexibility.

Indeed!

Unfortunately, the request struct was here since the beginning. But with all the functional options, in combination with the Run function, we are really close to start deprecating the usage of the request struct in favour of the Run function + options.

I'd keep it as is, so that users don't have to worry about setting true/false: the Run functions (in modules and core) will always _run_ the container, so the default will be true.

options.go
mdelapenya: Nope, we added a log package for the logging of the library, but the logconsumers, and the config, it's still in the core.

options.go
mdelapenya: Containers must come without names, by default, so users would just want to add a name, not cleaning up an existing one.

This related to the discussion we had about booleans: if at some point we have all fields private, only accessible via functional opts, then we can deprecate the GenericContainerRequest struct and "clone" it into an internal/private struct that will hold the state for the user request. Then, the opinionated semantics for not accepting true/false (for booleans) or changing the container name (here) would enforce the expressiveness of the API

network/network.go
mdelapenya: Good question: the docs say:
> In case the network name is `bridge`, no aliases are set. This is because network-scoped alias is supported only for containers in user defined networks.

And in the code we are: first, setting the networks; second, not adding the aliases.

Do you consider we should raise an error here if the user wants to attach a container to the bridge network? I'm not opposed to it, just want to confirm with you first.

mdelapenya: We can add a `WithBridgeNetwork()` that resolves the cases where users want to explicitly attach containers to that network

mdelapenya:
@stevenh I added the option for the bridge network, hope it's good for review now

stevenh (approved):

Issue #2945, opened 2025-01-22
[Bug]: "wait for reaper 0743154e: no port to wait for"


srenatus:
### Testcontainers version

0.35.0

### Using the latest Testcontainers version?

Yes

### Host OS

Linux

```
    OS Version: 22.04.5 LTS
    Kernel Version: 6.5.0-1025-azure
    Image Version: 20250105.1.0
    Systemd version: 249.11-0ubuntu3.12
```

### Host arch

x86

### Go version

1.23.5

### Docker version

```shell
Docker Amazon ECR Credential Helper 0.9.0
Docker Compose v2 2.27.1
Docker-Buildx 0.19.3
Docker Client 26.1.3
Docker Server 26.1.3
```

### Docker info

```shell
Github Runner :/
```

### What happened?

Suddenly, a new type of failure spuriously hits our CI runs:

```
    e2e_test.go:46: generic container: create container: reaper: from container "0743154e": wait for reaper 0743154e: no port to wait for
```

The code in question is a helper method to spin up minio:

```go
func minioContainer(ctx context.Context, t *testing.T) (testcontainers.Container, string) {
	t.Helper()
	tc, err := tc_minio.Run(ctx, "minio/minio:latest", testcontainers.CustomizeRequestOption(func(r *testcontainers.GenericContainerRequest) error {
		r.ContainerRequest.Env["MINIO_ROOT_USER"] = minioRootUser
		r.ContainerRequest.Env["MINIO_ROOT_PASSWORD"] = minioRootPassword
		return nil
	}))
	if err != nil {
		t.Fatal(err)
	}
	cs, err := tc.ConnectionString(ctx)
	if err != nil {
		t.Fatal(err)
	}
	return tc, cs
}
```

where `tc_minio` is "github.com/testcontainers/testcontainers-go/modules/minio".

I've never seen this before bumping TC from 0.34.0 to 0.35.0 last week. This used to run smoothly for a long while.

### Relevant log output

```shell

```

### Additional information

I'm sorry there aren't better logs -- but the test run uses TC _a lot_ and this is the only failure in it. So previous and subsequent calls to reaper must have succeeded.

stevenh:
Using the following test, I can't reproduce this.
```go
package testcontainers_test

import (
	"context"
	"testing"

	"github.com/stretchr/testify/require"
	"github.com/testcontainers/testcontainers-go"
	tc_minio "github.com/testcontainers/testcontainers-go/modules/minio"
)

[...]

PR #3158, opened 2025-05-08
fix(wait): no port to wait for

stevenh:
Fix a race condition when processing `HostPortStrategy.WaitUntilReady` which could result in it returning "no port to wait for" when no port is specified as is the case for `ForExposedPort`.

This could happen if the call happens very quickly after the container is started and the port is not yet available.

We now retry the internal port detection until found or the internal context is cancelled.

Fixes: #2945

### DIFF
--- a/wait/host_port.go
+++ b/wait/host_port.go
@@ -95,6 +95,25 @@ func (hp *HostPortStrategy) Timeout() *time.Duration {
 	return hp.timeout
 }
 
+// detectInternalPort returns the lowest internal port that is currently bound.
+// If no internal port is found, it returns the zero nat.Port value which
+// can be checked against an empty string.
+func (hp *HostPortStrategy) detectInternalPort(ctx context.Context, target StrategyTarget) (nat.Port, error) {
+	var internalPort nat.Port
+	inspect, err := target.Inspect(ctx)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


wait/host_port.go
```diff
@@ -95,6 +95,25 @@ func (hp *HostPortStrategy) Timeout() *time.Duration {
...
+func (hp *HostPortStrategy) detectInternalPort(ctx context.Context, target StrategyTarget) (nat.Port, error) {
+	var internalPort nat.Port
+	inspect, err := target.Inspect(ctx)
+	if err != nil {
+		return internalPort, fmt.Errorf("inspect: %w", err)
+	}
+
+	for port := range inspect.NetworkSettings.Ports {
+		if internalPort == "" || port.Int() < internalPort.Int() {
+			internalPort = port
```
mdelapenya, line 110: suggestion: can we just break after setting the internal port to avoid entering the loop again?

mdelapenya (approved):
LGTM, thanks!

One simple comment regarding for loops 😅

wait/host_port.go
stevenh: We're looking for the lowest port which might not be first port, so not if we want to maintain the that current behaviour. This was introduced in this fix https://github.com/testcontainers/testcontainers-go/pull/2641

PR #3161, opened 2025-05-12
feat: support adding wait strategies as functional option

mdelapenya:
- **feat: add a new functional option to append wait strategies to modules**
- **chore: update modules**

## What does this PR do?
It adds a new functional option to add additional wait strategies to the container.

## Why is it important?
It's needed to allow users to append new wait strategies instead of replacing them all. At the moment a user used a module, and customised it with `WithWaitStrategy` the module strategies were overriden.

With this change, users can choose if overriding or appending.

### DIFF
--- a/docs/features/common_functional_options.md
+++ b/docs/features/common_functional_options.md
@@ -249,6 +249,8 @@ If you need to set a different wait strategy for the container, you can use `tes
 
 At the same time, it's possible to set a wait strategy and a custom deadline with `testcontainers.WithWaitStrategyAndDeadline`.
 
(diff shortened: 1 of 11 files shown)


mdelapenya:
@stevenh this PR would resolve the issue we discussed the other day about losing the original wait strategies. Instead of changing the behaviour of the current option, I'm adding a new one that makes explicit the addition of strategies, like we did with the life cycle hooks

mdelapenya:
Merging, as I need this to continue with the refactor of the docs in #3163

If you have any concern, please let's discuss it in a separate issue/PR 🙏

Issue #2748, opened 2024-08-26
[Bug]: Regression for Kafka


mixnblend:
### Testcontainers version

0.33.0

### Using the latest Testcontainers version?

Yes

### Host OS

Mac

### Host arch

ARM

### Go version

1.22

### Docker version

```shell
docker version
Client:
 Version:           25.0.4-rd
 API version:       1.44
 Go version:        go1.21.8
 Git commit:        c4cd0a9
 Built:             Fri Mar  8 09:09:46 2024
 OS/Arch:           darwin/arm64
 Context:           default

Server: Docker Engine - Community
 Engine:
  Version:          26.1.1
  API version:      1.45 (minimum version 1.24)
  Go version:       go1.21.9
  Git commit:       ac2de55
  Built:            Tue Apr 30 11:48:47 2024
  OS/Arch:          linux/arm64
  Experimental:     false
 containerd:
  Version:          1.6.31
  GitCommit:        e377cd56a71523140ca6ae87e30244719194a521
 runc:
  Version:          1.1.12
  GitCommit:        v1.1.12-0-g51d5e94
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### Docker info

```shell
docker info
Client: Docker Engine - Community
 Version:    27.1.2
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.16.2
    Path:     /Users/jonathanar/.docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  2.29.2
    Path:     /Users/jonathanar/.docker/cli-plugins/docker-compose

Server:
 Containers: 5
  Running: 5
  Paused: 0
  Stopped: 0
 Images: 57
 Server Version: 26.1.1
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: e377cd56a71523140ca6ae87e30244719194a521
[...]

Issue #2670, opened 2024-07-26
[Bug]: intermittent "port not found" error preventing Kafka and generic containers from starting


aakso:
### Testcontainers version

0.32.0

### Using the latest Testcontainers version?

Yes

### Host OS

MacOS

### Host arch

ARM

### Go version

1.21.12

### Docker version

```shell
Client:
 Version:           27.1.1
 API version:       1.46
 Go version:        go1.21.12
 Git commit:        6312585
 Built:             Tue Jul 23 19:54:12 2024
 OS/Arch:           darwin/arm64
 Context:           desktop-linux

Server: Docker Desktop 4.33.0 (160616)
 Engine:
  Version:          27.1.1
  API version:      1.46 (minimum version 1.24)
  Go version:       go1.21.12
  Git commit:       cc13f95
  Built:            Tue Jul 23 19:57:14 2024
  OS/Arch:          linux/arm64
  Experimental:     true
 containerd:
  Version:          1.7.19
  GitCommit:        2bf793ef6dc9a18e00cb12efb64355c2c9d5eb41
 runc:
  Version:          1.7.19
  GitCommit:        v1.1.13-0-g58aa920
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### Docker info

```shell
Client:
 Version:    27.1.1
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.16.1-desktop.1
    Path:     /Users/aakso/.docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.29.1-desktop.1
    Path:     /Users/aakso/.docker/cli-plugins/docker-compose
  debug: Get a shell into any image or container (Docker Inc.)
    Version:  0.0.34
    Path:     /Users/aakso/.docker/cli-plugins/docker-debug
  desktop: Docker Desktop commands (Alpha) (Docker Inc.)
    Version:  v0.0.14
    Path:     /Users/aakso/.docker/cli-plugins/docker-desktop
  dev: Docker Dev Environments (Docker Inc.)
    Version:  v0.1.2
    Path:     /Users/aakso/.docker/cli-plugins/docker-dev
  extension: Manages Docker extensions (Docker Inc.)
    Version:  v0.2.25
    Path:     /Users/aakso/.docker/cli-plugins/docker-extension
[...]

PR #3165, opened 2025-05-15
fix(kafka,redpanda): deadlock with remote Docker Host

mabrarov:
## Changes

* fix(kafka): waiting for mapped port without trying to connect to it (which is not possible due to Kafka is not started at that point of time) in container post start hook which prepares configuration for Kafka (fix for https://github.com/testcontainers/testcontainers-go/issues/2748).
* fix(redpanda): waiting for mapped ports without trying to connect to them (which is not possible due to Redpanda is not started at that point of time) in container post start hook before preparing configuration for Redpanda which requires completion of port mapping (fix for https://github.com/testcontainers/testcontainers-go/issues/2748).
* fix(redpanda): UNIX path for target file when copying files into container with Redpanda (which is Linux based) to run correctly when host OS is Windows.
* fix(redpanda): support of remote Docker Engine in tests utilizing TLS.

## What does this PR do?

1. Fixed deadlock at the start of container in Kafka and Redpanda modules when remote Docker Engine is used.
1. Fixed Redpanda module when host OS is Windows.

## Why is it important?

1. Usage of remote Docker Engine can be required for CI/CD.
1. Support of Windows host (e.g. using local Docker Desktop or remote Docker Engine) in Redpanda module.

## Related issues

- Closes https://github.com/testcontainers/testcontainers-go/issues/2748
- Relates https://github.com/testcontainers/testcontainers-go/issues/2670

## How to test this PR

Refer to:

1. https://github.com/mabrarov/testcontainers-go-kafka-2748 (feat/kafka_redpanda_wait_for_mapped_port branch of that repo uses this PR source branch) - to test https://github.com/testcontainers/testcontainers-go/issues/2748
1. https://github.com/mabrarov/testcontainers-go-kafka-2670 (feat/kafka_redpanda_wait_for_mapped_port branch of that repo uses this PR source branch) - to ensure there is no regression for https://github.com/testcontainers/testcontainers-go/issues/2670

## Follow-ups

Refer to https://github.com/testcontainers/testcontainers-go/issues/2670#issuecomment-2884488508.


mapped_port.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/docker/go-connections/nat"
+)
+
+// WaitForMappedPort waits for given local port to be mapped by checking mapping
```
stevenh, line 11: question: To confirm my understanding, what we need to support is the scenario where the port to be exposed but nothing in the container is listening on the internal port yet?

While this works using local docker desktop, arguably it shouldn't, it doesn't work in your environment. I would really be interested in the output from your environment for `nc -vz <ip> <exposed port>` when the container is in startup.

If this is the case we should add a new option to `HostPortStrategy` `SkipExternalCheck()` mirroring the existing [SkipInternalCheck()](https://github.com/testcontainers/testcontainers-go/blob/main/wait/host_port.go#L76) instead of this WaitForMappedPort function.

Is this something you can take on?

stevenh (commented):
I think I understand the issue you're experiencing now, so many thanks for all your effort on this.

If you could review the comment below to confirm we're on the same page and if so you're thoughts taking on the suggestion of how to proceed?

mapped_port.go
mdelapenya: > I like this approach (I never liked this new function). Yes, I can do this.

@mabrarov thanks for this! We appreciate your hard work with the kafka module, all the comments you did in many other PRs, the deep research about the root cause (which made us to suggest reusing the existing wait-strategy), so please feel free to ping us with your suggestions on how to make this complex module easier to use 🙏

modules/redpanda/redpanda_test.go
```diff
@@ -509,19 +514,21 @@ func TestRedpandaWithTLS(t *testing.T) {
...
 
 func TestRedpandaWithTLSAndSASL(t *testing.T) {
-	tmp := t.TempDir()
+	ctx := context.Background()
 
+	containerHostAddress, err := detectContainerHostAddress(testImage, ctx)
+	require.NoError(t, err, "failed to detect container host address")
+
+	tmp := t.TempDir()
 	cert := tlscert.SelfSignedFromRequest(tlscert.Request{
```
stevenh, line 523: fyi: there's now a new method SelfSignedFromRequestE which returns an error, which is more idiomatic.

No need to change but wanted to make you aware.

modules/redpanda/redpanda_test.go
```diff
@@ -698,3 +705,23 @@ func TestRedpandaBootstrapConfig(t *testing.T) {
 		require.False(t, needsRestart)
 	}
 }
+
+func detectContainerHostAddress(image string, ctx context.Context) (string, error) {
```
stevenh, line 709: suggestion: pass in the test so you can use require, eliminating the need to return an error.

modules/redpanda/redpanda_test.go
```diff
@@ -698,3 +705,23 @@ func TestRedpandaBootstrapConfig(t *testing.T) {
...
+	c, err := testcontainers.GenericContainer(ctx, testcontainers.GenericContainerRequest{
+		ContainerRequest: testcontainers.ContainerRequest{
+			Image: image,
+		},
+		Started: false,
+	})
```
stevenh, line 715: suggestion: use [testcontainers.CleanupContainer](https://pkg.go.dev/github.com/testcontainers/testcontainers-go#CleanupContainer)

wait/host_port.go
```diff
@@ -178,6 +187,15 @@ func (hp *HostPortStrategy) WaitUntilReady(ctx context.Context, target StrategyT
 		}
 	}
 
+	if hp.skipExternalCheck {
+		return nil
+	}
+
+	ipAddress, err := target.Host(ctx)
```
stevenh, line 194: praise: nice catch on this wasn't needed if skipping.

wait/host_port.go
```diff
@@ -178,6 +187,15 @@ func (hp *HostPortStrategy) WaitUntilReady(ctx context.Context, target StrategyT
...
+	if err != nil {
+		return err
```
stevenh, line 196: suggestion: I know this was just moved but we're making sure all new code wraps returned errors to make it easier to identify the source.
```suggestion
		return fmt.Error("host: %w", err)
```

modules/kafka/kafka.go
```diff
@@ -125,6 +125,7 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
 func copyStarterScript(ctx context.Context, c testcontainers.Container) error {
 	if err := wait.ForListeningPort(publicPort).
 		SkipInternalCheck().
+		SkipExternalCheck().
```
stevenh, line 128: suggestion: looking at this I think a helper would be nice e.g. `ForMappedPort(<port>)` which sets skip internal and external.

For example add to `host_port.go`
```go
func ForMappedPort(port nat.Port) *HostPortStrategy {
	return NewHostPortStrategy(port).SkipInternalCheck().SkipExternalCheck()
}
```

stevenh (commented):
Thanks for the updates, getting there.

Have put some suggestions for you, if you aren't aware we use [conventional comments](https://conventionalcomments.org/) to help folks understand the intent better.

stevenh (changes requested):
Missed clicking request changes, see above.

mapped_port.go
stevenh: > Hi @stevenh,
> > I would really be interested in the output from your environment for `nc -vz <ip> <exposed port>` when the container is in startup.
> 
> Depending on tool (I've tried curl with `telent://` and telnet client in Linux) the results are different and confusing. There are 3 test environments listed in description of this pull request. Please let me know what environment (I guess, all) and what part of environment (in case of remote Docker Engine) you want me to try. I'm not sure what is the best replacement for `nc` on Windows (I can try to install `nc` as part of MSYS2, but such test can introduce additional behavior specific to MSYS) - maybe PowerShell [Test-NetConnection](https://learn.microsoft.com/en-us/powershell/module/nettcpip/test-netconnection?view=windowsserver2025-ps) commandlet... Please note that #2748 (which this pull request fixes) seems to be actual only for remote Docker Engine (I failed to reproduce it with local Docker Engine on Linux or with local Docker Desktop on Windows).

Yer interested in the different docker targets, the source can be anything that you have nc on. I'm interested if we see different results from a common source i.e. is there variation depending on when / how docker is hosted.

modules/redpanda/redpanda_test.go
mabrarov: IMHO, it is better to remove this (not started, temporary) container immediately (within this function) without waiting for completion of test (if I understand [testcontainers.CleanupContainer](https://pkg.go.dev/github.com/testcontainers/testcontainers-go#CleanupContainer) correctly).

mabrarov:
Hi @stevenh,

Regarding your [ask](https://github.com/testcontainers/testcontainers-go/pull/3165#discussion_r2097416297) for details:

> While this works using local docker desktop, arguably it shouldn't, it doesn't work in your environment. I would really be interested in the output from your environment for `nc -vz <ip> <exposed port>` when the container is in startup.

I updated https://github.com/mabrarov/testcontainers-go-kafka-2748/blob/master/README.md and the last screenshot (at the bottom of file) confuses me - it looks like wrong port is checked there (33509 instead of 33508) - so I'm going to investigate it deeper.

Thank you.

stevenh:
> Hi @stevenh,
> 
> Regarding your [ask](https://github.com/testcontainers/testcontainers-go/pull/3165#discussion_r2097416297) for details:
> 
> > While this works using local docker desktop, arguably it shouldn't, it doesn't work in your environment. I would really be interested in the output from your environment for `nc -vz <ip> <exposed port>` when the container is in startup.
> 
> I updated https://github.com/mabrarov/testcontainers-go-kafka-2748/blob/master/README.md and the last screenshot (at the bottom of file) confuses me - it looks like wrong port is checked there (33509 instead of 33508) - so I'm going to investigate it deeper.
> 
> Thank you.

Not sure I understand 33508 is the ryuk port, 33509 is the kafka ports, so its what I would expect?

mabrarov:
Hi @stevenh,

Yes, my test was wrong (was checking wrong port from PowerShell). Updating it right now.

Thank you.

modules/redpanda/redpanda.go
```diff
@@ -158,7 +158,7 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
 		},
 		testcontainers.ContainerFile{
 			Reader:            bytes.NewReader(bootstrapConfig),
-			ContainerFilePath: filepath.Join(redpandaDir, bootstrapConfigFile),
+			ContainerFilePath: filepath.ToSlash(filepath.Join(redpandaDir, bootstrapConfigFile)),
```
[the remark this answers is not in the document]
mabrarov: Done

modules/redpanda/redpanda_test.go
mabrarov: Done

modules/redpanda/redpanda_test.go
mabrarov: Done

modules/redpanda/redpanda_test.go
mabrarov: Done (it makes the test simpler).

wait/host_port.go
mabrarov: Done

modules/kafka/kafka.go
mabrarov: Done

mapped_port.go
mabrarov: Hi @mdelapenya,

My team uses Testcontainers for Go and is happy somebody created (and published) it for us. When we have issue and need a change we can open pull request to get it implemented sooner, so "nothing personal, just business" :wink:

To be honest, neither this pull request nor the root cause of https://github.com/testcontainers/testcontainers-go/issues/2748 are specific to Kafka or Redpanda. The only "specific" part is inconvenient requirement to know the exposed port before starting Kafka for generation of Kafka configuration.

The only suggestion for Kafka module I have is described in https://github.com/testcontainers/testcontainers-go/issues/2206#issuecomment-2887726469 and is demonstrated in https://github.com/mabrarov/testcontainers-go-kafka-2206/pull/1 (it can be enhanced if / after this pull request is taken into some release of Testcontainers for Go).

Thank you and the team supporting Testcontainers for Go - it is really a game changer in the world of (automated) testing.

modules/redpanda/redpanda_test.go
```diff
@@ -698,3 +701,19 @@ func TestRedpandaBootstrapConfig(t *testing.T) {
...
 }
+
+func detectContainerHostAddress(t *testing.T, image string, ctx context.Context) string {
+	c, err := testcontainers.GenericContainer(ctx, testcontainers.GenericContainerRequest{
+		ContainerRequest: testcontainers.ContainerRequest{
+			Image: image,
+		},
+		Started: false,
+	})
+	require.NoErrorf(t, err, "failed to create stopped container")
```
stevenh, line 712: suggestion: just use NoError, we don't care what type it is just that it isn't an error.

More below...

wait/host_port.go
```diff
@@ -124,15 +144,11 @@ func (hp *HostPortStrategy) WaitUntilReady(ctx context.Context, target StrategyT
...
-	ipAddress, err := target.Host(ctx)
-	if err != nil {
-		return err
-	}
-
 	waitInterval := hp.PollInterval
 
 	internalPort := hp.Port
 	i := 0
+	var err error
```
stevenh, line 151: suggestion: move err inside the if as its not needed outside with the addition of the port, err :=

stevenh (changes requested):
Getting close, just a few bits to fix.

stevenh:
> Hi @stevenh,
> 
> https://github.com/mabrarov/testcontainers-go-kafka-2748/blob/master/README.md is fixed now and is ready for your review ([Local Docker Engine](https://github.com/mabrarov/testcontainers-go-kafka-2748/blob/master/README.md#local-docker-engine) and [Remote Docker Engine](https://github.com/mabrarov/testcontainers-go-kafka-2748/blob/master/README.md#remote-docker-engine) sections). In the meantime I'm going to address this pull request comments.
> 
> Thank you for your patience.

So it looks like nc is getting the right results, could you also run a test with go:
```go
package main

import (
	"context"
	"fmt"
	"net"
	"os"
	"time"
)

func main() {
	if len(os.Args) < 2 {
		fmt.Println("Usage: go run main.go <host:port>")
		return
	}
	address := os.Args[1]

	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
	defer cancel()

	dialer := &net.Dialer{}
	conn, err := dialer.DialContext(ctx, "tcp", address)
	if err != nil {
		fmt.Printf("Dial error: %#v\n", err)
		return
	}
	defer conn.Close()

	fmt.Printf("Connected to: %q\n", address)
}
```

wait/host_port.go
mabrarov: I can do only

```go
	i := 0
	if internalPort == "" {
		var err error
		// Port is not specified, so we need to detect it.
		internalPort, err = hp.detectInternalPort(ctx, target)
	...
	}
```

but not

```go
	i := 0
	if internalPort == "" {
		// Port is not specified, so we need to detect it.
		internalPort, err := hp.detectInternalPort(ctx, target)
	...
	}
```

because the last one declares another `internalPort` variable inside `if` which shadows `internalPort` variable used outside of `if`.

I can do

```go
	i := 0
	if internalPort == "" {
		// Port is not specified, so we need to detect it.
		port, err := hp.detectInternalPort(ctx, target)
		if err != nil {
			return fmt.Errorf("detect internal port: %w", err)
		}

		for port == "" {
			select {
			case <-ctx.Done():
				return fmt.Errorf("detect internal port: retries: %d, last err: %w, ctx err: %w", i, err, ctx.Err())
			case <-time.After(waitInterval):
				if err := checkTarget(ctx, target); err != nil {
					return fmt.Errorf("detect internal port: check target: retries: %d, last err: %w", i, err)
				}

				port, err = hp.detectInternalPort(ctx, target)
				if err != nil {
					return fmt.Errorf("detect internal port: %w", err)
				}
			}
		}
		
		internalPort = port
	}
```

but it seems to be worse than the first case.

mabrarov: Done. Used first option.

modules/redpanda/redpanda_test.go
mabrarov: Done

modules/redpanda/redpanda_test.go
```diff
@@ -509,19 +512,19 @@ func TestRedpandaWithTLS(t *testing.T) {
 }
 
 func TestRedpandaWithTLSAndSASL(t *testing.T) {
-	tmp := t.TempDir()
+	ctx := context.Background()
 
-	cert := tlscert.SelfSignedFromRequest(tlscert.Request{
+	containerHostAddress := detectContainerHostAddress(t, testImage, ctx)
+	tmp := t.TempDir()
```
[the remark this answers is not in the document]
mabrarov: Done.

modules/redpanda/redpanda_test.go
```diff
@@ -698,3 +701,19 @@ func TestRedpandaBootstrapConfig(t *testing.T) {
...
+
+func detectContainerHostAddress(t *testing.T, image string, ctx context.Context) string {
+	c, err := testcontainers.GenericContainer(ctx, testcontainers.GenericContainerRequest{
+		ContainerRequest: testcontainers.ContainerRequest{
+			Image: image,
+		},
+		Started: false,
+	})
+	require.NoErrorf(t, err, "failed to create stopped container")
+	testcontainers.CleanupContainer(t, c)
```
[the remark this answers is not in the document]
mabrarov: Fixed.

stevenh (changes requested):
I think this is now an elegant solution, thanks for all your hard work on this.

The only thing missing which @mdelapenya would pick up on is it needs docs for the new methods, sorry I always forget too as go docs are generated from the code, but testcontainers-go also has its own site, which needs manually added docs unfortunately.

stevenh:
> Hi @stevenh,
>
> Done - updated https://github.com/mabrarov/testcontainers-go-kafka-2748/blob/master/README.md
> 
> Thank you.
Grr, sorry we need to unpack the error a bit:
```go
package main

import (
	"context"
	"errors"
	"fmt"
	"net"
	"os"
	"time"
)

func main() {
	if len(os.Args) < 2 {
		fmt.Println("Usage: go run main.go <host:port>")
		return
	}
	address := os.Args[1]

	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
	defer cancel()

	var dialer net.Dialer
	conn, err := dialer.DialContext(ctx, "tcp", address)
	if err != nil {
		var opErr *net.OpError
		if errors.As(err, &opErr) {
			var syscallErr *os.SyscallError
			if errors.As(opErr.Err, &syscallErr) {
				fmt.Printf("Dial error: %#v\n", syscallErr)
				return
			}
			fmt.Printf("Dial error: %#v\n", opErr)
			return
		}
		fmt.Printf("Dial error: %#v\n", err)
		return
	}
	defer conn.Close()

	fmt.Printf("Connected to: %q\n", address)
}
```

stevenh:
can we also see the result of `nc -vz` from the remote host, the same as the go test as I'm trying to see if there is something in the go causing a different behaviour when connected to remotely compared to locally, so want to compare that with a native `nc`.

stevenh:
@mabrarov can I also see the result of `nc -vz` from the remote host, the same as the go test as I'm trying to see if there is something in the go causing a different behaviour when connected to remotely compared to locally, so want to compare that with a native `nc`

mabrarov:
Hi @stevenh,

> the result of `nc -vz` from the remote host

My host OS is Windows 11 which doesn't have `nc`. I can try `nc` from another VM (comparing to VM where Docker Engine runs). Will it be sufficient?

Thank you.

stevenh:
You can or it should run just fine under WSL.

mabrarov:
Hi @stevenh,

I have updated https://github.com/mabrarov/testcontainers-go-kafka-2748/blob/master/README.md with results of `nc` and with tests with new version of `dial` app.

Thank you.

stevenh:
> I have updated https://github.com/mabrarov/testcontainers-go-kafka-2748/blob/master/README.md with results of `nc` and with tests with new version of `dial` app.

Thanks, so to confirm my understanding, if you use nc on host you see:
```shell
nc -vz 127.0.0.1 33114
Ncat: Version 7.92 ( https://nmap.org/ncat )
Ncat: Connected to 127.0.0.1:33114.
Ncat: 0 bytes sent, 0 bytes received in 0.01 seconds.
```

But from the remote host you see:
```shell
nc -vz 192.168.204.133 33114
nc: connect to 192.168.204.133 port 33114 (tcp) failed: Connection refused
```
Does the on result change if you use the external IP not the loopback address on host?
```shell
nc -vz 192.168.204.133 33114
```

mabrarov:
Hi @stevenh 

> if you use nc on host you see:
> 
> ```shell
> nc -vz 127.0.0.1 33114
> Ncat: Version 7.92 ( https://nmap.org/ncat )
> Ncat: Connected to 127.0.0.1:33114.
> Ncat: 0 bytes sent, 0 bytes received in 0.01 seconds.
> ```
> 
> But from the remote host you see:
> 
> ```shell
> nc -vz 192.168.204.133 33114
> nc: connect to 192.168.204.133 port 33114 (tcp) failed: Connection refused
> ```

Yes, that's correct.

> Does the on result change if you use the external IP not the loopback address on host?
> 
> ```shell
> nc -vz 192.168.204.133 33114
> ```

Yes, the result changes - I get "Connection refused" - if I use external IP for checking connection from the same host where Docker Engine runs. I included this test into https://github.com/mabrarov/testcontainers-go-kafka-2748/blob/master/README.md#remote-docker-engine.

Thank you.

stevenh:
> I have updated https://github.com/mabrarov/testcontainers-go-kafka-2748/blob/master/README.md with results of `nc` and with tests with new version of `dial` app.
Now that is interesting, could you check how the ports are really bound , for example:
```shell
netstat -anv | grep <port>
```

mabrarov:
Hi @stevenh,

> Now that is interesting, could you check how the ports are really bound , for example:
> 
> ```shell
> netstat -anv | grep <port>
> ```

Just to ensure you read the results of test correctly: please note that test is stopped at breakpoint at the time I check state of containers in https://github.com/mabrarov/testcontainers-go-kafka-2748/blob/master/README.md

Here is output of `netstat`:

```text
$ docker ps -a
CONTAINER ID   IMAGE                                COMMAND                  CREATED         STATUS         PORTS                                                               NAMES
4ac21d122004   testcontainers/ryuk:0.11.0           "/bin/ryuk"              2 minutes ago   Up 2 minutes   0.0.0.0:32768->8080/tcp, [::]:32768->8080/tcp                       reaper_536606a398bc0a941320379d2bda17eda930b52a136151eed5e87c3cee9d13c3
62cd650f0579   confluentinc/confluent-local:7.5.0   "sh -c 'while [ ! -f…"   2 minutes ago   Up 2 minutes   8082/tcp, 9092/tcp, 0.0.0.0:32769->9093/tcp, [::]:32769->9093/tcp   cool_thompson
$ docker top 62cd650f0579
UID                 PID                 PPID                C                   STIME               TTY                 TIME                CMD
abrarovm            2500                2478                0                   17:09               ?                   00:00:01            sh -c while [ ! -f /usr/sbin/testcontainers_start.sh ]; do sleep 0.1; done; bash /usr/sbin/testcontainers_start.sh
abrarovm            4321                2500                0                   17:12               ?                   00:00:00            /usr/bin/coreutils --coreutils-prog-shebang=sleep /usr/bin/sleep 0.1
$ docker logs 62cd650f0579
$ nc -vz 127.0.0.1 32769
Ncat: Version 7.92 ( https://nmap.org/ncat )
Ncat: Connected to 127.0.0.1:32769.
Ncat: 0 bytes sent, 0 bytes received in 0.01 seconds.
$ ./.build/dial 127.0.0.1:32769
Connected to: "127.0.0.1:32769"
$ ip address show | grep 192
    inet 192.168.204.133/24 brd 192.168.204.255 scope global dynamic noprefixroute ens160
$ nc -vz 192.168.204.133 32769
Ncat: Version 7.92 ( https://nmap.org/ncat )
Ncat: Connection refused.
$ ./.build/dial  192.168.204.133:32769
Dial error: &os.SyscallError{Syscall:"connect", Err:0x6f}
$ netstat -anv | grep 32769
netstat: no support for `AF INET (sctp)' on this system.
netstat: no support for `AF INET (sctp)' on this system.
tcp        0      0 0.0.0.0:32769           0.0.0.0:*               LISTEN
tcp        0      0 127.0.0.1:59368         127.0.0.1:32769         TIME_WAIT
tcp6       0      0 :::32769                :::*                    LISTEN
netstat: no support for `AF IPX' on this system.
netstat: no support for `AF AX25' on this system.
netstat: no support for `AF X25' on this system.
netstat: no support for `AF NETROM' on this system.
netstat: no support for `BTPROTO L2CAP' on this system.
netstat: no support for `BTPROTO RFCOMM' on this system.
$ netstat -anv | grep 32769
netstat: no support for `AF INET (sctp)' on this system.
netstat: no support for `AF INET (sctp)' on this system.
tcp        0      0 0.0.0.0:32769           0.0.0.0:*               LISTEN
tcp        0      0 127.0.0.1:59368         127.0.0.1:32769         TIME_WAIT
tcp6       0      0 :::32769                :::*                    LISTEN
netstat: no support for `AF IPX' on this system.
netstat: no support for `AF AX25' on this system.
netstat: no support for `AF X25' on this system.
netstat: no support for `AF NETROM' on this system.
netstat: no support for `BTPROTO L2CAP' on this system.
netstat: no support for `BTPROTO RFCOMM' on this system.
```

Thank you.

stevenh:
Ok after researching this, what we're seeing where `127.0.0.1:32769` connects but `192.168.x.x:32769` doesn’t is expected behaviour, despite Docker mapping the port.

This is explained by how the Linux networking stack handles loopback vs external connections for ports forwarded by Docker when the target container isn’t yet listening on the internal port.

**Key Points**
1.	Docker Port Mapping (NAT) and iptables:
	•	Docker sets up NAT rules via iptables to forward traffic from host ports (e.g., `32769`) to the container’s internal port (e.g., 9093).
	•	These rules redirect traffic only when there’s a process in the container listening on the mapped port.
2.	Loopback Shortcut:
	•	When connecting via `127.0.0.1:32769`, Linux often uses a loopback optimization: it bypasses full NAT and routing logic. Since the port appears to be listening (LISTEN from Docker proxy or similar), the connection is established at the TCP layer, even if the internal container process hasn’t yet started listening.
3.	Non-loopback Interfaces Are Routed Fully:
	•	When using your external IP (`192.168.204.133:32769`), the connection must pass through full routing and NAT rules. Since the container process isn’t listening yet, the NAT rule cannot complete the destination translation properly, and the kernel returns `ECONNREFUSED` (error code 0x6f).
4.	netstat Shows Listening Because of Docker’s Userland Proxy or socket activation:
	•	Docker may use a userland proxy or set up a socket to accept connections (even when the internal service isn’t ready), especially on loopback.
	•	However, for external addresses, the connection still requires actual routing to the container’s listener, which fails when the process isn’t ready.

stevenh (changes requested):
Code is all good, just needs the docs updating, in case you missed that from my earlier comment.

wait/host_port.go
```diff
@@ -178,6 +193,15 @@ func (hp *HostPortStrategy) WaitUntilReady(ctx context.Context, target StrategyT
 		}
 	}
 
+	if hp.skipExternalCheck {
+		return nil
```
mabrarov, line 197: What if `SkipExternalCheck()` is used without `SkipInternalCheck()`? Doesn't user expect internal check to happen in this case?

stevenh (dismissed):
Looks good to me 😍

modules/redpanda/redpanda_test.go
```diff
@@ -698,3 +697,20 @@ func TestRedpandaBootstrapConfig(t *testing.T) {
 		require.False(t, needsRestart)
 	}
 }
+
+func detectContainerHostAddress(ctx context.Context, t *testing.T, image string) string {
+	t.Helper()
+	c, err := testcontainers.GenericContainer(ctx, testcontainers.GenericContainerRequest{
```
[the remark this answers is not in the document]
mabrarov: Hi @mdelapenya,

Your suggestion matches my initial thought, but I decided to not rely on `testcontainers.DockerProvider`, because:

1. It makes this test more coupled with specific implementation (provider).
1. This address should match the one returned by [redpanda.Container.KafkaSeedBroker](https://github.com/testcontainers/testcontainers-go/blob/main/modules/redpanda/redpanda.go#L287) method. This method doesn't rely on `testcontainers.DockerProvider.DaemonHost` (like, for example, [pulsar.Container.BrokerURL](https://github.com/testcontainers/testcontainers-go/blob/main/modules/pulsar/pulsar.go#L37) method [does](https://github.com/testcontainers/testcontainers-go/blob/main/modules/pulsar/pulsar.go#L52)), but relies on [testcontainers.Container.PortEndpoint](https://github.com/testcontainers/testcontainers-go/blob/main/container.go#L43) method which [uses](https://github.com/testcontainers/testcontainers-go/blob/main/docker.go#L144) [testcontainers.Container.Host](https://github.com/testcontainers/testcontainers-go/blob/main/container.go#L44) to determine address for communication with container. **Implementation** of `testcontainers.Container.Host` [uses](https://github.com/testcontainers/testcontainers-go/blob/main/docker.go#L166) `testcontainers.DockerProvider.DaemonHost`. I agree that creation of container is not an elegant solution, but my intention was to not rely on internals as much as possible (stop the dive into internals as soon as possible, to make this test more resilient to changes) and I expected that the choice will be made during review.

There is [k3s.getContainerHost](https://github.com/testcontainers/testcontainers-go/blob/main/modules/k3s/k3s.go#L114) method which I can re-implement / adopt in this test as alternative to your suggestion.

What is your opinion (choice)?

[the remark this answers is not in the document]
mdelapenya: Fine with the k3s implementation. We could even move it to the core so it's reusable by other modules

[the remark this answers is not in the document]
mabrarov: Hi @mdelapenya,

I included `testcontainers.GetContainerHost` function (moved from [k3s.getContainerHost](https://github.com/testcontainers/testcontainers-go/blob/main/modules/k3s/k3s.go#L114)) into this pull request.

Additionally, I revisited all places, where `testcontainers.DockerProvider.DaemonHost` is used, and replaced calls of this method with call of `testcontainers.DockerContainer.Host` method if possible or with call of (new) `testcontainers.GetContainerHost` function otherwise.

The only missing parts are:

1. Tests for the new `testcontainers.GetContainerHost` function (sorry, it is too hard / time consuming for me).
1. Documentation for `testcontainers.GetContainerHost` (I'm not sure what file to add this documentation).

Due to more files are impacted by this pull request now, I appreciate if you and @stevenh could make one more round of review. Sorry for increasing the scope of this pull request, but IMHO it feels to be correct direction.

Thank you.

generic.go
```diff
@@ -148,3 +148,33 @@ func Run(ctx context.Context, img string, opts ...ContainerCustomizer) (*DockerC
 
 	return c, nil
 }
+
+// GetContainerHost gets host (IP or name) of the docker daemon where container port is exposed.
+// Note: prefer using Container.Host when it is possible.
+// Warning: this function is based on Docker host setting, it will fail if using an SSH tunnel.
+// "TESTCONTAINERS_HOST_OVERRIDE" env variable can be used to customize host returned by this function.
+func GetContainerHost(ctx context.Context, opts ...ContainerCustomizer) (string, error) {
```
stevenh, line 156: suggestion: it's idiomatic to avoid Get prefix used by other languages unless there's a conflict.

@mdelapenya how does this play with the incoming context support, are we introducing something that will be redundant or need changing for that?

I suspect this was introduced in response to PR feedback, but its bloating the surface area of this, might be better to do it as a follow up, @mdelapenya thoughts?

stevenh (changes requested):

generic.go
mabrarov: Renamed to `ContainerHost`.

mabrarov:
Hi @mdelapenya,

Regarding my [comment](https://github.com/testcontainers/testcontainers-go/pull/3165#issuecomment-2910326600) about unstable tests in reuse_test.go:

1. I still think that configurable timeout will be helpful (I prefer all timings to be configurable because they depend on hardware usually).
1. The issue is not caused by too small timeout, but is caused by Docker client reporting that previously created container is running while it is stopped.

I tried main branch on Rocky Linux 9.5 with Docker Engine 28.1.1 and found that sometimes test passes and sometimes test fails. The failed test looks like:

```text
$ go test -v -count 1 -run TestGenericContainer_stop_start_withReuse . || docker ps -a
=== RUN   TestGenericContainer_stop_start_withReuse
2025/05/27 20:28:09 github.com/testcontainers/testcontainers-go - Connected to docker:
  Server Version: 28.1.1
  API Version: 1.49
  Operating System: Rocky Linux 9.5 (Blue Onyx)
  Total Memory: 15704 MB
  Testcontainers for Go Version: v0.38.0
  Resolved Docker Host: unix:///var/run/docker.sock
  Resolved Docker Socket Path: /var/run/docker.sock
  Test SessionID: bcf6e12b4cdec61c01215ade3310f5e142a1646bca0e87e7505c551c8f70a5fe
  Test ProcessID: 276c500e-c9e1-49b8-8e0d-26fcd4644951
2025/05/27 20:28:09 🐳 Creating container for image nginx:alpine
2025/05/27 20:28:09 🐳 Creating container for image testcontainers/ryuk:0.11.0
2025/05/27 20:28:09 ✅ Container created: 931b882451ca
2025/05/27 20:28:09 🐳 Starting container: 931b882451ca
2025/05/27 20:28:09 ✅ Container started: 931b882451ca
2025/05/27 20:28:09 ⏳ Waiting for container id 931b882451ca image: testcontainers/ryuk:0.11.0. Waiting for: &{Port:8080/tcp timeout:<nil> PollInterval:100ms skipInternalCheck:false}
2025/05/27 20:28:09 🔔 Container is ready: 931b882451ca
2025/05/27 20:28:09 ✅ Container created: cf711b2a8ec9
2025/05/27 20:28:09 🐳 Starting container: cf711b2a8ec9
2025/05/27 20:28:09 ✅ Container started: cf711b2a8ec9
2025/05/27 20:28:09 🔔 Container is ready: cf711b2a8ec9
2025/05/27 20:28:09 🐳 Stopping container: cf711b2a8ec9
2025/05/27 20:28:19 ✅ Container stopped: cf711b2a8ec9
2025/05/27 20:28:19 ✅ Container started: cf711b2a8ec9
2025/05/27 20:28:19 All requested ports were not exposed: port 8080/tcp is not mapped yet
2025/05/27 20:28:19 All requested ports were not exposed: port 8080/tcp is not mapped yet
2025/05/27 20:28:19 All requested ports were not exposed: port 8080/tcp is not mapped yet
2025/05/27 20:28:19 All requested ports were not exposed: port 8080/tcp is not mapped yet
2025/05/27 20:28:20 All requested ports were not exposed: port 8080/tcp is not mapped yet
2025/05/27 20:28:20 All requested ports were not exposed: port 8080/tcp is not mapped yet
2025/05/27 20:28:20 All requested ports were not exposed: port 8080/tcp is not mapped yet
2025/05/27 20:28:21 All requested ports were not exposed: port 8080/tcp is not mapped yet
2025/05/27 20:28:21 All requested ports were not exposed: port 8080/tcp is not mapped yet
2025/05/27 20:28:21 All requested ports were not exposed: port 8080/tcp is not mapped yet
2025/05/27 20:28:22 All requested ports were not exposed: port 8080/tcp is not mapped yet
2025/05/27 20:28:22 All requested ports were not exposed: port 8080/tcp is not mapped yet
2025/05/27 20:28:23 All requested ports were not exposed: port 8080/tcp is not mapped yet
2025/05/27 20:28:23 All requested ports were not exposed: port 8080/tcp is not mapped yet
2025/05/27 20:28:24 container logs (all exposed ports, [8080/tcp], were not mapped in 5s: port 8080/tcp is not mapped yet):
/docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration
/docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/
/docker-entrypoint.sh: Launching /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh
10-listen-on-ipv6-by-default.sh: info: Getting the checksum of /etc/nginx/conf.d/default.conf
10-listen-on-ipv6-by-default.sh: info: Enabled listen on IPv6 in /etc/nginx/conf.d/default.conf
/docker-entrypoint.sh: Sourcing /docker-entrypoint.d/15-local-resolvers.envsh
/docker-entrypoint.sh: Launching /docker-entrypoint.d/20-envsubst-on-templates.sh
/docker-entrypoint.sh: Launching /docker-entrypoint.d/30-tune-worker-processes.sh
/docker-entrypoint.sh: Configuration complete; ready for start up
2025/05/27 17:28:09 [notice] 1#1: using the "epoll" event method
2025/05/27 17:28:09 [notice] 1#1: nginx/1.27.5
2025/05/27 17:28:09 [notice] 1#1: built by gcc 14.2.0 (Alpine 14.2.0)
2025/05/27 17:28:09 [notice] 1#1: OS: Linux 5.14.0-503.40.1.el9_5.x86_64
2025/05/27 17:28:09 [notice] 1#1: getrlimit(RLIMIT_NOFILE): 4000000:4000000
2025/05/27 17:28:09 [notice] 1#1: start worker processes
2025/05/27 17:28:09 [notice] 1#1: start worker process 30
2025/05/27 17:28:09 [notice] 1#1: start worker process 31
2025/05/27 17:28:09 [notice] 1#1: start worker process 32
2025/05/27 17:28:09 [notice] 1#1: start worker process 33
2025/05/27 17:28:09 [notice] 1#1: start worker process 34
2025/05/27 17:28:09 [notice] 1#1: start worker process 35
2025/05/27 17:28:09 [notice] 1#1: start worker process 36
2025/05/27 17:28:09 [notice] 1#1: start worker process 37

    reuse_test.go:39:
                Error Trace:    /ws/github/mabrarov/testcontainers-go/reuse_test.go:39
                Error:          Received unexpected error:
                                create container: all exposed ports, [8080/tcp], were not mapped in 5s: port 8080/tcp is not mapped yet
                Test:           TestGenericContainer_stop_start_withReuse
2025/05/27 20:28:24 🐳 Stopping container: cf711b2a8ec9
2025/05/27 20:28:24 ✅ Container stopped: cf711b2a8ec9
2025/05/27 20:28:24 🐳 Terminating container: cf711b2a8ec9
2025/05/27 20:28:24 🚫 Container terminated: cf711b2a8ec9
--- FAIL: TestGenericContainer_stop_start_withReuse (15.36s)
FAIL
FAIL    github.com/testcontainers/testcontainers-go     15.375s
FAIL
```

Here are screenshots from debugger for another run and the same fail of the same test (in order of screenshot creation):

Screenshot from 2025-05-27 20-26-14
Screenshot from 2025-05-27 20-26-37

I suspect Docker Daemon caches state of container or uses asynchronous logic when container is stopped which causes Docker client used by Testcontainers for Go to report about container running.

It seems to be a serious issue which requires dedicated investigation (including Docker Engine bug-tracker review) and workaround / fix.

Thank you.

mdelapenya:
@mabrarov 

> I still think that configurable timeout will be helpful (I prefer all timings to be configurable because they depend on hardware usually).

Working on this now

mabrarov:
Hi @mdelapenya,

I opened https://github.com/testcontainers/testcontainers-go/issues/3177 for the issue described in https://github.com/testcontainers/testcontainers-go/pull/3165#issuecomment-2913422933. Sorry, but I failed to find time to make a deeper investigation (like checking Docker Engine issues or checking if some sort of caching happens inside Testcontainers for Go).

Thank you.

stevenh (approved):

mdelapenya:
@mabrarov I've renamed the PR title so that it's even clearer what we do when reading the release notes. Please let me know what you think

mdelapenya:
Merged, thank you so much!

modules/redpanda/redpanda_test.go
```diff
@@ -698,3 +700,29 @@ func TestRedpandaBootstrapConfig(t *testing.T) {
...
+
+	logging := req.Logger
+	if logging == nil {
+		logging = log.Default()
+	}
+	p, err := req.ProviderType.GetProvider(testcontainers.WithLogger(logging))
+	if err != nil {
+		return "", err
+	}
+
```
mabrarov, line 721:
```go
defer p.Close()
```

?

mabrarov: Refer to https://github.com/testcontainers/testcontainers-go/compare/main...mabrarov:testcontainers-go:fix/kafka_redpanda_wait_for_mapped_port

mdelapenya: @mabrarov could you submit a small fix for that? 🙏

mabrarov: Hi @mdelapenya,

Here is it: https://github.com/testcontainers/testcontainers-go/pull/3539.

PR #3168, opened 2025-05-22
feat!: add options when creating RawCommand

benja-M-1:
## What does this PR do?

This PR adds a second optional parameter to the `testcontainers.NewRawCommand` function to pass executable options to it.

## Why is it important?

This allows any startup command to be ran with options like a different user, working directory and any available executable option.

For example:

```go
testcontainers.Run(
	context.Background(),
	"rust:latest",
	tc.WithEntrypoint("tail"),
	tc.WithCmd("-f", "/dev/null"),
	tc.WithConfigModifier(func(c *container.Config) {
		c.WorkingDir = "/tmp"
	}),
	tc.WithStartupCommand(
		// This creates the /tmp/hello_world directory
		NewExecCommand([]string{"cargo", "new", "hello_world", "--vcs", "none"}),
		// This will add a dependency executing cargo from /tmp/hello_world
		NewExecCommand([]string{"cargo", "add", "rand"}, tcexec.WithWorkingDir("/tmp/hello_world")),
		NewExecCommand([]string{"cargo", "build"}, tcexec.WithWorkingDir("/tmp/hello_world")),
	),
	tc.WithWaitStrategy(wait.ForAll(
		wait.ForFile("/tmp/hello_world/target/debug/hello_world"),
	)),
```

## Related issues

-

### DIFF
--- a/options.go
+++ b/options.go
@@ -311,11 +311,11 @@ type RawCommand struct {
 	cmds []string
 }
 
-func NewRawCommand(cmds []string) RawCommand {
+func NewRawCommand(cmds []string, opts ...tcexec.ProcessOption) RawCommand {
 	return RawCommand{
 		cmds: cmds,
 		ExecOptions: ExecOptions{
-			opts: []tcexec.ProcessOption{},
+			opts: opts,
 		},
 	}
 }
--- a/options_test.go
+++ b/options_test.go
@@ -146,7 +146,7 @@ func TestWithStartupCommand(t *testing.T) {
 		Started: true,
 	}
 
-	testExec := testcontainers.NewRawCommand([]string{"touch", "/tmp/.testcontainers"})
+	testExec := testcontainers.NewRawCommand([]string{"touch", ".testcontainers"}, exec.WithWorkingDir("/tmp"))
 
 	err := testcontainers.WithStartupCommand(testExec)(&req)
 	require.NoError(t, err)


options.go
```diff
@@ -311,11 +311,11 @@ type RawCommand struct {
 	cmds []string
 }
 
-func NewRawCommand(cmds []string) RawCommand {
+func NewRawCommand(cmds []string, opts ...tcexec.ProcessOption) RawCommand {
```
mdelapenya, line 314: bug: this is a breaking change as users of this exported type would receive compilation errors. Have you considered adding a functional option, or adding a new function?

benja-M-1: If there was a compilation error wouldn't the linter fail and the test not start?
I think it is fine because it is a spreaded arguments, making it optional. If nothing is provided then `opts` is an empty slice of `tcexec.ProcessOption`.

benja-M-1: Also, both cases are covered by a test in `options_test.go` line 149 and 178.

benja-M-1: Nonetheless, happy to consider other ways to achieve that!

benja-M-1: @mdelapenya as I was currious about that edge case I tried a reproduction case in that [playground](https://go.dev/play/p/kuGQI7isstS). I didn't manage to make it fail, did I miss something?

mdelapenya: It would happen in code using a previous release, that is assigning the _function_ (not the result) to a variable. https://go.dev/play/p/5LsWTXgv_iU

Imagine this code in your project, using a previous release:
```go
fn := testcontainers.NewRawCommand
...
fn(cmds)
```

After this change, your `fn` variable would have changed its signature, so this would fail:

```go
func TestRawCommandExecution(t *testing.T) {
    // This type definition would break because it expects the old signature
    type commandFactory func([]string) testcontainers.RawCommand

    tests := []struct {
        name    string
        factory commandFactory
        cmd     []string
        want    testcontainers.RawCommand
    }{
        {
            name:    "create touch command",
            // This assignment would break because NewRawCommand's signature changed to include variadic opts
            factory: testcontainers.NewRawCommand, // This line would fail to compile
            cmd:     []string{"touch", "/tmp/test"},
            want: testcontainers.RawCommand{
                ExecOptions: testcontainers.ExecOptions{
                    opts: []tcexec.ProcessOption{},
                },
                cmds: []string{"touch", "/tmp/test"},
            },
        },
    }

    for _, tt := range tests {
        t.Run(tt.name, func(t *testing.T) {
            got := tt.factory(tt.cmd)
            if !reflect.DeepEqual(got, tt.want) {
                t.Errorf("factory() = %v, want %v", got, tt.want)
            }
        })
    }
}
```

Yeah, a very edge case, but still a breaking change

mdelapenya (approved):
As we discussed, this is a very light BC, so I'm changing the title to `feat!`.

Other than that, LGTM, thanks for your time here!

PR #3174, opened 2025-05-26
chore!: use testcontainers.Run function everywhere

mdelapenya:
- **chore: deprecate GenericContainer**
- **chore: migrate tests to Run**
- **docs: update quickstart**
- **chore: update more tests**
- **chore: update more tests**
- **chore(aerospike): use Run function**
- **chore(socat): use Run function**
- **chore: update more tests**
- **chore: update tests in the wait package**
- **chore: use Run in the network package**
- **chore: move test to the network package**
- **chore: use Run in the core**
- **chore(nginx): use Run in nginx example**
- **chore(modulegen): use testcontainers.Run**
- **chore(azurite): use testcontainers.Run**
- **break(artemis): use testcontainers.Run**
- **break(eventhubs): use testcontainers.Run**
- **break(arangodb): use testcontainers.Run**
- **break(servicebus): use testcontainers.Run**
- **chore(cassandra): use testcontainers.Run**
- **chore(chroma): use testcontainers.Run**
- **break(clickhouse): use testcontainers.Run**
- **chore(consul): use testcontainers.Run**
- **chore(couchbase): use testcontainers.Run**
- **break(databend): use testcontainers.Run**
- **chore(dind): use testcontainers.Run**
- **break(dolt): use testcontainers.Run**
- **break(dynamodb): use testcontainers.Run**
- **chore(etcd): use testcontainers.Run**
- **break(etcd): make options return errors**
- **chore(gcloud): use testcontainers.Run in deprecated constructors**
- **chore(bigquery): use testcontainers.Run**
- **chore(bigtable): use testcontainers.Run**
- **chore(datastore): use testcontainers.Run**
- **chore(firestore): use testcontainers.Run**
- **chore(pubsub): use testcontainers.Run**
- **chore(spanner): use testcontainers.Run**
- **chore(graafanalgtm): use testcontainers.Run**
- **chore(inbucket): use testcontainers.Run**
- **chore(influxdb): use testcontainers.Run**
- **chore(k3s): use testcontainers.Run**
- **chore(k6): use testcontainers.Run**
- **break(kafka): use testcontainers.Run**
- **chore(localstack): use testcontainers.Run**
- **break(mariadb): use testcontainers.Run**
- **break(meilisearch): use testcontainers.Run**
- **chore(memcached): use testcontainers.Run**
- **chore(milvus): use testcontainers.Run**
- **break(minio): use testcontainers.Run**
- **chore(mockserver): use testcontainers.Run**
- **break(mongodb): use testcontainers.Run**
- **docs(dockermodelrunner): fix comment**
- **break(mssql): use testcontainers.Run**
- **docs(mssql): describe MSSQL issue with negative certificates**
- **chore(mssql): bump images to address the negative serial number of the certificate**
- **break(mysql): use testcontainers.Run**
- **break(nats): use testcontainers.Run**
- **chore(neo4j): use testcontainers.Run**
- **chore(openfga): use testcontainers.Run**
- **break(openldap): use testcontainers.Run**
- **chore(pinecone): use testcontainers.Run**
- **break(postgres): use testcontainers.Run**
- **chore(pulsar): use testcontainers.Run**
- **chore(qdrant): use testcontainers.Run**
- **break(rabbitmq): use testcontainers.Run**
- **chore(rabbitmq): consistency in credentials as env vars**
- **chore(redis): use testcontainers.Run**
- **break(redpanda): use testcontainers.Run**
- **chore(registry): use testcontainers.Run**
- **chore(scylladb): use testcontainers.Run**
- **fixup socat**
- **chore(surrealdb): use testcontainers.Run**
- **fixup socat**
- **chore(toxiproxy): use testcontainers.Run**
- **chore(valkey): use testcontainers.Run**
- **chore(vault): use testcontainers.Run**
- **chore(vearch): use testcontainers.Run**
- **chore(weaviate): use testcontainers.Run**
- **break(yugabyte): use testcontainers.Run**
- **chore(opensearch): use testcontainers.Run**

## What does this PR do?
It applies the recently added `testcontainers.Run` function in the core and in all the modules, deprecating the usage of the `GenericContainer` exposed function.

For that, we needed to review all the modules, and:

1. define an slice of `moduleOptions` representing the values in the original container request
2. append all the passed-in options from the module's Run function to the `moduleOpts` slice
3. check if the module consumed values from the container request to fill any state in the container: e.g. reading env vars from the container request's Env field.
  - if yes: look for an options.go file in the module with all the options, verifying there is not `options` struct, and a custom `Option` type representing the module's own customisation types. If that struct and the Option type existed, we added a map[string]string for the env vars, changing all the existing options that could set env vars into the container request, into an Option type writing the passed-in values to the new `env` field in the options. Why? Because with this new approach, the module has no access to the container request within the scope of the Run method, so we need to change the functional opts and read those env vars from the options instead. The module's Run function now has to loop through the passed-in options and apply its own options.
  - if no, proceed
4. remove the code to process the core options (e.g. `testcontainers.WithEnv`), as they are processed internally in the `testcontainers.Run` function.
5. if the module needed a change to read from the internal options, append a new option to the `moduleOptions`.

The result is that the `testcontainers.Run` function receives a slice of core options.

**This is considered a breaking change**, as for those modules where we changed the existing options to the new custom type, the function signature changed. In practice, users that simply passed the options to the Run function has nothing to change, but we acknowledge that there could be users that are assigning these options to a variable, and here they will receive the breaking change. This could happen mostly in tests and test tables.

Apart from this, we noticed that several modules had their internal `Option` type not returning errors, which was not consistent for our pattern to validate the options. Given this PR is already a BC for many modules, we took the chance to update those modules' options to return error, causing **another breaking change**.

The first commit in this PR is deprecating the `GenericContainer` function, so consumer would need to update to the `Run` function whenever possible. Their lint process should warn them at build/CI time.

### Modules that receive a Breaking Change

- ArangoDB
- Artemis
- Azure: eventhubs, servicebus
- Clickhouse
- Databend
- Dolt
- DynamoDB
- Etcd
- Kafka
- MariaDB
- Meilisearch
- Minio
- MongoDB
- MSSQL
- MySQL
- NATS
- OpenLDAP
- Postgres
- RabbitMQ
- Redpanda
- Yugabyte

## Why is it important?
`testcontainers.GenericContainer` has been an important function in the library since the beginning, but we consider that having `testcontainers.Run` is much more idiomatic to the operations we want to do, which is no other than running a container. Also, it pairs with the `docker run` command.

With the addition of functional options for all the container request fields, we consider it's now possible to even deprecate the struct, move it to an internal package in the library, and always customise containers through the functional options APIs.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -53,7 +53,7 @@ jobs:
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@28deaeda66b76a05916b6923827895f2b14ab387 # v3.28.16
+      uses: github/codeql-action/init@181d5eefc20863364f96762470ba6f862bdef56b # v3.29.2
       with:
         languages: ${{ matrix.language }}
         # If you wish to specify custom queries, you can do so here or in a config file.
@@ -67,7 +67,7 @@ jobs:
     # Autobuild attempts to build any compiled languages (C/C++, C#, Go, Java, or Swift).
     # If this step fails, then you should remove it and run the build manually (see below)
     - name: Autobuild
-      uses: github/codeql-action/autobuild@28deaeda66b76a05916b6923827895f2b14ab387 # v3.28.16
+      uses: github/codeql-action/autobuild@181d5eefc20863364f96762470ba6f862bdef56b # v3.29.2
 
     # ℹ️ Command-line programs to run using the OS shell.
     # 📚 See https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsrun
@@ -80,6 +80,6 @@ jobs:
     #     ./location_of_script_within_repo/buildscript.sh
 
     - name: Perform CodeQL Analysis
-      uses: github/codeql-action/analyze@28deaeda66b76a05916b6923827895f2b14ab387 # v3.28.16
+      uses: github/codeql-action/analyze@181d5eefc20863364f96762470ba6f862bdef56b # v3.29.2
       with:
         category: "/language:${{matrix.language}}"
--- a/.github/workflows/scorecards.yml
+++ b/.github/workflows/scorecards.yml
@@ -51,6 +51,6 @@ jobs:
 
       # required for Code scanning alerts
       - name: "Upload SARIF results to code scanning"
-        uses: github/codeql-action/upload-sarif@28deaeda66b76a05916b6923827895f2b14ab387 # v3.28.16
+        uses: github/codeql-action/upload-sarif@181d5eefc20863364f96762470ba6f862bdef56b # v3.29.2
         with:
           sarif_file: results.sarif
--- a/Pipfile
+++ b/Pipfile
@@ -8,7 +8,7 @@ verify_ssl = true
 [packages]
 mkdocs = "==1.5.3"
 mkdocs-codeinclude-plugin = "==0.2.1"
-mkdocs-include-markdown-plugin = "==7.1.5"
+mkdocs-include-markdown-plugin = "==7.1.6"
 mkdocs-material = "==9.5.18"
 mkdocs-markdownextradata-plugin = "==0.2.6"
 
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "4a9599a9c2db79998493adaaa691752f995ed409e15840df5aae155c83963d51"
+            "sha256": "160280c0c2d9beaa296833b5f5e754123020e5a15854a5ee6201df9cf5761554"
         },
         "pipfile-spec": 6,
         "requires": {
@@ -26,11 +26,11 @@
         },
         "bracex": {
             "hashes": [
-                "sha256:12c50952415bfa773d2d9ccb8e79651b8cdb1f31a42f6091b804f6ba2b4a66b6",
-                "sha256:13e5732fec27828d6af308628285ad358047cec36801598368cb28bc631dbaf6"
+                "sha256:0b0049264e7340b3ec782b5cb99beb325f36c3782a32e36e876452fd49a09952",
+                "sha256:98f1347cd77e22ee8d967a30ad4e310b233f7754dbf31ff3fceb76145ba47dc7"
             ],
-            "markers": "python_version >= '3.8'",
-            "version": "==2.5.post1"
+            "markers": "python_version >= '3.9'",
+            "version": "==2.6"
         },
         "certifi": {
             "hashes": [
@@ -139,11 +139,11 @@
         },
         "click": {
             "hashes": [
-                "sha256:63c132bbbed01578a06712a2d1f497bb62d9c1c0d329b7903a866228027263b2",
-                "sha256:ed53c9d8990d83c2a27deae68e4ee337473f6330c040a31d4225c9574d16096a"
+                "sha256:27c491cc05d968d271d5a1db13e3b5a184636d9d930f148c50b038f0d0646202",
+                "sha256:61a3265b914e850b85317d0b3109c7f8cd35a670f963866005d6ef1d5175a12b"
             ],
-            "markers": "python_version >= '3.7'",
-            "version": "==8.1.8"
+            "markers": "python_version >= '3.10'",
+            "version": "==8.2.1"
         },
         "colorama": {
             "hashes": [
@@ -186,11 +186,11 @@
         },
         "markdown": {
             "hashes": [
-                "sha256:794a929b79c5af141ef5ab0f2f642d0f7b1872981250230e72682346f7cc90dc",
-                "sha256:7df81e63f0df5c4b24b7d156eb81e4690595239b7d70937d0409f1b0de319c6f"
+                "sha256:247b9a70dd12e27f67431ce62523e675b866d254f900c4fe75ce3dda62237c45",
+                "sha256:5c83764dbd4e00bdd94d85a19b8d55ccca20fe35b2e678a1422b380324dd5f24"
             ],
             "markers": "python_version >= '3.9'",
-            "version": "==3.8"
+            "version": "==3.8.2"
         },
         "markupsafe": {
             "hashes": [
@@ -287,12 +287,12 @@
         },
         "mkdocs-include-markdown-plugin": {
             "hashes": [
-                "sha256:a986967594da6789226798e3c41c70bc17130fadb92b4313f42bd3defdac0adc",
-                "sha256:d0b96edee45e7fda5eb189e63331cfaf1bf1fbdbebbd08371f1daa77045d3ae9"
+                "sha256:7975a593514887c18ecb68e11e35c074c5499cfa3e51b18cd16323862e1f7345",
+                "sha256:a0753cb82704c10a287f1e789fc9848f82b6beb8749814b24b03dd9f67816677"
             ],
             "index": "pypi",
             "markers": "python_version >= '3.9'",
-            "version": "==7.1.5"
+            "version": "==7.1.6"
         },
         "mkdocs-markdownextradata-plugin": {
             "hashes": [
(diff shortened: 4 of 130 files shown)


mdelapenya:
@stevenh I think this is ready to review. I know it's a large one, but it applies consistently all the changes commit by commit.

I decided to send it in one single PR instead of multiple ones because I'm deprecating the `GenericContainer` function, but thinking out loud, I could split this PR in multiple batches, with 5-10 modules each, so that the last one includes the deprecation 🤔  Whatever works for you I'm happy to do so.

internal/core/labels.go
```diff
@@ -33,6 +34,50 @@ const (
...
+
+// Labels returns a copy of the current labels
+func (m *labelMerger) Labels() map[string]string {
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+
+	labels := make(map[string]string, len(m.labels))
+	for k, v := range m.labels {
+		labels[k] = v
+	}
```
stevenh, line 76: suggestion: use [maps.Copy](https://pkg.go.dev/maps#Copy)

internal/core/labels.go
```diff
@@ -33,6 +34,50 @@ const (
 	LabelReap = LabelBase + ".reap"
 )
 
+// labelMerger provides thread-safe operations for merging labels
+type labelMerger struct {
+	mu     sync.RWMutex
```
stevenh, line 39: question: why does this need to be thread safe?

modules/arangodb/options.go
```diff
@@ -2,10 +2,34 @@ package arangodb
 
 import "github.com/testcontainers/testcontainers-go"
 
+type options struct {
+	env map[string]string
+}
+
+func defaultOptions() options {
```
stevenh, line 9: question: could this just be WithEnv?

modules/arangodb/arangodb.go
```diff
@@ -43,47 +43,46 @@ func (c *Container) HTTPEndpoint(ctx context.Context) (string, error) {
...
+
+	// module options take precedence over default options
+	moduleOpts = append(moduleOpts, testcontainers.WithEnv(defaultOptions.env))
 
-	container, err := testcontainers.GenericContainer(ctx, genericContainerReq)
+	container, err := testcontainers.Run(ctx, img, moduleOpts...)
 	var c *Container
 	if container != nil {
-		c = &Container{Container: container, password: req.Env["ARANGO_ROOT_PASSWORD"]}
+		c = &Container{Container: container, password: defaultOptions.env["ARANGO_ROOT_PASSWORD"]}
```
stevenh, line 81: bug: this will always return the default value not the the overridden one, if the user customised it.

modules/arangodb/options.go
```diff
@@ -2,10 +2,34 @@ package arangodb
...
+func (o Option) Customize(*testcontainers.GenericContainerRequest) error {
+	// NOOP to satisfy interface.
+	return nil
+}
+
 // WithRootPassword sets the password for the ArangoDB root user
-func WithRootPassword(password string) testcontainers.CustomizeRequestOption {
-	return func(req *testcontainers.GenericContainerRequest) error {
-		req.Env["ARANGO_ROOT_PASSWORD"] = password
+func WithRootPassword(password string) Option {
```
stevenh, line 30: suggestion: delegate to WithEnv

modules/artemis/artemis.go
```diff
@@ -86,39 +55,36 @@ func RunContainer(ctx context.Context, opts ...testcontainers.ContainerCustomize
...
-	container, err := testcontainers.GenericContainer(ctx, req)
+	// module options take precedence over default options
+	moduleOpts = append(moduleOpts, testcontainers.WithEnv(defaultOptions.env))
+
+	ctr, err := testcontainers.Run(ctx, img, moduleOpts...)
 	var c *Container
-	if container != nil {
-		c = &Container{Container: container}
+	if ctr != nil {
+		c = &Container{Container: ctr, user: defaultOptions.env["ARTEMIS_USER"], password: defaultOptions.env["ARTEMIS_PASSWORD"]}
```
stevenh, line 83: bug: doesn't get the updated values.

modules/artemis/options.go
```diff
@@ -0,0 +1,59 @@
...
+
+// Customize is a NOOP. It's defined to satisfy the testcontainers.ContainerCustomizer interface.
+func (o Option) Customize(*testcontainers.GenericContainerRequest) error {
+	// NOOP to satisfy interface.
+	return nil
+}
+
+// WithCredentials sets the administrator credentials. The default is artemis:artemis.
+func WithCredentials(user, password string) Option {
+	return func(o *options) error {
```
stevenh, line 32: suggestion: delegate to WithEnv, more below

modules/azure/eventhubs/options.go
```diff
@@ -44,9 +47,9 @@ func WithAzurite(img string, opts ...testcontainers.ContainerCustomizer) Option
 }
 
 // WithAcceptEULA sets the ACCEPT_EULA environment variable to "Y" for the eventhubs container.
-func WithAcceptEULA() testcontainers.CustomizeRequestOption {
-	return func(req *testcontainers.GenericContainerRequest) error {
-		req.Env["ACCEPT_EULA"] = "Y"
+func WithAcceptEULA() Option {
+	return func(o *options) error {
```
stevenh, line 50: suggestion: delegate to to WithEnv

modules/azure/eventhubs/options.go
```diff
@@ -55,9 +58,9 @@ func WithAcceptEULA() testcontainers.CustomizeRequestOption {
 // WithConfig sets the eventhubs config file for the eventhubs container,
 // copying the content of the reader to the container file at
 // "/Eventhubs_Emulator/ConfigFiles/Config.json".
-func WithConfig(r io.Reader) testcontainers.CustomizeRequestOption {
-	return func(req *testcontainers.GenericContainerRequest) error {
-		req.Files = append(req.Files, testcontainers.ContainerFile{
+func WithConfig(r io.Reader) Option {
```
stevenh: question: could you clarify why this change is needed?

stevenh (changes requested):
Partial review, as I might be missing something with the switch to custom options vs Customisers? So thought it best to stop and get the answer before I continnue.

modules/arangodb/arangodb.go
mdelapenya: I've debugged this and the opts.env receives the passed-in password from the option. All the modules that I'm adding an env field to the opts struct need access to one or more of those key/values to complete the state of the container (e.g. a password, a user), and with the WithEnv approach we simply has no access to the underlying map of env-vars. That's why I'm defining this intermediate map in the module opts struct, which is later contributed to the container opts using WithEnv. Does it make sense?

modules/azure/eventhubs/options.go
mdelapenya: Mmm I think it was probably because of inertia... I double checked, and it can be kept as the core option.

Good catch

modules/arangodb/arangodb.go
stevenh: Right now I get it.

The problem you have is you're changing the order of the options which can change the result, for example if you have a password option, followed by a WithEnv option which changes the value. With the current code the password option will get moved after the WithEnv and the resulting configuration would be incorrect. It's not going to be common case, but still breaking behaviour.

Instead we could make it easy to access the created container environment. It's currently possible via Inspect so a helper such as:
```go
// Env gets the environment variables for the container.
func (c *DockerContainer) Env(ctx context.Context) (map[string]string, error) {
	inspect, err := c.Inspect(ctx)
	if err != nil {
		return nil, fmt.Errorf("inspect: %w", err)
	}

	env := make(map[string]string, len(inspect.Config.Env))
	for _, e := range inspect.Config.Env {
		// Split the environment variable into key and value
		parts := strings.SplitN(e, "=", 2)
		if len(parts) == 2 {
			// Add the key-value pair to the map
			env[parts[0]] = parts[1]
		} else {
			// If no value is provided, set the value to an empty string
			env[parts[0]] = ""
		}
	}

	return env, nil
}
```

Alternatively we could store the value from create request and expose by a simple method. The special case however would be a loaded container. We could still handle that.
```go
// Env gets the environment variables for the container.
func (c *DockerContainer) Env() map[string]string {
	return maps.Clone(c.env)
}
```

Thoughts?

mdelapenya:
I'm converting this PR into draft as reference to sent smaller chunks and make progress with it.

I'm listing all the smaller chunk PRs below:

- [x] https://github.com/testcontainers/testcontainers-go/pull/3304
- [x] https://github.com/testcontainers/testcontainers-go/pull/3305
- [x] https://github.com/testcontainers/testcontainers-go/pull/3307
- [x] https://github.com/testcontainers/testcontainers-go/pull/3311
- [x] https://github.com/testcontainers/testcontainers-go/pull/3312
- [x] https://github.com/testcontainers/testcontainers-go/pull/3318

mdelapenya:
Closing as complete

PR #3176, opened 2025-05-30
feat: make port mapping checks timeout configurable

mdelapenya:
- **feat: make exposing ports configurable by config**
- **chore: apply configuration to the port mapping check**
- **chore: refactor checkPortsMapped to make it testable**

## What does this PR do?
This PR adds a new setting to the properties file: `tc.port.mapping.timeout`, with a default value of 5s.

Users can configure it via properties, or with the `TESTCONTAINERS_PORT_MAPPING_TIMEOUT` env var.

This new property is used in the check for exposed ports.

## Why is it important?
Before these changes, the library always used 5s to check for all the exposed ports, and depending on the underlying hardware it could lead to flakiness.

Now, users can set the env var, or the property in the ~/.testcontainers.properties file to change that timeout.
 

## Related issues

- Related to #3165

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -85,6 +85,11 @@ type Config struct {
 	//
 	// Environment variable: TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE
 	TestcontainersHost string `properties:"tc.host,default="`
+
+	// TestcontainersPortMappingTimeout is the time to wait before all the exposed ports are mapped.
+	//
+	// Environment variable: TESTCONTAINERS_PORT_MAPPING_TIMEOUT
+	TestcontainersPortMappingTimeout time.Duration `properties:"tc.port.mapping.timeout,default=5s"`
 }
 
 // }
@@ -139,6 +144,11 @@ func read() Config {
 		ryukConnectionTimeoutEnv := readTestcontainersEnv("RYUK_CONNECTION_TIMEOUT")
 		if timeout, err := time.ParseDuration(ryukConnectionTimeoutEnv); err == nil {
 			config.RyukConnectionTimeout = timeout
+		}
+
+		testcontainersPortMappingTimeoutEnv := readTestcontainersEnv("TESTCONTAINERS_PORT_MAPPING_TIMEOUT")
+		if timeout, err := time.ParseDuration(testcontainersPortMappingTimeoutEnv); err == nil {
+			config.TestcontainersPortMappingTimeout = timeout
 		}
 
 		return config
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -26,6 +26,7 @@ func resetTestEnv(t *testing.T) {
 	t.Setenv("RYUK_VERBOSE", "")
 	t.Setenv("RYUK_RECONNECTION_TIMEOUT", "")
 	t.Setenv("RYUK_CONNECTION_TIMEOUT", "")
+	t.Setenv("TESTCONTAINERS_PORT_MAPPING_TIMEOUT", "")
 }
 
 func TestReadConfig(t *testing.T) {
@@ -79,16 +80,17 @@ func TestReadTCConfig(t *testing.T) {
 		t.Setenv("TESTCONTAINERS_RYUK_CONTAINER_PRIVILEGED", "true")
 		t.Setenv("RYUK_RECONNECTION_TIMEOUT", "13s")
 		t.Setenv("RYUK_CONNECTION_TIMEOUT", "12s")
-
+		t.Setenv("TESTCONTAINERS_PORT_MAPPING_TIMEOUT", "11s")
 		config := read()
 
 		expected := Config{
-			HubImageNamePrefix:      defaultHubPrefix,
-			RyukDisabled:            true,
-			RyukPrivileged:          true,
-			Host:                    "", // docker socket is empty at the properties file
-			RyukReconnectionTimeout: 13 * time.Second,
-			RyukConnectionTimeout:   12 * time.Second,
+			HubImageNamePrefix:               defaultHubPrefix,
+			RyukDisabled:                     true,
+			RyukPrivileged:                   true,
+			Host:                             "", // docker socket is empty at the properties file
+			RyukReconnectionTimeout:          13 * time.Second,
+			RyukConnectionTimeout:            12 * time.Second,
+			TestcontainersPortMappingTimeout: 11 * time.Second,
 		}
 
 		assert.Equal(t, expected, config)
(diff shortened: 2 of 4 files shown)


mabrarov:
Hi @mdelapenya,

> Need to investigate more in depth, as all tests pass locally, but on CI, they hang for ever (30m, GitHub workers default timeout). It only happens for the socat module, rootless Docker, and Ryuk disabled, so I guess there is something related to the ports that is causing the "deadlock"??

If you talk about https://github.com/testcontainers/testcontainers-go/actions/runs/15347196816/job/43186220491?pr=3176:
```text
     reuse_test.go:39: 
        	Error Trace:	/home/runner/work/testcontainers-go/testcontainers-go/reuse_test.go:39
        	Error:      	Received unexpected error:
        	            	create container: all exposed ports, [8080/tcp], were not mapped in 5s: port 8080/tcp is not mapped yet
        	Test:       	TestGenericContainer_stop_start_withReuse
2025/05/30 13:09:28 🐳 Stopping container: fcb600a0f955
2025/05/30 13:09:28 ✅ Container stopped: fcb600a0f955
2025/05/30 13:09:28 🐳 Terminating container: fcb600a0f955
2025/05/30 13:09:28 🚫 Container terminated: fcb600a0f955

=== FAIL: . TestGenericContainer_stop_start_withReuse (re-run 5) (14.98s)
```
then please note that this PR is not going to fix this issue, because it is not related to timeout value (which is made configurable in this PR). Refer to the 2nd note in https://github.com/testcontainers/testcontainers-go/pull/3165#issuecomment-2913422933 for the root cause of this test failure.

Thank you.

mdelapenya:
@mabrarov the CI builds I mentioned: 

- socat build: hangs https://github.com/testcontainers/testcontainers-go/actions/runs/15339849072/job/43173511093
- rootless build: hangs https://github.com/testcontainers/testcontainers-go/actions/runs/15339849072/job/43173511077
- reaper disabled: hangs https://github.com/testcontainers/testcontainers-go/actions/runs/15339849072/job/43173511093

mabrarov:
Hi @stevenh and @mdelapenya,

> Not new, but this check seems to duplicate what wait already handles, can we just remove this check totally allow wait definitions to handle it. We could have containers waiting for ports which the user doesn't care about making things slower than they might be?

I like this idea, but please note that not all parts (mostly modules) of Testcontainers for Go can be ready for this, because some existing code can expect that port mapping completes immediately after container start. If we remove this readiness hook, then we risk to break such code.

Here is a quick & dirty research for such code (code which inspects mapped ports after container started without making multiple tries): https://github.com/testcontainers/testcontainers-go/issues/2670#issuecomment-2884488508.

Thank you.

mdelapenya:
> The failed tests in CI behave same way. The only difference is larger timeout (if I get it right).

I think the larger timeout is caused because the CI does not have a testcontainers.properties and our config code fails to define the new timeout (we need to fix it, as mentioned in the follow-up section of the PR), as a result, there is no timeout, and the retry hangs forever. This is what I observe from those builds.

In any case, I'm more interested in removing the code, and enforce defining best practices with the wait strategies. A validation warning, maybe?

stevenh:
> > Hi @mdelapenya,
> > > If we finally remove this check, we must make sure we inform client code that they must create a robust wait strategy
> > 
> > 
> > Yes, and the users may need `wait.ForMappedPort` from #3176 to do that 😄 (note that `wait.ForMappedPort` waits for a single port while `defaultReadinessHook` waits for multiples ports, so `wait.ForMappedPort` is not one-to-one replacement for `defaultReadinessHook`).
> 
> Mmm why not adding a new wait strategy for all the mapped ports, using your code as buildng block? It would satisfy the readinessCheck, and could be optional or even required (added as a wait strategy in the `testcontainers.Run` function).

On the face that sounds like a good idea, but might not be the best solution, why? Containers could be exposing lots of ports, and the user might not care about all of them. As we know expose is a specific edge case for them to be usable is something different so I would suggest that being explicit about the ones you care about would result in better behaviour.

Sure the wait for all might be convenient but result in lazy application and slower start up times. Thooughts?

mabrarov:
Hi @mdelapenya,

> why not adding a new wait strategy for all the mapped ports, using your code as buildng block? It would satisfy the readinessCheck, and could be optional or even required (added as a wait strategy in the `testcontainers.Run` function).

Do you plan to revisit this PR and instead of making `defaultReadinessHook` timeout configurable:

1. Plan to remove `defaultReadinessHook`
1. Plan to add new wait strategy (where user can choose either wait for mapping of all exposed ports or wait for mapping of explicitly provided ports) and recommend it in documentation

?

It is not clear for me what we are going to do with https://github.com/testcontainers/testcontainers-go/pull/3165 in that case (what will be the order of PRs, what is the future of https://github.com/testcontainers/testcontainers-go/pull/3165 in general).

Thank you.

stevenh:
https://github.com/testcontainers/testcontainers-go/pull/3165 should be merged when ready, that could be good building block for containers that need to know the port details for container configuration, that's an edge case in my mind, but still solid core change.

The more standard workflow is waiting for a listening port.

It might be useful to have a "for all exposed" option for port wait strategies, but I think we should explore the impact of that before adding.

stevenh:
> > Sure the wait for all might be convenient but result in lazy application and slower start up times. Thooughts?
> 
> We have two options:
> 
> 1. create an option for waiting for all the mapped ports, document it, and let users decide, not applying to by default, as I suggested above. It's basically what we are doing in the readiness check atm (not consistent with the wait strategies approach). We must explicitly mention it in the release notes, that the behaviour has changed (means a breaking change)
> 2. keep the code as is, making it configurable (proper naming included)

I like option 1 as it removes the duplication and simplifies the configration for all cases. It might be a breaking change, but should only be for modules which have aren't correctly waiting for their required ports, so in the long run will result in more stability.

One could argue its not a breaking change, it's just one which will expose bugs in other areas, so not something to shy away from IMO.

mabrarov:
Hi,

If my opinion is needed / helpful, then I vote for option 1 from https://github.com/testcontainers/testcontainers-go/pull/3176#issuecomment-2922822104. I was surprised by `defaultReadinessHook`. It would be nice to check if Testcontainers for Java has such feature (if yes, then, maybe maintainers of Testcontainers for Java can speak about this feature or want to remove / replace it too).

Thank you.

mabrarov:
FYI @mdelapenya,

I opened https://github.com/testcontainers/testcontainers-go/issues/3177 for the issue described in https://github.com/testcontainers/testcontainers-go/pull/3176#issuecomment-2922755534 (in https://github.com/testcontainers/testcontainers-go/pull/3165#issuecomment-2913422933).

Thank you.

mdelapenya:
Closing, we are going to remove the internal check for the exposed ports:
1. the library offers a robust set of wait strategies that allows user configure the process of waiting for a good state
2. given 1) we cannot over protect our users with this check. If we recommend defining a proper wait strategy, it means exactly that, and waiting for ports must be defined by the user. We are adding a new `wait.ForMappedPort` strategy (see #3165)

For that, )'m closing this one and will send a new one removing the check.

PR #3201, opened 2025-06-05
fix: fix memory leak in reaper connect function

### LINKED ISSUE #2878: [Bug]: leak in reaper connect function with terminationSignal
### Testcontainers version

0.34.0

Yes

### Host OS

linux

### Host arch

amd64

### Go version

1.23.2

### Docker version

 Version:           26.1.4
 API version:       1.45
 Go version:        go1.21.11
 Git commit:        5650f9b
 Built:             Wed Jun  5 11:29:19 2024
 Context:           default

 Engine:
  Version:          26.1.4
  API version:      1.45 (minimum version 1.24)
  Go version:       go1.21.11
  Git commit:       de5c9cf
  Built:            Wed Jun  5 11:29:19 2024
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.33
  GitCommit:        d2d58213f83a351ca8f528a95fbd145f5654e957
 runc:
  Version:          1.1.12
  GitCommit:        v1.1.12-0-g51d5e94
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0

### Docker info

 Version:    26.1.4
 Context:    default
 Debug Mode: false
 Plugins:
    Version:  v0.14.1
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
WARNING: No swap limit support
    Version:  v2.27.1
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 5
 Server Version: 26.1.4
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
 Swarm: inactive
 Runtimes: runc io.containerd.runc.v2
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: d2d58213f83a351ca8f528a95fbd145f5654e957
 runc version: v1.1.12-0-g51d5e94
 init version: de40ad0
 Security Options:
  apparmor
 seccomp
   Profile: builtin
 Kernel Version: 5.4.0-186-generic
 Operating System: Ubuntu 20.04.6 LTS
 OSType: linux
 Architecture: x86_64
 CPUs: 8
 Total Memory: 15.32GiB
 Name: troy
 ID: 2WIM:7GHM:XDAV:VDAR:XRVJ:WGK7:YU4F:JHWE:J25C:CPE2:YE5U:5EMH
 Debug Mode: false
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false

### What happened?

im testing with `goleak.VerifyTestMain(m)` and with the new version of testcontainers there is a goroutine leak in reaper. it looks like the terminationSignal chan wont gets closed so the goroutine hangs. but its just a wild guess here.

### Relevant log output

```
goleak: Errors on successful test run: found unexpected goroutines:
[Goroutine 99 in state chan receive, with github.com/testcontainers/testcontainers-go.(*Reaper).connect.func1 on top of the stack:
github.com/testcontainers/testcontainers-go.(*Reaper).connect.func1()
        /home/marv/wip/hemingway/vendor/github.com/testcontainers/testcontainers-go/reaper.go:527 +0x1d3
created by github.com/testcontainers/testcontainers-go.(*Reaper).connect in goroutine 22
        /home/marv/wip/hemingway/vendor/github.com/testcontainers/testcontainers-go/reaper.go:522 +0x311
 Goroutine 195 in state chan receive, with github.com/testcontainers/testcontainers-go.(*Reaper).connect.func1 on top of the stack:
github.com/testcontainers/testcontainers-go.(*Reaper).connect.func1()
        /home/marv/wip/hemingway/vendor/github.com/testcontainers/testcontainers-go/reaper.go:527 +0x1d3
created by github.com/testcontainers/testcontainers-go.(*Reaper).connect in goroutine 141
        /home/marv/wip/hemingway/vendor/github.com/testcontainers/testcontainers-go/reaper.go:522 +0x311
 Goroutine 206 in state chan receive, with github.com/testcontainers/testcontainers-go.(*Reaper).connect.func1 on top of the stack:
github.com/testcontainers/testcontainers-go.(*Reaper).connect.func1()
        /home/marv/wip/hemingway/vendor/github.com/testcontainers/testcontainers-go/reaper.go:527 +0x1d3
created by github.com/testcontainers/testcontainers-go.(*Reaper).connect in goroutine 166
        /home/marv/wip/hemingway/vendor/github.com/testcontainers/testcontainers-go/reaper.go:522 +0x311
]
```

### Additional information

_No response_

ionicc:
## What does this PR do?

Adds a fix to pass termination signal to reaper in the context of an error. The previous implementation had a memory leak issue due to the usage of a redundant check on an error which was returned before reaching the statement of passing the termination signal.

## Why is it important?

This will make sure a memory leak does not happen if reaper.connect returns an error and properly sends a termination signal.

## Related issues

- Closes #2878

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -1335,15 +1335,14 @@ func (p *DockerProvider) ReuseOrCreateContainer(ctx context.Context, req Contain
 
 		termSignal, err := r.Connect()
 		if err != nil {
+			if termSignal != nil {
+				// If a termination signal is present with an error, Pass along to cleanup.
+				// to the reaper to ensure it cleans up the container.
+				termSignal <- true
+			}
 			return nil, fmt.Errorf("reaper connect: %w", err)
 		}
 
-		// Cleanup on error.
-		defer func() {
-			if err != nil {
-				termSignal <- true
-			}
-		}()
 	}
 
 	// default hooks include logger hook and pre-create hook
@@ -1575,15 +1574,13 @@ func (p *DockerProvider) CreateNetwork(ctx context.Context, req NetworkRequest) 
 
 		termSignal, err := r.Connect()
 		if err != nil {
-			return nil, fmt.Errorf("reaper connect: %w", err)
-		}
-
-		// Cleanup on error.
-		defer func() {
-			if err != nil {
+			if termSignal != nil {
+				// If a termination signal is present with an error, Pass along to cleanup.
+				// to the reaper to ensure it cleans up the container.
 				termSignal <- true
 			}
-		}()
+			return nil, fmt.Errorf("reaper connect: %w", err)
+		}
 	}
 
 	// add the labels that the reaper will use to terminate the network to the request


docker.go
```diff
@@ -1335,15 +1335,14 @@ func (p *DockerProvider) ReuseOrCreateContainer(ctx context.Context, req Contain
 
 		termSignal, err := r.Connect()
 		if err != nil {
+			if termSignal != nil {
```
stevenh, line 1338: question: I don't see a flow where r.Connect can return an error and termSignal, it appears to be one or the other. Can you clarify where this would come from?

ionicc: @stevenh Hmm, From my observation of term signal handling in these methods, The defer func written to pass the term signal is redundant as the method returns with an error before the defer func is set. See below:
From: https://github.com/testcontainers/testcontainers-go/blob/main/docker.go#L1336-L1347
```go
termSignal, err := r.Connect()
		if err != nil {
			return nil, fmt.Errorf("reaper connect: %w", err) <-- Returns
		}

		// Cleanup on error.
		defer func() { <-- Never gets set in the case of error present
			if err != nil {
				termSignal <- true
			}
		}()
```

Due to this, There's a possibility of a memory leak in the case both, a term signal and an error is passed. This is also the conclusive point of the issue linked.

stevenh: That's not the case as err and termHandker can't both be set and err is the captured err so any error which is returned from then on will result in the term signal being handled before return.

What evidence of memory leak do you have?

stevenh: @ionicc any thoughts on the above?

ionicc: @stevenh Sorry for my late reply on this, I didn't get a chance to dig deeper into the mutually exclusive nature of r.Connect() for termSignal and err. I understand your point, the shadowing of the function signature err by original errors that are being instantiated through the func will not apply to the captured error case, Is that the intention? For ex in https://github.com/testcontainers/testcontainers-go/blob/main/docker.go#L1371, The function signature will be nil and would not be caught by the defer func.

stevenh: That's actually not shadowed, new vars are instantiated left to right only if needed, so where err is already defined a new one is not created.

In a defer statement where a named variable is in play, the value is always that of return value not what was last assigned to that variable. You can [see this behaviour here](https://go.dev/play/p/CB9MHf1WzPd).

So what the `defer` is doing is ensuring that the `termSignal` is triggered if the function encounters an error before returning.

ionicc: Sorry for the late reply @stevenh. Got it, Thank you! Then I believe we can close this PR.

stevenh: Thanks @ionicc glad we could clear that up

PR #3211, opened 2025-06-17
exec: remove uses of ExecOptions.Detach

thaJeztah:
This option is not actually used by the client, and not handled by the API, so removing its use.

## What does this PR do?

The `ExecOptions.Detach` field was a leftover from an old implementation of the CLI (see https://github.com/moby/moby/pull/50218), and is not used by the API client, nor is it part of the API, so we can skip using this field.

## Why is it important?

See above

## Related issues

- relates to https://github.com/moby/moby/pull/50218

-


mdelapenya (dismissed):
LGTM, thanks!

Could you update the Ollama module to remove the uses of Detach in there too? 🙏 

Approving the PR anyway

PR #3216, opened 2025-06-20
feat: add Dex IdP module

prskr:
## What does this PR do?

Add a [Dex](https://dexidp.io) module, similar to the already existing [testcontainers-java](https://testcontainers.com/modules/dex/) module.

## Why is it important?

This allows to test OAuth2/OpenID Connect authentication implementations.

## Related issues

No issue was created for this so far.

## How to test this PR

Hopefully the implemented tests and documentation is sufficient, if not, please let me know, so I can extend them 😊

### DIFF
--- a/modules/dex/dex.go
+++ b/modules/dex/dex.go
@@ -13,7 +13,6 @@ import (
 	"strings"
 
 	dexapi "github.com/dexidp/dex/api/v2"
-	"github.com/google/uuid"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
 
@@ -49,6 +48,11 @@ type Container struct {
 // CreateClientApp registers a new client application in Dex.
 // If ClientID or ClientSecret are empty, they will be generated and returned.
 func (c Container) CreateClientApp(ctx context.Context, opts ...CreateClientAppOption) (*dexapi.Client, error) {
+	clientApp, err := combineOptionsToClientApp(opts)
+	if err != nil {
+		return nil, fmt.Errorf("combine options to client app: %w", err)
+	}
+
 	apiClient, connCloser, err := c.createDexAPIClient(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("prepare Dex API client: %w", err)
@@ -60,13 +64,7 @@ func (c Container) CreateClientApp(ctx context.Context, opts ...CreateClientAppO
 		}
 	}()
 
-	var dexClientApp dexapi.Client
-
-	for _, opt := range opts {
-		opt(&dexClientApp)
-	}
-
-	resp, err := apiClient.CreateClient(ctx, &dexapi.CreateClientReq{Client: &dexClientApp})
+	resp, err := apiClient.CreateClient(ctx, &dexapi.CreateClientReq{Client: clientApp})
 	if err != nil {
 		return nil, fmt.Errorf("register client application: %w", err)
 	}
@@ -83,6 +81,16 @@ func (c Container) CreatePassword(
 	credential CreatePasswordCredential,
 	opts ...CreatePasswordOption,
 ) (err error) {
+	email, hash, err := credential()
+	if err != nil {
+		return fmt.Errorf("create password credential: %w", err)
+	}
+
+	dexPassword, err := combineOptionsToPassword(email, hash, opts)
+	if err != nil {
+		return fmt.Errorf("combine options to password: %w", err)
+	}
+
 	apiClient, connCloser, err := c.createDexAPIClient(ctx)
 	if err != nil {
 		return fmt.Errorf("create Dex API client: %w", err)
@@ -94,22 +102,7 @@ func (c Container) CreatePassword(
 		}
 	}()
 
-	dexPassword := dexapi.Password{
-		// defaulting user ID
-		// can be overridden by using WithUserID
-		UserId: uuid.New().String(),
-	}
-
-	dexPassword.Email, dexPassword.Hash, err = credential()
-	if err != nil {
-		return fmt.Errorf("create password credential: %w", err)
-	}
-
-	for _, opt := range opts {
-		opt(&dexPassword)
-	}
-
-	if _, err = apiClient.CreatePassword(ctx, &dexapi.CreatePasswordReq{Password: &dexPassword}); err != nil {
+	if _, err = apiClient.CreatePassword(ctx, &dexapi.CreatePasswordReq{Password: dexPassword}); err != nil {
 		return fmt.Errorf("create password in Dex: %w", err)
 	}
 
@@ -121,19 +114,19 @@ func (c Container) CreatePassword(
 // OpenIDConfiguration returns the OpenID configuration for the Dex instance.
 // It retrieves the raw configuration, unmarshals it into an OpenIDConfiguration struct,
 // and returns any error that occurs during the process.
-func (c Container) OpenIDConfiguration(ctx context.Context) (OpenIDConfiguration, error) {
+func (c Container) OpenIDConfiguration(ctx context.Context) (*OpenIDConfiguration, error) {
 	rawCfg, err := c.RawOpenIDConfiguration(ctx)
 	if err != nil {
-		return OpenIDConfiguration{}, fmt.Errorf("fetching raw OpenID configuration: %w", err)
+		return nil, fmt.Errorf("fetching raw OpenID configuration: %w", err)
 	}
 
 	var cfg OpenIDConfiguration
 	err = json.Unmarshal(rawCfg, &cfg)
 	if err != nil {
-		return OpenIDConfiguration{}, fmt.Errorf("unmarshal OpenID configuration: %w", err)
+		return nil, fmt.Errorf("unmarshal OpenID configuration: %w", err)
 	}
 
-	return cfg, nil
+	return &cfg, nil
 }
 
 // RawOpenIDConfiguration retrieves the raw OpenID configuration from the Dex instance.
@@ -175,7 +168,7 @@ func (c Container) RawOpenIDConfiguration(ctx context.Context) (rawCfg []byte, e
 			continue
 		}
 
-		patched, err := patchEndpoint(strings.Trim(string(jsonValue), "\""), httpEndpoint)
+		patched, err := patchEndpoint(strings.Trim(string(jsonValue), `"`), httpEndpoint)
 		if err != nil {
 			return nil, err
 		}
--- a/modules/dex/options.go
+++ b/modules/dex/options.go
@@ -1,11 +1,16 @@
 package dex
 
 import (
+	"errors"
+
 	dexapi "github.com/dexidp/dex/api/v2"
+	"github.com/google/uuid"
 	"golang.org/x/crypto/bcrypt"
 
 	"github.com/testcontainers/testcontainers-go"
 )
+
+var errEmptyUserID = errors.New("user ID cannot be empty")
 
 // WithIssuer sets the issuer for the Dex container.
 // In most cases, the issuer should be set to the URL of the Dex container.
@@ -30,7 +35,6 @@ type CreatePasswordCredential func() (email string, passwordHash []byte, err err
 
 func PlainTextCredential(email, password string) CreatePasswordCredential {
 	hashedPassword, err := bcrypt.GenerateFromPassword([]byte(password), bcrypt.DefaultCost)
-
 	return func() (string, []byte, error) {
 		return email, hashedPassword, err
 	}
@@ -42,60 +46,127 @@ func HashedCredential(email string, passwordHash []byte) CreatePasswordCredentia
 	}
 }
 
-type CreatePasswordOption func(pw *dexapi.Password)
+func combineOptionsToPassword(email string, hash []byte, opts []CreatePasswordOption) (*dexapi.Password, error) {
+	options := createPasswordOptions{
+		UserID: uuid.New().String(),
+	}
+
+	for _, opt := range opts {
+		if err := opt(&options); err != nil {
+			return nil, err
+		}
+	}
+
+	return &dexapi.Password{
+		Email:    email,
+		Hash:     hash,
+		Username: options.Username,
+		UserId:   options.UserID,
+	}, nil
+}
+
+type createPasswordOptions struct {
+	Username string
+	UserID   string
+}
+
+type CreatePasswordOption func(pw *createPasswordOptions) error
 
 func WithUserID(userID string) CreatePasswordOption {
-	return func(pw *dexapi.Password) {
-		pw.UserId = userID
+	return func(pw *createPasswordOptions) error {
+		if userID == "" {
+			return errEmptyUserID
+		}
+
+		pw.UserID = userID
+		return nil
 	}
 }
 
 func WithUsername(username string) CreatePasswordOption {
-	return func(pw *dexapi.Password) {
+	return func(pw *createPasswordOptions) error {
 		pw.Username = username
+		return nil
 	}
 }
 
-type CreateClientAppOption func(ca *dexapi.Client)
+func combineOptionsToClientApp(opts []CreateClientAppOption) (*dexapi.Client, error) {
+	var options createClientAppOptions
+
+	for _, o := range opts {
+		if err := o(&options); err != nil {
+			return nil, err
+		}
+	}
+
+	return &dexapi.Client{
+		Id:           options.ID,
+		Secret:       options.Secret,
+		RedirectUris: options.RedirectUris,
+		TrustedPeers: options.TrustedPeers,
+		Public:       options.Public,
+		Name:         options.Name,
+		LogoUrl:      options.LogoURL,
+	}, nil
+}
+
+type createClientAppOptions struct {
+	ID           string
+	Secret       string
+	RedirectUris []string
+	TrustedPeers []string
+	Public       bool
+	Name         string
+	LogoURL      string
+}
+
+type CreateClientAppOption func(ca *createClientAppOptions) error
 
 func WithClientName(clientName string) CreateClientAppOption {
-	return func(ca *dexapi.Client) {
+	return func(ca *createClientAppOptions) error {
 		ca.Name = clientName
+		return nil
 	}
 }
 
 func WithClientID(clientID string) CreateClientAppOption {
-	return func(ca *dexapi.Client) {
-		ca.Id = clientID
+	return func(ca *createClientAppOptions) error {
+		ca.ID = clientID
+		return nil
 	}
 }
 
 func WithClientSecret(clientSecret string) CreateClientAppOption {
-	return func(ca *dexapi.Client) {
+	return func(ca *createClientAppOptions) error {
 		ca.Secret = clientSecret
+		return nil
 	}
 }
 
-func WithRedirectURIs(redirectURIs []string) CreateClientAppOption {
-	return func(ca *dexapi.Client) {
+func WithRedirectURIs(redirectURIs ...string) CreateClientAppOption {
+	return func(ca *createClientAppOptions) error {
 		ca.RedirectUris = redirectURIs
+		return nil
 	}
 }
 
 func WithPublicClient(public bool) CreateClientAppOption {
-	return func(ca *dexapi.Client) {
+	return func(ca *createClientAppOptions) error {
 		ca.Public = public
+		return nil
 	}
 }
 
-func WithTrustedPeers(trustedPeers []string) CreateClientAppOption {
-	return func(ca *dexapi.Client) {
+func WithTrustedPeers(trustedPeers ...string) CreateClientAppOption {
+	return func(ca *createClientAppOptions) error {
 		ca.TrustedPeers = trustedPeers
+		return nil
 	}
 }
 
 func WithLogoURL(logoURL string) CreateClientAppOption {
-	return func(ca *dexapi.Client) {
-		ca.LogoUrl = logoURL
+	return func(ca *createClientAppOptions) error {
+		ca.LogoURL = logoURL
+		return nil
 	}
 }


modules/dex/dex.go
```diff
@@ -0,0 +1,347 @@
...
+	testcontainers.Container
+	Client *http.Client
+}
+
+// CreateClientApp registers a new client application in Dex.
+// If ClientID or ClientSecret are empty, they will be generated and returned.
+func (c Container) CreateClientApp(ctx context.Context, req CreateClientAppRequest) (*dexapi.Client, error) {
+	apiClient, connCloser, err := c.createDexAPIClient(ctx)
+	if err != nil {
+		return nil, err
```
stevenh: suggestion: wrap the error so we know where it came from, more below and in other files.

prskr: Thanks for the reminder! 🙈
Apparently, I rushed a bit too fast through the prototype 😅

prskr: I hope I caught all returned errors, tbh. I wasn't sure about calls to TC APIs like `PortEndpoint(...)` but as they also don't always return wrapped errors, I figured, I'll wrap all for consistency and a better DX.

If you'd prefer it any other way, just let me know 😊

modules/dex/dex.go
```diff
@@ -0,0 +1,349 @@
...
+	HTTPPort = "5556/tcp"
+	GRPCPort = "5557/tcp"
+)
+
+const (
+	bcryptCost    = 10
+	defaultIssuer = "http://localhost:5556"
+)
+
+var ErrPasswordRequired = errors.New("plaintext password or hash is required")
```
stevenh: suggestion: keep this private, no need to expand the API surface area by exporting.

modules/dex/dex.go
```diff
@@ -0,0 +1,349 @@
...
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+
+	"github.com/testcontainers/testcontainers-go"
+	"github.com/testcontainers/testcontainers-go/wait"
+)
+
+const (
+	HTTPPort = "5556/tcp"
+	GRPCPort = "5557/tcp"
```
stevenh, line 25: question: do these need to be exported, seems like they could be private and reduce the API surface area.

modules/dex/dex.go
```diff
@@ -0,0 +1,349 @@
...
+
+// Container represents the Dex container type used in the module
+type Container struct {
+	testcontainers.Container
+	Client *http.Client
+}
+
+// CreateClientApp registers a new client application in Dex.
+// If ClientID or ClientSecret are empty, they will be generated and returned.
+func (c Container) CreateClientApp(ctx context.Context, req CreateClientAppRequest) (*dexapi.Client, error) {
```
stevenh: suggestion: use parameters instead of structures as that makes it clear what the user needs to pass, using a struct the caller can easily pass an empty struct with likely unexpected results.

modules/dex/dex.go
```diff
@@ -0,0 +1,349 @@
...
+	return resp.Client, nil
+}
+
+// CreatePassword registers a new 'password' (login) in Dex. Either Password or Hash must be provided.
+// If Hash is provided, Password will be ignored. If UserID is not set, a UUID will be generated.
+// The password will be hashed using bcrypt if provided. Returns ErrPasswordRequired if neither
+// Password nor Hash are provided, or an error from bcrypt if hashing fails.
+func (c Container) CreatePassword(
+	ctx context.Context,
+	req CreatePasswordRequest,
```
stevenh: suggestion: as above parameters instead of struct, also solves the password / hash as you can have a parameter which is either, but not both.

modules/dex/dex.go
```diff
@@ -0,0 +1,349 @@
...
+) (err error) {
+	apiClient, connCloser, err := c.createDexAPIClient(ctx)
+	if err != nil {
+		return fmt.Errorf("prepare Dex API client: %w", err)
```
stevenh: suggestion: match the function naming
```suggestion
		return fmt.Errorf("create Dex API client: %w", err)
```

modules/dex/dex.go
```diff
@@ -0,0 +1,349 @@
...
+
+	defer func() {
+		if closeErr := connCloser.Close(); closeErr != nil {
+			err = errors.Join(err, fmt.Errorf("close GRPC connection: %w", closeErr))
+		}
+	}()
+
+	if req.Password == "" && len(req.Hash) == 0 {
+		return ErrPasswordRequired
+	}
```
stevenh: suggestion: this can be eliminated with parameter approach.

modules/dex/dex.go
```diff
@@ -0,0 +1,349 @@
...
+
+	if req.UserID == "" {
+		req.UserID = uuid.New().String()
+	}
```
stevenh: suggestion: use functional options for optional parameters.

modules/dex/dex.go
```diff
@@ -0,0 +1,349 @@
...
+	return nil
+}
+
+// OpenIDConfiguration returns the OpenID configuration for the Dex instance.
+// It retrieves the raw configuration, unmarshals it into an OpenIDConfiguration struct,
+// and returns any error that occurs during the process.
+func (c Container) OpenIDConfiguration(ctx context.Context) (cfg OpenIDConfiguration, err error) {
+	rawCfg, err := c.RawOpenIDConfiguration(ctx)
+	if err != nil {
+		return cfg, err
```
stevenh: suggestion: wrap error

modules/dex/dex.go
```diff
@@ -0,0 +1,349 @@
...
+		return fmt.Errorf("create password in Dex: %w", err)
+	}
+
+	return nil
+}
+
+// OpenIDConfiguration returns the OpenID configuration for the Dex instance.
+// It retrieves the raw configuration, unmarshals it into an OpenIDConfiguration struct,
+// and returns any error that occurs during the process.
+func (c Container) OpenIDConfiguration(ctx context.Context) (cfg OpenIDConfiguration, err error) {
```
stevenh: suggestion: return a pointer so it can be nil on error.

stevenh: suggestion: no need to name these.

modules/dex/dex.go
```diff
@@ -0,0 +1,349 @@
...
+
+	return cfg, nil
+}
+
+// RawOpenIDConfiguration retrieves the raw OpenID configuration from the Dex instance.
+// It fetches the configuration from the .well-known/openid-configuration endpoint,
+// patches all endpoint URLs to use the container's HTTP endpoint, and returns the
+// modified configuration as JSON bytes. Returns an error if the request fails
+// or if parsing/marshaling the configuration fails.
+func (c Container) RawOpenIDConfiguration(ctx context.Context) (rawCfg []byte, err error) {
```
stevenh, line 137: question: whats the use case for the raw version outside of the this package? I'm wondering if this should be private?

modules/dex/dex.go
```diff
@@ -0,0 +1,349 @@
...
+	httpEndpoint, err := c.PortEndpoint(ctx, HTTPPort, "http")
+	if err != nil {
+		return nil, fmt.Errorf("get Dex HTTP endpoint: %w", err)
+	}
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, httpEndpoint+"/.well-known/openid-configuration", http.NoBody)
+	if err != nil {
+		return nil, fmt.Errorf("prepare OIDC discovery requrest: %w", err)
```
stevenh: suggestion: follow the function call naming, and fix the typo.

modules/dex/dex.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+
+	httpClient := c.Client
+	if httpClient == nil {
+		httpClient = http.DefaultClient
+	}
```
stevenh: question: could the default be set on c and overridden with a functional option if needed?

modules/dex/dex.go
```diff
@@ -0,0 +1,349 @@
...
+	parsedOriginalURL.Scheme = parsedPatchedURL.Scheme
+	parsedOriginalURL.Host = parsedPatchedURL.Host
+
+	patched = parsedOriginalURL.String()
+	return patched, nil
+}
+
+// randomSecret generates a random password for identities.
+// Based on https://pkg.go.dev/crypto/rand@go1.24.0#Text
+// Can be replaced as soon as testcontainers-go is updated to Go 1.24 or higher.
```
stevenh: suggestion: mark with a TODO so its easy to find.

modules/dex/openid.go
```diff
@@ -0,0 +1,19 @@
+package dex
+
+type OpenIDConfiguration struct {
```
stevenh, line 6: suggestion: add documentation comment, required for all public methods, variables, constants and types.

modules/dex/options.go
```diff
@@ -0,0 +1,22 @@
...
+func WithIssuer(issuer string) testcontainers.ContainerCustomizer {
+	return testcontainers.CustomizeRequestOption(func(req *testcontainers.GenericContainerRequest) error {
+		req.Env["DEX_ISSUER"] = issuer
+		return nil
+	})
+}
+
+// WithLogLevel overrides the default log level for the Dex container.
+// See [log/slog](log/slog#Level) for possible values.
+func WithLogLevel(level string) testcontainers.ContainerCustomizer {
```
stevenh, line 27: question: should this take an https://pkg.go.dev/log/slog#Level if that's what it requires?

stevenh (changes requested):
thanks for the quick changes, a bit more of an in depth pass this time.

modules/dex/dex.go
prskr: Making this error private would mean, you cannot check against it and in consequence it'd make no sense to declare it as a variable in the first place?
I'd prefer to keep it that way to allow users to check for `errors.Is(err, dex.ErrPasswordRequired)`

modules/dex/dex.go
prskr: yes and no, if a user wants to have a matching `issuer` (in the OIDC discovery document) to the actual exposed port, it makes things a lot easier when the ports are part of the public API, because it is then possible to use the constant to construct a `WithExposedPorts(...)` option:

```go
// ...
  dex.WithIssuer("http://localhost:15556"),
  testcontainers.WithExposedPorts(fmt.Sprintf("15556:%s", dex.HTTPPort)),
//...
```

(see also the example)

Of course that is also possible without the ports being part of the public API but in the end, it potentially makes the code even more reliable because in case of a change, it's only necessary to update the constants as well and theoretically, user that stick to the (testcontainers-go) constants are less likely to experience a breaking change than users using their own constants/hardcoded strings

modules/dex/dex.go
prskr: I disagree, when using parameters instead of a struct, this would mean:

- either `clientID` and `clientSecret` **must** be set or it is allowed to pass empty strings and the function is checking for it and generating them (as it is now) but IMHO that does not improve the readability of the code in any way (empty strings vs. empty struct?!)
- every time Dex adds new fields to the client app this would potentially mean a breaking change of the public API of this module (at least for mandatory fields) or it would be necessary to add a variadic options API for this function as well to pass optional parameters, making it more confusing to consume

on the other hand (see also `CreatePassword` ):

- if mandatory fields are missing, it's easy to return an `error`
- if more fields are becoming mandatory, it would also be possible to add a `Validate()` function to the request structs to allow users of the module to check up front if their request is valid
- generally, it is always possible to add new fields **without** a breaking change in the API

Also, too many parameters in a public API are making it hard to use, for the `CreateClientApp` we'd be talking about 6 parameters, with the `context.Context` actually being the only one that is mandatory ( I double checked, actually even the generation of the `Id` in my function is not even necessary, as Dex is generating it if necessary) so how would you communicate that to the caller?

modules/dex/dex.go
prskr: see also previous comment.

Also for the password or hash question, it does not really make sense to me?
The `Hash` needs to be a proper Bcrypt hash, whereas the `Password` can be some arbitrary `string`.

If unifying it to a single parameter I'd have to rely on some heuristics to determine whether the string is a Bcrypt `Hash` or if it is meant to be the `Password` for no good reason because it's perfectly possible to pass either a `Password` or a `Hash` in their corresponding fields.

Also, passing the `Password` as `string` and the `Hash` as `[]byte` clarifies and distinguishes their individual characteristics to the caller. Apart from being able to clear the `Hash` after usage as a nice best practice even though this is not really security critical in this context.

So, I'd rather keep it as is for the mentioned reasons.

modules/dex/dex.go
prskr: see comment above, this does not make the API of the module clearer in any way if there are at least 4 different option types:

- general testcontainers-go container options
- dex module specific container options
- options for `CreateClientApp`
- options for `CreatePassword`

I prefer to stick to the struct parameter approach for this quite simple use cases

modules/dex/dex.go
prskr: all errors returned from `RawOpenIDConfiguration(...)` are already wrapped, I don't see what additional information that would add for the caller that makes the underlying issue easier to understand/debug

modules/dex/dex.go
prskr: Why would that be better?
It is anyway an anti-pattern to use the result of a function if the returned `error` was not `nil` and I tend to return value objects rather as structs than as pointers to avoid bugs with unintentional modifications.
Also I don't think that saving a few bytes here, is really an improvement 😅

prskr: well, it makes the code a bit shorter but you got it 👍

modules/dex/dex.go
prskr: It allows to use the patched configuration to be parsed for instance as [ProviderConfig](https://pkg.go.dev/github.com/coreos/go-oidc/v3/oidc#ProviderConfig) to interact with the provider in an easier way.

Or it can also be used in a mocked `http.Client` to return the patched discovery document as `http.Request#Body` (with a `io.NopCloser` and a `bytes.Reader` ).

If I'd have to tidy the public API I'd rather remove the non-raw version than the raw version 😅  because it's a lot more flexible

modules/dex/dex.go
prskr: theoretically yes but I don't think it would improve anything.

I won't ague again about option pattern APIs 😅 but I also don't see a realistic use case.

My bet would be that 99.9% of the users won't ever touch or even think about using a custom`http.Client` here and if they have some use case for that (like...transparently retrying the discovery document retrieval call in their test suite 😅 ) I would even argue that having to pass different `http.Client` instances to multiple calls is rather a code flaw than a 'feature'.

And even if they would need something like this, they could simply copy the `dex.Container` struct and use different `http.Client` instances on the struct level rather than on function level...

So, if you don't have a hard reason for that or insist on changing that, I'd prefer to keep it as is

modules/dex/options.go
prskr: I'd rather not to.
I thought about it and even though Dex is currently using `log/slog` it could of course change in the future.

Without any kind of compiler checking between this function and the underlying Dex version and also no guarantee about which version of Dex is used (because the image is a `string` parameter), it is perfectly possible that there's a breaking change without any testcontainers-go maintainer noticing it.
With the `logLevel` being a string parameter it's always possible to work around that and rely on the log level parsing on Dex side, whereas if we force a `log/slog.Level` here, it could become completely useless until a maintained fixes it.

But if you prefer a `log/slog.Level` I can change it of course

modules/dex/dex.go
prskr: I'd rather keep it together with the struct approach as described above

modules/dex/dex.go
stevenh: Yes, but the question is why would a user need to check the error, it failed because the code didn't pass the correct values in, so will need to be fixed to address the issue.

What scenario are you envisaging that checking the error would help with?

modules/dex/dex.go
stevenh: sounds like you're suggesting the customisation of client isn't needed, as it doesn't have a well defined use case. If thats the case I'd remove. We can add in later if a case requires it.

I've see so many cases where a developer, myself included, thinks lets add this option because its not hard to do, only for that option to never be used. So with all of the years under my belt now I always ask, but is that actually needed.

modules/dex/options.go
stevenh: thats a good call, avoid exposing the implementation detail to the user.

modules/dex/dex.go
stevenh: Sounds like the same direction which was take with the container request type, which is something we're in the process of changing, because it became unmanageable and very hard for the user to understand.
 
Could you clarify how each of the fields is used, CreateClientAppRequest and CreatePasswordRequest, specifically which are optional, and which may have and exclusive requirement. For example it sounds like Password should never be set if Hash is.

If you've not read it this article describes the benefits of [functional options](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis), which is what I'm thinking would be the better option here and is the direction we're going with all testcontainers core.

modules/dex/dex.go
prskr: Alright, will make it private

prskr: Alright, will make it private

modules/dex/dex.go
prskr: Sure, with all the experience, who am I to question that, will remove it

modules/dex/dex.go
prskr: I am well aware of the pattern, thank you.

As for the usage:

- for the CreatePasswort request email and password or hash are required for a user to login
- for the CreateClientApp, all fields are optional and will be generated on demand if necessary ( ClientID and ClientSecret)

Honestly, I'm this I am not willing to discuss, if you insist on the options pattern, I will just remove the functions altogether or probably close the PR

modules/dex/dex.go
mdelapenya: @prskr our job as maintainers is to keep consistency across the codebase, because we want to make sure contributors create a consistent experience while using the project.

I'm sorry if we are adding too many questions here or in any other PR, so please forgive us if we are many times too verbose. Also, we follow conventional comments, which if not used to, could be seen as very "direct", but I do confirm that once familiar with the style it's pretty simple to understand that we are talking just about code, not about anything else.

Said that, we are questionable, so we are really glad to see you interact with us explaining your concerns. The resulting code will be richer and with more context.

modules/dex/dex.go
mdelapenya: @prskr we are transitioning to functional opts when/if possible, so I'd suggest using them here. If you see other modules in the repo, we usually define an options struct receiving the optional values, including custom Opts setting them. I think this pattern, which could be seen as "boring", makes the API pretty predictable.

You are always able to create your own module under your account, with your own APIs, and we can link to it in the modules catalog as a community module, but having it here will mean automatic releases, CI execution, and more visibility, among others.

In any case, we'd love to see your contribution in the repo, if possible.

modules/dex/dex.go
```diff
@@ -0,0 +1,352 @@
...
+	}
+
+	if _, err = apiClient.CreatePassword(ctx, apiReq); err != nil {
+		return fmt.Errorf("create password in Dex: %w", err)
+	}
+
+	clear(req.Hash)
+
+	return nil
+}
```
stevenh: I wanted to take the time to provide an example of what I had in mind, to make sure we're aligned on what I suggested and how it could help users.

As @mdelapenya mentioned this is a approach we're looking to standardise on to provide a consistent user experience across all testcontainer modules.

While this is more initial code, it makes it obvious at a glance what the user needs to use the method, specifically:
1. Email is required
2. Either a password or a hash are required
3. UserID and Username are optional

Here's it in use:
```go
if err := dexContainer.CreateLogin(ctx, email, dex.WithPassword(password), dex.WithUsername("ted.tester")); err != nil {
	log.Printf("failed to create login: %s", err)
	return
}
```

This approach helps prevent more issues at compile time instead of runtime, for example it enforces an auth option is provided. Also if the underlying API changes, adding or removing options we are more likely, can update in a way which maintains compatibility.

Hope this helps.
```suggestion
// createPasswordRequest represents a request to register an identity in Dex.
type createPasswordRequest struct {
	userID   string
	email    string
	username string
	// Password is a plain text password.
	password string
	// Hash is a bcrypt hash of the password.
	hash []byte
}

// CreatePasswordOption represents a create password optional parameter.
type CreatePasswordOption func(*createPasswordRequest) error

// CreatePasswordAuth represents a create password authentication parameter.
type CreatePasswordAuth func(*createPasswordRequest) error

// WithPassword sets the plaintext password for the password request.
func WithPassword(password string) CreatePasswordAuth {
	return func(req *createPasswordRequest) error {
		if password == "" {
			return errors.New("password cannot be empty")
		}
		req.password = password
		return nil
	}
}

// WithHash sets the bcrypt hash for the password request.
func WithHash(hash []byte) CreatePasswordAuth {
	return func(req *createPasswordRequest) error {
		if len(hash) == 0 {
			return errors.New("hash cannot be empty")
		}
		req.hash = hash
		return nil
	}
}

// WithUserID sets the user ID for the password request.
func WithUserID(userID string) CreatePasswordOption {
	return func(req *createPasswordRequest) error {
		if userID == "" {
			return fmt.Errorf("userID cannot be empty")
		}
		req.userID = userID
		return nil
	}
}

// WithUsername sets the username for the password request.
func WithUsername(username string) CreatePasswordOption {
	return func(req *createPasswordRequest) error {
		if username == "" {
			return errors.New("username cannot be empty")
		}
		req.username = username
		return nil
	}
}

// CreateLogin registers a new login in Dex.
// It requires either a plaintext password or a bcrypt hash of the password.
// If the password is provided, it will be hashed using bcrypt before being sent to Dex.
// If the hash is provided, it will be used directly.
// The user ID is generated if not provided, and the username can be set optionally.
func (c Container) CreateLogin(ctx context.Context, email string, auth CreatePasswordAuth, options ...CreatePasswordOption) (err error) {
	req := &createPasswordRequest{
		email:  email,
		userID: uuid.New().String(),
	}

	for _, opt := range options {
		if err = opt(req); err != nil {
			return fmt.Errorf("apply option: %w", err)
		}
	}

	apiClient, connCloser, err := c.createDexAPIClient(ctx)
	if err != nil {
		return fmt.Errorf("create Dex API client: %w", err)
	}

	defer func() {
		if closeErr := connCloser.Close(); closeErr != nil {
			err = errors.Join(err, fmt.Errorf("close GRPC connection: %w", closeErr))
		}
	}()

	if len(req.hash) == 0 {
		req.hash, err = bcrypt.GenerateFromPassword([]byte(req.password), bcryptCost)
		if err != nil {
			return fmt.Errorf("hash plaintext password with bcrypt: %w", err)
		}
	}

	apiReq := &dexapi.CreatePasswordReq{
		Password: &dexapi.Password{
			UserId:   req.userID,
			Username: req.username,
			Email:    req.email,
			Hash:     req.hash,
		},
	}

	if _, err = apiClient.CreatePassword(ctx, apiReq); err != nil {
		return fmt.Errorf("create password in Dex: %w", err)
	}

	return nil
}
```

modules/dex/dex.go
prskr: I totally get what you're doing for the "containers" API and I don't want to question that or whatever. My only problem with the approach - especially in this context - is that if you have multiple option types in a single package all starting with 'WithXXX' in my opinion does not improve the developer experience in any way because (AFAIK) gopls for instance does not filter for functions/variables matching the option type but it filters solely for functions/variables matching 'With' so it's up to the developer to check which options can be used with the current function. Apart from that I continuously find new options in APIs that I'm using probably because they are in other packages or because they don't follow the naming scheme or because I was confronted with too many god damn options so I couldn't easily browse them all.

The point where you manually create documentation for all possible options because Godoc isn't helping should alert you that something's off.

And this problem is getting even worse when you have multiple option types because some option might sound correctly but the type does not match (or not obviously match because some "smart" type hacks were applied).

I'm not saying you should not use this for the container API, I even created some options for this module, I just don't see the point of claiming there's this single hammer and everything is a nail. CreatePassword has 2 mandatory parameters and...4 optional ones, CreateClientApp only has optional parameters... I'm currently on vacation and don't have a laptop with me but I'm happy to create an example of how the API would Look Like to illustrate how inconvenient that could become... no

Edit: [CreateCollection](https://golang.testcontainers.org/modules/chroma/#container-methods) is actually a great example, is this really easier to understand and review than passing a struct with named fields that would tell you what this 'true' means or that 'map[string]any'? No? I thought so too...probably I should close this PR and implement the one integration test I need with a generic container...

modules/dex/dex.go
prskr: I get that, but all I saw is "suggestion: " which implies I can do that or not, combined with "requested changes" so it's not a suggestion, it is "change this or it won't be merged" which is also fine but then make it consistent please 😊

Also, sorry to be frank but, it's quite common to make it possible to switch the http.Client for various reasons, I don't want to start explaining every common design decision of the Go Community over and over again...

Edit: I'm German, I'm all in for direct comments, I'm just getting annoyed if the comments aren't leading anywhere but just extend the discussion forever for no good reason

modules/dex/dex.go
prskr: I am aware what you have in mind and how the options pattern is implemented (although I prefer interfaces over pure functions because that is slightly more flexible) but I still fail to see how that catches anything at compile time unless you replace the option for password or hash with an actual 'credential' type with two implementations (e.g. PlainTextCredential and HashCredential) but the options pattern does not solve this in any way 🤷‍♂️

Plus - but that is quite personal - I don't Like how that forces (in most cases) multi-line function calls that also don't improve error handling readability...

prskr:
I'll change everything as you requested when I'm back from vacation, hopefully this will then at least guarantee a quick merge although I'll detest my own API then

modules/dex/dex.go
stevenh: I'm not sure I follow, when I look at the docs all the options are grouped under the type:
image
Also grouped together in the index, so if you're looking for say an CreatePasswordAuth option, the docs make it clear the options you have:
image

I'm interested to hear more about the issue you're seeing with options spread across packages, have you got any examples of that?

I just tried in VS code using the options in the example and prompted with the relevant options, so again interested to hear about that issue more, so we can understand what you're experiencing.

Definitely interested in what it might look like, when you get back, in the mean time enjoy your holiday and home you have some 🌞

modules/dex/dex.go
stevenh: Let me try to clarify, with the struct there is no compile time validation, if you didn't set either a password or hash on the struct you will only know when the code runs. The user has to read the comments to know they have to set one or the other.

With the options version, first it's self documenting, as there is `auth` parameter which must be provided, if you don't, or provide an invalid option it won't even compile. Yes you could still pass a blank string to the methods, but that would be a conscious action vs just passing in a default struct with no values set.

Below is the options version, as you can see it enforces you have to pass `auth` which has two valid options, `WithPassword` and `WithHash`, the other truly optional values are seperate.
```go
func (c Container) CreateLogin(ctx context.Context, email string, auth CreatePasswordAuth, options ...CreatePasswordOption) (err error) {
```

modules/dex/dex.go
mdelapenya: @prskr I'm Spanish and I'm more friendly in my communication style. And I do appreciate the same no matter the nationality.

modules/dex/dex.go
prskr: I already commented I will do everything as requested. Just for the record as also already mentioned @mdelapenya there are no  "suggestions" here, most of the time it does not even look like you'd even read what I explained, it's 100% pushing into "do this like I want, or fuck off" and as a consequence this will certainly be the last PR I will do for test containers. I already had a similar experience with testcontainers-dotnet where I discussed for at least 3 months with the maintainer about every tiny API specific (and back then I didn't even discuss much, I just directly changed everything as requested) and nevertheless the PR was actually never merged at all because I eventually moved on and solved my problem in a different way. If that is how you want to live open source, alright, just my 2 cents.

mdelapenya: @prskr I would like to be candid for a moment. Discussing the technical details for every tiny API specific is the right thing to do when contributing to a project that is so widely used. If not, we can get to an API that is super difficult to use/read/maintain, like the API for the Podman support you and me introduced a couple of years ago. We are paying the added complexity at the provider layer so hard, that I'm removing the entire provider layer the soonest. And at that time I was not picky for any tiny API specific, but encouraging you to contribute more. The result? You merged what you wanted and did not appear back until now. So it's not about being German, Spanish or British, it's about being polite with the OSS community here at Testcontainers Go. I can be paid by my employer to do this, and you can tell me that. But @stevenh is using is very reduced spare time to respond, read all the tiny API specifics of any PR he takes, and do so much amazing work that he does not deserve how you are responding to him.

As I said in https://github.com/testcontainers/testcontainers-go/pull/3216#discussion_r2161904777, if you're not really convinced, like it already happened here, and the .NET example you brought, you can always  use your repo with your own API, like the wiremocks or microcks folks did, although for obvious different reasons.

I'll let you do what you consider it's fine, we are not gonna close this as we see the value in adding modules to the catalog.

I want to finish saying sorry for being rude, but I had to defend the work of the community here.

prskr: No need to apologize, I rather get honest critic/feedback than polite tip toeing!
I'm honestly sorry that the Podman support caused you so many troubles and I wasn't available to at least share that pain. For what it's worth: back then I switched job and my new employer does not use go or testcontainers and I barely had some time left to contribute to anything OSS in the last years. In fact, I spent the last 4 years convincing my superiors that a company that big should definitely contribute to open source...It wasn't for the lack of interest but simply time. So, sorry!

I get that the API needs to be good as it's hard/impossible to change once it's published and I'm as much concerned about the API as you, are otherwise I'd have blindly accepted what was proposed without pushing back. If the options pattern is your only way to go, alright, but please document it then also accordingly and don't claim it's a suggestion when it's a requirement really. With the little time I have to do open source stuff, I'd like to be as efficient as possible, that also means, the fewer iterations I have to incorporate review comments and such, the better, if that makes sense.

modules/dex/dex.go
stevenh: The rule I tend to use is wrap errors returned from public functions, private which already contain the relevant context are ok. I'll look at this example when back at a PC, but wanted to let you know the criteria which triggered the original comment

modules/dex/dex.go
stevenh: I believe it's been mentioned before but the prefixes are defined by the [conventional comments standard](https://conventionalcomments.org/), which we follow for this project. I would definitely encourage you to familiarise with that, if you aren't already.

We never know what contributors have considered when developing a potential change, so one of our roles is to ask the questions, make suggestions with goal of making testcontainers-go better for all. This can involve helping to ensure we provide a consistent style and approach across the codebase, so users can apply knowledge from using one module, or the core, to another resulting in a better developer experience.

We always do so with best intent, so I would encourage you to engage with the team in a respectful and courteous manor. Opinions may differ, we may not start out aligned, we may have missed a key fact when comments are made, if you believe this is the case I would encourage you to present the details to build that shared understanding. A well formed case for a different approach will always be considered.

My explanation of how functional options could be applied to the user creation, is hopefully a good example of that, in that I presented what the code might look like and explained how it can help users avoid runtime issues. You suggested that these might have a discoverability issue, and I'm interested to understand more about that, as I've not experienced that personally. That doesn't mean I'm right and you're wrong, just that we need to share our collective experiences so the other person can understand the alternative perspective.

We might have strong opinions but they are loosely held, and we want to understand and learn from others.

Most importantly we're all here to make testcontainers-go better, every contribution is valuable, it's at the heart of open source so thank you for you work so far on this and I hope we can work together to get it over the line.

prskr: Okay: it is quite common to create some kind of "escape hatch" to explicitly use a provided http.Client instead if the global one for multiple reasons.
Due to the nature of the default http.Client being a global one and the arguably good or bad practice (I don't want to argue about that) that certain libraries are patching the global http.Client with certain behaviors, many developers prefer to specify the http.Client (for instance with an "empty" default). This ensures that the client behaves as expected and does not for instance automatically retry requests, tries to submit telemetry or emit log lines, that could make test debugging a lot easier. Hence, I figured, that would be some valuable option. Of course, if you think different, I'm happy to remove this.

modules/dex/dex.go
prskr: I can certainly wrap the error if you prefer it that way.
Normally, I'm trying to add relevant context to understand the underlying issue, whereas always wrapping the error feels to me more like a stack trace "emulation".
In this case, I don't see how that would help to fix anything, probably you explain your view on this?
I'm fine either way, just trying to understand the pattern.

stevenh: yes RawOpenIDConfiguration wraps the errors internally, but as you mentioned its to provide the user with the full context, without that those errors could be ambiguous. Instead of manually checking if that might be the case, we're just going with always wrap the returns of public functions.

We've had a few cases of issues reported where we couldn't tell if that was path A or path B, so we're just making it easier to maintain by sticking with wrap them all, if you would be so kind.

modules/dex/dex.go
prskr: One thing is the following:

image

There are 3 different option types in the package:

- for the container itself
- for the client app
- for the password

apparently, they fixed at least the sorting now - which is great - but nevertheless I get all options although the last 4 don't apply for this function.

Similarly, but different:

image

When searching for options for the container.
Yes, surprisingly, I get some options from both packages where I already imported options from, but the full list of all options I only get when I correctly prefix it with a matching (and imported) package.

Also, for the godoc  to work as you previously showcased by you, the options themselves also have to be public, making it harder to change the structure of the option themselves later. If the option **types** are not public, the functions will not be grouped under type.

Just mentioning it, for a common understanding and as you already mentioned concerns about the size of the public API.

As for examples of options spreading across packages: one obvious example are the various container options coming from the `testcontainers` package + the options for the individual module for instance `dex` in this case.
Which is why you have this include in the docs for every module to also show all common options 😅

Having said that, I'm currently reworking the API as requested and will push an update soon (probably, without doc comments, just for a first review and when all API details are settled, I'll add the missing comments).

modules/dex/dex.go
prskr: I now removed the `http.Client` option completely as well

modules/dex/dex.go
prskr: removed

modules/dex/dex.go
prskr: Alright, wrapped it

modules/dex/dex.go
prskr: is this resolved or shall I remove it?

prskr:
Not sure why the pipeline now fails to be honest, I did validate golangci-lint locally and it worked, will check it in the upcoming days

modules/dex/dex.go
prskr: Updated for options

modules/dex/dex.go
prskr: Done

modules/dex/dex.go
prskr: Is this still a concern? Just trying to follow up on all change requests.

modules/dex/dex.go
stevenh: That's a good catch on the private types vs public for grouping, definitely something to consider when adding these in the future, thanks for taking the time to demo that.

It's interesting that the intellisense doesn't take the type information into account, but it does show the type and the will flag if you chose an incompatible option, so you do know instantly that it's an invalid option thankfully.

modules/dex/dex.go
```diff
@@ -0,0 +1,279 @@
...
+		if closeErr := connCloser.Close(); closeErr != nil {
+			err = errors.Join(err, fmt.Errorf("close GRPC connection: %w", closeErr))
+		}
+	}()
+
+	var dexClientApp dexapi.Client
+
+	for _, opt := range opts {
+		opt(&dexClientApp)
+	}
```
stevenh: question: can you imagine a scenario where the user could pass an option which is invalid, for example a blank value?

I ask as we've had to make breaking changes to the API in the past because an error return wasn't included from the start, so I would look to add this to all options even if it's currently not used.

modules/dex/dex.go
```diff
@@ -0,0 +1,279 @@
...
+		return nil, fmt.Errorf("prepare Dex API client: %w", err)
+	}
+
+	defer func() {
+		if closeErr := connCloser.Close(); closeErr != nil {
+			err = errors.Join(err, fmt.Errorf("close GRPC connection: %w", closeErr))
+		}
+	}()
+
+	var dexClientApp dexapi.Client
```
stevenh: praise: nice reuse of the client here, however I would use a local struct with a private dexapi.Client to prevent leaking internal details.

modules/dex/dex.go
```diff
@@ -0,0 +1,279 @@
...
+	apiClient, connCloser, err := c.createDexAPIClient(ctx)
+	if err != nil {
+		return fmt.Errorf("create Dex API client: %w", err)
+	}
+
+	defer func() {
+		if closeErr := connCloser.Close(); closeErr != nil {
+			err = errors.Join(err, fmt.Errorf("close GRPC connection: %w", closeErr))
+		}
+	}()
```
stevenh, line 103: suggestion: move this to after the option processing so we don't create, only to trigger an error and never use.

modules/dex/dex.go
```diff
@@ -0,0 +1,279 @@
...
+		return nil, fmt.Errorf("decode OIDC discovery response: %w", err)
+	}
+
+	for _, property := range propertiesToPatch {
+		jsonValue, ok := partiallyDecoded[property]
+		if !ok {
+			continue
+		}
+
+		patched, err := patchEndpoint(strings.Trim(string(jsonValue), "\""), httpEndpoint)
```
stevenh: nit: use backticks to avoid escaping, as it makes it easier to read.
```suggestion
		patched, err := patchEndpoint(strings.Trim(string(jsonValue), `"`), httpEndpoint)
```

modules/dex/options.go
```diff
@@ -0,0 +1,101 @@
...
+	return testcontainers.CustomizeRequestOption(func(req *testcontainers.GenericContainerRequest) error {
+		req.Env["DEX_LOG_LEVEL"] = level
+		return nil
+	})
+}
+
+type CreatePasswordCredential func() (email string, passwordHash []byte, err error)
```
stevenh, line 34: suggestion: add comments to all public types and methods.

modules/dex/options.go
```diff
@@ -0,0 +1,101 @@
...
+
+func PlainTextCredential(email, password string) CreatePasswordCredential {
```
stevenh, line 36: question: should we use a different names here, as it currently suggests that this is less secure than the hashed version which is no the case as they both result in hashed credentials.

modules/dex/options.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+}
+
+func HashedCredential(email string, passwordHash []byte) CreatePasswordCredential {
+	return func() (string, []byte, error) {
+		return email, passwordHash, nil
+	}
+}
+
+type CreatePasswordOption func(pw *dexapi.Password)
```
stevenh: question: as above, should we return an error here to allow it to validate user input, for example I would expect an error if I passed in a blank userID, username etc.

modules/dex/dex.go
```diff
@@ -0,0 +1,279 @@
...
+		return fmt.Errorf("create Dex API client: %w", err)
+	}
+
+	defer func() {
+		if closeErr := connCloser.Close(); closeErr != nil {
+			err = errors.Join(err, fmt.Errorf("close GRPC connection: %w", closeErr))
+		}
+	}()
+
+	dexPassword := dexapi.Password{
```
stevenh: suggestion: similar to above, local struct with private dexapi.Password, which would simplify the code below allowing it to just be a opts loop.

modules/dex/options.go
```diff
@@ -0,0 +1,101 @@
...
+
+func PlainTextCredential(email, password string) CreatePasswordCredential {
+	hashedPassword, err := bcrypt.GenerateFromPassword([]byte(password), bcrypt.DefaultCost)
+
+	return func() (string, []byte, error) {
+		return email, hashedPassword, err
+	}
+}
+
+func HashedCredential(email string, passwordHash []byte) CreatePasswordCredential {
```
stevenh, line 43: question: is there any restrictions on passwordHash, sometimes then need to be a fixed length?

modules/dex/options.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+}
+
+func WithClientSecret(clientSecret string) CreateClientAppOption {
+	return func(ca *dexapi.Client) {
+		ca.Secret = clientSecret
+	}
+}
+
+func WithRedirectURIs(redirectURIs []string) CreateClientAppOption {
```
stevenh: suggestion: use a variadic here so users can just pass in values without having to wrap with a slice.

modules/dex/options.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+}
+
+func WithPublicClient(public bool) CreateClientAppOption {
+	return func(ca *dexapi.Client) {
+		ca.Public = public
+	}
+}
+
+func WithTrustedPeers(trustedPeers []string) CreateClientAppOption {
```
stevenh: suggestion: as above, use a variadic.

modules/dex/dex.go
prskr: Well, in theory most ™️  default values are blank, the only option that could be problematic is `WithUserID` as Dex (apparently) requires a `UserID` to be set, which I defaulted to `uuid.New().String()`. If a user passes `WithUserID("")` it would become invalid and the gRPC call would return a matching error (if I remember correctly from testing).

I can certainly add an `error` return value if you prefer it that way. It's a bit unfortunate that the option types themselves need to be public for the grouping otherwise we could omit the `error` for now and add it later if necessary. With the option types themselves being public it would be safer to add the `error` return from the beginning although it's probably unused.

Your call, what you prefer

modules/dex/dex.go
prskr: you mean something like:

```go
// createClientAppOptions - copy of dexapi.Client
type createClientAppOptions struct {
	Id           string
	Secret       string
	RedirectUris []string
	TrustedPeers []string
	Public       bool
	Name         string
	LogoUrl      string
}
```

and then

```go
type CreateClientAppOption func(cao *createClientAppOptions) error
```

and similarly for the `CreatePassword` function?

modules/dex/options.go
prskr: will do, as said, I wanted to align on the API structure first before having to rewrite the API docs multiple times 😊

modules/dex/options.go
prskr: Well it is kind of less secure because it exposes the plain text password in the memory whereas if the hash is read from an embedded file or so, the password is not exposed.
But I'm open for suggestions for a different naming if you prefer it that way

modules/dex/options.go
prskr: sure, will do

modules/dex/dex.go
prskr: for the record: I don't think this is improving anything but sure, will do

modules/dex/options.go
prskr: from what I can tell, there's a limitation of the password length that is enforced when the hash is calculated already by the `bcrypt` package. I could not find any max length for the hash itself?
A basic heuristic would be to check how long the hash would be with max 72 characters but I could imagine that this might be changing?

Is it worth it to try to catch this kind of errors on this level or should we rely on the provider API implementation to catch it?
I'm mostly concerned about inconsistent limits between this library (testcontainers-go), the Dex API and the bcrypt implementation in Go.

modules/dex/dex.go
prskr: I updated it accordingly, I hope this suits your expectations

modules/dex/dex.go
stevenh: Agreed, public type with private fields and a error return is the best way to go to avoid breaking changes and improved documentation visibility.

modules/dex/dex.go
prskr: thought: with a custom struct instead of `dexapi` types (`Client` and `Password`) the option types are more closed and less flexible.
If you seek a way to force library users to wait for you to add new options to the API because of new options in the Dex API, this is the way.
If you want to allow users of the library to use new options in the Dex API without a testcontainers-go maintainer, the previous approach would be more flexible because it allows users to create their own options when necessary to further customize the behavior.

I don't want further discussions, I just wanted to clarify this, so we are all on the same page. I already implemented the requested change.

modules/dex/dex.go
stevenh: not quite, combining with your suggestion above to keep the type public to improve doc grouping:
```golang
// CreateClientAppOptions represents ... see ....
type CreateClientAppOptions struct {
    client dexapi.Client
}
```

modules/dex/options.go
stevenh: If there's no hard restriction on the underlying API currently, as you suggested better to defer to it to the underlying SDK than introduce artificial restrictions here.

modules/dex/dex.go
prskr: already implemented the `error` return

modules/dex/dex.go
prskr: note: although I appreciate `praise` comments, in this case a `suggestion` or an `issue` would've been better (IMHO) in case you want to ensure that the comment is actually considered during review. Probably also only me, but I tend to focus on stuff that needs attention during the feedback phase, and `praise` is probably something I'd keep for later

modules/dex/dex.go
prskr: Is there something further to discuss here or can this be resolved as the requested approach is already implemented and I would expect further discussion to take place at the individual options?

modules/dex/dex.go
prskr: I guess this discussion is concluded as all requested changes are implemented?

modules/dex/options.go
prskr: From what I can tell after checking also the code of the Dex IdP, there's no limit other than on the storage which is...eventually 1GB for instance on PostgreSQL (the column type is `bytea`, same for SQLite).

While checking that, I noticed that Dex devs recommend 12 as cost instead of the default cost of the bcrypt package (which is 10). I'm not really concerned about the difference, especially as we're only talking about testing scenarios and I'd rather stick to the bcrypt package defined constant than putting a custom constant into testcontainers-go but I wanted to at least bring this to your attention for potential future discussions.

modules/dex/dex.go
prskr: > not quite, combining with your suggestion above to keep the type public to improve doc grouping:
> 
> ```golang
> 
> // CreateClientAppOptions represents ... see ....
> 
> type CreateClientAppOptions struct {
> 
>     client dexapi.Client
> 
> }
> 
> ```

That one I don't get, the grouping is unaffected by the way how I now changed the accumulator of the option. Do you want to avoid a dependency to the dexapi.Client I the options (I.e. also in the parameters) or what are you trying to tell me here?

prskr: See also:

image

and

image

even though I don't see...any real advantage of embedding the `dexapi` type in the corresponding option struct, I can of course certainly do that if you like

Edit: updated screenshot of docs to match the options in the example

prskr:
@stevenh  @mdelapenya I thought about this again and came to the conclusion that I will not progress with this PR or any other PR with testcontainers-go in the future.

I do understand that, you do what you think you must and I hope you understand that I have to do, what I must.
I don't get any fun out of contributing to testcontainers-go anymore this way.

Even without further discussions and simply accepting all "suggestions" as change requests, every round of adopted changes only brought up another, the last one even without ANY change to the API of the package OR maintenance implications but - as I see it - purely a different taste of things.

If that is how you successfully collaborate with other contributors, great! But I'd have wished for a more efficient and focused way of reviewing to ensure consistency and quality without sacrificing individuality.

I wish you all the best

mdelapenya:
@prskr our main goal has always been and will always be to think in the "common good" for the project and the community first.

The take away after this for us is to revisit the contribution guidelines, because the design process of any software is continuously evolving, and they could be outdated soon. What we cannot do is to write down all the rules and patterns for each block in the project, as they would be outdated the next day a new pattern emerges. Our job as maintainers is to transmit that "tribal knowledge" in any manner: updating the docs (we seem to fail here?) or during the review (as we also failed here?). Of course, it never rains to everybody's desires (spanglish), so we perfectly understand your decision. We advised you to create your module, with your rules under your own account, as many other people do.

It's always hard to cut relationships with a project, so I really believe you did a pretty conscious decision.

PR #3218, opened 2025-06-26
feat: allow saving specific platforms for an image

LaurentGoderre:
## What does this PR do?

This introduces a way to pass image save options (like platform) to the image client. This allows saving a specific architecture for an image.

## Why is it important?

With OCI multi-arch images, it is sometimes needed to save image for a different architecture than the current one. This changes allows to control the behavior instead of defaulting to the current platform.

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -1786,11 +1786,11 @@ func (p *DockerProvider) ListImages(ctx context.Context) ([]ImageInfo, error) {
 
 // SaveImages exports a list of images as an uncompressed tar
 func (p *DockerProvider) SaveImages(ctx context.Context, output string, images ...string) error {
-	return p.SaveImagesWithOps(ctx, output, images)
+	return p.SaveImagesWithOpts(ctx, output, images)
 }
 
-// SaveImagesWithOpts exports a list of images as an uncompressed tar, passiong options to the provider
-func (p *DockerProvider) SaveImagesWithOps(ctx context.Context, output string, images []string, opts ...SaveImageOption) error {
+// SaveImagesWithOpts exports a list of images as an uncompressed tar, passing options to the provider
+func (p *DockerProvider) SaveImagesWithOpts(ctx context.Context, output string, images []string, opts ...SaveImageOption) error {
 	saveOpts := saveImageOptions{}
 
 	for _, opt := range opts {
--- a/image.go
+++ b/image.go
@@ -22,6 +22,6 @@ type SaveImageOption func(*saveImageOptions) error
 type ImageProvider interface {
 	ListImages(context.Context) ([]ImageInfo, error)
 	SaveImages(context.Context, string, ...string) error
-	SaveImagesWithOps(context.Context, string, []string, ...SaveImageOption) error
+	SaveImagesWithOpts(context.Context, string, []string, ...SaveImageOption) error
 	PullImage(context.Context, string) error
 }
--- a/image_test.go
+++ b/image_test.go
@@ -99,7 +99,7 @@ func TestSaveImagesWithOpts(t *testing.T) {
 	require.NoErrorf(t, err, "creating test container")
 
 	output := filepath.Join(t.TempDir(), "images.tar")
-	err = provider.SaveImagesWithOps(
+	err = provider.SaveImagesWithOpts(
 		context.Background(), output, []string{req.Image}, SaveDockerImageWithPlatforms(p...),
 	)
 	require.NoErrorf(t, err, "saving image %q", req.Image)
--- a/modules/k3s/k3s.go
+++ b/modules/k3s/k3s.go
@@ -213,7 +213,7 @@ func (c *K3sContainer) LoadImagesWithOpts(ctx context.Context, images []string, 
 		_ = os.Remove(imagesTar.Name())
 	}()
 
-	err = provider.SaveImagesWithOps(context.Background(), imagesTar.Name(), images, opts...)
+	err = provider.SaveImagesWithOpts(context.Background(), imagesTar.Name(), images, opts...)
 	if err != nil {
 		return fmt.Errorf("saving images %w", err)
 	}


docker.go
```diff
@@ -1784,8 +1784,30 @@ func (p *DockerProvider) ListImages(ctx context.Context) ([]ImageInfo, error) {
 	return images, nil
 }
 
+func SaveImageOptionWithPlatforms(plaforms ...specs.Platform) SaveImageOption {
```
stevenh: suggestion: add comment for all public methods, more in other files.

question: could this have a simpler name, WithPlatforms or SaveImagesWithPlatforms

docker.go
```diff
@@ -1784,8 +1784,30 @@ func (p *DockerProvider) ListImages(ctx context.Context) ([]ImageInfo, error) {
...
+		opts.Platforms = plaforms
+	}
+}
+
 // SaveImages exports a list of images as an uncompressed tar
 func (p *DockerProvider) SaveImages(ctx context.Context, output string, images ...string) error {
+	return p.SaveImagesWithOps(ctx, output, images)
+}
+
+// SaveImages exports a list of images as an uncompressed tar
```
stevenh: suggestion: correct description which refers to the old name

image.go
```diff
@@ -10,9 +12,16 @@ type ImageInfo struct {
 	Name string
 }
 
+type SaveOptions struct {
+	Platforms []specs.Platform
+}
+
+type SaveImageOption func(*SaveOptions)
```
stevenh: suggestion: return an error so its future proof for extensions.

image.go
```diff
@@ -10,9 +12,16 @@ type ImageInfo struct {
 	Name string
 }
 
+type SaveOptions struct {
+	Platforms []specs.Platform
```
stevenh: suggestion: keep details private so changes can be used without breaking the API.

image_test.go
```diff
@@ -71,3 +72,37 @@ func TestSaveImages(t *testing.T) {
...
 }
+
+func TestSaveImagesWithOpts(t *testing.T) {
+	t.Setenv("DOCKER_HOST", core.MustExtractDockerHost(context.Background()))
+
+	provider, err := ProviderDocker.GetProvider()
+	require.NoErrorf(t, err, "failed to get provider")
+
+	defer func() {
+		_ = provider.Close()
```
stevenh: suggestion: check the error.

stevenh (commented):
Thanks for this, I've done a quick pass with a few improvement suggestions.

stevenh (changes requested):
Clicking the right option this time 😄

image.go
LaurentGoderre: The reason I need it is for the K3S module so I would have to implement a duplicate type in that module

docker.go
LaurentGoderre: Done!

image.go
LaurentGoderre: Actually, I made it work while keeping it private.

LaurentGoderre:
@stevenh updated! Thanks for the review!

LaurentGoderre:
FYI, we are using a simplified version of this change internally for our tests without a problem

docker.go
```diff
@@ -1786,6 +1786,22 @@ func (p *DockerProvider) ListImages(ctx context.Context) ([]ImageInfo, error) {
...
+	return p.SaveImagesWithOps(ctx, output, images)
+}
+
+// SaveImagesWithOpts exports a list of images as an uncompressed tar, passiong options to thw provider
+func (p *DockerProvider) SaveImagesWithOps(ctx context.Context, output string, images []string, opts ...SaveImageOption) error {
+	saveOpts := saveOptions{}
+	dockerOpts := []client.ImageSaveOption{}
+
+	for _, o := range opts {
+		o(&saveOpts)
```
stevenh: bug: not checking the error return

image.go
```diff
@@ -10,9 +12,25 @@ type ImageInfo struct {
 	Name string
 }
 
+type saveOptions struct {
+	Platforms []specs.Platform
```
stevenh: suggestion: if you change the type to use the underlying target this simplifies the overall code e.g.

```golang

type saveImageOptions struct {
	options []client.ImageSaveOption
}

// SaveImageOption represents an option that can be passed to SaveImagesWithOps
type SaveImageOption func(*saveImageOptions) error

// SaveImageWithPlatforms allows specifying which platform(s) to save.
func SaveImageWithPlatforms(platforms ...specs.Platform) SaveImageOption {
	return func(opts *saveImageOptions) error {
		opts.options = append(opts.options, client.ImageSaveWithPlatforms(platforms...))

		return nil
	}
}

// SaveImages exports a list of images as an uncompressed tar
func (p *DockerProvider) SaveImages(ctx context.Context, output string, images ...string) error {
	return p.SaveImagesWithOps(ctx, output, images)
}

// SaveImagesWithOpts exports a list of images as an uncompressed tar, passing options to thw provider.
func (p *DockerProvider) SaveImagesWithOps(ctx context.Context, output string, images []string, opts ...SaveImageOption) error {
	saveOpts := saveImageOptions{}
	for _, opt := range opts {
		if err := opt(&saveOpts); err != nil {
			return fmt.Errorf("applying save image option: %w", err)
		}
	}

	outputFile, err := os.Create(output)
	if err != nil {
		return fmt.Errorf("opening output file %w", err)
	}
	defer func() {
		_ = outputFile.Close()
	}()

	imageReader, err := p.client.ImageSave(ctx, images, saveOpts.options...)
	if err != nil {
		return fmt.Errorf("saving images %w", err)
	}
```

stevenh (changes requested):

neersighted (changes requested):
Beyond @stevenh's review which I concur with, I'd like to see this PR use string-specs for the platform (like the rest of the TC/Docker API) and use `github.com/containerd/platforms.ParseAll()` to transform them.

mdelapenya:
For reference, we addressed this in the go-sdk too: https://pkg.go.dev/github.com/docker/go-sdk/image#SaveOption

image.go
LaurentGoderre: @stevenh doing it this way ties it to the Docker provider though no? This makes it harder to expand it to the other potential image provider.

LaurentGoderre:
@neersighted implemented your suggestion!

image.go
stevenh: As the internal is private we can still make any needed changes

modules/k3s/k3s.go
```diff
@@ -210,7 +213,7 @@ func (c *K3sContainer) LoadImages(ctx context.Context, images ...string) error {
 		_ = os.Remove(imagesTar.Name())
 	}()
 
-	err = provider.SaveImages(context.Background(), imagesTar.Name(), images...)
+	err = provider.SaveImagesWithOps(context.Background(), imagesTar.Name(), images, opts...)
```
mdelapenya:
```suggestion
	err = provider.SaveImagesWithOpts(context.Background(), imagesTar.Name(), images, opts...)
```

docker.go
```diff
@@ -1786,6 +1786,19 @@ func (p *DockerProvider) ListImages(ctx context.Context) ([]ImageInfo, error) {
 
 // SaveImages exports a list of images as an uncompressed tar
 func (p *DockerProvider) SaveImages(ctx context.Context, output string, images ...string) error {
+	return p.SaveImagesWithOps(ctx, output, images)
+}
+
+// SaveImagesWithOpts exports a list of images as an uncompressed tar, passiong options to the provider
+func (p *DockerProvider) SaveImagesWithOps(ctx context.Context, output string, images []string, opts ...SaveImageOption) error {
```
mdelapenya:
```suggestion
	return p.SaveImagesWithOpts(ctx, output, images)
}

// SaveImagesWithOpts exports a list of images as an uncompressed tar, passing options to the provider
func (p *DockerProvider) SaveImagesWithOpts(ctx context.Context, output string, images []string, opts ...SaveImageOption) error {
```

image.go
```diff
@@ -10,9 +12,16 @@ type ImageInfo struct {
...
+	dockerSaveOpts []client.ImageSaveOption
+}
+
+type SaveImageOption func(*saveImageOptions) error
+
 // ImageProvider allows manipulating images
 type ImageProvider interface {
 	ListImages(context.Context) ([]ImageInfo, error)
 	SaveImages(context.Context, string, ...string) error
+	SaveImagesWithOps(context.Context, string, []string, ...SaveImageOption) error
```
mdelapenya:
```suggestion
	SaveImagesWithOpts(context.Context, string, []string, ...SaveImageOption) error
```

image_test.go
```diff
@@ -71,3 +74,38 @@ func TestSaveImages(t *testing.T) {
...
+
+	p, err := platforms.ParseAll([]string{"linux/amd64"})
+	require.NoError(t, err)
+
+	ctr, err := provider.CreateContainer(context.Background(), req)
+	CleanupContainer(t, ctr)
+	require.NoErrorf(t, err, "creating test container")
+
+	output := filepath.Join(t.TempDir(), "images.tar")
+	err = provider.SaveImagesWithOps(
```
mdelapenya:
```suggestion
	err = provider.SaveImagesWithOpts(
```

image_test.go
```diff
@@ -51,7 +53,8 @@ func TestSaveImages(t *testing.T) {
 	require.NoErrorf(t, err, "failed to get provider")
 
 	defer func() {
-		_ = provider.Close()
+		err = provider.Close()
```
mdelapenya, line 56: praise: good catch!

mdelapenya (dismissed):
I added a few comments, just typos.

Other than that, LGTM, thanks for this! And sorry for the late review, my fault 🙏

PR #454, opened 2022-06-13
fix: strip stream headers from logs

funvit:
Closes #365 

`DockerContainer` func `Logs` now strips header for each line.

This fix cannot be used with TTY mode.

Why strip header?  Due to setting TTY breaks some tests (ex: Test_ShouldRecognizeLogTypes) ...

PR #3226, opened 2025-07-02
fix: strip headers from logs using log stream specification 

LaurentGoderre:
## What does this PR do?

According to the log stream specification, the last 4 bytes of the header indicate the length of the log frame. There is an issue when a log message is exactly 10 character long, the header ends with `\00\00\00\10`. The current implementation uses `ReadLine` so it doesn't correctly read the header which results in the header being output. This is what the it looks like with the new test:

```
Error:      	Not equal: 
				expected: "abcdefghi\nfoo"
				actual  : "\x01\x00\x00\x00\x00\x00\x00\ni\nfoo"
				
				Diff:
				--- Expected
				+++ Actual
				@@ -1,2 +1,3 @@
				-abcdefghi
				+
				+i
					foo
Test:       	TestContainerLogsShouldBeWithoutStreamHeader
```

## Related issues

Relates to #454

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -368,7 +368,6 @@ func (c *DockerContainer) inspectRawContainer(ctx context.Context) (*container.I
 // Logs will fetch both STDOUT and STDERR from the current container. Returns a
 // ReadCloser and leaves it up to the caller to extract what it wants.
 func (c *DockerContainer) Logs(ctx context.Context) (io.ReadCloser, error) {
-
 	options := container.LogsOptions{
 		ShowStdout: true,
 		ShowStderr: true,
--- a/docker.go
+++ b/docker.go
@@ -5,6 +5,7 @@ import (
 	"bufio"
 	"context"
 	"encoding/base64"
+	"encoding/binary"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -367,7 +368,6 @@ func (c *DockerContainer) inspectRawContainer(ctx context.Context) (*container.I
 // Logs will fetch both STDOUT and STDERR from the current container. Returns a
 // ReadCloser and leaves it up to the caller to extract what it wants.
 func (c *DockerContainer) Logs(ctx context.Context) (io.ReadCloser, error) {
-	const streamHeaderSize = 8
 
 	options := container.LogsOptions{
 		ShowStdout: true,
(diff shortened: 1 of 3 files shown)


docker.go
```diff
@@ -384,34 +385,23 @@ func (c *DockerContainer) Logs(ctx context.Context) (io.ReadCloser, error) {
...
-				line = line[streamHeaderSize:] // trim stream header
-				lineStarted = false
-			}
-			if !isPrefix {
-				lineStarted = true
-			}
-
-			_, errW := pw.Write(line)
-			if errW != nil {
+			header := make([]byte, 8)
```
stevenh: suggestion: use the streamHeaderSize constant here, which addresses the compile failure too.

Also move the buffer out of the loop as this could cause lots of GC.

docker.go
```diff
@@ -384,34 +385,23 @@ func (c *DockerContainer) Logs(ctx context.Context) (io.ReadCloser, error) {
...
-
+			frameSizeBuf := header[4:]
+			frameSize := binary.BigEndian.Uint32(frameSizeBuf)
+			frame := make([]byte, frameSize)
+			_, err = r.Read(frame)
 			if err != nil {
 				_ = pw.CloseWithError(err)
 				return
 			}
+			pw.Write(frame)
```
stevenh: bug: no error check on pw.Write.

docker.go
```diff
@@ -384,34 +385,23 @@ func (c *DockerContainer) Logs(ctx context.Context) (io.ReadCloser, error) {
...
-				_, errW := pw.Write([]byte("\n"))
-				if errW != nil {
-					return
-				}
-			}
-
+			frameSizeBuf := header[4:]
+			frameSize := binary.BigEndian.Uint32(frameSizeBuf)
+			frame := make([]byte, frameSize)
+			_, err = r.Read(frame)
```
stevenh: suggestion: no need for a fixed buffer or the temporary variable
```suggestion
			frameSize := binary.BigEndian.Uint32(header[4:])
			if _, err := io.CopyN(pw, r, frameSize); err != nil {
				pw.CloseWithError(err) //nolint: errcheck
				return
			}
```

Will need the code below removing too.

docker.go
LaurentGoderre: Done

LaurentGoderre:
@stevenh all fixed! Thanks!

stevenh (approved):
LGTM

stevenh:
Looks like it might have caused some regressions @LaurentGoderre do you want to have a look, as it might also be the tests need correcting?

mdelapenya:
@LaurentGoderre thanks for the fix! I've taken a deep look at it, and ported it to the in-progress docker/go-sdk (see https://github.com/docker/go-sdk/pull/56/files). If you take a look, that PR also addresses that the container is using TTY, and a more robust error handling. Could you apply that here too?

Also, regarding the build from dockerfile errors, I'm going to debug them locally using your PR to understand why they are happening now.

In any case, thank you so much for contributing to tc-go!

LaurentGoderre:
@mdelapenya I think the build failures are because the previous implementation appended new lines automatically and it seems like it may have auto-added one that wasn't from the actual logs.

logconsumer_test.go
```diff
@@ -378,7 +378,32 @@ func TestContainerLogsShouldBeWithoutStreamHeader(t *testing.T) {
...
+	})
+	CleanupContainer(t, ctr)
+	require.NoError(t, err)
+
+	r, err := ctr.Logs(ctx)
+	require.NoError(t, err)
+	defer r.Close()
+	b, err := io.ReadAll(r)
+	require.NoError(t, err)
+	assert.Equal(t, "abcdefghi\r\nfoo", strings.TrimSpace(string(b)))
```
LaurentGoderre: Not sure why a `\r` is present here

mdelapenya: No idea either. Asking Cursor about that:
> When TTY mode is enabled (cfg.Tty = true), Docker simulates a terminal environment. Terminals traditionally use \r\n (CRLF) line endings rather than just \n (LF).
The behavior difference is:
Normal container (no TTY): Uses \n line endings → "abcdefghi\nfoo"
TTY-enabled container: Uses \r\n line endings → "abcdefghi\r\nfoo"
This is why the test specifically checks for \r\n - it's validating that when TTY mode is enabled, the container logs correctly preserve the terminal-style line endings that would be expected in a real terminal session.
The TTY mode is commonly used when you want to run interactive commands or when applications expect to be running in a terminal environment (like when using docker run -it).

LaurentGoderre: I read somewhere else that it was expected behavior so all good for me

mdelapenya (dismissed):
LGTM, thanks!

logconsumer_test.go
mdelapenya:
```suggestion
	require.Equal(t, "abcdefghi\r\nfoo", strings.TrimSpace(string(b)))
```

logconsumer_test.go
```diff
@@ -378,7 +378,32 @@ func TestContainerLogsShouldBeWithoutStreamHeader(t *testing.T) {
 	defer r.Close()
 	b, err := io.ReadAll(r)
 	require.NoError(t, err)
-	assert.Equal(t, "0", strings.TrimSpace(string(b)))
+	assert.Equal(t, "abcdefghi\nfoo", strings.TrimSpace(string(b)))
```
mdelapenya:
```suggestion
	require.Equal(t, "abcdefghi\nfoo", strings.TrimSpace(string(b)))
```

mdelapenya (commented):
last minute changes, please consider using require instead of require. We are slowly moving towards that, although not documented anywhere.

PS: I'm committing the suggestions myself to avoid you doing the rework 🙏

mdelapenya (approved):
LGTM

mdelapenya:
Thanks @LaurentGoderre for your contribution! This was a very subtle bug 🐞 Good job!

Issue #2752, opened 2024-08-27
[Bug]: Failed to get image auth for https://index.docker.io/v1/. [...] Error is: credentials not found in native keychain


simon-connektica:
### Testcontainers version

0.33.0

### Using the latest Testcontainers version?

Yes

### Host OS

Linux

### Host arch

x86_64

### Go version

1.22.3

### Docker version

```shell
Client: Docker Engine - Community
 Version:           27.1.2
 API version:       1.46
 Go version:        go1.21.13
 Git commit:        d01f264
 Built:             Mon Aug 12 11:50:58 2024
 OS/Arch:           linux/amd64
 Context:           default

Server: Docker Engine - Community
 Engine:
  Version:          27.1.2
  API version:      1.46 (minimum version 1.24)
  Go version:       go1.21.13
  Git commit:       f9522e5
  Built:            Mon Aug 12 11:50:58 2024
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.7.20
  GitCommit:        8fc6bcff51318944179630522a095cc9dbf9f353
 runc:
  Version:          1.1.13
  GitCommit:        v1.1.13-0-g58aa920
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### Docker info

```shell
Client: Docker Engine - Community
 Version:    27.1.2
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.16.2
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.29.1
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 20
 Server Version: 27.1.2
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
[...]

PR #3228, opened 2025-07-03
chore: clarify image auth warning message for public images

diegolatorre16:
- chore: clarify image auth warning message for public images

## What does this PR do?

This PR updates the log message shown when Docker image authentication credentials are not found during an image pull.
The new message clarifies that missing credentials is expected for public images and is not a critical error.
No functional or behavioral changes are introduced; only the log output is improved for clarity.

## Why is it important?

The previous log message used the word "Failed", which could be interpreted as a critical error by users, even though missing credentials is normal for public images.
This change helps avoid confusion and reduces unnecessary concern for users pulling public images, improving the developer experience and log readability.

## Related issues

- Relates #2752


stevenh (approved):
Nice one have seen this and thought the same on many occasions, thanks for the PR

PR #3230, opened 2025-07-11
feat: add Solace pubsub+ module

### LINKED ISSUE #3229: [Enhancement]: Add solace pubsub+ module
### Proposal

Hello, 
I added a small [solace pubsub+](https://testcontainers.com/modules/solace/) module for my own use, it currently doesn't have all the advance features but enough for creating queue, subscribing topics as well as making the solace specific configurations when spinning up the container. 

Raised a quick PR, let me know if everything is crafted as expected in the PR.

unicod3:
## What does this PR do?

Adds a new module for [Solace pubsub+](https://testcontainers.com/modules/solace/)

## Why is it important?

Exposes a ready-to-use module for Solace Pubsub+

## Related issues

- Closes #3229


coverage.out
```diff
@@ -0,0 +1,296 @@
+mode: atomic
```
mdelapenya: bug: please not include this file into the commit

modules/solace/coverage.out
```diff
@@ -0,0 +1,69 @@
+mode: atomic
```
mdelapenya: bug: please not add this coverage file to the commit

modules/solace/go.mod
```diff
@@ -0,0 +1,68 @@
+module github.com/testcontainers/testcontainers-go/modules/solace
+
+go 1.23.0
+
+require (
+	github.com/docker/docker v28.2.2+incompatible
+	github.com/docker/go-connections v0.5.0
+	github.com/stretchr/testify v1.10.0
+	github.com/testcontainers/testcontainers-go v0.37.0
```
mdelapenya: Suggestion: we released a new version yesterday. Please run mod tidy after that.
```suggestion
	github.com/testcontainers/testcontainers-go v0.38.0
```

modules/solace/solace.go
```diff
@@ -0,0 +1,291 @@
...
+	Username       string
+	Password       string
+	Vpn            string
+	queues         map[string][]string // queueName -> topics
+	WithClientCert bool
+	Image          string
+	ctx            context.Context
+	exposedPorts   []string
+	envVars        map[string]string
+	shmSize        int64
```
mdelapenya: suggestion: I'd say all the properties must be private, adding accessor container methods if needed: e.g. `username` would have a `Username() string` accessor.

Else, client could could change the internal state of the container at runtime

modules/solace/solace.go
```diff
@@ -0,0 +1,291 @@
...
+}
+
+// WithCredentials sets the client credentials (username, password)
+func (s *SolaceContainer) WithCredentials(username, password string) *SolaceContainer {
```
mdelapenya: issue: if the options are added to the container type, then consumers of this module would need to create a struct and run it. If you check the rest of the modules, we have a completely different pattern: the options customise the container request, or an intermediate options struct, and then the Run function applies those values.

As an example, in https://github.com/testcontainers/testcontainers-go/pull/3230/files#diff-dd60fb936e6c398b98448b75da35178a52f2de76bd7277a3c3de2c5043fd0775R20 you needed a `New` function to create the container, and then run it. For the user experience, we think it's better to hide the implementation details of building the container request, and instead just expose the Run function and the functional options that configure the container.

Same pattern would apply to all the options added here. I recommend you taking a look at #3232, which adds a new module using options.

Does it sound reasonable to you?

modules/solace/solace.go
```diff
@@ -0,0 +1,291 @@
...
+}
+
+const (
+	DefaultImage = "solace/solace-pubsub-standard:latest"
+	DefaultVpn   = "default"
+	DefaultUser  = "root"
+	DefaultPass  = "password"
+)
+
+func NewSolaceContainer(ctx context.Context, image string) *SolaceContainer {
```
mdelapenya: issue: as per my comment above, this constructor function is not needed anymore.

modules/solace/solace.go
```diff
@@ -0,0 +1,291 @@
...
+		}
+
+		// Wait before next attempt
+		time.Sleep(intervalSeconds * time.Second)
+	}
+
+	return fmt.Errorf("timeout waiting for Solace broker to become active after %d seconds", maxAttempts)
+}
+
+func (s *SolaceContainer) Run(ctx context.Context) error {
```
mdelapenya: issue: this pattern breaks the consistency with the rest of the library modules. Please check my previous comment.

This function should initialise a container request, and apply the passed in options. Also not being attached to the container type but as a public function in the package.

modules/solace/solace.go
```diff
@@ -0,0 +1,291 @@
...
+
+	// --- CLI script generation for queue/topic config ---
+	// Reference: https://docs.solace.com/Admin-Ref/CLI-Reference/VMR_CLI_Commands.html
+	var cliScript string
+	if len(s.queues) > 0 {
+		cliScript += "enable\nconfigure\n"
+
+		// Create VPN if not default
+		if s.Vpn != DefaultVpn {
+			cliScript += fmt.Sprintf("create message-vpn %s\n", s.Vpn)
```
mdelapenya: suggestion: use a multi-line string with backticks, maybe? Or a template file that is copied. See redpanda module for a similar approach.

modules/solace/solace.go
```diff
@@ -0,0 +1,291 @@
...
+		Started:          true,
+	})
+	if err != nil {
+		return err
+	}
+
+	s.Container = container // assign before using s.Container
+
+	// Wait for Solace broker to be fully active before configuring
+	if err := s.waitForSolaceActive(ctx); err != nil {
```
mdelapenya: suggestion: you should use the wait package in the testcontainers core project, to add it in the WaitingFor field of the container request

modules/solace/solace.go
```diff
@@ -0,0 +1,291 @@
...
+		if code != 0 {
+			return fmt.Errorf("CLI script execution failed with exit code %d: %s", code, output)
+		}
+	}
+
+	return nil
+}
+
+// BrokerURLFor returns the origin URL for a given service
+func (s *SolaceContainer) BrokerURLFor(service Service) (string, error) {
```
mdelapenya: praise: this is a proper example of a container method

modules/solace/solace.go
```diff
@@ -0,0 +1,291 @@
...
+	host, err := s.Container.Host(s.ctx)
+	if err != nil {
+		return "", err
+	}
+	port, err := s.Container.MappedPort(s.ctx, nat.Port(fmt.Sprintf("%d/tcp", service.Port)))
+	if err != nil {
+		return "", err
+	}
+	return fmt.Sprintf("%s://%s:%s", service.Protocol, host, port.Port()), nil
```
mdelapenya: issue: use ctr.PortEndpoint instead. See the rest of the modules that calculate an endpoint/URL

modules/solace/solace.go
```diff
@@ -0,0 +1,291 @@
...
+}
+
+
+func (s *SolaceContainer) Terminate() error {
```
mdelapenya: issue: we won't need this, as the testcontainers.Container type should be embedded into this module's container type

modules/solace/solace.go
```diff
@@ -0,0 +1,291 @@
...
+
+	"github.com/docker/docker/api/types/container"
+	"github.com/docker/go-connections/nat"
+
+	"github.com/testcontainers/testcontainers-go"
+	"github.com/testcontainers/testcontainers-go/wait"
+)
+
+type SolaceContainer struct {
+	Container      testcontainers.Container
```
mdelapenya: issue: for consistency across the modules, please use the same pattern, in which the `testcontainers.Container` type is embedded into the module's type.

mdelapenya (changes requested):
@unicod3 thanks for submitting this PR. I did an initial review and found some issues that need to be addressed, mostly related to the consistent patterns we apply when building the modules.

TL;DR;
- no constructor for the Container type
- options must be exposed at the package level, not added to the container type
- Run function using functional opts

In any case, we can work with you in making it consistent, and merge it into the project. Thank you so much for your contribution

coverage.out
unicod3: done

modules/solace/coverage.out
unicod3: done

modules/solace/go.mod
unicod3: done

modules/solace/solace.go
unicod3: done.

modules/solace/solace.go
unicod3: done

modules/solace/solace.go
unicod3: done

modules/solace/solace.go
unicod3: @mdelapenya thank you for the detailed review, that make sense indeed,  I'll update the PR accordingly.

modules/solace/solace.go
unicod3: done, encapsulated them.

modules/solace/solace.go
unicod3: burned with fire 🔥

modules/solace/solace.go
unicod3: updated to align with the rest of the modules.

modules/solace/solace.go
unicod3: this is done as well with [87a7297](https://github.com/testcontainers/testcontainers-go/pull/3230/commits/87a7297704be209e607270e88debfef942042fe5)

modules/solace/solace.go
unicod3: done with [939236d](https://github.com/testcontainers/testcontainers-go/pull/3230/commits/939236d93b2f019c9a8da64af4b108b7ddd622ea)

modules/solace/solace.go
unicod3: addressed that, moved into a separate file with explicit methods

unicod3:
> @unicod3 thanks for submitting this PR. I did an initial review and found some issues that need to be addressed, mostly related to the consistent patterns we apply when building the modules.
> 
> TL;DR;
> 
>     * no constructor for the Container type
> 
>     * options must be exposed at the package level, not added to the container type
> 
>     * Run function using functional opts
> 
> 
> In any case, we can work with you in making it consistent, and merge it into the project. Thank you so much for your contribution

@mdelapenya  I've addressed them all, let me know if I missed anything, thank you for the review.

modules/solace/options.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/testcontainers/testcontainers-go"
+)
+
+type options struct {
+	vpn          string
+	username     string
+	password     string
+	exposedPorts []string
+	queues       map[string][]string // queueName -> topics
+	envVars      map[string]string
```
mdelapenya: suggestion: I think we do not need the envVars here, as we have a `testcontainers.WithEnv` option in the core.

modules/solace/service.go
```diff
@@ -0,0 +1,47 @@
+package solace
+
+type Service struct {
```
mdelapenya, line 4: suggestion: let's add comments to the types 🙏

modules/solace/solace.go
```diff
@@ -0,0 +1,125 @@
...
+
+	"github.com/testcontainers/testcontainers-go"
+	"github.com/testcontainers/testcontainers-go/wait"
+)
+
+const (
+	defaultVpn = "default"
+)
+
+type SolaceContainer struct {
```
mdelapenya: suggestion: let's use a simpler name. I know there are other modules using the "old" name (including the module name), but we are slowly moving to a place where the package/module adds the context: `solace.Container` reads better than `solace.SolaceContainer`.

```suggestion
type Container struct {
```

modules/solace/solace.go
```diff
@@ -0,0 +1,125 @@
...
+	settings := defaultOptions()
+	for _, opt := range opts {
+		if apply, ok := opt.(Option); ok {
+			if err := apply(&settings); err != nil {
+				return nil, err
+			}
+		}
+	}
+
+	req := testcontainers.ContainerRequest{
```
mdelapenya: suggestion: we are slowly moving to the pattern for directly use the Run function, instead of the ContainerRequest. See my last addition in the `dockermcpgateway` module [here](https://github.com/testcontainers/testcontainers-go/blob/main/modules/dockermcpgateway/dockermcpgateway.go#L76-L79). If you feel this can be done in a follow up, just let me know and I can do it afterwards. I'd prefer start establishing this pattern more and more from now on (I need to update the module generator 🤔 )

mdelapenya (commented):
Looks pretty good, thanks for your hard work! Just a few comments, and we are done 😊

modules/solace/solace.go
unicod3: done

modules/solace/service.go
unicod3: done

modules/solace/options.go
unicod3: done

modules/solace/solace.go
unicod3: @mdelapenya tried to switch to the Run method as you suggested, could you take a look if everything looks alright?

modules/solace/examples_test.go
```diff
@@ -0,0 +1,210 @@
...
+
+	"github.com/testcontainers/testcontainers-go"
+	sc "github.com/testcontainers/testcontainers-go/modules/solace"
+)
+
+func ExampleRun() {
+	ctx := context.Background()
+	ctr, err := sc.Run(ctx, "solace-pubsub-standard:latest",
+		sc.WithCredentials("admin", "admin"),
+		sc.WithExposedPorts("5672/tcp", "8080/tcp"),
```
mdelapenya: suggestion: I think I now understand the purpose of the Services type. Wdyt if we add a `solace.WithServices(srv ...Service)` option that:
1. adds an exposed port for the given service
2. adds a wait strategy for the listening port of the given service

Then, we can simplify the options struct, removing the ports. In the end, keeping it would mean that users could set any port, but I guess the intention is to just expose a Service's port. Is this assumption correct?

unicod3: yes your understanding is correct, the main idea is to let user chose which solace ports (aka service in solace) to be exposed to world (but they are limited), I like your idea of encapsulating them, will take a look at it 👍

modules/solace/script.go
```diff
@@ -0,0 +1,100 @@
...
+
+	// Configure queues and topic subscriptions
+	script += generateQueueConfig(settings.queues, settings.vpn)
+
+	return script
+}
+
+// generateVPNConfig creates the CLI commands for setting up a custom VPN
+func generateVPNConfig(vpn string) string {
+	return fmt.Sprintf(`create message-vpn %s
```
stevenh: question: would all of these be better done with a single template instead of multiple functions and internal logic?

I'm wondering if a WithScript might be the logical conclusion?

modules/solace/solace.go
```diff
@@ -0,0 +1,124 @@
...
+		return c, fmt.Errorf("generic container: %w", err)
+	}
+
+	// Generate CLI script for queue/topic configuration
+	cliScript := generateCLIScript(settings)
+	if cliScript == "" {
+		return c, nil
+	}
+	// Write the CLI script to a temp file
+	tmpFile, err := os.CreateTemp("", "solace-queue-setup-*.cli")
```
stevenh: suggestion: use a reader to avoid temporary file

modules/solace/examples_test.go
mdelapenya: Thanks for being so supportive 🙇 Looking forward for reviewing it, as we are really close to merge

modules/solace/solace.go
```diff
@@ -0,0 +1,124 @@
...
+
+func (c *Container) Username() string {
+	return c.settings.username
+}
+
+func (c *Container) Password() string {
+	return c.settings.password
+}
+
+func (c *Container) Vpn() string {
```
stevenh: suggestion: VPN instead of Vpn as it's an acronym

modules/solace/solace.go
```diff
@@ -0,0 +1,124 @@
...
+	return c, nil
+}
+
+// BrokerURLFor returns the origin URL for a given service
+func (s *Container) BrokerURLFor(ctx context.Context, service Service) (string, error) {
+	p := nat.Port(fmt.Sprintf("%d/tcp", service.Port))
+	return s.PortEndpoint(ctx, p, service.Protocol)
+}
+
+func (c *Container) Username() string {
```
stevenh, line 119: Suggestion: all methods should be documented

stevenh (changes requested):

modules/solace/solace.go
unicod3: done.

modules/solace/solace.go
unicod3: done

modules/solace/solace.go
unicod3: Thanks for the comment. I noticed that there is a `CopyToContainer` method, which allows me to avoid creating a temporary file. Let me know if that can be further simplified.

modules/solace/script.go
unicod3: If I understand you correctly, you suggest introducing a `WithScript` method where module users can provide their solace script themselves? 

That might be good flexibility indeed; however, I don't think the users would want to know/write the Solace script to just introduce a basic queue/subscription pair as it can be frustrating to get it working, so maintaining a well tested logic inside the module itself wouldn't hurt imho, that logic might also provide an example/starting point to other devs that want to contribute this solace module.

That being said, I'd be happy to hear more on that as myself had a hard time to get this script working for solace 😜

stevenh: That sounds sensible, do you think using text/template internally would be good move?

modules/solace/examples_test.go
unicod3: The change is done, thanks for the suggestion, it became cleaner with the change.

modules/solace/script.go
unicod3: I'll take a look at the template 👍

modules/solace/solace.go
```diff
@@ -35,14 +35,32 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
...
 	}
 
+	// Build exposed ports from services
+	var exposedPorts []string
+	for _, service := range settings.services {
+		exposedPorts = append(exposedPorts, fmt.Sprintf("%d/tcp", service.Port))
+	}
+
+	// Build wait strategies - combine the default wait strategy with port-based wait strategies for each service
+	var waitStrategies []wait.Strategy
```
mdelapenya: suggestion: pre-allocate the size of the slice, as we know it's len(settings.services + 1)

modules/solace/solace.go
```diff
@@ -35,14 +35,32 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
...
+
+	// Primary wait strategy for Solace to be ready
+	waitStrategies = append(waitStrategies, wait.ForExec([]string{"grep", "-q", "Primary Virtual Router is now active", "/usr/sw/jail/logs/system.log"}).
+		WithStartupTimeout(1*time.Minute).
+		WithPollInterval(1*time.Second))
+
+	// Add port-based wait strategies for each service
+	for _, service := range settings.services {
```
mdelapenya: suggestion: I think we can do this in the above for loop, just once.

modules/solace/options.go
```diff
@@ -44,10 +38,11 @@ func (o Option) Customize(*testcontainers.GenericContainerRequest) error {
 	return nil
 }
 
-// WithExposedPorts allows adding extra exposed ports
-func WithExposedPorts(ports ...string) Option {
+// WithServices configures the services to be exposed with their wait strategies
+func WithServices(srv ...Service) Option {
 	return func(o *options) error {
-		o.exposedPorts = append(o.exposedPorts, ports...)
+		// Clear existing services and use only the specified ones
```
mdelapenya, line 46: bug: if the passed slice is empty, then the container will not expose any service. I'd return an error here for that case

mdelapenya (commented):
Waiting for the template changes proposed by @stevenh, LGTM.

I added some comments too. But we are in the good track, thank you so much for your work

modules/solace/script.go
unicod3: Implemented the template approach, I _heavily inspired_ from the rabbitmq module where they used `embed` package. 
Let me know if that looks alright. 
https://github.com/testcontainers/testcontainers-go/blob/f5c13097653b3fbe4bdd6b980804bacc4fb20b88/modules/rabbitmq/rabbitmq.go#L204

fyi @mdelapenya

modules/solace/solace.go
unicod3: done.

modules/solace/options.go
unicod3: yeah, you are right, updated this part and covered with a test

modules/solace/solace.go
unicod3: done

modules/solace/options_test.go
```diff
@@ -0,0 +1,337 @@
...
+import (
+	"reflect"
+	"testing"
+)
+
+func TestDefaultOptions(t *testing.T) {
+	opts := defaultOptions()
+
+	// Test default values
+	if opts.vpn != "default" {
```
mdelapenya: suggestion: prefer testify's require over stdlib. We encourage this change mostly because of its readability.

mdelapenya (approved):
LGTM! Even though I think we need to use testify in the options_test file, I'm approving this PR before hand.

Thank you for all your hard work during the review 🙇

modules/solace/options_test.go
unicod3: done.

unicod3:
> LGTM! Even though I think we need to use testify in the options_test file, I'm approving this PR before hand.
> 
> Thank you for all your hard work during the review 🙇

You're welcome, enjoyed the collaboration. In my last commit, I also addressed linter issues, there is a [CodeQL check](https://github.com/testcontainers/testcontainers-go/pull/3230/checks?check_run_id=47005453438) failing however I am not really sure if that's because of my branch.

modules/solace/options_test.go
```diff
@@ -1,35 +1,24 @@
 package solace
 
 import (
-	"reflect"
 	"testing"
+
+	"github.com/stretchr/testify/assert"
```
mdelapenya: suggestion: let's use require everywhere; if an assertion fails, then we probably do not want to continue the execution

unicod3: done

unicod3:
@mdelapenya just fixed the failing test issue fyi

mdelapenya:
@unicod3 I added a commit on top of yours, just to refine the docs. PLMK what you think. We can merge this PR right after that.

unicod3:
> @unicod3 I added a commit on top of yours, just to refine the docs. PLMK what you think. We can merge this PR right after that.

LGTM.

modules/solace/examples_test.go
```diff
@@ -0,0 +1,212 @@
...
+
+	// Create topic resource
+	topic := resource.TopicOf(topicName)
+
+	// Publish the message
+	if err := directPublisher.Publish(message, topic); err != nil {
+		return fmt.Errorf("failed to publish message: %w", err)
+	}
+
+	fmt.Printf("Published message to topic: %s\n", topicName)
```
mdelapenya, line 160: ~~bug: let's not write to stdout out of a testable example. I'm resolving this locally, will push an update asap~~

Nevermind, I've just checked it WAS in a testable example... 🤦

modules/solace/examples_test.go
```diff
@@ -0,0 +1,214 @@
...
+	}()
+
+	// Test message publishing
+	err = publishTestMessage(messagingService, topicName)
+	if err != nil {
+		return fmt.Errorf("failed to publish message: %w", err)
+	}
+
+	// Wait a moment for message to be delivered to queue
+	time.Sleep(1 * time.Second)
```
mdelapenya: question: is there any way of removing this time.Sleep? This value could change depending on the host running the tests

unicod3: I think i can use a channel to wait, would be more consistent

unicod3: After examining the test in detail, I noticed that `consumeTestMessage` already implements a channel to listen to the message. I simply removed the sleep, which was initially added for debugging purposes.

mdelapenya: good catch!

mdelapenya:
@unicod3 thanks for your contribution! I'm adding it to the modules catalog right now: https://github.com/testcontainers/community-module-registry/pull/147

PR #3241, opened 2025-07-24
feat(options): add WithProvider

dfioravanti:
## What does this PR do?
The generic container allows to set a provider for a container but this is not exposed via the options. This commit adds this capability so that it is possible to set the provider to Podman on MacOS when the auto detection fails

## Why is it important?
It is part of what allows using Ryuk on MacOS. One still has to set the VM to run rootless but without this change even with root it does not work because the wrong network is detected

## Related issues

 - Relates [#2264](https://github.com/testcontainers/testcontainers-go/issues/2264)


mdelapenya:
@dfioravanti thanks for this. I mentioned that we are working on the Docker SDK, so the concept of Provider will be automatically discarded with that (the SDK simply uses the current docker context to get a docker client), so the configuration would be much much easier, so I'd say that this PR won't be not needed.

On the other hand, thanks to the functional options pattern, this new option is harmless, so it's fine for me adding it and let users of the library choose.

Said that, could you please check if the podman docs are up-to-date with these changes? 🙏

dfioravanti:
@mdelapenya thanks for been open to merge this

> Said that, could you please check if the podman docs are up-to-date with these changes? 🙏

I updated the documentation and added the option to the list of options. Does it work now?

docs/features/common_functional_options.md
```diff
@@ -437,6 +437,13 @@ ctr, err := mymodule.Run(ctx, "docker.io/myservice:1.2.3",
 
 If you need to prevent the container from being started after creation, you can use the `testcontainers.WithNoStart` option.
 
+##### WithProvider
+
+- Since <a href="https://github.com/testcontainers/testcontainers-go/releases/tag/v0.39.0"><span class="tc-version">:material-tag: v0.39.0</span></a>
```
mdelapenya: suggestion: the release process will change this to the right release. Imagine we release a patch, or even a major!
```suggestion
- Not available until the next release <a href="https://github.com/testcontainers/testcontainers-go"><span class="tc-version">:material-tag: main</span></a>
```

docs/system_requirements/using_podman.md
```diff
@@ -44,6 +44,44 @@ The reaper container needs to connect to the docker daemon to reap containers, s
 > systemctl --user start podman.socket
 ```
 
+## MacOS
+
+Currently, in MacOS the autodetection of podman does not work as intended, which leads to Ryku failing at boot-up.
+In order to use testcontainer then either
+1. Disable Ryku (not recommended): see [here](../features/garbage_collector.md#ryuk)
+2. If you want to use Ryku then you need
```
mdelapenya:
```suggestion
Currently, in MacOS the autodetection of podman does not work as intended, which leads to Ryuk failing at boot-up.
In order to use Testcontainers then either
1. Disable Ryuk (not recommended): see [here](../features/garbage_collector.md#ryuk)
2. If you want to use Ryuk then you need
```

mdelapenya (commented):
Just a couple of refinements, other than that, LGTM

dfioravanti:
@mdelapenya I fixed the two problems, thanks for the review

docs/features/common_functional_options_list.md
```diff
@@ -60,6 +60,8 @@ The following options are exposed by the `testcontainers` package.
 - [`CustomizeRequest`](/features/creating_container/#customizerequest) Since <a href="https://github.com/testcontainers/testcontainers-go/releases/tag/v0.20.0"><span class="tc-version">:material-tag: v0.20.0</span></a>
 - [`WithName`](/features/creating_container/#withname) Since <a href="https://github.com/testcontainers/testcontainers-go/releases/tag/v0.38.0"><span class="tc-version">:material-tag: v0.38.0</span></a>
 - [`WithNoStart`](/features/creating_container/#withnostart) Since <a href="https://github.com/testcontainers/testcontainers-go/releases/tag/v0.38.0"><span class="tc-version">:material-tag: v0.38.0</span></a>
+- [`WithProvider`](/features/creating_container/#withprovider) Since <a href="https://github.com/testcontainers/testcontainers-go/releases/tag/v0.39.0"><span class="tc-version">:material-tag: v0.39.0</span></a>
```
mdelapenya: Sorry, I did not find it earlier 😢 
```suggestion
- [`WithProvider`](/features/creating_container/#withprovider) Not available until the next release <a href="https://github.com/testcontainers/testcontainers-go"><span class="tc-version">:material-tag: main</span></a>
```

mdelapenya (approved):
LGTM, thanks!

docs/features/common_functional_options.md
```diff
@@ -437,6 +437,13 @@ ctr, err := mymodule.Run(ctx, "docker.io/myservice:1.2.3",
 
 If you need to prevent the container from being started after creation, you can use the `testcontainers.WithNoStart` option.
 
+##### WithProvider
```
mdelapenya, line 440: question: wdyt if we make this even more explicit, naming the option as `WithPodmanProvider`? Given Docker is the default runtime, podman users just pass the new option when needed. Thoughts?

dfioravanti: Actually I like that you can choose because for example I have a function that detects if the test is running on Mac or linux and uses the right provider. I guess I could do it even with your approach but it gets a bit more convoluted. Ultimately I do not care but I would be happy if it stays as it is

mdelapenya (approved):

PR #3243, opened 2025-07-25
pipenv update requests

piotrkowalczuk:
## What does this PR do?

Update python dependencies.

```shell
pipenv update requests
// manually removed requests from Pipfile
pipenv lock
```

## Why is it important?

image

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -5,7 +5,7 @@
         },
         "pipfile-spec": 6,
         "requires": {
-            "python_version": "3.8"
+            "python_version": "3.13"
         },
         "sources": [
             {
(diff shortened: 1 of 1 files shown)


mdelapenya:
Hi @piotrkowalczuk thanks for taking the time to check this. I'd like to know the motivation on this PR, as we recently merged #3204, which seems related.

Could you elaborate a bit more on the need for this? 🙏

piotrkowalczuk:
Hi @mdelapenya, possibly motivation is the same. To remove security vulnerability and silence security scanner. Since the package was already applied I'll close this PR. When it is possible to see https://github.com/testcontainers/testcontainers-go/pull/3204 being part of a release?

mdelapenya:
I'm not concerned about that Pipfile, as it's used just for the docs site, which is deployed on Netlify. And the docs site is deployed with every merge to main.

piotrkowalczuk:
I see. In our monorepo this causes constant security check failures. Having a patch release would be super welcome.

PR #3248, opened 2025-08-03
feat(kafka_native): new module

strowk:
## What does this PR do?

Adding new module: kafka_native, that should be an alternative to existing kafka module based on confluent kafka.

## Why is it important?

Kafka native is starting faster and can fix issues such as this: https://github.com/testcontainers/testcontainers-go/issues/2206

## Related issues

- Relates #2206 

## How to test this PR

Explain here how this PR will be tested by the reviewer: commands, dependencies, steps, etc.

```bash
cd modules/kafka_native && go test -timeout 30s -count 1 -run '^TestKafka$' github.com/testcontainers/testcontainers-go/modules/kafka_native ; cd -
```

## Additional Context

There is also this PR, that attempts to modify existing module: https://github.com/testcontainers/testcontainers-go/pull/2683

Here is a brief list of differences in approaches:
- this PR does not attempt to add SASL support and instead only focuses on using Kafka Native
- this PR does not modify existing module to avoid potentially breaking backwards compatibility for users depending on details of confluent implementation
- this PR does not remove starting script (advertising needs to detect correct hostname, at least for docker in Windows) and instead just adds "exec" before starting Kafka Native process, this _should_ help with related issue linked above because exec would replace process instead of starting a new one and piping inputs, so signals would go to the right process properly


strowk:
closed this, because I did not use the right branch apparently

Issue #2980, opened 2025-02-10
[Bug]: Unable to start kafka container when using an image that has a subfix after version number.


ouranders:
### Testcontainers version

0.35.0

### Using the latest Testcontainers version?

Yes

### Host OS

macOs Sonoma

### Host arch

ARM

### Go version

1.23.6

### Docker version

```shell
Client:
 Cloud integration: v1.0.22
 Version:           20.10.12
 API version:       1.41
 Go version:        go1.16.12
 Git commit:        e91ed57
 Built:             Mon Dec 13 11:46:56 2021
 OS/Arch:           darwin/arm64
 Context:           desktop-linux
 Experimental:      true

Server: Docker Desktop 4.37.2 (179585)
 Engine:
  Version:          27.4.0
  API version:      1.47 (minimum version 1.24)
  Go version:       go1.22.10
  Git commit:       92a8393
  Built:            Sat Dec  7 10:38:33 2024
  OS/Arch:          linux/arm64
  Experimental:     false
 containerd:
  Version:          1.7.21
  GitCommit:        472731909fa34bd7bc9c087e4c27943f9835f111
 runc:
  Version:          1.1.13
  GitCommit:        v1.1.13-0-g58aa920
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### Docker info

```shell
Client:
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
  ai: Ask Gordon - Docker Agent (Docker Inc., v0.5.1)
  buildx: Docker Buildx (Docker Inc., v0.19.2-desktop.1)
  compose: Docker Compose (Docker Inc., v2.31.0-desktop.2)
  debug: Get a shell into any image or container (Docker Inc., 0.0.37)
  desktop: Docker Desktop commands (Beta) (Docker Inc., v0.1.0)
  dev: Docker Dev Environments (Docker Inc., v0.1.2)
  extension: Manages Docker extensions (Docker Inc., v0.2.27)
  feedback: Provide feedback, right in your terminal! (Docker Inc., v1.0.5)
  init: Creates Docker-related starter files for your project (Docker Inc., v1.4.0)
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc., 0.6.0)
  scan: Docker Scan (Docker Inc., v0.17.0)
  scout: Docker Scout (Docker Inc., v1.15.1)

Server:
 Containers: 36
  Running: 1
[...]

Issue #2630, opened 2024-07-05
[Feature]: Kafka (KRaft) - customizable startup script/advertised listeners


seeroush:
### Problem

When building a network with multiple containers that depend on the Kafka container, the startup script assumes that connections are only being attempted between the host and the Kafka cluster directly. Because the startup script has a hard-coded `KAFKA_ADVERTISED_LISTENERS` environment variable exported, other dependent services (such as Debezium) are not able to communicate with the container because the advertised listeners are only accessible from the host and are not directly reachable in the container network.

My simple usecase setup is as follows:

1. Create a Kafka container and give it an alias
2. Create a Debezium container and have it connect to the Kafka container

Here is the sample Debezium container setup that I am building, using a generic container request

```go
debeziumContainer, err := testcontainers.GenericContainer(ctx, testcontainers.GenericContainerRequest{
	ContainerRequest: testcontainers.ContainerRequest{
		Image:          "debezium/connect:1.6",
		ExposedPorts:   []string{"8083"},
		WaitingFor:     wait.ForLog("Finished starting connectors and tasks").WithStartupTimeout(30 * time.Second),
		Networks:       network,
		NetworkAliases: aliases,
		Env: map[string]string{
			"BOOTSTRAP_SERVERS":                      "kafka:9093",
			"GROUP_ID":                               "1",
			"CONFIG_STORAGE_TOPIC":                   "debezium_connect_config",
			"OFFSET_STORAGE_TOPIC":                   "debezium_connect_offsets",
			"STATUS_STORAGE_TOPIC":                   "debezium_connect_status",
			"CONNECT_KEY_CONVERTER_SCHEMAS_ENABLE":   "false",
			"CONNECT_VALUE_CONVERTER_SCHEMAS_ENABLE": "false",
		},
	},
	Started: true,
})
```

### Solution

There are a few ways this could potentially be handled.

1. Provide a customization option for the advertised listeners environment variable that appends to the existing listeners list.
```go
kafkaContainer, err := kafka.RunContainer(ctx, "confluentinc/confluent-local:7.5.0", 
   kafka.WithClusterID("test-cluster"),
   kafka.WithAdvertisedListeners([]string{"PLAINTEXT://kafka:9093"}), // append to existing environment variable defaults
   network.WithNewNetwork(ctx, []string{"kafka"}, network.WithLabels(map[string]string{
       "name": "kafka",
   }), network.WithDriver("bridge")),
)
```

[...]

PR #3249, opened 2025-08-03
feat(kafka_native): new module

strowk:
## What does this PR do?

Adding new module: kafka_native, that should be an alternative to existing kafka module based on confluent kafka.

## Why is it important?

Kafka native is starting faster and can fix issues such as this: https://github.com/testcontainers/testcontainers-go/issues/2206

## Related issues

- Relates #2206 

## How to test this PR

Explain here how this PR will be tested by the reviewer: commands, dependencies, steps, etc.

```bash
cd modules/kafka_native && go test -timeout 30s -count 1 -run '^TestKafka$' github.com/testcontainers/testcontainers-go/modules/kafka_native ; cd -
```

## Additional Context

There is also this PR, that attempts to modify existing module: https://github.com/testcontainers/testcontainers-go/pull/2683

Here is a brief list of differences in approaches:
- this PR does not attempt to add SASL support and instead only focuses on using Kafka Native
- this PR does not modify existing module to avoid potentially breaking backwards compatibility for users depending on details of confluent implementation
- this PR does not remove starting script (advertising needs to detect correct hostname, at least for docker in Windows) and instead just adds "exec" before starting Kafka Native process, this _should_ help with related issue linked above because exec would replace process instead of starting a new one and piping inputs, so signals would go to the right process properly


strowk:
For comparison: on my machine currently, `TestKafka` in kafka_native runs in 3.6 seconds, while original module test runs in 16.6 seconds. On big codebases and weaker machines this could translate in significant speed up!

```bash
$ cd modules/kafka_native && go test -timeout 30s -count 1 -run '^TestKafka$' github.com/testcontainers/testcontainers-go/modules/kafka_native ; cd -
ok      github.com/testcontainers/testcontainers-go/modules/kafka_native        3.693s

$ cd modules/kafka && go test -timeout 30s -count 1 -run '^TestKafka$' github.com/testcontainers/testcontainers-go/modules/kafka ; cd -
ok      github.com/testcontainers/testcontainers-go/modules/kafka       16.654s
```

Tests themselves are identical, only different images and less confluent-dependant configs.

Upd: Commands above are outdated, check PR description for up to date way to measure.

mdelapenya:
Hi @strowk thanks for adding this! I'm currently taking a look, also discussing internally with the other TC maintainers if this code should live as a new sub-package under the kafka module. What are your thoughts on this?

tim-reply:
Sounds fine to me. Just wondering what would interface then look like

strowk:
Sounds ok, though not sure what would interface look like

mdelapenya:
Hmmm thinking more in depth about the submodule, will the existing kafka options apply to this native image? If so, then we should probably find them, although it can be probably done in a follow-up.

mabrarov:
Hi @mdelapenya and @strowk,

Are we ensured we need a new module instead of extending existing one (kafka)? I planned to prepare that change - refer to https://github.com/testcontainers/testcontainers-go/compare/main...mabrarov:testcontainers-go:feat/kafka_alt_images - since May 2025 (refer to https://github.com/testcontainers/testcontainers-go/pull/2683#issuecomment-2912345737 and https://github.com/mabrarov/testcontainers-go-kafka-2206/pull/2), but I got time for that just recently.

Please note that (my case) usage of apache/kafka-native image can be problematic due to this image doesn't contain tools (like kafka-topics) which can be required for Kafka health check (e.g. in Docker Compose or Kubernetes). Using different images for tests (Testcontainers: apache/kafka-native) and local / dev or QA environments (Docker Compose, Kubernetes: apache/kafka) can bring additional issues.

Thank you.

strowk:
@mabrarov , if I understand correctly some earlier concerns from maintainers were that removing confluent image from existing module would be not backwards compatible change for users that rely on something confluent specific (I for sure am one of those, albeit I know how to transition). 

I.e it would for those users look such that the new version is pulled by something like dependabot or renovate and then they have to figure out how to transition to new image and have to stop updating testcontainers until the solution is found, which is somewhat undesirable situation. 
I base this line of thinking on the discussion in the other PR introducing kafka native that I linked in description.

https://github.com/testcontainers/testcontainers-go/pull/2683#issuecomment-2856920122

As to bitnami.. seeing what Broadcom did recently to images and how much the secured ones cost.. I sort of doubt it would find very many users TBH. This is just my opinion of course, but I would avoid bitnami as much as possible from now on.

mabrarov:
Hi @strowk,

You are right about binami/kafka image (the reason I got back to this work).

As for confluentinc/confluent-local image - we can keep it too while adding support for apache/kafka and apache/kafka-native images to kafka module. Feel free to check [feat/kafka_alt_images](https://github.com/testcontainers/testcontainers-go/compare/main...mabrarov:testcontainers-go:feat/kafka_alt_images) branch I referenced in my previous comment. Anyway, I'm happy (and appreciate) the work on kafka module is going on even if it will be a dedicated module. My need in that case is to support not just apache/kafka-native but apache/kafka image too.

Thank you.

strowk:
@mabrarov 

Ooh, thanks for clarification, I did not know that there separate apache/kafka is somehow significantly better than the native one.

This does mean that it would be better for users to have access to all three possible images -confluent/apache-not-native/apache-native via same module and pick between "flavors" by some sort of an option probably?
I feel like it should be more explicit than just giving different image and module generating different configs based on that image..
Though of course configs derived from image would be easier to use probably, assuming it will not break :)

mabrarov:
@strowk,

> it would be better for users to have access to all three possible images -confluent/apache-not-native/apache-native via same module and pick between "flavors" by some sort of an option probably?

It is a question I have too. Maybe maintainers of Testcontainers for Go can answer it?

IMHO: only limited set of docker images (including tags) can be supported because we cannot predict what changes will be introduced in new version of image. Considering this limitation I'd just  recommend to use specific (including tag) image with Testcontainers.

Thank you.

strowk:
@mabrarov , I suppose that deriving what to do from parsing the image is a common answer in codebase as I can see:

https://github.com/testcontainers/testcontainers-go/blob/main/modules/elasticsearch/version.go

https://github.com/testcontainers/testcontainers-go/blob/ec51c6715cca41d30daacc7519118111a15ea623/modules/localstack/localstack.go#L32

https://github.com/testcontainers/testcontainers-go/blob/ec51c6715cca41d30daacc7519118111a15ea623/modules/redpanda/redpanda.go#L439

Though it is usually about version of same "flavor", but I dont see that much difference, except we probably would need to allow user to override whatever we define..

mabrarov:
Hi @strowk,

Examples you referenced in https://github.com/testcontainers/testcontainers-go/pull/3249#issuecomment-3506835063 cover only lower bound while I am worried about upper bound too. Anyway, we can start from:

1. Check for the minimum version for each image repository/flavor.
1. Use some specific images in tests to have them as recommendation for the users (maybe need to cover that explicitly in documentation).

I feel it will be still the "Testcontainers way".

Thank you.

strowk:
@mabrarov , I have made an attempt at this based on a very simple logic that looks like this:

```go
const (
	apacheKafkaImagePrefix  = "apache/kafka"
	confluentincImagePrefix = "confluentinc/"
	dockerIoPrefix          = "docker.io/"
)

func isApache(image string) bool {
	return strings.HasPrefix(image, apacheKafkaImagePrefix) || strings.HasPrefix(image, dockerIoPrefix+apacheKafkaImagePrefix)
}

func isConfluentinc(image string) bool {
	return strings.HasPrefix(image, confluentincImagePrefix) || strings.HasPrefix(image, dockerIoPrefix+confluentincImagePrefix)
}

func getStarterScriptContent(image string) string {
	if isApache(image) {
		return apacheStarterScriptContent
	} else if isConfluentinc(image) {
		return confluentincStarterScriptContent
	} else {
		// Default to confluentinc for backward compatibility
		// in situations when image was custom specified based on confluentinc
		return confluentincStarterScriptContent
	}
}

```

This does not care about specific version as such, only whether or not image is confluent vs apache. In this "apache/kafka-native" and "apache/kafka" would work the same (starter script has no issues with either so far as I can tell).

It seems to work on and I tried with confluent, native and usual apache with this times:
- 14.77s for confluentinc
- 2.13s for native
- 5.65s for apache not native

This is just one run of course, so not super representative, but kind of interesting that apache/kafka shows good timing vs confluentinc one, though still not quite as good as native.

mabrarov:
Hi @strowk,

Just to complete my work with some milestone I opened #3488. Please feel free to discuss with maintainers what way to proceed - copy some parts of that PR (e.g. links to GitHub issues, test for graceful shutdown of Kafka) into this PR (you was the first 😄) or just complete with this PR... or maybe complete/merge #3488 instead of this PR.

Thank you for your help and work.

strowk:
Hey @mdelapenya  , I have made a lot of changes to this:

- new stuff is all added into existing "kafka" module, "kafka_native" is removed
- user just needs to give apache/kafka-native or apache/kafka image (those are all explained in docs)
- I had to merge from main, due to conflicts, tell me if you'd rather I made another PR (contributing doc says that force push is a no-no).

@mabrarov  , as requested this should work for "apache/kafka" as well as for "apache/kafka-native" , there are tests and examples to confirm.

modules/kafka/kafka_test.go
mabrarov, line 1: This PR description mentions #2206. Please consider adding test for Kafka graceful shutdown (e.g. from #3488 where I copied the test from #2206).

modules/kafka/kafka_test.go
```diff
@@ -66,6 +66,32 @@ func TestKafka(t *testing.T) {
...
 }
 
+func TestKafka(t *testing.T) {
+	testCases := []struct {
+		name  string
+		image string
+	}{
+		{
+			name:  "confluentinc",
+			image: "confluentinc/confluent-local:7.5.0",
```
mabrarov, line 84: Minimal supported version is 7.4.0 so it is reasonable to add a test for 7.4.0 too.

modules/kafka/kafka_test.go
```diff
@@ -66,6 +66,32 @@ func TestKafka(t *testing.T) {
...
+		},
+		{
+			name:  "apache native",
+			image: "apache/kafka-native:4.0.1",
```
mabrarov, line 88: 4.0.1 is pretty new version. Can we have tests for 3.9.1 (as the last 3.x) too?

modules/kafka/version.go
mabrarov, line 1: If this PR extends existing parsing of docker image name, then is it possible to fix #2980 here too?

modules/kafka/kafka.go
```diff
@@ -30,6 +30,13 @@ echo '' > /etc/confluent/docker/ensure
 /etc/confluent/docker/configure
 /etc/confluent/docker/launch`
```
mabrarov, line 31:
```suggestion
exec /etc/confluent/docker/launch`
```
?

modules/kafka_native/kafka.go
```diff
@@ -0,0 +1,172 @@
...
+const (
+	starterScript = "/usr/sbin/testcontainers_start.sh"
+
+	// starterScript {
+	starterScriptContent = `#!/bin/bash
+export KAFKA_ADVERTISED_LISTENERS=%s,BROKER://%s:9092
+echo Starting Kafka Native
+exec /etc/kafka/docker/run
+`
+	// }
```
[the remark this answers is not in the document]
strowk: That script is only used when image starts with `apache/kafka`, as "isApache" method does

modules/kafka/kafka_test.go
strowk: Nice idea, copied from your PR: https://github.com/testcontainers/testcontainers-go/pull/3249/commits/5a57c72dc0a7fd1e85f19eacb6f6a5492bfeb726
Only do this for apache images, though I did not see it necessary for more than one version.

modules/kafka/kafka_test.go
strowk: easy enough: https://github.com/testcontainers/testcontainers-go/pull/3249/commits/ea0521247a8c5d51be93eaff20ffa80c4e1a165e

modules/kafka/kafka_test.go
strowk: sure: https://github.com/testcontainers/testcontainers-go/pull/3249/commits/ea0521247a8c5d51be93eaff20ffa80c4e1a165e

modules/kafka/kafka_test.go
mabrarov: I'm checking some confluentinc/confluent-local images right now too, because https://github.com/confluentinc/kafka-images/pull/401 was merged. I doubt I will succeed.

modules/kafka/kafka.go
strowk: You'd think it would help, but your "TestKafkaGracefulShutdown" still fails on confluent even with this.

I made 

```go
func TestKafkaGracefulShutdown(t *testing.T) {
	testCases := []struct {
		name  string
		image string
	}{
		{
			name:  "apache native",
			image: "apache/kafka-native:4.0.1",
		},
		{
			name:  "apache not-native",
			image: "apache/kafka:4.0.1",
		},
		{
			name:  "confluentinc",
			image: "confluentinc/confluent-local:7.5.0",
		},
	}
```

Changed script:

```go
	// starterScriptConfluentinc {
	confluentincStarterScriptContent = `#!/bin/bash
source /etc/confluent/docker/bash-config
export KAFKA_ADVERTISED_LISTENERS=%s,BROKER://%s:9092
echo Starting Kafka KRaft mode
sed -i '/KAFKA_ZOOKEEPER_CONNECT/d' /etc/confluent/docker/configure
echo 'kafka-storage format --ignore-formatted -t "$(kafka-storage random-uuid)" -c /etc/kafka/kafka.properties' >> /etc/confluent/docker/configure
echo '' > /etc/confluent/docker/ensure
/etc/confluent/docker/configure
exec /etc/confluent/docker/launch`
	// }
```

, then from "modules/kafka" run this:

```bash
go test -timeout 180s -run '^TestKafkaGracefulShutdown$' github.com/testcontainers/testcontainers-go/modules/kafka
```

It still says 
```
--- FAIL: TestKafkaGracefulShutdown (72.19s)
    --- FAIL: TestKafkaGracefulShutdown/confluentinc (64.10s)
        kafka_test.go:169:
                Error Trace:    C:/work/testcontainers-go/modules/kafka/kafka_test.go:169
                Error:          Kafka did not gracefully exit
                Test:           TestKafkaGracefulShutdown/confluentinc
                Messages:       Kafka did not gracefully exit in 1m0s
```

Launch itself does "exec" too: https://github.com/confluentinc/kafka-images/blob/7.5.0-post/kafka/include/etc/confluent/docker/launch#L37
, but my guess is that down the line somewhere it breaks, but I don't have much patience debugging confluent bash scripts TBH :) and since "exec" does not help in this script, I am not sure I see the point. 
Users should just switch to "apache" images if they can.

mabrarov: Please check https://github.com/confluentinc/kafka-images/pull/401. 7.5.0 and 7.9.1 are broken (https://github.com/testcontainers/testcontainers-go/issues/2206#issuecomment-2887726469).

modules/kafka/version.go
strowk: No actual parsing occurs per-se, it just checks string for couple prefixes. This parsing is in quite other place. Feels like it should be separate PR. Plus TBH I am not even sure where to start with this one. `:7.5.0.arm64` is not semver and I am not quite sure how to correctly identify all possible corner cases here. Unless it would be ok to list explicitly all possible architectures, which I don't find a compelling approach honestly.

It seems to me that users might find it much easier to just switch to Apache (which would not have that issue)?

mabrarov: After a closer look at this PR I agree with all from https://github.com/testcontainers/testcontainers-go/pull/3249#discussion_r2507346663

modules/kafka/kafka.go
strowk: Which one works?

mabrarov: I found that confluentinc/confluent-local:8.1.0 works with #3488 but with a major note: it looks like the start of container is slow (it is hard to detect when it completes) and until start completes shutdown signal is ignored, so the test fails if I don't give container some time holding the test at breakpoint at modules/kafka/kafka_test.go:94 (from #3488). It will take more time to understand what is wrong with start of confluentinc/confluent-local:8.1.0. Let's consider that confluentinc/confluent-local is legacy supported for backward compatibility only and not make change which I suggested in https://github.com/testcontainers/testcontainers-go/pull/3249#discussion_r2507284252. The more I work with confluentinc/confluent-local, the more I don't like it.

strowk: > The more I work with confluentinc/confluent-local, the more I don't like it
hah, very true

> Let's consider that confluentinc/confluent-local is legacy supported for backward compatibility only

Works for me! Resolving this then, thanks for your help!

modules/kafka/kafka.go
```diff
@@ -30,6 +30,13 @@ echo '' > /etc/confluent/docker/ensure
...
 	// }
+
+	// starterScriptApache {
+	apacheStarterScriptContent = `#!/bin/bash
+export KAFKA_ADVERTISED_LISTENERS=%s,BROKER://%s:9092
```
mabrarov: I would also add a listener for localhost and advertise it here (as well as for the case of confluentinc/confluent-local image). Feel free to check #3488. Such listener can be used by the user in custom wait strategy checking localhost (e.g. using kafka-topics tool) from command executed within container.

modules/kafka/kafka.go
```diff
@@ -30,6 +30,13 @@ echo '' > /etc/confluent/docker/ensure
...
+echo Starting Apache Kafka
+exec /etc/kafka/docker/run`
```
mabrarov, line 38: Can we avoid hard-coded value like it is done in #3488?

docs/modules/kafka.md
```diff
@@ -1,10 +1,12 @@
...
+# Kafka
 
 Since <a href="https://github.com/testcontainers/testcontainers-go/releases/tag/v0.24.0"><span class="tc-version">:material-tag: v0.24.0</span></a>
 
 ## Introduction
 
-The Testcontainers module for KRaft: [Apache Kafka Without ZooKeeper](https://developer.confluent.io/learn/kraft).
+The Testcontainers module for Kafka.
+
+This module would run Kafka in Kraft mode: [Apache Kafka Without ZooKeeper](https://developer.confluent.io/learn/kraft/) and it supports both [Apache Kafka](https://kafka.apache.org/) and [Confluent](https://docs.confluent.io/kafka/overview.html) images.
```
mabrarov: It can be helpful for the users if we notice here (an outcome of https://github.com/testcontainers/testcontainers-go/pull/3249#discussion_r2507351306):

1. Confluent images don't support (stably) graceful shutdown.
2. Confluent images are considered legacy (supported for backward compatibility) and Apache Kafka images are recommended.

modules/kafka/kafka.go
strowk: There is a certain benefit in keeping hardcoded value tho, it is a bit clearer what is going on when looking at docs:
https://golang.testcontainers.org/modules/kafka/#init-script

Docs basically take out snippets of code to include inside built static website. The way my PR does this, it looks like this:

image

image


Plus I kind of doubt that increasing complexity of this script would make it better. There is some chance that entrypoint would change in the future Apache Kafka versions, but is this probability plus losses from users facing related problems higher than efforts of testcontainers maintainers understanding what is going on in that script-building logic?

mabrarov: Well, I can be too worried of (old) docker images being removed from Docker Hub (like the recent case with JDK images). It is the reason I tried to make #3488 a little bit more forward-compatible. Considering documentation code snippet, I suppose you (@strowk) is the best decision maker for this comment thread.

Thank you for clarification.

docs/modules/kafka.md
strowk: I am not sure about calling them legacy, TBH, this sounds a bit like an insult.
I will mention graceful shutdown though, totally makes sense. Confluent images also are somewhat larger:

```
confluentinc/confluent-local 8.1.0 592MB
apache/kafka 4.0.1 393MB
apache/kafka-native  4.0.1 137MB
```

modules/kafka/kafka.go
strowk: This does sound interesting, but I am not quite sure if this PR should fit this part. 
This change should probably come with some kind of example that would demonstrate this technique?

docs/modules/kafka.md
strowk: here : https://github.com/testcontainers/testcontainers-go/pull/3249/commits/9c948d2661d69a5f186f4ca95e354f866f0bf1f7

modules/kafka/kafka.go
strowk: I'll keep this open until there is maintainer to weigh-in..

modules/kafka/version.go
mabrarov, line 1: Sorry, I forgot the reason I don't rely on image name in #3488 - custom docker registry (e.g. the corporate one) where docker image name can have different prefix or even completely different name.

strowk: Ah, infamous corporate registries. Yeah, I see this as a possible issue. I have added this to the list of differences I summarized in PR description.

strowk: I feel like then, maybe it would make sense if user could pick (i.e override image-based detection) if they want to use "confluent" script vs "apache" script? 
This would solve this problem without introducing too much complexity in the generic starting script?
Though I see why it would be less convenient approach.

Sort of.. 
`kafka.WithStarterScript(kafka.ConfluentStarterScript)`
or 
`kafka.WithStarterScript(kafka.ApacheStarterScript)` ?

mabrarov: `kafka.WithStarterScript` (a new kind of `ContainerCustomizer`?) seems to be good, but what will be default value? Will default value be deduced from image name?

modules/kafka/kafka.go
mabrarov: I added `TestKafkaLocalhostListener` into #3488 to demonstrate possible usage of localhost listener.

modules/kafka/version.go
strowk: Default is deduced, yeah.

Yeah, I see that ContainerCustomizer can have fake implementation and then options are processed separately.
Something sort of like that: https://github.com/testcontainers/testcontainers-go/blob/ec51c6715cca41d30daacc7519118111a15ea623/modules/etcd/options.go#L9-L35

strowk: So, something like this can work for this case: https://github.com/testcontainers/testcontainers-go/pull/3249/commits/99b4833436b06670341e9fb78550ff90c01d25da

strowk: You can see usage example here : https://deploy-preview-3249--testcontainers-go.netlify.app/modules/kafka/#__tabbed_3_3

modules/kafka/options.go
```diff
@@ -0,0 +1,41 @@
...
+func (o RunOption) Customize(_ *testcontainers.GenericContainerRequest) error {
+	return nil
+}
+
+// WithStarterScript is an option to set a custom starter script content for the Kafka container.
+//
+// You would typically use this option when the image you are using is different from
+// the standard ones or the image is in your custom registry and automatic inference
+// of the starter script does not work as expected.
+func WithStarterScript(content string) RunOption {
```
mabrarov, line 30: Looks like a step forward to implement #2630. Is it reasonable to mention #2630 in description of this PR ("Related issues" section)? If this PR will mention #2630, then do we need to provide an example (e.g. for localhost listener - refer to https://github.com/testcontainers/testcontainers-go/pull/3249#discussion_r2507359423)?

modules/kafka/kafka.go
strowk: Copied here : https://github.com/testcontainers/testcontainers-go/pull/3249/commits/b21194e6ed5b8d899f17362e7a0dc1c153c4ae78
:)
Thanks!

This is just initial, needs some kind of documentation probably to let users know how this works too :)
(and when it does not, since native images do not include those tools)

modules/kafka/version.go
mabrarov: Hi @strowk, 
I believe we can resolve this thread (I don't have such option).
Thank you.

modules/kafka/options.go
strowk: Huh, interesting. 

On one side it could be cool to capitalize on ability of user to override custom script, by improving the interface there from using "fmt" to maybe "text/template" so that it would be somewhat less likely to break in the future. Of course changes to default templates (and expectations from other configs that script does something) would be potentially a breaking change for anyone who provided a custom script template, but as it goes.. "with great power comes great responsibility"

image

From the other side there could be argument against this though and while using "text/template" internally is a good idea, it might be better to restructure WithStarterScript to not take the whole script (actually template of a script..), but rather just enum like ConfluentScript = 1, ApacheScript = 2, so users could not decide to provide any script (and therefore experience issues with new updates), in which case customizable listeners would have to come in a more strict (and hopefully convenient) way, say as `WithExtraListeners` to take slice of strings and then pass it on for provider-specific script template to construct the whole starter script.

Of course combination of `WithStarterScript(scriptTemplate string)` and `WithExtraListeners(listeners []string)` is also possible, while first option might be potentially overriding second one (depending on whether provided scriptTemplate supports taking something like `{{ range .listeners }}..` or not.

strowk: I'd like to see if maintainers have any opinions to those possible directions.

@mdelapenya , would be nice to have some input

mdelapenya: @strowk @mabrarov thanks for working on the kafka module, it's great to see you working on this. First, I'm not a Kafka expert nor user, so I'd delegate to what makes sense to you (easy answer). The difficult one comes regarding what API we want to provide to kafka users. In Java, the module offers two different containers:

- ConfluentContainer: https://github.com/testcontainers/testcontainers-java/blob/main/modules/kafka/src/main/java/org/testcontainers/kafka/ConfluentKafkaContainer.java
- KafkaContainer: https://github.com/testcontainers/testcontainers-java/blob/main/modules/kafka/src/main/java/org/testcontainers/kafka/KafkaContainer.java

I think we can provide the same containers following what we did with MongoDB and Atlas: have a separate Go package inside the kafka module for confluent, or what ever makes sense. Then (probably as a follow up) we can find commonalities to both implementations and refactor their internals. It would be something like this:

```
- kafka
    - apache/options.go
    - confluent/options.go
    - internal/common.go
    - kafka.go
```

We could place the confluent-specific options in that package for users to do something like:

```go
ctr, err := kafka.Run(ctx, "image", apache.WithXXX())
ctr, err := kafka.Run(ctx, "image", confluent.WithXXX())
```

Of course, I'm happy to hear your voice for finding the proper design.

stevenh: To clarify, do you expect all options to be compatible with both flavours, or is the intent of apache.WithXXX to be a apache specific option?

strowk: ATM in this PR all options are "shared". One option overrides "flavor" detection (by passing script that is the difference between flavors), the other configures cluster id. 

At the current moment I do not imagine options that would be specific to apache or confluent, albeit it is not impossible as such and not impeded by design of this PR in any way. If and when such specific options would appear, I agree that placing them in apache and confluent packages would make sense, though it might be also a bit weird since the interface then would possibly allow passing confluent option together with apache image potentially causing unexpected outcomes.

Also at the current moment replicating java approach with two classes is somewhat unnecessary, as currently shared container structure does its job well enough.

modules/kafka/kafka.go
```diff
@@ -17,19 +17,26 @@ import (
 
 const publicPort = nat.Port("9093/tcp")
 const (
-	starterScript = "/usr/sbin/testcontainers_start.sh"
+	starterScriptPath = "/usr/sbin/testcontainers_start.sh"
 
-	// starterScript {
-	starterScriptContent = `#!/bin/bash
+	// starterScriptConfluentinc {
+	ConfluentStarterScript = `#!/bin/bash
```
stevenh: question: Is there a reason this needs to be exported?

I ask as we try to keep the API surface area as small as possible by not exporting unless its needed.

modules/kafka/kafka.go
```diff
@@ -72,15 +90,15 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
 		}),
 		testcontainers.WithEntrypoint("sh"),
 		// this CMD will wait for the starter script to be copied into the container and then execute it
-		testcontainers.WithCmd("-c", "while [ ! -f "+starterScript+" ]; do sleep 0.1; done; bash "+starterScript),
+		testcontainers.WithCmd("-c", "while [ ! -f "+starterScriptPath+" ]; do sleep 0.1; done; bash "+starterScriptPath),
```
stevenh, line 93: question: I know this isn't new but why do we need the file check loop?

I ask as the might be that this could be done before the container is started.

Could be mute point if we do move to templated command.

modules/kafka/options.go
stevenh: I think I'm hearing that the difference between the two flavours can handled, such that we could have:
```go
ctr, err := kafka.Run(ctx, "image", kafka.WithFlavourApache())
ctr, err := kafka.Run(ctx, "image", kafka.WithFlavourConfluent())
```

We could pick the default as the current flavour to maintain compatibility or switch to the best performance by default as a potential breaking change.

Underneath this could use an internal, for now, `withStarterScript(template)` then internally it would build a template with various options from other methods allowing for easy future expansion.

Did I understand both perspectives correctly?

modules/kafka/kafka.go
mabrarov: Hi @stevenh,

Do you mean:

1. Create container, but do not start it.
1. Copy custom starter script into created container.
1. Start container created at step 1.

?

The starter script needs to know:

1. Address of host where container runs (will run).
1. Mapped port of container (port of container host which specific local port of container is mapped to).

I'm not that good in Docker API, but according to Docker CLI, mapped port is unknown until container starts:

```text
$ docker create --name kafka apache/kafka-native:4.0.1
...
$ docker inspect kafka
...
        "NetworkSettings": {
            "Bridge": "",
            "SandboxID": "",
            "SandboxKey": "",
            "Ports": {},
...
$ docker start kafka
$ docker inspect kafka
...
        "NetworkSettings": {
            "Bridge": "",
            "SandboxID": "b15ec42e25b4910dc7153ecf428b31fc27f6db73f549046b09ed0ccbd6849987",
            "SandboxKey": "/var/run/docker/netns/b15ec42e25b4",
            "Ports": {
                "9092/tcp": null
            },
... 
```

It means that we cannot prepare content of starter script, until container starts.

Thank you.

modules/kafka/kafka.go
strowk: Yeah, this could be used in conjuction with option to override starter script selection. This is needed for example when user configures different registry (not docker.io) and has there f.e a mirror of kafka images. Then detection of apache or confluent would not work and user need to pass starter scrip explicitly

modules/kafka/kafka.go
strowk: Yeah, to clarify, this is due to the fact that kafka (unlike most other testcontainers) needs to know how exactly clients would be calling it (to include these addresses in the list of advertised listeners).

modules/kafka/kafka.go
stevenh: Could that just be an option instead of exposing the config file, as [per this comment](https://github.com/testcontainers/testcontainers-go/pull/3249#discussion_r2512317095)?

strowk: It could! 
I would think on how this would work a bit later. One question I would consider is what happens if user gives both WithApacheFlavor and WithConfluentFlavor. Simple would be that last wins, but probably clearer would be to return error, though maybe user has some one package that pre-configures container (i.e generates  slice of options) and then they might want to override that (i.e by appending to the slice) then it would be more convenient to just use last one.

Of course it would remove possibility of giving arbitrary script, which is possibly something that has likely chance of being broken anyway, and for things that would benefit from that (i.e extra advertised listeners), there could be better typed approaches.

stevenh: I would just return an error, it ensures the author knows they made mistake.

You could still have a default type to maintain compatibility by adding an internal flag set by the functional options.

I think there's definitely a benefit to using a more strictly typed approach, but if we did ultimately want to provide the option for a custom script this could be its own flavour.

Thanks for bearing with me on the discussion here, as I know was a little late to the party, hopefully it helps us align on the best solution 😄

mdelapenya: Returning an error if +1 flavor is passed seems sensible to me.

Also, if the user is trying to break its own kafka passing both flavors... well, then I'd say it's their responsibility to use the module correctly, even though it does not come with the guardrails to prevent multiple flavors... Docs must be clear about that even though the guardrail does not exist.

Regarding the arbitrary script, we many times add that to a module: postgres has a WithInitScript option. Why not having a WithStarterScript option then? Of course, if that script is broken, we cannot do anything apart from raising the error because kafka did not start correctly

strowk: Ah, so there is already such practice. It would probably make sense then to name it WithInitScript as well..

strowk: Hey, so I have provided now `WithApacheFlavor` and `WithConfluentFlavor`, which allowed to not export default starter scripts. I still kept `WithStarterScript`, but just documented that user might face possible incompatibility issues with it.

There is an error returned if both `With..Flavor` options are passed.

I did not rename `WithStarterScript` to `WithInitScript`, because sematics are somewhat different (starter is what is called from CMD directly and is supposed to control main process, while init scripts are just called from entrypoint before the main process start).

I also do not give error if user provided both With..Flavor and WithStarterScript, though it sort of does not make sense too in current implementation, but my view of those is that flavor options obviously conflict with each other, but it might be not so obvious that they also conflict with custom starter script and in future there could be more specifics added (if confluent and apache would continue to diverge) that would make it sensible to pass WithStarterScript with one of With..Flavor ones. So in case of those, whichever is passed last would win.

docs/modules/kafka.md
```diff
@@ -42,12 +61,12 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
...
-In example: `Run(context.Background(), "confluentinc/confluent-local:7.5.0")`.
+In example: `Run(context.Background(), "apache/kafka-native:4.0.1")`.
 
 !!! warning
-    The minimal required version of Kafka for KRaft mode is `confluentinc/confluent-local:7.4.0`. If you are using an image that
-    is different from the official one, please make sure that it's compatible with KRaft mode, as the module won't check
-    the version for you.
+    Module expects that the image in use supports Kraft mode (Kafka without ZooKeeper).
+    The minimal required version of Confluent images for KRaft mode is `confluentinc/confluent-local:7.4.0`.
+    All Apache images support Kraft mode.
```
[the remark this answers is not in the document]
strowk: I highly doubt that it would change in future and if it will, documentation could be updated.

modules/kafka/kafka.go
strowk: @stevenh  , @mdelapenya  , do you guys have any opinion on this? Should we try to derive the executable from inspecting the image or just keep them hardcoded until/if they change?

modules/kafka/options.go
strowk: I have rewritten how user can choose flavor using specific flavor options as requested. I still keep `WithStarterScript` option available though, as I think it can be useful in some situations. Though I am thinking it might need some better parameterization in future, but for now I think it can be used as is.. Tell me if any of you would rather remove it for now, I am not particularly set on this, just thought it could be useful.

modules/kafka/kafka.go
strowk: FYI, I have also documented localhost listener usage in md file: https://github.com/testcontainers/testcontainers-go/pull/3249/commits/f6f75b2e918be850b357a5ba2fe496c26f19ec4f

modules/kafka/benchmark_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+func startStopBenchmark(b *testing.B, image string) {
+	b.Helper()
+	for b.Loop() {
+		kafkaContainer, err := kafka.Run(context.Background(),
+			image,
+		)
+		if err != nil {
+			b.Fatalf("failed to start container: %s", err)
+		}
```
stevenh: suggestion: use testify require to simplify, more below

```suggestion
		require.NoError(b, err)
```

modules/kafka/kafka_test.go
```diff
@@ -89,3 +131,93 @@ func assertAdvertisedListeners(t *testing.T, container testcontainers.Container)
...
+
+func TestFailOnBothFlavors(t *testing.T) {
+	ctx := context.Background()
+
+	_, err := kafka.Run(ctx, "apache/kafka-native:4.0.1",
+		kafka.WithApacheFlavor(),
+		kafka.WithConfluentFlavor(),
+	)
+	require.Error(t, err)
+	require.EqualError(t, err, "apply option: flavor was already set, provide only one of WithApacheFlavor or WithConfluentFlavor")
```
stevenh: nit: No need to change, but the test for an error is typically good enough and avoids fragile tests if the actual error is changed either directly or via additional chaning.

modules/kafka/options.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// RunOption is an option that configures how Kafka container is started.
+type RunOption func(*runOptions) error
+
+var _ testcontainers.ContainerCustomizer = (RunOption)(nil)
+
+func (o RunOption) Customize(_ *testcontainers.GenericContainerRequest) error {
+	return nil
+}
```
stevenh, line 22: @mdelapenya this seems to be a common pattern, maybe we should consider a NoopCustomizer which can be embedded to avoid repetition of this code everywhere.

modules/kafka/options.go
```diff
@@ -0,0 +1,88 @@
...
+
+// WithStarterScript is an option to set a custom starter script content for the Kafka container.
+//
+// You would typically use this option when the image you are using is different from
```
stevenh, line 26: suggestion: follow the same pattern as WithApacheFlavor etc and return an error if already set.

modules/kafka/options_test.go
```diff
@@ -0,0 +1,71 @@
...
+			image: "custom/kafka:latest",
+			want:  confluentStarterScript,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			opts := &runOptions{
+				image: tt.image,
+			}
+			if got := opts.getStarterScriptContent(); got != tt.want {
```
stevenh: suggestion: use require.Equal to simplify, not we typically don't specify a custom error to keep it simple.

modules/kafka/version.go
```diff
@@ -0,0 +1,17 @@
...
+
+import "strings"
+
+const (
+	apacheKafkaImagePrefix  = "apache/kafka"
+	confluentincImagePrefix = "confluentinc/"
+	dockerIoPrefix          = "docker.io/"
+)
+
+func isApache(image string) bool {
```
stevenh, line 11: question: do we need both? If not each use would usually be a switch not an if which I saw above.

Alternatively it could be a imageType call returning an enum type
```go
type flavour int
const (
    unknownFlavour flavour = iota
    apacheFlavour 
    confluentFlavour
    customFlavour
)
```

modules/kafka/version_test.go
```diff
@@ -0,0 +1,115 @@
...
+		},
+		{
+			name:  "custom image",
+			image: "custom/kafka:latest",
+			want:  false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := isApache(tt.image); got != tt.want {
```
stevenh: suggestion: use require.Equal, more below.

Will need to be refactored if the we change to typed as per above.

modules/kafka/version.go
stevenh, line 1: suggestion: align filename with behaviour of the methods, type, flavour or other depending on the outcome of the other comments.

modules/kafka/kafka.go
stevenh: @mabrarov to confirm my understanding you're suggesting using the detail from inspect to identify the location of binary to exec, so that if that changed in the future it would remain compatible?

modules/kafka/options.go
mdelapenya: Good idea, we need that to be applied elsewhere.

docs/modules/kafka.md
```diff
@@ -84,3 +151,17 @@ The `Brokers(ctx)` method returns the Kafka brokers as a string slice, containin
 <!--codeinclude-->
 [Get Kafka brokers](../../modules/kafka/kafka_test.go) inside_block:getBrokers
 <!--/codeinclude-->
+
+## Localhost listener
```
mdelapenya:
```suggestion
#### Localhost listener
```

mdelapenya (commented):
This PR looks good to me, and it comes with a lot of work, thank you so much for dedicating your time for this. We'll make sure to make it into the project the soonest 🙇 

Let's address the existing review comments, and let's merge it on green CI.

modules/kafka/kafka.go
mabrarov: Hi @stevenh,

Yes, your understanding is correct (e.g. `copyStarterScript` in modules/kafka/kafka.go from #3488), but I am not sure if it matches the way Testcontainers for Go prefer (e.g. `getDockerProvider` from #3488 can be considered an issue making Testcontainers less decoupled with Docker).

Thank you.

modules/kafka/kafka_test.go
strowk: decided to drop it - https://github.com/testcontainers/testcontainers-go/pull/3249/commits

modules/kafka/benchmark_test.go
strowk: done: https://github.com/testcontainers/testcontainers-go/pull/3249/commits/c6a2b1dc65f3bd3f3718b7ef467c931f384699aa

modules/kafka/options.go
strowk: I assume you mean another PR? If not - "unresolve" this or just ping me

modules/kafka/options.go
strowk: done: https://github.com/testcontainers/testcontainers-go/pull/3249/commits/81b106567a6e87c9bbc6f103565cfd8fb46f5ebe

docs/modules/kafka.md
strowk: Hey, I've applied this without looking, but results do not seem correct to me..

This looks like "Localhost listener" section is under "Container Methods", but it does not actually describe container method, it just tells that there is a configuration that allows to do particular things from inside of container (via some options, but not via container methods).

image

Did you mean to put it under "Module Reference" maybe? I.e with three `#` symbols?

modules/kafka/options_test.go
strowk: done: https://github.com/testcontainers/testcontainers-go/pull/3249/commits/4627d1b86eadc82f431a49eb9a8a2004047ea97c

modules/kafka/version_test.go
strowk: for now just simplified assertion: https://github.com/testcontainers/testcontainers-go/pull/3249/commits/4627d1b86eadc82f431a49eb9a8a2004047ea97c

will see about other thing later

modules/kafka/kafka.go
strowk: yeah, so basically my feeling about this was that I am not sure if this technique would work reliably enough, i.e it won't break in some setups that would make resulting code less stable, not more, i.e trade off is extra complexity (and chance of bug) vs likeliness of ENTRYPOINT or CMD becoming different in one of flavors (or maybe users providing their own images without their own starter scripts.. in which case inspection might help too).

docs/modules/kafka.md
```diff
@@ -57,18 +88,58 @@ The environment variables that are already set by default are:
...
+#### Starter script
+
+The Kafka container will be started using a custom shell script.
+
+Module would vary the starter script depending on the image in use, using following logic:
 
-The Kafka container will be started using a custom shell script:
+- image starts with `apache/kafka`: use Apache Kafka starter script.
+- image starts with `confluentinc/`: use Confluent starter script.
+- otherwise: use Confluent starter script (for backward compatibility).
```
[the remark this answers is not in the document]
strowk: done: https://github.com/testcontainers/testcontainers-go/pull/3249/commits/3034eea0e4763f58318c6736d790244cced5e488

modules/kafka/options_test.go
```diff
@@ -0,0 +1,66 @@
...
+			want:  confluentStarterScript,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			opts := &runOptions{
+				image: tt.image,
+			}
+			require.Equal(t, tt.want, opts.getStarterScriptContent())
+			require.NoError(t, WithStarterScript("mytestsript")(opts))
```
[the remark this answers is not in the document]
strowk: done: https://github.com/testcontainers/testcontainers-go/pull/3249/commits/1fc9b85fa1edbfa10461e3cb3b5381302e863df0

modules/kafka/options.go
mdelapenya: Yes, sorry I was not clear enough. This is for us in a separate PR. I created #3492 for that

mabrarov:
Hi @strowk,

Could you please resolve merge conflicts?

Is there a plan for the further work on this PR?

Thank you.

PR #3250, opened 2025-08-04
Set correct version of azure in azurite

EzequielAyzenberg:
## What does this PR do?

It fixes the version of azure module in azurite module

## Why is it important?

Right now is not working the command
```
go get github.com/testcontainers/testcontainers-go/modules/azurite@v0.38.0 

> go: github.com/testcontainers/testcontainers-go/modules/azurite imports
        github.com/testcontainers/testcontainers-go/modules/azure/azurite: reading github.com/testcontainers/testcontainers-go/modules/azure/go.mod at revision modules/azure/v0.35.0: unknown revision modules/azure/v0.35.0
```

## Related issues

-

### DIFF
--- a/modules/azurite/go.mod
+++ b/modules/azurite/go.mod
@@ -7,7 +7,7 @@ toolchain go1.23.6
 require (
 	github.com/stretchr/testify v1.10.0
 	github.com/testcontainers/testcontainers-go v0.38.0
-	github.com/testcontainers/testcontainers-go/modules/azure v0.35.0
+	github.com/testcontainers/testcontainers-go/modules/azure v0.38.0
 )
 
 require (


modules/azurite/go.mod
```diff
@@ -7,7 +7,7 @@ toolchain go1.23.6
 require (
 	github.com/stretchr/testify v1.10.0
 	github.com/testcontainers/testcontainers-go v0.38.0
-	github.com/testcontainers/testcontainers-go/modules/azure v0.35.0
```
mdelapenya, line 10: praise: good catch! The release process is not capturing this, as it only updates the dependencies with the core, but does not consider that a module depends on another, which is the case for the deprecated azurite module.

Should we change this from draft to ready?

EzequielAyzenberg: Hey @mdelapenya thanks for replying and taking the time. I think I'll close this PR because in latest versions I understand: azurite as dependency is discouraged and azure should be imported instead. See my closed issue to confirm: https://github.com/testcontainers/testcontainers-go/issues/3251
By my side I confirm I could fix my project doing that

If you still think it's an error I'll work further in this PR
Cheers!

mdelapenya: Oh I missed the issue! Thanks for the reminder

Then, as the azurite module is deprecated, and will eventually removed, I think it's not worth it to maintain its set of dependencies, apart from the automated process.

So I'll close this PR because of that.

Thank you so much for confirming the current state of the modules!

BTW I'm pretty interested in learning how you use it: use cases, benefits of using it, missing features, etc. Would you like to share your use case as a comment here, even though the PR is closed?

EzequielAyzenberg: Sorry the delay, I'm back with the answer.
We mainly use `azurite.Run("mcr.microsfot.com/azure-storage/azurite:x.y.z")` so we can containerize a blob storage in our tests. Any suggestion to comply with the future deprecation?

mdelapenya: @EzequielAyzenberg just migrating to the new `azure/azurite` module would do it, so I think you have to change the import paths, only 🤔

PR #3252, opened 2025-08-06
chore(postgres): use require.NotNil instead of assert.NotNil

nolandseigler:
## What does this PR do?

This uses `require.NotNil` instead of `assert.NotNil` when the db variable will be used in the remainder of the test.

## Why is it important?

If the db variable were nil the next line that calls defer db.Close would panic and the following call to db.Exec that we are setting up could panic too.

## Related issues

- Relates #2808

### DIFF
--- a/modules/postgres/postgres_test.go
+++ b/modules/postgres/postgres_test.go
@@ -237,7 +237,7 @@ func TestWithSSL(t *testing.T) {
 
 	db, err := sql.Open("postgres", connStr)
 	require.NoError(t, err)
-	assert.NotNil(t, db)
+	require.NotNil(t, db)
 	defer db.Close()
 
 	result, err := db.Exec("SELECT * FROM testdb;")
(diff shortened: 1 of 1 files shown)


modules/postgres/postgres_test.go
```diff
@@ -237,7 +237,7 @@ func TestWithSSL(t *testing.T) {
 
 	db, err := sql.Open("postgres", connStr)
 	require.NoError(t, err)
-	assert.NotNil(t, db)
+	require.NotNil(t, db)
```
mdelapenya, line 239: suggestion: could you remove the entire `assert` import and replace it with `require`? 🙏 

Other than than, LGTM

nolandseigler: absolutely. will knock that out today. thanks!

nolandseigler: heres the update. [588bdd0](https://github.com/testcontainers/testcontainers-go/pull/3252/commits/588bdd0de060161a4aafab380ca3a89fcf16172d).

mdelapenya (approved):
LGTM, thank you so much for your contribution @nolandseigler. Even the smallest of them can be very impactful 💪 

Cheers!

PR #3254, opened 2025-08-08
 feat(modules/atlaslocal): add MongoDB Atlas Local module 

prestonvasquez:
## Related issues

-


modules/atlaslocal/atlaslocal.go
```diff
@@ -0,0 +1,360 @@
...
+		}
+
+		if info.IsDir() {
+			return fmt.Errorf("username file must be a file, got a directory: %s", usernameFile)
+		}
+
+		req.Env["MONGODB_INITDB_ROOT_USERNAME_FILE"] = usernameFile
+
+		// Mount username file as read-only in the container.
+		prev := req.HostConfigModifier
```
mdelapenya: suggestion: let's use the copy files APIs here, which are more deterministic. Please add the file to the req.Files attribute, or use the `WithFiles` option.

modules/atlaslocal/atlaslocal.go
```diff
@@ -0,0 +1,360 @@
...
+		}
+
+		if info.IsDir() {
+			return fmt.Errorf("password file must be a file, got a directory: %s", passwordFile)
+		}
+
+		req.Env["MONGODB_INITDB_ROOT_PASSWORD_FILE"] = passwordFile
+
+		// Mount password file as read-only in the container.
+		prev := req.HostConfigModifier
```
mdelapenya: suggestion: same here, let's use the Files attribute or WithFiles option

modules/atlaslocal/atlaslocal.go
```diff
@@ -0,0 +1,360 @@
...
+			}
+		}
+
+		return nil
+	}
+}
+
+// WithDisableTelemetry opts out of telemetry for the MongoDB Atlas Local
+// container by setting the DO_NOT_TRACK environment variable to 1.
+func WithDisableTelemetry() testcontainers.CustomizeRequestOption {
```
mdelapenya: not a blocker
```suggestion
func WithNoTelemetry() testcontainers.CustomizeRequestOption {
```

modules/atlaslocal/atlaslocal.go
```diff
@@ -0,0 +1,360 @@
...
+
+	return func(req *testcontainers.GenericContainerRequest) error {
+		prev := req.HostConfigModifier
+		req.HostConfigModifier = func(hostConfig *container.HostConfig) {
+			if prev != nil {
+				prev(hostConfig)
+			}
+
+			// Remove any old /docker-entrypoint-initdb.d bind mounts.
+			filtered := hostConfig.Binds[:0]
```
mdelapenya: suggestion: same here, let's use the Files APIs

modules/atlaslocal/atlaslocal.go
```diff
@@ -0,0 +1,360 @@
...
+		}
+
+		return nil
+	}
+}
+
+// WithMongotLogFile sets the path to the file where you want to store the logs
+// of Atlas Search (mongot) by setting the MONGOT_LOG_FILE environment variable.
+// Note that this can be set to /dev/stdout or /dev/stderr for convenience.
+func WithMongotLogFile(logFile string) testcontainers.CustomizeRequestOption {
```
mdelapenya: question: should we simplify the API adding types for stdout/stderr so users don't configure a non existing path? same for the runner log file

modules/atlaslocal/atlaslocal_test.go
```diff
@@ -0,0 +1,626 @@
...
+
+	outBytes, err := io.ReadAll(reader)
+	require.NoError(t, err)
+
+	if len(outBytes) < 8 {
+		t.Fatalf("Exec output too short: less than 8 bytes!")
+	}
+
+	out := strings.TrimSpace(string(outBytes[8:]))
+	assert.Equal(t, expected, out, "DO_NOT_TRACK env var value mismatch")
```
mdelapenya: suggestion: we prefer using require instead of assert. Please apply everywhere

modules/atlaslocal/atlaslocal_test.go
```diff
@@ -0,0 +1,626 @@
...
+	t.Helper()
+
+	exitCode, reader, err := ctr.Exec(context.Background(), []string{"sh", "-c", fmt.Sprintf("echo $%s", envVarName)})
+	require.NoError(t, err)
+	require.Equal(t, 0, exitCode)
+
+	outBytes, err := io.ReadAll(reader)
+	require.NoError(t, err)
+
+	if len(outBytes) < 8 {
```
mdelapenya: suggestion: use require.Less

modules/atlaslocal/atlaslocal_test.go
```diff
@@ -0,0 +1,626 @@
...
+
+	out := strings.TrimSpace(string(outBytes[8:]))
+	assert.Equal(t, expected, out, "DO_NOT_TRACK env var value mismatch")
+}
+
+func assertFile(t *testing.T, ctr testcontainers.Container, filename string, empty bool) {
+	t.Helper()
+
+	// Pull the log file and assert non-empty.
+	reader, _ := ctr.CopyFileFromContainer(context.Background(), filename)
```
mdelapenya: bug: check for the error here

prestonvasquez:
@mdelapenya Thanks for the thoughtful feedback! 

> do we need a separate module for Atlas, or is it fine to have an specialised container in the MongoDB module? 

I have the following concerns with combining `atlaslocal` with the `mongodb` module:

1. This would introduce namespace conflicts, i.e. `mongodb.WithUsername` does something different than `atlaslocal.WithUsername` and for safety I would recommend not combining the behavior:

```go
func WithUsername(username string) testcontainers.CustomizeRequestOption {
		if username != "" {
			req.Env["MONGODB_INITDB_ROOT_USERNAME"] = username // atlaslocal
			req.Env["MONGO_INITDB_ROOT_USERNAME"] = username   // mongodb
		}

		return nil
	}
}
```
2.  New `mongodb.WithAtlasLocal*` functional options are somewhat awkward and could be confusing: “Why doesn’t `mongodb.WithAtlasLocalUsername` work with `mongodb.Run()`?”
3. Having `mongodb.RunAtlasLocal()` pull a different image than `mongodb.Run()` is surprising and breaks user expectations. 
4. Atlas Local will likely be popular, and burying it inside the existing MongoDB documentation may risk hiding it from users.

I agree that `atlaslocal` could be more obviously represented as a `mongodb` product. Would adding a module to `mongodb` be an acceptable solution? This seems to have [precedent](https://github.com/testcontainers/testcontainers-go/tree/main/modules/gcloud/bigquery).

```go
package main 

import (
​​	"github.com/testcontainers/testcontainers-go/modules/mongodb/atlaslocal"
)

// ...
``` 
If this is acceptable, could we retain the existing documentation for this at `docs/modules/atlaslocal`? If this isn’t straightforward (e.g. embedded code examples), then I’d be happy to update the existing `mongodb` documentation to include `atlaslocal`.

mdelapenya:
@prestonvasquez thanks for your detailed response. Indeed, I think following the gcloud module pattern would be of interest, having a `mongodb/atlas` or `mongodb/atlaslocal` submodule would work. I'd start without reusing any from the mongodb module, and let the usage discover shared components.

About the docs, I'm totally fine if the entry is `MongoDB Atlas Local`. Just name it in the mkdocs.yml with that title and in the proper order, next to the existing MongoDB module. You can test the docs with `make serve-docs`.

This PR looks promising!

modules/atlaslocal/atlaslocal.go
prestonvasquez: Good idea. Suggest we use the following functions:

```go
WithMongotLogToStdout()
WithMongotLogToStderr()
WithMongotLogToFile() // always logs to "/tmp/mongot.log"
```

And then a convenience to copy (wrapping `CopyFileFromContainer`) the file from container:
```go
(*Container) CopyMongotLog(ctx context.Context) (io.ReadCloser, error) {} 
```

modules/mongodb/atlaslocal/atlaslocal.go
```diff
@@ -0,0 +1,401 @@
...
+// Container represents the MongoDBAtlasLocal container type used in the module
+type Container struct {
+	testcontainers.Container
+	username string
+	password string
+}
+
+// Run creates an instance of the MongoDBAtlasLocal container type
+func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustomizer) (*Container, error) {
+	req := testcontainers.ContainerRequest{
```
mdelapenya: suggestion: we are moving away from defining the container request here in the modules, but instead using the more modern `testcontainers.Run` function, which needs an slice of options. You basically need to convert the current req struct into the options slice.

PS: I can do this for you if you feel it's too much work for the review. I would do it adding commits on top of yours, or in a follow-up

modules/mongodb/go.mod
```diff
@@ -7,7 +7,8 @@ toolchain go1.23.6
 require (
 	github.com/stretchr/testify v1.10.0
 	github.com/testcontainers/testcontainers-go v0.38.0
-	go.mongodb.org/mongo-driver v1.13.1
+	go.mongodb.org/mongo-driver v1.17.4
+	go.mongodb.org/mongo-driver/v2 v2.3.0
```
prestonvasquez: @mdelapenya Combining atlaslocal into modules/mongodb requires that we maintain both v1 and v2 in the go.mod file. I anticipate that the Go toolchain will prune this data from the dependency topology if a user is only using modules/mongodb. Perhaps we should open a ticket to create a v2 for the mongodb testcontainer?

mdelapenya: @stevenh wdyt? I really don't like how Go manages major versions, as it comes with a lot of maintenance work

mdelapenya: @prestonvasquez sorry for the delay, I'm in a conference. Let's make progress by keeping the original approach of a separate `atlaslocal` module. The fact that the project name is relevant enough and we bypass the dependencies issue you raised, makes me think it deserves to have its own namespace. Sorry for the back and forth 🙏

matthewdale: Hi, I'm another member of the MongoDB Go Driver team. It's easy to update the MongoDB Go Driver to v2 in the `mongodb` module (see https://github.com/testcontainers/testcontainers-go/pull/3278). I'm in favor of keeping `atlaslocal` in the `mongodb` module.

@mdelapenya are there other reasons to separate the `atlaslocal` module, or just the v1/v2 driver dependency?

mdelapenya: Hi @matthewdale thanks for the bump PR. Given the driver version matches, we can keep this one as a submodule of the parent mongodb module. I just wanted to make progress and not block you with this one.

matthewdale:
@mdelapenya Preston is going to be out for a few weeks, is it possible to move this PR along while he's out?

I see that merging https://github.com/testcontainers/testcontainers-go/pull/3278 created merge conflicts in this PR. I've created a merge commit https://github.com/matthewdale/testcontainers-go/commit/ce8a9e406d7c56934fe9fc0442d5af7982525562 that resolves the conflicts. I made that commit by running:
```bash
gh pr checkout 3254
git merge main
git checkout main modules/mongodb/go.* modules/mongodb/*.go
git merge --continue
```

Are you able to resolve the merge conflicts? Are there any other changes needed before merging this PR?

mdelapenya:
@matthewdale sure, I can take over and resolve the conflicts by myself, and push the changes to this branch.

Thanks for letting me know 🙏

.github/dependabot.yml
```diff
@@ -18,6 +18,7 @@ updates:
         - /modules/aerospike
         - /modules/arangodb
         - /modules/artemis
+        - /modules/atlaslocal
```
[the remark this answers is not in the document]
mdelapenya: @coderabbitai good catch! Although the proposed solution is not valid, as atlas is part of the mongodb Go module, so simply removing the atlas entry is enough. Thanks!

mdelapenya:
@prestonvasquez @matthewdale thanks again for your work here. If possible, checking that the mongodb module also follows the best practices recommended by the vendor would be really helpful 🙏

mdelapenya:
@prestonvasquez @matthewdale the module is present in our modules catalog here: https://testcontainers.com/modules/mongodb-atlas/?language=go

Thanks!

PR #3256, opened 2025-08-11
fix(nats): use wait for listening port instead of wait for log

own2pwn:
## What does this PR do?

Updates the WaitingFor condition in test containers from

```
wait.ForLog("Listening for client connections on 0.0.0.0:4222")
```

to

```
wait.ForListeningPort(defaultClientPort),
```

## Why is it important?

The previous log message indicated only that the server had started, which might occur before the listener is ready to handle all operations.
`wait.ForListeningPort(defaultClientPort),` is a clearer signal that the nats server startup process has completed, providing a more accurate readiness check.

### DIFF
--- a/modules/nats/nats.go
+++ b/modules/nats/nats.go
@@ -33,10 +33,7 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
 		Image:        img,
 		ExposedPorts: []string{defaultClientPort, defaultRoutingPort, defaultMonitoringPort},
 		Cmd:          []string{"-DV", "-js"},
-		WaitingFor: wait.ForAll(
-			wait.ForListeningPort(defaultClientPort),
-			wait.ForLog("Server is ready"),
-		),
+		WaitingFor:   wait.ForListeningPort(defaultClientPort),
 	}
 
 	genericContainerReq := testcontainers.GenericContainerRequest{


modules/nats/nats.go
```diff
@@ -33,7 +33,7 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
 		Image:        img,
 		ExposedPorts: []string{defaultClientPort, defaultRoutingPort, defaultMonitoringPort},
 		Cmd:          []string{"-DV", "-js"},
-		WaitingFor:   wait.ForLog("Listening for client connections on 0.0.0.0:4222"),
+		WaitingFor:   wait.ForLog("Server is ready"),
```
mdelapenya: suggestion: instead of relying on the log message, we should replace it with a wait for the listening port, which is more deterministic; besides we are more into using the wait for listening port strategy more and more. Thoughts?

own2pwn: yeah, i noticed the listen port strategy is being used more and more in the project. still, i can imagine (hypothetically) a case where the listener comes up for a second and then the server crashes. do you think we better combine both or just keep wait for port?

```
WaitingFor: wait.ForAll(
  wait.ForListeningPort(defaultClientPort),
  wait.ForLog("Server is ready"),
)
```

ovyan:
good catch!

stevenh (commented):
Thanks for this, waiting on the port is positive change which we've been doing in other places. However, do we actually need the ForLog still or is ForListeningPort good enough?

own2pwn:
> Thanks for this, waiting on the port is positive change which we've been doing in other places. However, do we actually need the ForLog still or is ForListeningPort good enough?

yup, just checked and just the port check seems to be sufficient!

mdelapenya (approved):
@own2pwn LGTM, I'll merge it once the CI passes.

Thank you so much for your contribution, it makes the project much better 👏

PR #3266, opened 2025-08-21
feat(nebulagraph): add NebulaGraph module

### LINKED ISSUE #3265: [Feature]: Add NebulaGraph Module
### Problem

I’d like to request support for [NebulaGraph](https://github.com/vesoft-inc/nebula)
 as a new module in testcontainers-go.

### Solution

Add support for NebulaGraph database

### Benefit

NebulaGraph Database users will able to utilize testcontainers in their projects

### Alternatives

Nope

Yes

egasimov:
## What does this PR do?
This PR introduces a new NebulaGraph module for testcontainers-go.

## Why is it important?

It will enable NebulaGraph users utilize database as testcontainers in their project

## Related issues

- Closes #3265

### How to test this PR
Check written [tests](https://github.com/egasimov/testcontainers-go/blob/feat/nebulagraph/modules/nebulagraph/nebulagraph_test.go)

e.g
```

func TestNebulaGraphContainer(t *testing.T) {
	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Minute)
	defer cancel()

	container, err := nebulagraph.RunContainer(ctx)
	require.NoError(t, err)
	t.Cleanup(func() { _ = container.Terminate(ctx) })

	conn, err := container.ConnectionString(ctx)
	require.NoError(t, err)
	require.NotEmpty(t, conn)

	// Parse the connection string to get host and port
	host, portt, err := net.SplitHostPort(conn)
	require.NoError(t, err)

	portInt, err := strconv.Atoi(portt)
	require.NoError(t, err)
     ......
}

````

### DIFF
--- a/modules/nebulagraph/nebulagraph_test.go
+++ b/modules/nebulagraph/nebulagraph_test.go
@@ -107,6 +107,6 @@ func TestNebulaGraphContainer(t *testing.T) {
 
 		vals := row.GetValues()
 		require.NotEmpty(t, vals, "Row values should not be empty")
-		require.Equal(t, vals[0].GetIVal(), int64(1), "Expected first column to be 1")
+		require.Equal(t, int64(1), vals[0].GetIVal())
 	})
 }

### CODE REVIEW

File: modules/nebulagraph/go.mod [hunk 1]
```diff
@@ -0,0 +1,63 @@
+module github.com/testcontainers/testcontainers-go/modules/nebulagraph
```
mdelapenya, line 1: bug: I think we need to run `go mod tidy` to sync this file 🤔

mdelapenya (commented):
Hi @egasimov thanks for your work in this PR, it will bring a lot of value to the community by adding a new module. First we probably need to massage it a bit to follow some patterns in the project. Let me go through them.

First, this PR adds a `RunContainer` function that starts a default cluster. I'd rework it to have a separate `Run` function for each container type: (I intentionally removed the `d` but feel free to re-add it)
- RunGraph 
- RunStorage
- RunMeta
- RunActivator/RunConsole

This way users of the module can start each module in isolation, if needed. Please consider that I'm not an expert in the technology and you probably know much more than me about it, so it could make no sense to that. Please let me know if that's the case.

Then, we could add a `RunCluster` function that would represent the `Run` function in this PR. I'd try to follow the `Run` signature (ctx, img, options):
```go
func RunCluster(
   ctx Context,
   graphdImg string, graphdCustomizers []testcontainers.ContainerCustomizer,
   storagedImg string, storagedCustomizers []testcontainers.ContainerCustomizer,
   metadImg string, metadCustomizers []testcontainers.ContainerCustomizer,
)
```

This API structure would be more familiar to users of the project.

I'd also try to use the new `testcontainers.Run` function: instead of creating a container request to pass it to the GenericContainer function, I'd define a slice of customiser options that will be passed to the Run function. This a more modern API, as we probably deprecate GenericContainer in a short period of time.

Please let me know what you think about these changes 🙏 

PS: we have another module using a cluster approach, although it uses the same image for all the nodes. You can find it in the ETCD module.

Apart from that, I'd try to avoid replacing a docker compose setup if it brings a lot of complexity, like defining a defined dependency order

egasimov:
Hi, @mdelapenya .
Thank you for such a detailed review and sharing valuable thoughts!
Here are my answers for the given feedbacks/comments.

> First, this PR adds a RunContainer function that starts a default cluster. I'd rework it to have a separate Run function for each container type: (I intentionally removed the d but feel free to re-add it)
RunGraph
RunStorage
RunMeta
RunActivator/RunConsole

Sure, we might give a user option to run each component of NebulaGraphCluster individually, but it will require a bit of knowledge of nebula configuration it after spinning up the containers(metad, graphd, storaged), and but might bring flexibility in the cost of manual works on the user side. 🤔 We may consider it with another PR that introduces advanced Nebula Cluster as testcontainers(e.g 2 graphd, 3 storaged, 1 metad), enabling users to test specific usecases. 

> we could add a RunCluster function that would represent the Run function in this PR. I'd try to follow the Run signature (ctx, img, options):

Agree with you. it will look more familiar from of testcontainers user perspective. 👍 

> I'd also try to use the new `testcontainers.Run` function: instead of creating a container request to pass it to the GenericContainer function, I'd define a slice of customiser options that will be passed to the Run function. This a more modern API, as we probably deprecate GenericContainer in a short period of time.

Sure, definitely, if there is an modern API, we would prefer to use it. 👍 

> Apart from that, I'd try to avoid replacing a docker compose setup if it brings a lot of complexity, like defining a defined dependency order

If understood correctly, we can declare dependencies among containers and handover managing of dependencies to docker engine rather application code to take care of it.

modules/nebulagraph/options.go
```diff
@@ -0,0 +1,123 @@
...
+
+	defaultGraphdImage   = "vesoft/nebula-graphd:v3.8.0"
+	defaultMetadImage    = "vesoft/nebula-metad:v3.8.0"
+	defaultStoragedImage = "vesoft/nebula-storaged:v3.8.0"
+)
+
+const (
+	defaultMetadContainerName    = "metad0"
+	defaultGraphdContainerName   = "graphd0"
+	defaultStoragedContainerName = "storaged0"
```
mdelapenya: question: do we need container names here? I'd advocate for not setting up names, and instead define network aliases for each node. That way, you can resolve by the alias inside the defined docker network.

Container names lead to potential failures when running containers in parallel

egasimov: Yes, setting network alias is enough for containers in the same network communicate each other. 

UPDATE:
Removed static name for containers.

modules/nebulagraph/nebulagraph_test.go
```diff
@@ -0,0 +1,122 @@
...
+		ctx,
+		clientFactory,
+		&pool.ObjectPoolConfig{
+			MaxIdle:  5,
+			MaxTotal: 10,
+		},
+	)
+
+	// Test client connection and basic queries
+	t.Run("basic operations", func(t *testing.T) {
```
mdelapenya:
```suggestion
	t.Run("basic-operations", func(t *testing.T) {
```

modules/nebulagraph/options.go
```diff
@@ -0,0 +1,120 @@
...
+	graphdPort   = "9669"
+	metadPort    = "9559"
+	storagedPort = "9779"
+
+	graphdPortHTTP   = "19669"
+	metadPortHTTP    = "19559"
+	storagedPortHTTP = "19779"
+)
+
+func defaultGrapdContainerCustomizers(nw *testcontainers.DockerNetwork) []testcontainers.ContainerCustomizer {
```
mdelapenya: typo, let's change it elsewhere
```suggestion
func defaultGraphdContainerCustomizers(nw *testcontainers.DockerNetwork) []testcontainers.ContainerCustomizer {
```

modules/nebulagraph/options.go
```diff
@@ -0,0 +1,120 @@
...
+			"--local_ip=" + graphdNetworkAlias,
+			"--ws_ip=" + graphdNetworkAlias,
+			"--ws_http_port=" + graphdPortHTTP,
+			"--logtostderr=true",
+			"--redirect_stdout=false",
+			"--v=0",
+			"--minloglevel=0",
+		}...),
+		testcontainers.WithEnv(map[string]string{"USER": user}),
+		testcontainers.WithWaitStrategy(wait.ForHTTP("/status").WithPort(graphdPortHTTP + "/tcp").WithStartupTimeout(2 * time.Minute)),
```
mdelapenya: suggestion: let's define a expected status code (200 OK?) for the wait for HTTP. More in the other wait strategies.

mdelapenya: question: isn't 2 minutes too much? Not opposed, just trying to find a balanced number

modules/nebulagraph/nebulagraph.go
```diff
@@ -0,0 +1,123 @@
...
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/testcontainers/testcontainers-go"
+	"github.com/testcontainers/testcontainers-go/network"
+)
+
+// NebulaGraphCluster represents a running NebulaGraph cluster for testing
+type NebulaGraphCluster struct {
```
mdelapenya: suggestion: let's call this type simply Cluster, as the package will add `nebulagraph` as namespace:

`var cluster nebulagraph.Cluster`

modules/nebulagraph/nebulagraph.go
```diff
@@ -0,0 +1,123 @@
...
+	Graphd    testcontainers.Container
+	Metad     testcontainers.Container
+	Storaged  testcontainers.Container
+	Activator testcontainers.Container
+	Network   *testcontainers.DockerNetwork
```
mdelapenya: question: do we need all the attributes public? If we expose them, users would be able to alter the cluster state on their own, which could be undesired. As far as I see, the attributes are accessed only internally, so there is no real need to expose them, although I could be wrong.

modules/nebulagraph/nebulagraph.go
```diff
@@ -0,0 +1,123 @@
...
+	activator, err := testcontainers.Run(ctx, defaultNebulaConsoleImage, defaultActivatorContainerCustomizers(netRes)...)
+	if err != nil {
+		errs := []error{fmt.Errorf("failed to start activator container: %w", err)}
+		errs2 := terminateContainersAndRemoveNetwork(ctx, netRes, storaged, graphd, metad)
+		aggErrs := append(errs, errs2...)
+		return nil, errors.Join(aggErrs...)
+	}
+
+	// Give meta service a moment to process the registration
+	time.Sleep(10 * time.Second)
```
mdelapenya: question: is there any way to avoid this sleep, using a wait strategy on any of the nodes, probably the activator? It would be more deterministic that using 10 seconds, as this could be a different depending on the host.

modules/nebulagraph/nebulagraph.go
```diff
@@ -0,0 +1,123 @@
...
+		Metad:     metad,
+		Storaged:  storaged,
+		Activator: activator,
+		Network:   netRes,
+	}, nil
+}
+
+// ConnectionString returns the host:port for connecting to NebulaGraph graphd
+func (c *NebulaGraphCluster) ConnectionString(ctx context.Context) (string, error) {
+	host, err := c.Graphd.Host(ctx)
```
mdelapenya: question: should we check that the graphd node is not nil?

mdelapenya (commented):
Good progress! I left a few comments, but we are really close to have the PR in a good state.

Thanks for your hard work

modules/nebulagraph/options.go
egasimov: Agree, I added wait strategies and make it having more deterministic behaviour.

modules/nebulagraph/nebulagraph.go
egasimov: Agree with you,  we can make all variables inside struct as private.

modules/nebulagraph/nebulagraph.go
egasimov: Made changes to make it more deterministic.

modules/nebulagraph/nebulagraph.go
egasimov: Couldnt think a scenario where it is being null 🤔

mdelapenya: ahh users being users using the API... 😅 :

```go
c := nebulagrapth.Cluster{}
c.ConnectionString()
```

But this is defensive programming style, so not a blocker at all

mdelapenya:
@egasimov if you agree, I can send a commit updating the docs without forcing you to do more work on it. I'd add the commit on top of yours, complementing your work. wdyt?

mdelapenya:
@egasimov I added a few commits more to fix the lint, please let me know what you think about them before merging it 🙏 

I'd also add testable examples, and this gives me the sense the module was not generated using the `modulegen` tool, as recommended in https://golang.testcontainers.org/modules/ (the docs were missing some entries too). But anyway, I can help you out so we can merge this the soonest.

egasimov:
@mdelapenya Thanks for such a great contribution. 👏  I have really appreciated it. 🙏 
If you have any suggestions to make MR better, let me know, please.

mdelapenya:
@egasimov for some reason, the tests are failing:

```
FAIL TestNebulaGraphContainer/basic-operations (0.00s)
  2025/09/03 22:53:17 [DEBUG] [NebulaClient_440562a5fe] - client was not open, going to open transport before activated...
  2025/09/03 22:53:17 [DEBUG] [NebulaClient_440562a5fe] - client is opened transport, activated successfully
  2025/09/03 22:53:17 [DEBUG] [NebulaClient_[44](https://github.com/testcontainers/testcontainers-go/actions/runs/17433825810/job/49547123495?pr=3266#step:8:45)0562a5fe] - client opened transport
      nebulagraph_test.go:120: 
          	Error Trace:	/home/runner/work/testcontainers-go/testcontainers-go/modules/nebulagraph/nebulagraph_test.go:120
          	Error:      	Should be true
          	Test:       	TestNebulaGraphContainer/basic-operations
          	Messages:   	Expected to find storaged0 in ONLINE state
  2025/09/03 22:53:17 🐳 Stopping container: bf1ddf6e8158
  2025/09/03 22:53:17 🐳 Stopping container: 7bd6991f1c8a
  2025/09/03 22:53:17 🐳 Stopping container: 3de73a01d92c
```

Could you take a look? 🙏

egasimov:
@mdelapenya No worries. I have replaced query used in tests with more deterministic.
 `YIELD 1` is alternative to  `SELECT 1` in the SQL databases.
 
 Previous tests with `SHOW HOSTS` are not deterministic -- meant might behaviour itself as flaky tests in CI/CD.
 Because of nebulagraph is distributed systems, storaged's status is eventually restored.

mdelapenya:
@egasimov could you update this assertion 🙏 ? BTW no need to add a message

```
 Error: modules/nebulagraph/modules/nebulagraph/nebulagraph_test.go:110:3: expected-actual: need to reverse actual and expected values (testifylint)
  		require.Equal(t, vals[0].GetIVal(), int64(1), "Expected first column to be 1")
```

egasimov:
@mdelapenya Fixed testifylint issues, feel free to check it pls 🙏

mdelapenya:
@egasimov we are adding your contribution to the modules catalog here: https://github.com/testcontainers/community-module-registry/pull/151/files

It ~~'ll appear here~~ has been deployed here https://testcontainers.com/modules/?language=go&category=nosql-database ~~soon~~

PR #3272, opened 2025-09-02
fix: update `WithHtpasswd` to use `os.CreateTemp` instead of `os.Create` with `filepath.Join`.

jakobmoellerdev:
## What does this PR do?

when starting many distribution registries all with different `WithHtpasswd` they currently accidentally share all the same file mount that is overwritten because we always mount to a static `htpasswd`.

## Why is it important?

This makes it so every call to the credentials actually causes the creation of a truly random file mount so the files are unique per GenericContainerRequest

## Related issues

Start two containers for distribution registry with differing names and `WithHtpasswd` that differs. Observe that actually two different credentials are used and the first call is not accidentally overwritten.


jakobmoellerdev:
np, added a test case to cover this :) thanks for taking a quick look!

modules/registry/registry_test.go
```diff
@@ -9,6 +9,7 @@ import (
 
 	"github.com/cpuguy83/dockercfg"
 	"github.com/stretchr/testify/require"
+	"golang.org/x/crypto/bcrypt"
```
mdelapenya, line 12: bug: could you run `go mod tidy` from the root of the registry module? The lint phase detected the entry is missing in the go.mod file

jakobmoellerdev: sure, do you want me to bump the dep too? (were on v0.37.0 and could go to v0.41.0)

mdelapenya: Whatever works for you sounds good to me 😊

PR #3271, opened 2025-08-28
feat(modules/registry): add helper functions to pull and tag images

alegrey91:
## What does this PR do?

This PR adds a couple of helper functions to interact with the registry module:
* `PullImage`
* `TagImage`

## Why is it important?

Currently, the registry module does not support these two simple operations.
I've found them useful to have in the library, since most tests that rely on `modules/registry` will probably need to use them.

PR #3275, opened 2025-09-05
feat(registry): add helper functions to pull and tag images

mdelapenya:
## What does this PR do?

This PR adds a couple of helper functions to interact with the registry module:
* `PullImage`
* `TagImage`

When pulling images, it will use the same architecture as the image used to run the registry, as it could be using and arm-based image. Important to notice that, when inspecting an image, the field `ImageManifestDescriptor` response field is only available with containerd integration enabled. This happens on recent installations of Docker Desktop. If that field is `nil`, the pulled image will use `linux` as OS and the host's architecture.

## Why is it important?

Currently, the registry module does not support these two simple operations.
I've found them useful to have in the library, since most tests that rely on `modules/registry` will probably need to use them.

## Related issues

- Supersedes and Closes #3271

### DIFF
--- a/modules/registry/registry.go
+++ b/modules/registry/registry.go
@@ -18,7 +18,6 @@ import (
 	"github.com/docker/docker/api/types/registry"
 
 	"github.com/testcontainers/testcontainers-go"
-	"github.com/testcontainers/testcontainers-go/log"
 	"github.com/testcontainers/testcontainers-go/wait"
 )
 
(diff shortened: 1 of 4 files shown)


modules/registry/registry.go
```diff
@@ -199,12 +199,72 @@ func (c *RegistryContainer) PushImage(ctx context.Context, ref string) error {
...
+
+	inspect, err := c.Inspect(ctx)
+	if err != nil {
+		return fmt.Errorf("inspect registry container: %w", err)
+	}
+
+	platform := runtime.GOARCH
+	if inspect.ImageManifestDescriptor != nil && inspect.ImageManifestDescriptor.Platform != nil {
+		platform = inspect.ImageManifestDescriptor.Platform.Architecture
+	}
```
mdelapenya, line 228: @alegrey91 I know you're on PTO, just pinging you for future reference.

alegrey91: Appreciated!

mdelapenya:
@alegrey91 thanks for your great work on this. I was about to merge this but I thought "if tc-go already exposes the docker client, then it's probably not needed to add this to the module". But I've checked the upstream code in the rancher repo, which tells me that users would need it, and because the scope for this feature is narrowed to the module, then it's fine to provide this API at the module level.

In any case, LGTM!

PR #3276, opened 2025-09-10
fix(kafka): strip architecture suffix from Kafka image tags for semver parsing

asahasrabuddhe:
## What does this PR do?

This PR strips the architecture suffix (e.g. .arm64) from Kafka image tags.

## Why is it important?

Newer Kafka Docker images have started including architecture suffixes in
their tags (e.g. `7.5.9-1-ubi8.arm64`, `7.5.9.arm64`). These tags are
problematic because the `.arm64` suffix is not valid according to the
[Semantic Versioning](https://semver.org) specification.

- `7.5.9-1-ubi8.arm64` is **semver-compatible** because everything before
  `.arm64` (`7.5.9-1-ubi8`) is already a valid semver string with a
  pre-release identifier (`-1-ubi8`). The `.arm64` suffix, while not valid
  semver itself, is layered on top of an otherwise valid version.
- `7.5.9.arm64`, on the other hand, is **not semver-compatible** because
  the `.arm64` directly follows the patch version, making the whole string
  invalid under semver rules.

This change ensures compatibility with `golang.org/x/mod/semver` by
normalising the tag — removing the trailing architecture suffix — before
parsing. That way, we consistently fall back to a valid semver core
(e.g. `7.5.9` or `7.5.9-1-ubi8`), while still supporting images with suffixes.

## Related issues

Closes #2890 

## How to test this PR

We can try to use a newer Kafka image with a tag that contains the architecture suffix (e.g. `confluentinc/confluent-local:7.5.0.arm64` or `confluentinc/confluent-local:7.5.9.arm64`). We should no longer see the `KRaft mode is only available since version 7.4.0` error message.

### DIFF
--- a/modules/kafka/kafka_helpers_test.go
+++ b/modules/kafka/kafka_helpers_test.go
@@ -95,13 +95,19 @@ func TestValidateKRaftVersion(t *testing.T) {
 			wantErr: false,
 		},
 		{
-			name:    "Official: valid, with the amd64 architecture suffix",
+			name:    "Official: invalid, with the amd64 architecture suffix",
 			image:   "confluentinc/confluent-local:7.5.9.amd64",
-			wantErr: false,
+			wantErr: true,
 		},
 		{
-			name:  "Official: valid, with the arm64 architecture suffix",
-			image: "confluentinc/confluent-local:7.5.9.arm64",
+			name:    "Official: invalid, with the arm64 architecture suffix",
+			image:   "confluentinc/confluent-local:7.5.9.arm64",
+			wantErr: true,
+		},
+		{
+			name:    "Official: valid, without the architecture suffix",
+			image:   "confluentinc/confluent-local:7.5.9",
+			wantErr: false,
 		},
 		{
 			name:    "lacks tag",


mdelapenya (dismissed):
LGTM, thanks!

My only question would be if we should notify the Kafka folks about releasing multi-arch images to avoid breaking semver 🤔

@eddumelendez @kiview wdyt?

asahasrabuddhe:
> LGTM, thanks!
> 
> My only question would be if we should notify the Kafka folks about releasing multi-arch images to avoid breaking semver 🤔
> 
> @eddumelendez @kiview wdyt?

If not multi-arch images, they should try to fix their tagging to be semver compatible.

modules/kafka/kafka.go
```diff
@@ -212,6 +212,11 @@ func validateKRaftVersion(fqName string) error {
 		version = "v" + version
 	}
 
+	// remove the architecture suffix
+	if strings.HasSuffix(version, ".amd64") || strings.HasSuffix(version, ".arm64") {
```
mdelapenya, line 223: @asahasrabuddhe I just noticed the confluent folks do release multi-arch images, see https://hub.docker.com/r/confluentinc/cp-kafka/tags and https://hub.docker.com/r/confluentinc/confluent-local/tags

So it's a matter of choice whether passing an image which includes the architecture or the multi-arch one. I.e. `confluentinc/confluent-local:7.5.9` works before these changes.

Although these changes are harmless, I'm not sure if we should handle that "complexity" here, and let user choose the right platform 🤔  What are your thoughts on this?

asahasrabuddhe: Good point, @mdelapenya. At the very least, we should improve the error message for cases where the image tag doesn’t conform to `semver`. Currently, it’s sending an incorrect message.

asahasrabuddhe: Good point, @mdelapenya. At the very least, we should improve the error message for cases where the image tag doesn’t conform to `semver`. Right now, it’s sending an incorrect message.

mdelapenya: I'm open for suggestions. Could you elaborate on that? 🙏

modules/kafka/kafka.go
```diff
@@ -212,6 +212,11 @@ func validateKRaftVersion(fqName string) error {
...
+		version = version[:strings.LastIndex(version, ".")]
```
asahasrabuddhe: @mdelapenya How about something like this? But at this point, we might as well let them use the image by fixing the version.

```suggestion
		return fmt.Errorf("invalid image tag %q: architecture suffixes like .arm64 or .amd64 are not valid semver; please use a multi-architecture image instead", tag)
```

mdelapenya: Given we are using semver libraries to get a valid version, then we should honor it. I think this is a good compromise, failing if the version is not following semver.

mdelapenya: @asahasrabuddhe let's apply that fix, so we can merge this PR

Thanks!

PR #3302, opened 2025-09-19
feat: add network driver options

### LINKED ISSUE #3301: [Feature]: Network Driver Options
### Problem

When creating a network it's not possible to add driver options.

In our use case we want to inject a specific interface in the config of the container.
Given we attach the container to 3 networks we can't rely on it always being `eth0` as the order changes each time for the corresponding networks.
There is also no stable order given a random docker network name is used, docker uses a map for the networks where the hashing will always result in different ordering with different docker instances, etc.

There is also not the option to exec into the container to discover the interface (as we can get to know the IP address via the container networks in an inspect and then search that way) as then we already are too late (the config is needed at startup).

So to work around all of this we want to set the driver option `com.docker.network.container_iface_prefix` so the interface name is determined beforehand.

### Solution

A new `NetworkCustomizer` `WithDriverOptions(driverOpts map[string]string)` which sets a new field `DriverOptions` in `NetworkRequest`.

In the docker cli it's internally called `driverOpts`, but on the docker client type `NetworkCreateOptions` it's just called `Options` (i.e. without driver). So not sure what name you want to go for there.

### Benefit

It allows for more control over network creation.

### Alternatives

I don't immediately see any alternative implementation.

As a work-around I can create the docker network outside of testcontainers with the labels and create the reaper manually, but this is deprecated.

Yes

Wielewout:
## What does this PR do?

Used the field `Options` in `network.CreateOptions` (docker).
Also added a `DriverOptions` field to the (deprecated) `NetworkRequest` since we convert from `CreateOptions` to `NetworkRequest` and then back to `CreateOptions` in the docker provider.

## Why is it important?

This gives more control over network creation.

For example it allows to set a container interface prefix with `com.docker.network.container_iface_prefix`.
This can give stability to the interface names when needing to inject one in the config (before the container is running) when the container attaches to multiple network.
Without it there is no way of predicting which interface (`eth0`, `eth1`, ...) would correspond with which network.

## Related issues

- Closes #3301

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -1575,6 +1575,7 @@ func (p *DockerProvider) CreateNetwork(ctx context.Context, req NetworkRequest) 
 
 	nc := network.CreateOptions{
 		Driver:     req.Driver,
+		Options:    req.DriverOptions,
 		Internal:   req.Internal,
 		EnableIPv6: req.EnableIPv6,
 		Attachable: req.Attachable,
--- a/docs/features/networking.md
+++ b/docs/features/networking.md
@@ -17,6 +17,7 @@ Then, you can create a network using the `network.New` function. This function r
 - `WithAttachable()`
 - `WithCheckDuplicate()`
 - `WithDriver(driver string)`
+- `WithDriverOptions(options map[string]string)`
 - `WithEnableIPv6()`
 - `WithInternal()`
 - `WithLabels(labels map[string]string)`
--- a/network.go
+++ b/network.go
@@ -37,6 +37,7 @@ func (n DefaultNetwork) ApplyDockerTo(opts *DockerProviderOptions) {
 // NetworkRequest represents the parameters used to get a network
 type NetworkRequest struct {
 	Driver         string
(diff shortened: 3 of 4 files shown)


mdelapenya (approved):
LGTM! Before merging, I'd suggest we add a test for this, although the code is pretty simply: it just passes the options to the Docker type.

Wielewout:
> LGTM! Before merging, I'd suggest we add a test for this, although the code is pretty simply: it just passes the options to the Docker type.

@mdelapenya The only way to test that I see currently is to use the driver option `com.docker.network.container_iface_prefix` to set a custom interface name (e.g. `test_if`). We then create a small go application which provides all the interface names over http (`GET /interfaces`) and put it in a container. If `test_if0` is among the interface names, the test passes. Any thoughts?

(I'm not immediately sure when I'll find some time to try this out)

lnanctx:
Might I suggest that the test simply confirm the driver option is passed through on the request..?

PR #3308, opened 2025-09-26
fix: use path.Join instead of url.JoinPath when prepending a custom registry to an image

### LINKED ISSUE #3306: [Bug]: Incorrect TESTCONTAINERS_HUB_IMAGE_NAME_PREFIX parsing causes 404 on image pull
### Testcontainers version

v0.39.0

Yes

### Host OS

Linux

### Host arch

amd64

### Go version

1.25

### Docker version

<details>
  <summary>Collapsed</summary>

  ```shell
Client:
 Version:           28.4.0
 API version:       1.51
 Go version:        go1.24.7
 Git commit:        d8eb465
 Built:             Wed Sep  3 20:56:28 2025
 Context:           default
 Engine:
  Version:          28.4.0
  API version:      1.51 (minimum version 1.24)
  Go version:       go1.24.7
  Git commit:       249d679
  Built:            Wed Sep  3 20:58:50 2025
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v1.7.28
  GitCommit:        b98a3aace656320842a23f4a392a33f46af97866
 runc:
  Version:          1.3.0
  GitCommit:        v1.3.0-0-g4ca628d
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

</details>

### Docker info

<details>
  <summary>Collapsed</summary>

```shell
Client:
 Version:    28.4.0
 Context:    default
 Debug Mode: false
 Plugins:
    Version:  v0.28.0
    Path:     /usr/local/libexec/docker/cli-plugins/docker-buildx
    Version:  v2.39.2
    Path:     /usr/local/libexec/docker/cli-plugins/docker-compose
Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 0
 Server Version: 28.4.0
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
```

</details>

### What happened?

Regression caused by https://github.com/testcontainers/testcontainers-go/pull/2747.

## TL;DR

[`url.JoinPath` used to concat](https://github.com/testcontainers/testcontainers-go/blob/21b248e5e26506bf492fd544c5904105fd43773f/options.go#L247) `TESTCONTAINERS_HUB_IMAGE_NAME_PREFIX` and image tag results in tag being silently dropped if prefix contains a port. [Go Playground example](https://go.dev/play/p/Qyvby6z3FZ6). 

I then get 404 errors, because testcontainers-go pulls just `TESTCONTAINERS_HUB_IMAGE_NAME_PREFIX` with no tag. Example of the error:

```
Error response from daemon: error parsing HTTP 404 response body: unexpected end of JSON input: ""
```

## Long version

I'm trying to use [GitLab dependency proxy](https://docs.gitlab.com/user/packages/dependency_proxy/) for container images, so my GitLab CI doesn't exhaust Docker Hub rate limits. GitLab offers several [pre-defined variables](https://docs.gitlab.com/ci/variables/predefined_variables/) to access the proxy, including `CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX` which looks approximately like this: `gitlab.mycompany.com:443/mygroup/dependency_proxy/containers`.

> `CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX` – the direct group image prefix for pulling images through the Dependency Proxy.

The critical part is that this variable **contains a port** and **does not contain a scheme**.

- `url.Parse` parses urls with a port and no scheme wrongly. [GitHub issue](https://github.com/golang/go/issues/47955). This then results in `url.JoinPath` ignoring the second argument without an error. Repro in [Go Playground](https://go.dev/play/p/8HjldoU6qyG).
- Adding an empty `//` scheme or `https` scheme in the beginning makes Docker unhappy, as it doesn't expect a scheme. [Docker image reference format docs](https://docs.docker.com/reference/cli/docker/image/tag/).

## Solution

Use [`path.Join`](https://pkg.go.dev/path#Join) instead of `url.JoinPath`, as `url` package does not consider `gitlab.mycompany.com:443` to be a valid url.

I will be happy to create a PR.

fedorkanin:
## What does this PR do?

This MR swaps `url.JoinPath` with `path.Join` in image substitutors.

## Why is it important?

`url.JoinPath` uses `url.Parse` internally, which doesn't correctly parse URLs like `example.com:443` without a schema. Docker image reference cannot have a schema [\[docs\]](https://docs.docker.com/reference/cli/docker/image/tag/). Incorrectly parsed URL inside `url.JoinPath` causes the image tag (second argument) to be completely ignored [\[go playground\]](https://go.dev/play/p/8HjldoU6qyG), which results in 404 errors when pulling an image.

## Related issues

- Closes  #3306.
- Relates https://github.com/testcontainers/testcontainers-go/issues/2746.

## How to test this PR

I've added a test case to `image_substitutors_test.go`:

```sh
go test -v -run TestPrependHubRegistrySubstitutor
```

### DIFF
--- a/image_substitutors_test.go
+++ b/image_substitutors_test.go
@@ -67,7 +67,7 @@ func TestPrependHubRegistrySubstitutor(t *testing.T) {
 			require.Equalf(t, "my-registry/org/user/foo:latest", img, "expected my-registry/org/foo:latest, got %s", img)
 		})
 
-		t.Run("registry with port", func(t *testing.T) {
+		t.Run("registry-with-port", func(t *testing.T) {
 			s := newPrependHubRegistry("my-registry:5000")
 
 			img, err := s.Substitute("foo:latest")
--- a/image_substitutors_test.go
+++ b/image_substitutors_test.go
@@ -66,6 +66,14 @@ func TestPrependHubRegistrySubstitutor(t *testing.T) {
 
 			require.Equalf(t, "my-registry/org/user/foo:latest", img, "expected my-registry/org/foo:latest, got %s", img)
 		})
+
+		t.Run("registry with port", func(t *testing.T) {
+			s := newPrependHubRegistry("my-registry:5000")
+
+			img, err := s.Substitute("foo:latest")
+			require.NoError(t, err)
+			require.Equalf(t, "my-registry:5000/foo:latest", img, "expected my-registry:5000/foo:latest, got %s", img)
+		})
 	})
 
 	t.Run("should not prepend the hub registry to the image name", func(t *testing.T) {
--- a/options.go
+++ b/options.go
@@ -5,7 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"maps"
-	"net/url"
+	"path"
 	"time"
 
 	"dario.cat/mergo"
@@ -196,12 +196,7 @@ func (c CustomHubSubstitutor) Substitute(image string) (string, error) {
 		}
 	}
 
-	result, err := url.JoinPath(c.hub, image)
-	if err != nil {
-		return "", err
-	}
-
-	return result, nil
+	return path.Join(c.hub, image), nil
 }
 
 // prependHubRegistry represents a way to prepend a custom Hub registry to the image name,
@@ -244,12 +239,7 @@ func (p prependHubRegistry) Substitute(image string) (string, error) {
 		}
 	}
 
-	result, err := url.JoinPath(p.prefix, image)
-	if err != nil {
-		return "", err
-	}
-
-	return result, nil
+	return path.Join(p.prefix, image), nil
 }
 
 // WithImageSubstitutors sets the image substitutors for a container


image_substitutors_test.go
```diff
@@ -66,6 +66,14 @@ func TestPrependHubRegistrySubstitutor(t *testing.T) {
 
 			require.Equalf(t, "my-registry/org/user/foo:latest", img, "expected my-registry/org/foo:latest, got %s", img)
 		})
+
+		t.Run("registry with port", func(t *testing.T) {
```
mdelapenya: NIT: we are moving the subtest to this format (without spaces). Existing tests are untouched, yet, but we encourage new ones to not contain them.
```suggestion
		t.Run("registry-with-port", func(t *testing.T) {
```

mdelapenya:
Merged, thank you so much for your work here 🙇

PR #3317, opened 2025-09-27
chore: update relative-path-mode to "gitroot" in golangci configuration

mmorel-35:
## What does this PR do?

update relative-path-mode to "gitroot" in golangci configuration

## Why is it important?

Provide a clean path to files in output

## Related issues

-

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -85,5 +85,6 @@ output:
   formats:
     text:
       path: stdout
-  path-prefix: .
+run:
+  relative-path-mode: gitroot
 version: "2"


.golangci.yml
```diff
@@ -85,5 +85,6 @@ output:
   formats:
     text:
       path: stdout
-  path-prefix: .
+run:
+  relative-path-mode: gitroot
```
mdelapenya, line 89: question: according to the docs, the default value is `cfg`, the parent of the `.git` dir, which is the same as `gitroot`.

Can you share a link where I can see the issue with the file paths? 🙏

mmorel-35: When you have a linting error in module toto you a path with module/toto/module/toto/mypath.go instead of having just, module/toto/mypath.go . This is probably related to the use working-directory in
https://github.com/mmorel-35/testcontainers-go/blob/main/.github/workflows/ci-lint-go.yml

PR #3318, opened 2025-09-29
chore(azurite): use Run function

mdelapenya:
- **chore(azurite): use Run function**
- **chore(eventhubs): use Run function**
- **fix: do not fail the termination if the container is nil**
- **chore(servicebus): use Run function**

## What does this PR do?
Use Run function in the Azure modules

## Why is it important?
Migrate modules to the new API

## Related issues

- Relates #3174

### DIFF
--- a/modules/azure/servicebus/options.go
+++ b/modules/azure/servicebus/options.go
@@ -1,7 +1,9 @@
 package servicebus
 
 import (
+	"errors"
 	"io"
+	"strings"
 
 	"github.com/testcontainers/testcontainers-go"
 	"github.com/testcontainers/testcontainers-go/modules/mssql"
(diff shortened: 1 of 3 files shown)


docker.go
```diff
@@ -311,6 +311,10 @@ func (c *DockerContainer) Stop(ctx context.Context, timeout *time.Duration) erro
 //
 // Default: timeout is 10 seconds.
 func (c *DockerContainer) Terminate(ctx context.Context, opts ...TerminateOption) error {
+	if c == nil {
+		return nil
```
mdelapenya, line 315: I added this nil-guard after finding that we have hit the classic “typed-nil inside an interface” pitfall: in eventhubs (and probably other modules) `c.Container != nil` is true because the interface is non-nil, but it holds a nil *DockerContainer. Calling its method dereferences a nil receiver and panics inside the Terminate method.

PR #3420, opened 2025-10-07
chore(ollama): use Run function

mdelapenya:
## What does this PR do?

Use the Run function in ollama

## Why is it important?

Migrate modules to the new API, improving consistency and leveraging the latest testcontainers functionality.

## Related issues

- Relates to https://github.com/testcontainers/testcontainers-go/pull/3174

### DIFF
--- a/modules/ollama/local.go
+++ b/modules/ollama/local.go
@@ -90,8 +90,21 @@ type localProcess struct {
 	binary string
 }
 
-// runLocal returns an OllamaContainer that uses the local Ollama binary instead of using a Docker container.
-func (c *localProcess) run(ctx context.Context, req testcontainers.GenericContainerRequest) (*OllamaContainer, error) {
+// run returns an OllamaContainer that uses the local Ollama binary instead of using a Docker container.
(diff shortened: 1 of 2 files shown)


modules/ollama/local.go
```diff
@@ -714,7 +727,11 @@ func (c *localProcess) Customize(req *testcontainers.GenericContainerRequest) er
 	if req.WaitingFor == nil {
 		req.WaitingFor = logStrategy
 	} else {
-		req.WaitingFor = wait.ForAll(req.WaitingFor, logStrategy)
+		if w, ok := req.WaitingFor.(*wait.MultiStrategy); ok && len(w.Strategies) == 0 {
```
mdelapenya, line 730: This fix is for the case where the wait.Multistrategy's internal slice of strategies gets cleared but it's not setting the strategy to nil. We probably need to fix it in the Walk strategy

cc/ @stevenh

NathanBaulch: This looks like a bug - are you sure it's not supposed to be `len(w.Strategies) > 0`? Otherwise any existing strategies are lost...

PR #3437, opened 2025-10-09
fix(k3s): fix cross platform image loading

LaurentGoderre:
## What does this PR do?

This PR adds more robust error handling around image loading and replaces the LoadImagesWithOpt that uses `--all-platform` to a `LoadImagesWithPlatform` function that allows loading the platform correctly.

This is required because `--all-platform` fails when some digests for some platform are missing.

## Why is it important?

This is important because loading an image without specifying a platform is now broken.

### DIFF
--- a/image.go
+++ b/image.go
@@ -18,6 +18,12 @@ type saveImageOptions struct {
 
 type SaveImageOption func(*saveImageOptions) error
 
+type pullImageOptions struct {
+	dockerPullOpts client.ImagePullOptions
+}
+
+type PullImageOption func(*pullImageOptions) error
+
 // ImageProvider allows manipulating images
 type ImageProvider interface {
 	ListImages(context.Context) ([]ImageInfo, error)
--- a/modules/k3s/k3s.go
+++ b/modules/k3s/k3s.go
@@ -7,8 +7,10 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/containerd/platforms"
 	"github.com/moby/moby/api/types/container"
 	"github.com/moby/moby/api/types/mount"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"go.yaml.in/yaml/v3"
 
 	"github.com/testcontainers/testcontainers-go"
@@ -178,10 +180,12 @@ func unmarshal(bytes []byte) (*KubeConfigValue, error) {
 	return &kubeConfig, nil
 }
 
+// LoadImages imports local images into the cluster using containerd
 func (c *K3sContainer) LoadImages(ctx context.Context, images ...string) error {
-	return c.LoadImagesWithOpts(ctx, images)
+	return c.LoadImagesWithPlatform(ctx, images, nil)
 }
 
+// Deprecated: use LoadImagesWithPlatform instead
 func (c *K3sContainer) LoadImagesWithOpts(ctx context.Context, images []string, opts ...testcontainers.SaveImageOption) error {
 	provider, err := testcontainers.ProviderDocker.GetProvider()
 	if err != nil {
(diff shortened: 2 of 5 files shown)


modules/k3s/k3s.go
```diff
@@ -179,16 +181,23 @@ func unmarshal(bytes []byte) (*KubeConfigValue, error) {
 	return &kubeConfig, nil
 }
 
+// LoadImages imports a local image into the cluster using containerd
 func (c *K3sContainer) LoadImages(ctx context.Context, images ...string) error {
-	return c.LoadImagesWithOpts(ctx, images)
+	return c.LoadImagesWithPlatform(ctx, images, nil)
 }
 
-func (c *K3sContainer) LoadImagesWithOpts(ctx context.Context, images []string, opts ...testcontainers.SaveImageOption) error {
```
mdelapenya, line 185: breaking change: we cannot remove this from the API that easily. We can deprecate it and enforce the usage of the new function in the comments, including the explanation from this PR description.

LaurentGoderre: AAh gotcha!

LaurentGoderre: Fixed!

mdelapenya (commented):
LGTM, thanks! What I'd like to have is a test case for this if possible 🙏

mdelapenya (dismissed):
Actually, the code is tested by the `LoadImages` container method, so LGTM, although a test for the new function with a non-nil platform value would be desired 🙏 

In any case, I'm approving it right now.

image.go
```diff
@@ -18,10 +19,17 @@ type saveImageOptions struct {
...
+type PullImageOption func(*pullImageOptions) error
+
 // ImageProvider allows manipulating images
 type ImageProvider interface {
 	ListImages(context.Context) ([]ImageInfo, error)
 	SaveImages(context.Context, string, ...string) error
 	SaveImagesWithOpts(context.Context, string, []string, ...SaveImageOption) error
 	PullImage(context.Context, string) error
+	PullImageWithOpts(context.Context, string, ...PullImageOption) error
 }
```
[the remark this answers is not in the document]
mdelapenya: @LaurentGoderre is this breaking change needed for this PR?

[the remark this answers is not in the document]
LaurentGoderre: @mdelapenya it's needed for the test but we could not export it in the interface?

modules/k3s/k3s_test.go
```diff
@@ -39,17 +40,136 @@ func Test_LoadImages(t *testing.T) {
...
+	err = provider.PullImage(ctx, nginxImg)
 	require.NoError(t, err)
 
 	t.Run("Test load image not available", func(t *testing.T) {
 		err := k3sContainer.LoadImages(ctx, "fake.registry/fake:non-existing")
 		require.Error(t, err)
 	})
 
+	t.Run("Test load image with wrong architecture", func(t *testing.T) {
+		p, _ := platforms.Parse("linux/s390x")
```
mdelapenya, line 63:
```suggestion
		p, err := platforms.Parse("linux/s390x")
		require.NoError(t, err)
```

modules/k3s/k3s_test.go
```diff
@@ -39,17 +40,136 @@ func Test_LoadImages(t *testing.T) {
...
+
+	dockerProvider, ok := provider.(*testcontainers.DockerProvider)
+	require.True(t, ok, "provider must be a DockerProvider")
+
+	// ensure nginx image is available locally
+	err = provider.PullImage(ctx, "nginx")
+	require.NoError(t, err)
+
+	t.Run("Test load image not available", func(t *testing.T) {
+		p, _ := platforms.Parse("linux/amd64")
```
mdelapenya, line 145:
```suggestion
		p, err := platforms.Parse("linux/amd64")
		require.NoError(t, err)
```

modules/k3s/k3s_test.go
```diff
@@ -39,17 +40,136 @@ func Test_LoadImages(t *testing.T) {
...
+		err := k3sContainer.LoadImagesWithPlatform(ctx, []string{"fake.registry/fake:non-existing"}, &p)
```
mdelapenya, line 146:
```suggestion
		err = k3sContainer.LoadImagesWithPlatform(ctx, []string{"fake.registry/fake:non-existing"}, &p)
```

PR #3451, opened 2025-10-14
feat(azure): reduce time/memory by running specific sub-services

NathanBaulch:
## What does this PR do?

The default MS "azurite" container entrypoint runs the blob, queue and table services internally. The azurite Testcontainers module has an option to specify these services but it's not exposed to consumers and therefore always contains all 3 services.

This PR uses the options pattern established in the Redpanda module to allow consumers to choose the services they actually need. If only "blob" or "queue" are chosen, then the special `azurite-blob` or `azurite-queue` entrypoint is used respectively. Note that `azurite-table` is supported in the official container but is relatively new and appears to hang on shutdown for me so I'm not using that.

This also means fewer exposed ports and fewer wait strategies. FWIW, I'm seeing 30% faster startup time on my machine when running `azurite-blob` vs the full `azurite`.

## Why is it important?

Faster startup and lower memory usage, especially for the common scenario where only blob storage is needed.

### DIFF
--- a/modules/azure/azurite/azurite_test.go
+++ b/modules/azure/azurite/azurite_test.go
@@ -39,15 +39,11 @@ func TestAzurite_inMemoryPersistence(t *testing.T) {
 func TestAzurite_enabledServices(t *testing.T) {
 	ctx := context.Background()
 
-	services := []azurite.Service{azurite.BlobService, azurite.QueueService, azurite.TableService, "invalid"}
+	services := []azurite.Service{azurite.BlobService, azurite.QueueService, azurite.TableService}
 	for _, service := range services {
 		t.Run(string(service), func(t *testing.T) {
-			ctr, err := azurite.Run(ctx, "mcr.microsoft.com/azure-storage/azurite:3.33.0", azurite.WithInMemoryPersistence(0), azurite.WithEnabledServices(service))
+			ctr, err := azurite.Run(ctx, "mcr.microsoft.com/azure-storage/azurite:3.33.0", azurite.WithEnabledServices(service))
 			testcontainers.CleanupContainer(t, ctr)
-			if service == "invalid" {
-				require.Error(t, err)
-				return
-			}
 			require.NoError(t, err)
 
 			for _, srv := range services {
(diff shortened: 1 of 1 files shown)


NathanBaulch:
Goes without saying, but if this PR is accepted then the service constants in `services.go` will need to be un-deprecated. Not sure about the `ServiceURL` function, which I found useful in the unit tests.

modules/azure/azurite/azurite.go
```diff
@@ -33,74 +33,85 @@ type Container struct {
...
-				waitingFor = append(waitingFor, wait.ForListeningPort(TablePort))
+			port, err := servicePort(srv)
+			if err != nil {
+				return nil, err
 			}
+
+			cmd = append(cmd, fmt.Sprintf("--%sHost", srv), "0.0.0.0", fmt.Sprintf("--%sPort", srv), string(port))
+			exposedPorts = append(exposedPorts, string(port))
+			waitingFor = append(waitingFor, wait.ForListeningPort(port))
 		}
```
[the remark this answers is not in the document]
NathanBaulch: No need, `nat.Port` already has a `Port()` method.

mdelapenya:
Looks good to me thanks! Could you update the docs/azure file for this new option?

NathanBaulch:
Done.

docs/modules/azure.md
```diff
@@ -61,6 +61,12 @@ In example: `Run(context.Background(), "mcr.microsoft.com/azure-storage/azurite:
 
 When starting the Azurite container, you can pass options in a variadic way to configure it.
 
+#### WithEnabledServices
+
+- Since <a href="https://github.com/testcontainers/testcontainers-go/releases/tag/v0.40.0"><span class="tc-version">:material-tag: v0.40.0</span></a>
```
mdelapenya: As per https://golang.testcontainers.org/contributing/#adding-functional-options
```suggestion
- Not available until the next release <a href="https://github.com/testcontainers/testcontainers-go"><span class="tc-version">:material-tag: main</span></a>
```

modules/azure/azurite/options.go
```diff
@@ -8,12 +8,51 @@ import (
...
+type Option func(*options) error
+
+// Customize is a NOOP. It's defined to satisfy the testcontainers.ContainerCustomizer interface.
+func (o Option) Customize(*testcontainers.GenericContainerRequest) error {
+	// NOOP to satisfy interface.
+	return nil
+}
+
+// WithEnabledServices is a custom option to specify which services should be enabled.
+func WithEnabledServices(services ...Service) Option {
```
mdelapenya, line 33: suggestion: let's add a unit test for this option, i.e. to capture the error on duplicated services.

mdelapenya (approved):
LGTM! I'm approving this even though I added some comments, once addressed, and the CI pass (will pass 😉 ), this is ready to be merged.

Thank you so much for contributing to testcontainers-go, the library gets better day by day with your contributions!

Issue #1984, opened 2023-12-07
[Enhancement]: Reduce verbose log output


mfridman:
### Proposal

The default log output from this package is quite verbose. It would be ideal if the happy path (by default) logged as little as possible and reported only errors. And for those that do want verbose output, there could be a "verbose" flag or setting that could be enabled.

TL;DR - regardless if it's opt-in or opt-out, it'd be nice for this to be configurable.

---

1. I explicitly disabled this feature, no need for this output:

```
**********************************************************************************************
Ryuk has been disabled for the current execution. This can cause unexpected behavior in your environment.
More on this: https://golang.testcontainers.org/features/garbage_collector/
**********************************************************************************************
```

2. Not really interested in the Docker engine info:

```
2023/12/06 21:56:59 github.com/testcontainers/testcontainers-go - Connected to docker: 
  Server Version: 24.0.6
  API Version: 1.43
  Operating System: Docker Desktop
  Total Memory: 7944 MB
  Resolved Docker Host: unix:///var/run/docker.sock
  Resolved Docker Socket Path: /var/run/docker.sock
  Test SessionID: be5bcf090d5ac5b55046d7a35a670ab7341c71919218f6ee1e64ea4190322e69
  Test ProcessID: 469ede8f-9797-46fb-9ce8-eccb642a0d47
```

3. This is cute (hehe), but not strictly necessary:

```
2023/12/06 21:50:06 🐳 Creating container for image postgres:16-alpine
2023/12/06 21:50:06 ✅ Container created: f40dcefbd017
2023/12/06 21:50:06 🐳 Starting container: f40dcefbd017
2023/12/06 21:50:07 ✅ Container started: f40dcefbd017
2023/12/06 21:50:07 🚧 Waiting for container id f40dcefbd017 image: postgres:16-alpine. Waiting for: &{timeout:<nil> deadline:0xc0001a6e98 Strategies:[0xc0003b1a40]}

[...]

2023/12/06 22:15:03 🐳 Terminating container: f40dcefbd017
2023/12/06 22:15:03 🚫 Container terminated: f40dcefbd017
```

I added a custom logger, but it only suppressed items 2 and 3 above.

```go
type nopLogger struct{}

var _ testcontainers.Logging = nopLogger{}

func (nopLogger) Printf(string, ...interface{}) {}
```

ntin90:
I work around by setting the global Logger of tescontainers to ioutils.NopWriter, as following:
```golang
[...]

PR #3459, opened 2025-10-30
fix: should print max information by default

### LINKED ISSUE #2887: [Regression]: Missing container logs when container exited
### Testcontainers version

0.32.0

No

### Host OS

Linux

### Host arch

ARM

### Go version

1.22.4

### Docker version

```shell
Running on CI, cannot run this there
```

### Docker info

```shell
Running on CI, cannot run this there
```

### What happened?

Somewhere between versions v0.27.0 and v0.32.0 this library stopped printing out container logs whenever container failed and now I just have "failed to start container: container exited with code 1" in output. Please return previous behavior that helped troubleshooting.

### Relevant log output

```shell
failed to start container: container exited with code 1
```

### Additional information

_No response_

strowk:
When tests do not fail, what they printed rarely matters. However, when they do fail, it is important to see maximum information. This change switches default behavior back as was in 0.31, when failures could be quickly triaged without adding -v

label: fix, qol

## What does this PR do?

Simply switches default logging behavior to more convenient one in realworld scenarios.

## Why is it important?

Test output matters when we need to read it, we read it when tests fail, otherwise we might not even care if they run (if it is some pipeline).

## Related issues

- Closes #2887

## How to test this PR

https://github.com/testcontainers/testcontainers-go/issues/2887#issuecomment-3469359308

Ideally maybe it should buffer logs and print them out only when failure happened. This could affect memory usage though, so IMO defaulting to complete output would be preferred behavior in majority of the cases.


mdelapenya:
Hi @strowk I asked in #1984 about feedback, although I understand what you described in the issue and PR. This is key to me:

> Test output matters when we need to read it, we read it when tests fail,

I'm fine in merging the changes proposed in this PR, I just want some alignment in the community about it.

log/logger.go
```diff
@@ -19,16 +19,17 @@ type Logger interface {
...
 func init() {
-	// Enable default logger in the testing with a verbose flag.
+	// Disable default logger in testing mode if explicitly disabled via -test.v=false.
 	if testing.Testing() {
-		// Parse manually because testing.Verbose() panics unless flag.Parse() has done.
+		// Disable logging if explicitly disabled via -test.v=false
 		for _, arg := range os.Args {
-			if strings.EqualFold(arg, "-test.v=true") || strings.EqualFold(arg, "-v") {
-				defaultLogger = log.New(os.Stderr, "", log.LstdFlags)
+			if strings.EqualFold(arg, "-test.v=false") {
```
ar-sematell, line 34: suggestion: Create a public constructor for the `noopLogger`, or remove it and use `log.New(io.Discard, "", 0)`

mdelapenya: @strowk what are your thoughts on this?

strowk: Easy enough - https://github.com/testcontainers/testcontainers-go/pull/3459/commits/fba533ff23797b787a68bc8996a418827e419f95

braaar:
I agree with @strowk. The logs from testcontainers are noise while tests pass, but useful when tests fail.

ar-sematell:
> I agree with @strowk. The logs from testcontainers are noise while tests pass, but useful when tests fail.

It would be helpful to have an easy way to set the `testing.T` common func `Logf(format string, args ...any)` as a logger for the test containers, this way they would only print if the test failed.

```go
type testLogger interface {
	Logf(format string, args ...any)
}

type testLoggerWrapper struct {
	testLogger
}

func (w testLoggerWrapper) Printf(format string, v ...any) {
	w.Logf(format, v...)
}

func WithTestLogger(t testLogger)LoggerOption {
	return WithLogger(testLoggerWrapper{t})
}
```

using with

```go
func runNewPostgres(t *testing.T) (*testcontainersPostgres.PostgresContainer, error) {
	return testcontainersPostgres.Run(
		t.Context(),
		postgresImage,
		testcontainers.WithTestLogger(t),
		testcontainersPostgres.WithDatabase(PostgresDB),
		testcontainersPostgres.WithUsername(PostgresUser),
		testcontainersPostgres.WithPassword(PostgresPassword),
		testcontainersPostgres.BasicWaitStrategies(),
	)
}
```

mdelapenya:
> > I agree with @strowk. The logs from testcontainers are noise while tests pass, but useful when tests fail.
> 
> It would be helpful to have an easy way to set the `testing.T` common func `Logf(format string, args ...any)` as a logger for the test containers, this way they would only print if the test failed.

@ar-sematell I think this could be a nice addition. Could you send this in a different PR? 🙏

mdelapenya (approved):
LGTM. thanks for sending this and sorry for the delay in the review

PR #3482, opened 2025-11-03
fix: docker auth for docker.io images

LaurentGoderre:
## What does this PR do?

When doing a Docker pull on an image with the default registry `docker.io`, the auth helper doesn't match the image to the authentication data stored by the engine under `https://index.docker.io/v1` causing the error: `credentials not found in native keychain`

## Why is it important?

This is important to support the default registry prefix.

## How to test this PR

Unit test added


docker_auth.go
```diff
@@ -42,6 +43,11 @@ func dockerImageAuth(ctx context.Context, image string, configs map[string]regis
 	defaultRegistry := defaultRegistryFn(ctx)
 	reg := core.ExtractRegistry(image, defaultRegistry)
 
+	// Make comparison case-insensitive and handle both aliases
+	if strings.EqualFold(reg, "docker.io") || strings.EqualFold(reg, "registry.hub.docker.com") {
+		reg = defaultRegistry
+	}
+
```
stevenh: question: should this be handled by `core.ExtractRegistry`?

LaurentGoderre: It could. Up to y'all

stevenh: I think it should live there, @mdelapenya thoughts?

mdelapenya: Yeah, let's move it there, and once we use the go-sdk, it will be resolved and easier to replace

LaurentGoderre: @stevenh @mdelapenya done!

mdelapenya (approved):
LGTM, thanks!

PR #3484, opened 2025-11-04
fix(udp): map udp ports correctly

blueprismo:
## What does this PR do?

Fixes UDP port exposure.

Problem:
When using ExposedPorts: []string{"8080/udp"} in testcontainers-go, the MappedPort() function always returned "0/udp" instead of the actual mapped host port (e.g., "55051/udp"). This made it impossible to connect to UDP services running in containers.

Root Cause:
The issue was in the [mergePortBindings()](https://github.com/testcontainers/testcontainers-go/blob/75272033cf089a08dbeaeb65cb9b312c7fadd557/lifecycle.go#L600) function in lifecycle.go. When [nat.ParsePortSpecs()](https://github.com/testcontainers/testcontainers-go/blob/main/lifecycle.go#L537) processes exposed port specifications like `8080/udp`, it creates PortBinding structs with empty HostPort fields `({HostIP: "", HostPort: ""})`
  Docker interprets these differently:
  - HostPort: "" (empty) → "Don't bind this port to any host port" → results in port 0
  - HostPort: "0" → "Bind this port to a random available host port" → proper allocation

Solution:
  Modified mergePortBindings() to convert empty HostPort values to "0" for automatic port allocation:

## Why is it important?

To support UDP port bindings in testcontainers.

## Related issues
 
n/a

## How to test this PR

I added a test scenario for udp testing

PR #3485, opened 2025-11-04
fix(udp): expose UDP ports properly

blueprismo:
## What does this PR do?

Fixes UDP port exposure.

Problem:
When using ExposedPorts: []string{"8080/udp"} in testcontainers-go, the MappedPort() function always returned "0/udp" instead of the actual mapped host port (e.g., "55051/udp"). This made it impossible to connect to UDP services running in containers.

Root Cause:
The issue was in the [mergePortBindings()](https://github.com/testcontainers/testcontainers-go/blob/75272033cf089a08dbeaeb65cb9b312c7fadd557/lifecycle.go#L600) function in lifecycle.go. When [nat.ParsePortSpecs()](https://github.com/testcontainers/testcontainers-go/blob/main/lifecycle.go#L537) processes exposed port specifications like 8080/udp, it creates PortBinding structs with empty HostPort fields ({HostIP: "", HostPort: ""})
Docker interprets these differently:

HostPort: "" (empty) → "Don't bind this port to any host port" → results in port 0
HostPort: "0" → "Bind this port to a random available host port" → proper allocation
Solution:
Modified mergePortBindings() to convert empty HostPort values to "0" for automatic port allocation:

## Why is it important?

To support UDP connections correctly as now they aren't useful (at least with random port allocatio

## Related issues

Follows: https://github.com/testcontainers/testcontainers-go/pull/3484

### DIFF
--- a/lifecycle_test.go
+++ b/lifecycle_test.go
@@ -135,7 +135,7 @@ func TestPreCreateModifierHook(t *testing.T) {
 			"80/tcp": []nat.PortBinding{
 				{
 					HostIP:   "",
-					HostPort: "",
+					HostPort: "0",
 				},
 			},
 		}, inputHostConfig.PortBindings,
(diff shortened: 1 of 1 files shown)


mdelapenya:
@blueprismo i think we need to update the tests for the new condition. Can you take a look? 🙏

PR #3552, opened 2026-02-06
feat(chroma): update to chroma 1.x

tazarov:
What does this PR do?                                                                                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                           
Updates the Chroma module from chromadb/chroma:0.4.24 to chromadb/chroma:1.4.0 and bumps chroma-go from v0.1.2 to v0.3.2.                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                                                                           
Chroma 1.x ships a v2 REST API, so this PR:

- Switches the client import from github.com/amikos-tech/chroma-go to github.com/amikos-tech/chroma-go/pkg/api/v2
- Changes the heartbeat wait strategy to /api/v2/heartbeat and drops the wait.ForLog("Application startup complete") check (Chroma 1.x no longer emits that log)
- Rewrites tests and examples for the new NewHTTPClient/WithBaseURL constructor and updated method signatures
- Adds v1-config.yaml so the collections example can set allow_reset: true (Chroma 1.x reads config from a YAML file instead of env vars)
- Adds Close() calls to clean up client resources

Why is it important?

The module was pinned to Chroma 0.4.x, which is no longer maintained. Chroma 1.x changed the REST API and Go client in backwards-incompatible ways, so the module stopped working against current Chroma releases.

Related issues

N/A

### DIFF
--- a/modules/chroma/examples_test.go
+++ b/modules/chroma/examples_test.go
@@ -131,7 +131,6 @@ func ExampleChromaContainer_collections() {
 	fmt.Printf("Reset successful\n")
 
 	// createCollection {
-	// for testing we use a dummy hashing function NewConsistentHashEmbeddingFunction
 	col, err := chromaClient.GetOrCreateCollection(context.Background(), "test-collection",
 		chromago.WithCollectionMetadataCreate(
 			chromago.NewMetadata(
@@ -151,7 +150,6 @@ func ExampleChromaContainer_collections() {
 	// addData {
 	// verify it's possible to add data to the collection
 	err = col.Add(context.Background(),
-		// chroma.WithIDGenerator(chroma.NewULIDGenerator()),
 		chromago.WithIDs("1", "2"),
 		chromago.WithTexts("hello world", "goodbye world"),
 		chromago.WithMetadatas(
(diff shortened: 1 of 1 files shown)


modules/chroma/examples_test.go
```diff
@@ -60,29 +64,46 @@ func ExampleChromaContainer_connectWithClient() {
...
 	ctx := context.Background()
-
-	chromaContainer, err := chroma.Run(ctx, "chromadb/chroma:0.4.24", testcontainers.WithEnv(map[string]string{"ALLOW_RESET": "true"}))
+	cwd, err := os.Getwd()
+	if err != nil {
+		log.Printf("failed to get current working directory: %s", err)
+		return
+	}
+	chromaContainer, err := chroma.Run(ctx, "chromadb/chroma:1.4.0", testcontainers.WithEnv(map[string]string{"ALLOW_RESET": "true"}),
+		testcontainers.WithHostConfigModifier(func(hostConfig *container.HostConfig) {
```
mdelapenya: suggestion: instead of using the host config modifier to mount a file, we recommend using the WithFiles API instead. Copying files into the container makes it more reproducible that adding an externally modifiable file through a volume/mount 

Also, please add the config file inside testdata dir, instead of the current working dir.

mdelapenya (changes requested):

modules/chroma/examples_test.go
tazarov: Got used to the old ways of mounting stuff. Updating now :)

tazarov: Done

PR #3574, opened 2026-03-07
feat: improve container conflict detection

### LINKED ISSUE #3569: [Enhancement]: Improve container conflict detection compatibility
### Proposal

The container conflict detection used when creating the reaper container or re-using a container is based on the following [regex](https://github.com/testcontainers/testcontainers-go/blob/main/docker.go#L57):

```go
regexp.MustCompile("Conflict. The container name .* is already in use by container .*")
```

This is incompatible with Podman which outputs the following as part of the error:

```
creating container storage: the container name "reaper_26b1150e0c8a8c3d573ac1bbbdc037b514cb767dcc8d7216abf682ed16d5adfd" is already in use by 87d730cdfb5697a8b5379554ec47aa84ebb0003c2589e4f51bab8473a54092e6.
```

This is referenced in their [test suite](https://github.com/containers/podman/blob/v5.8.0/test/system/030-run.bats#L1706).

Considering the similarity between the two messages, it would be nice to slightly alter that regex for improved compatibility.

Desuuuu:
## What does this PR do?

Modify the container conflict detection regex for improved Podman compatibility.

The shape of the conflict error comes from Podman's [test suite](https://github.com/containers/podman/blob/v5.8.0/test/system/030-run.bats#L1706).

## Why is it important?

Container creation conflicts happen fairly often when creating the reaper container in parallel tests. Making the retry logic work with the Podman provider can prevent some of these errors from happening.

## Related issues

- Closes #3569

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -54,7 +54,7 @@ const (
 
 var (
 	// createContainerFailDueToNameConflictRegex is a regular expression that matches the container is already in use error.
-	createContainerFailDueToNameConflictRegex = regexp.MustCompile("Conflict. The container name .* is already in use by container .*")
+	createContainerFailDueToNameConflictRegex = regexp.MustCompile("[Tt]he container name .* is already in use by .*")
 
 	// minLogProductionTimeout is the minimum log production timeout.
 	minLogProductionTimeout = time.Duration(5 * time.Second)
--- a/reaper_test.go
+++ b/reaper_test.go
@@ -3,12 +3,15 @@ package testcontainers
 import (
 	"context"
 	"errors"
+	"fmt"
 	"os"
 	"strconv"
 	"sync"
+	"syscall"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


docker.go
```diff
@@ -54,7 +54,7 @@ const (
 
 var (
 	// createContainerFailDueToNameConflictRegex is a regular expression that matches the container is already in use error.
-	createContainerFailDueToNameConflictRegex = regexp.MustCompile("Conflict. The container name .* is already in use by container .*")
+	createContainerFailDueToNameConflictRegex = regexp.MustCompile("[Tt]he container name .* is already in use by .*")
```
mdelapenya, line 57: LGTM, although we should look for another way to detect the conflicts, logs are fragile. cc/ @stevenh

mdelapenya (approved):
LGTM, thanks!

PR #3591, opened 2026-03-10
chore!: migrate to moby modules

thaJeztah:
## What does this PR do?

Migrate the Docker client dependency from the deprecated `github.com/docker/docker` (v28, incompatible) to the new `github.com/moby/moby` modular packages (`github.com/moby/moby/api` + `github.com/moby/moby/client`). This is a **breaking change** — several public API types and method signatures change.

### Breaking changes to the Testcontainers API

**`Container` interface**

| Method | Old | New |
|--------|-----|-----|
| `PortEndpoint` | `PortEndpoint(ctx, port nat.Port, proto string)` | `PortEndpoint(ctx, port string, proto string)` |
| `MappedPort` | `MappedPort(ctx, nat.Port) (nat.Port, error)` | `MappedPort(ctx, string) (network.Port, error)` |
| `Ports` | `Ports(ctx) (nat.PortMap, error)` | `Ports(ctx) (network.PortMap, error)` |
| `LivenessCheckPorts` | returns `nat.PortSet` | returns `network.PortSet` |

**`ImageBuildInfo` interface & `FromDockerfile` struct**

| | Old | New |
|-|-----|-----|
| `BuildOptions()` | returns `build.ImageBuildOptions` | returns `client.ImageBuildOptions` |
| `BuildOptionsModifier` | `func(*build.ImageBuildOptions)` | `func(*client.ImageBuildOptions)` |

**`DockerClient` methods**

| Method | Old | New |
|--------|-----|-----|
| `Info` | `Info(ctx) (system.Info, error)` | `Info(ctx, client.InfoOptions) (client.SystemInfoResult, error)` |
| `Ping` | `Ping(ctx) (types.Ping, error)` | `Ping(ctx, client.PingOptions) (client.PingResult, error)` |
| `Events` | returns `(<-chan Message, <-chan error)` | returns `client.EventsResult` |
| `DiskUsage` | `(types.DiskUsageOptions) (types.DiskUsage, error)` | `(client.DiskUsageOptions) (client.DiskUsageResult, error)` |
| `RegistryLogin` | `(auth registry.AuthConfig) (registry.AuthenticateOKBody, error)` | `(client.RegistryLoginOptions) (client.RegistryLoginResult, error)` |

**Type replacements**

| Old (`docker/docker` + `docker/go-connections`) | New (`moby/moby`) |
|--|--|
| `nat.Port` (string) | `network.Port` (struct) or plain `string` in method args |
| `nat.PortMap` | `network.PortMap` |
| `nat.PortSet` | `network.PortSet` |
| `nat.PortBinding` | `network.PortBinding` (`.HostIP` is `netip.Addr`, not `string`) |
| `build.ImageBuildOptions` | `client.ImageBuildOptions` |

### Other changes

- Fix inverted port condition in HTTP wait strategy
- Fix `MergeCustomLabels` to validate all keys before merging (eliminates flaky test)
- Fix `mergePortBindings` to preserve user-defined `HostIP`/`HostPort` from `HostConfigModifier`
- Protect `DockerContainer.isRunning` with `atomic.Bool` to prevent data races
- Retry network removal on active endpoints race (nebulagraph)
- Restore stdcopy multiplexing for ollama local process
- Skip host network tests on colima (not just Docker Desktop)
- All 66 modules updated with new `go.mod` / `go.sum`

## Why is it important?

The `github.com/docker/docker` module has been deprecated in favor of the properly-versioned `github.com/moby/moby` modular packages. The new packages provide:

- Proper Go module versioning (no more `+incompatible`)
- Structured port types (`network.Port`) instead of string-based `nat.Port`, catching errors at compile time
- `netip.Addr` for host IPs instead of raw strings
- Cleaner API surfaces with result wrapper types

Staying on the deprecated module blocks upgrading to newer Docker Engine APIs and accumulates tech debt as the ecosystem moves to the moby packages.

## Related issues

- Supersedes #3617

## How to test this PR

1. Run the core test suite: `go test -race ./...`
2. Run module tests (each module has its own test suite requiring Docker)
3. Verify `make lint-all` passes across all 66 modules
4. Key tests to validate the migration:
   - `TestTwoContainersExposingTheSamePort`, `TestContainerCreation`, `TestContainerCreationWithName` — port handling
   - `TestPreCreateModifierHook` — port binding merge behavior
   - `TestDockerProvider_waitContainerCreation_retries` — container list result type
   - `TestGenericReusableContainerInSubprocess` — container list `.Items` usage
   - `TestUDPPortBinding` — port type changes

## Follow-ups

- **Blocking merge**: `TestContainerTerminationResetsState` and `Test_MultiContainerLogConsumer_CancelledContext` fail consistently with `-race` due to a data race **upstream in `moby/moby/client`** (`cancelReadCloser` in [`utils.go:139,152`](https://github.com/moby/moby/blob/76f6a0c21d54/client/utils.go#L139)). The `cancelReadCloser` struct has an unsynchronized field written during construction and read during the cancel callback's `Close()`, triggered via `ContainerLogs()`. This needs a fix in the `moby/moby/client` package before this PR can merge with `-race` enabled.


aausch:
@thaJeztah I had some extra cycles so went through as many of the outstanding issues as I could and tried to fix them. 

Hoping to help accelerate this migration.

Each change _should_ be in a self-contained commit just in case I got them wrong, to make it easy to review, cherry-pick, discard, etc.

udp_port_binding_test.go
```diff
@@ -54,7 +54,7 @@ func TestUDPPortBinding(t *testing.T) {
 
 		portNum := mappedPort.Num()
 		assert.Positive(t, portNum, "Port number should be greater than 0")
-		assert.LessOrEqual(t, portNum, 65535, "Port number should be valid UDP port range")
+		assert.LessOrEqual(t, portNum, uint16(65535), "Port number should be valid UDP port range")
```
thaJeztah: This assert can probably be dropped overall; it'd be extremely unlikely for a `uint16` to be larger than `65535` 😂

modules/chroma/go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/testcontainers/testcontainers-go/modules/chroma
 
 go 1.25.0
 
+toolchain go1.25.7
+
```
thaJeztah, line 6: Curious; was there a reason to enforce?

mdelapenya: Just consistency with the rest of the modules, but fine if we decide to remove them all

aausch: I thought this was a supply chain security requirement. 

Considering the volume of supply chain attacks recently, the industry is converging on pinning all required versions (and checking signatures wherever possible). Eg https://www.malwarebytes.com/blog/news/2026/03/axios-supply-chain-attack-chops-away-at-npm-trust - maintainer access token was compromised despite 2fa/mfa

aausch:
preemtive 🥂

mdelapenya (approved):
LGTM, I marked this as a breaking changes for the release notes, so that we include the migration path for the changed APIs

Thanks everybody participating in this, specially @aausch and @appleboy

appleboy:
@mdelapenya Waiting for next version. Thanks all.

wait/sql.go
```diff
@@ -111,7 +114,7 @@ func (w *waitForSQL) WaitUntilReady(ctx context.Context, target StrategyTarget)
 		}
 	}
 
-	db, err := sql.Open(w.Driver, w.URL(host, port))
+	db, err := sql.Open(w.Driver, w.URL(host, port.String()))
```
ShawnToubeau, line 117: Hi! I have a question about this change because my team is using the wait.ForSQL strategy.

Before, when `port` was `nat.Port` we could get the port number using `port.Port()` but now `port.String()` includes the protocol which means we have to parse the port ourselves. Does it make sense to either

1. Change the port type in the callback to `network.Port` 
or
1. Instead use `port.Port()` here to only pass the port number and exclude the protocol?

Thanks!

thaJeztah: Oh! Hm.. good question; I do recall that originally I changed some signatures to be more strongly typed (i.e., require `network.Port` as argument), but the number of places where a plain string was passed in, and now requiring to either use a `network.MustParsePort` or `network.ParsePort` was ... quite large, so I went back and made it loosely typed on the way in (and parse internally).

_However_ this part is a callback, so the user isn't bothered by having to do the conversion, so perhaps we should change the signature to accept a `network.Port` as type here; while I expect most cases the port to be a TCP port, theoretically it could be a different protocol; I see there's at least a couple of tests passing in a port with a protocol; so passing ONLY the `.Port()` would mean we'd lose that information https://github.com/testcontainers/testcontainers-go/blob/c251392bd68e906f4a8c351fef222ac5abb0e610/wait/sql_test.go#L15-L19


So perhaps changing it to a `network.Port` instead of a string would be best? That way the callback function wouldn't need to parse the port (again), and no information would be lost. Code that depended on `nat.Port` being a string (with extras) may need to be updated (either use `port.String()` or `port.Port()` depending on what's needed), but ... probably acceptable?

thaJeztah: - opened https://github.com/testcontainers/testcontainers-go/pull/3650 for consideration 😅

ShawnToubeau: Hey! Thanks for the quick response and PR! 

Changing the type to `network.Port` makes sense to me!

mdelapenya:
> Thanks for all the hard work 🙏
>
> Quick anecdote for your reference, but I spent the last 20 minutes fixing these breaking changes in main after a teammate merged a dependabot PR that just showed a simple minor version bump. Might be good to re-evaluate the semver usage strategy for this project

@alyosha we try to do our best, and because this project is not on v1, we from time to time consider adding a "breaking change" for the overall benefit of a better API.

Said that, I'm sorry to here that it took that long to resolve the conflicts, we tried (actually @thaJeztah did) to make them very explicit in the PR description, but it could be the case we missed something.

PR #3660, opened 2026-04-24
fix: wait for log production goroutine to drain on stop

### LINKED ISSUE #2887: [Regression]: Missing container logs when container exited

mdelapenya:
## What does this PR do?

Adds a logProductionDone channel that the goroutine closes on exit, and waits on it (with a bounded timeout) in stopLogProduction() before returning.

## Why is it important?

stopLogProduction() was cancelling the context and returning immediately, creating a race where in-flight log data could be dropped when Terminate() fired during active log streaming. Confirmed data loss of up to 770/1000 lines in reproduction tests.

## Related issues

Fixes #2887

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -87,6 +87,8 @@ type DockerContainer struct {
 	// logProductionCancel is used to signal the log production to stop.
 	logProductionCancel context.CancelCauseFunc
 	logProductionCtx    context.Context
+	// logProductionDone is closed when the log production goroutine exits.
+	logProductionDone chan struct{}
 
 	logProductionTimeout *time.Duration
 	logger               log.Logger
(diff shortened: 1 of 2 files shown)


mdelapenya:
Re: the P1 review comment on `stopLogProduction` (unbounded post-cancel wait):

Valid concern — fixed in 56f30d51.

Context cancellation propagates into the Docker transport and normally unblocks `stdcopy.StdCopy` within milliseconds (`timeoutCtx` passed to `ContainerLogs` is a child of `logProductionCtx`, so it is cancelled immediately). However, a stuck kernel socket read or daemon transport failure could prevent unblocking, leaving the wait truly unbounded.

Added a `select` with `time.After(minLogProductionTimeout)` (5 s) on the post-cancel receive, matching the pre-cancel timeout. Both waits are now bounded at 5 s each, so `stopLogProduction` can block at most ~10 s total under worst-case transport failure.

mdelapenya:
Re: CodeRabbit review comment on `docker.go:911-940` (pre-cancel 5 s wait regression):

The claim that "cancel first, then wait for drain" is equally safe is **incorrect**. Tested empirically with cancel-first: 4 out of 10 runs dropped logs (up to 850/1000 lines dropped).

Root cause is in `stdcopy.StdCopy` itself. When the context is cancelled, the Docker HTTP response body's `Read()` returns immediately with an error — `StdCopy` propagates that error and discards the remainder of the stream. It does **not** flush buffered data first. From the source:

```go
nr2, err = multiplexedSource.Read(buf[nr:])
// ...
if err != nil {
    return 0, err  // drops everything, no flush
}
```

The pre-cancel wait is load-bearing for correctness on the `Terminate()` path. The reviewer's `StopLogProducer()` regression concern is valid (5 s extra latency on still-running containers), but that method is already deprecated and the tradeoff favours zero data loss. Keeping the current wait-before-cancel approach.

PR #3672, opened 2026-05-07
fix: close temp file handle before removal 

acouvreur:
## What does this PR do?

When using LoadFile on `dind` or `k3s`, the temporary file descriptor is not closed.

On some OS (such as windows), this fails upon deletion of the temp file.

## Related issues

-


mdelapenya (approved):
LGTM! I recently saw something similar for other project on Windows, so this makes sense, thanks!

PR #3716, opened 2026-06-09
feat(k3s):  pull image opts

blueprismo:
## What does this PR do?

Allows the caller to use PullImageWithOpts (usually select a platform, but can also use any of the current (and future) `client.ImagePullOptions` options!)  so they can pull images

Refactors the recently introduced `PullImageWithPlatform` in favour of `PullImageWithOpts`

Adds the `PullDockerImageWithPlatform` 

## Why is it important?

Needed for some k3s tests, allowing poli-arch setups to work.

## Related issues

- Closes #3437 

`provider.PullImageWithOpts(ctx, img, testcontainers.PullDockerImageWithPlatform(p))`

### DIFF
--- a/docs/features/common_functional_options.md
+++ b/docs/features/common_functional_options.md
@@ -322,61 +322,6 @@ Using the `WithImageSubstitutors` options, you could define your own substitutio
 
 If you need to set the platform for a container, you can use `testcontainers.WithImagePlatform(platform string)`.
 
-##### PullDockerImageWithPlatform
-
-- Not available until the next release <a href="https://github.com/testcontainers/testcontainers-go"><span class="tc-version">:material-tag: main</span></a>
-
-If you need to pull an image for a specific OCI platform, you can use `testcontainers.PullDockerImageWithPlatform` with [DockerProvider.PullImageWithOpts](https://pkg.go.dev/github.com/testcontainers/testcontainers-go#DockerProvider.PullImageWithOpts).
-
-```golang
-platform, err := platforms.Parse("linux/amd64")
-if err != nil {
-    // handle error
-}
-
-provider, err := testcontainers.ProviderDocker.GetProvider()
-if err != nil {
-    // handle error
-}
-
-if err := provider.PullImageWithOpts(
-    ctx,
-    "nginx:mainline",
-    testcontainers.PullDockerImageWithPlatform(platform),
-); err != nil {
-    // handle error
-}
-```
-
-##### SaveDockerImageWithPlatforms
-
-- Since <a href="https://github.com/testcontainers/testcontainers-go/releases/tag/v0.42.0"><span class="tc-version">:material-tag: v0.42.0</span></a>
-
-If you need to export an image for specific OCI platforms, you can use `testcontainers.SaveDockerImageWithPlatforms` with [DockerProvider.SaveImagesWithOpts](https://pkg.go.dev/github.com/testcontainers/testcontainers-go#DockerProvider.SaveImagesWithOpts).
-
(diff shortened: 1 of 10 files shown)


image.go
```diff
@@ -18,11 +18,17 @@ type saveImageOptions struct {
...
+type PullImageOption func(*pullImageOptions) error
+
 // ImageProvider allows manipulating images
 type ImageProvider interface {
 	ListImages(context.Context) ([]ImageInfo, error)
 	SaveImages(context.Context, string, ...string) error
 	SaveImagesWithOpts(context.Context, string, []string, ...SaveImageOption) error
 	PullImage(context.Context, string) error
-	PullImageWithPlatform(context.Context, string, string) error
+	PullImageWithOpts(context.Context, string, ...PullImageOption) error
```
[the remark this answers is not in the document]
blueprismo: The unexported `pullImageOptions` only matters if 3rd parties needs new kinds of pull options that testcontainers hasn’t exported yet. So far we have the `PullDockerImageWithPlatform`.

`PullDockerImageWithRawOptions` would leak moby client types and isn't needed for our platform-specific pull + k3s load workflow. We can add more exported option helpers later if a concrete need arises (similar to save)

modules/k3s/k3s.go
```diff
@@ -178,10 +180,12 @@ func unmarshal(bytes []byte) (*KubeConfigValue, error) {
 	return &kubeConfig, nil
 }
 
+// LoadImages imports local images into the cluster using containerd
 func (c *K3sContainer) LoadImages(ctx context.Context, images ...string) error {
-	return c.LoadImagesWithOpts(ctx, images)
+	return c.LoadImagesWithPlatform(ctx, images, nil)
 }
 
+// Deprecated: use LoadImagesWithPlatform instead
```
mdelapenya: question: I'm not sure if we want to deprecate a better API (WithOpts) in favor of a narrower API (WithPlatform), Can we reverse the decision here? Wdyt?

blueprismo: The naming is a bit confusing, this isn't deprecating `docker.PullImageWithOpts`, we're adding that and removing `docker.PullImageWithPlatform`.

The deprecation is `k3s.LoadImagesWithOpts` (Docker -> k3s import). It always did `ctr import --all-platforms` regardless of save opts, which broke multi-arch setups. `k3s.LoadImagesWithPlatform` fixes that by aligning save + import.

Open to keeping LoadImagesWithOpts if we fix the import behaviour instead of deprecating, but right now I am not aware of any other loading options into k3s. We could leave the function for future-proofing and update the comment accordingly, WDYT?

docs/features/common_functional_options.md
```diff
@@ -322,6 +322,61 @@ Using the `WithImageSubstitutors` options, you could define your own substitutio
 
 If you need to set the platform for a container, you can use `testcontainers.WithImagePlatform(platform string)`.
 
+##### PullDockerImageWithPlatform
```
mdelapenya: suggestion: I'm not sure if this is the right place to locate this Pull option, as they are container options 🤔 

A reader could understand it can be used as a container customizer, when it's not.

Because there is no Pull Images section in the docs, let's skip that, and open a ticket from this comment to remember adding a Pulling Images section in the docs. Wdyt?

mdelapenya (approved):
LGTM! 

I'll fix the docs in the upstream, thanks!

docs/features/common_functional_options.md
blueprismo: Done!

PR #3726, opened 2026-06-15
feat(redis): move examples and test into a separate Go module

ash2k:
## What does this PR do?

Move examples and test for the Redis module into a separate Go module.

## Why is it important?

This avoids having github.com/redis/go-redis/v9 as a dependency. I use a different Redis client and would like to avoid an extra dependency. Even if its coming via tests, it still shows up in go.sum.

## Related issues

-


mdelapenya:
@ash2k I'd like to understand better the need for the separation, as the Go tool chain will strip any test dependency from the final binary you create, so no testcontainers-go dependency should land into the production binaries. Isn't that the case here?

ash2k:
@mdelapenya This is optional. I thought it's nicer to not have anything extra as a dependency. Also, I think (but not 100% sure) all of the modules (even test dependencies of dependencies) participate in version resolution. Not a problem for me though. Feel free to close this.

On the other hand, this repo did separate things into separate modules to cut down on dependency bloat, right? So this change basically follows the same idea.

modules/redis/example/redis_test.go
```diff
@@ -1,4 +1,4 @@
-package redis_test
+package example
```
mdelapenya, line 1: This change seems wrong to me, as we will be removing the tests from the redis module.

I'm happy to have the example package (better to name it v9?) but the tests must live in the module itself.

ash2k: All the tests use the go-redis client. I moved them because I want to remove the dependency. If I move them back then... the PR will be empty?

p.s. sorry for the slow response.

PR #3732, opened 2026-06-19
fix(compose): keep build-only services when Up starts a subset

davireis:
## Problem

`Up(RunServices(...))` (any partial-service `Up`) rebuilt `project.Services` down to just the selected services before calling the compose service. When a selected service builds from a `service:` **additional build context** (`build.additional_contexts.<name>: service:<dep>`), the referenced build-only service was dropped too, so the build failed with:

```
service "X" declares unknown service "Y" as additional contexts
```

…even though `docker compose up X` on the *same* project builds the context and starts only `X`.

### Repro
A compose project where a runtime service builds from a `service:` additional context:

```yaml
services:
  build-dep:                      # build-only, never started
    build: { dockerfile: dep.Dockerfile }
    image: example/build-dep:latest
  app:
    build:
      dockerfile: app.Dockerfile
      additional_contexts:
        build-dep: service:build-dep   # COPY --from=build-dep ...
    image: example/app:latest
```

`compose.Up(ctx, RunServices("app"))` → `compose up: no such service: build-dep`.

## Fix

Use `project.WithSelectedServices(services, types.IgnoreDependencies)` instead of rebuilding `project.Services`. It **disables** (not deletes) the non-selected services, keeping them in the project's disabled set, so compose can re-enable build dependencies for the build phase (`addBuildDependencies` + `WithServicesEnabled`) while still only creating/starting the selected services. `IgnoreDependencies` preserves the existing behavior of starting only the explicitly requested services (not their runtime `depends_on`). The returned (new) project is reassigned to the up options so create and start operate on the same scoped project.

## Test

Adds `TestDockerComposeAPIWithRunServicesAndServiceBuildContext` with a runtime service that builds from a `service:` additional context. It fails on the previous code (`no such service`) and passes with the fix. The existing `TestDockerComposeAPIWithRunServices` still passes, and the up-all path is unchanged.

### DIFF
--- a/modules/compose/compose_api.go
+++ b/modules/compose/compose_api.go
@@ -7,7 +7,6 @@ import (
 	"io"
 	"os"
 	"path/filepath"
-	"sort"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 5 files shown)


mdelapenya:
@davireis could you address copilot review comments? 🙏

PR #3733, opened 2026-06-19
feat: add ability to chain WithConfigModifier, WithEndpointSettingsModifier and WithHostConfigModifier

LaurentGoderre:
## What does this PR do?

This allows to pass more than one host config or config modifier function which are run sequentially.

## Why is it important?

This is particularly helpful when using modules like the k3s which defines its own host config function so if you need to add you own, you need to duplicate the parent one. This allows to create a chain of modifier instead.

## Related issues

None

## How to test this PR

Unit tests added


options.go
mdelapenya: @LaurentGoderre could you piggyback your work to also update the behavior for this option? 🙏 

I'd also update the docs regarding the functional options, with something like: 

```
You can combine multiple calls of this option, although last wins.
```

Use your clanker to refine the message, but keep the release marker so that the release process use the proper version at release time.

LaurentGoderre: Done!

options_test.go
```diff
@@ -904,3 +908,149 @@ func TestWithProvider(t *testing.T) {
...
+	})
+}
+
+func TestWithEndpointSettingsModifier(t *testing.T) {
+	net1, err := network.New(t.Context())
+	require.NoError(t, err)
+
+	net2, err := network.New(t.Context())
+	require.NoError(t, err)
+
```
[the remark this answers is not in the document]
LaurentGoderre: Done

mdelapenya:
Done — added a `Chainable since :material-tag: main` marker plus a one-liner explaining the chaining/last-wins semantics to each of the three sections in `common_functional_options.md`, and appended the same marker to the corresponding entries in `common_functional_options_list.md`.

mdelapenya:
@LaurentGoderre the AP sent a PR against this branch in your fork. If you merge it, those changes will land here

LaurentGoderre:
@mdelapenya done!

PR #3735, opened 2026-06-23
docs(site): add missing languages and reorder docs navigation

mdelapenya:
Align header dropdown and landing page card-grid with the canonical language list: Java, Go, .NET, Node.js, Python, Rust, Ruby, PHP, Haskell, Clojure, Elixir, Scala, Native.

- Add PHP, Clojure, Elixir, Scala, Native to header dropdown and card-grid
- Fix ordering to match canonical list
- Add missing logo SVGs


mdelapenya:
Addressed review comments:
- **Restore Go URL** (r3458383138): fixed in e376142 — the template renders `href="{{ child.url }}"` unconditionally, so the self-entry needs its URL back. Restored `url` on Go entry (no `external: true`).
- **Add alt text** (r3458368621): fixed in 61c3b4f — added `alt=""` to all card-grid img tags. Decorative icons with adjacent visible text should use empty alt per WCAG.
- **Scala intrinsic dimensions** (r3458383141): already fixed in 81cbe63 — added `width="30" height="30"` to `scala.svg`.

PR #3745, opened 2026-06-30
fix(openfga): remove playground wait from Run and up image to v1.18.0

### LINKED ISSUE #3744: [Enhancement]: Bring openfga testcontainer module up to date
### Proposal

A detailed sister issue has been raised in the OpenFGA repository: [Link to the issue](https://github.com/openfga/openfga/issues/3185)

#### Summary:
The healthcheck policies in the code depends on the playground endpoint that was disabled by default a few versions ago. This issue proposes an enhancement that refactors the healthcheck to not depend on that endpoint

selvakannanjr:
## What does this PR do?

Removes the unconditional wait.ForHTTP("/playground") wait strategy from the default Run() function in the OpenFGA module, and updates the default image version referenced in examples and tests from v1.5.0 to v1.18.0.

The playground wait strategy is now opt-in: the ExampleRun_connectToPlayground example demonstrates how to enable it explicitly using testcontainers.WithAdditionalWaitStrategy(...) alongside the required environment variables (OPENFGA_PLAYGROUND_ENABLED=true, OPENFGA_PLAYGROUND_ADDR=0.0.0.0:3000).

## Why is it important?

The playground is not enabled by default in OpenFGA — it requires explicit configuration. Having Run() unconditionally wait for the playground HTTP endpoint meant that containers started without playground support would either time out or rely on an endpoint that was never going to be available. This made Run() unnecessarily slow and fragile for users who don't need the playground.

The image bump to v1.18.0 brings the examples in line with a current, supported release of OpenFGA.

## Related issues

- Closes #3744 

## How to test this PR

The example tests use the recent v1.18.0 image and shows you how to enable and connect to the Playground if needed. The openfga_test.go tests the new version with the default container options.


selvakannanjr:
Will update the docs next

modules/openfga/openfga.go
```diff
@@ -62,9 +61,6 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
 
 				return (strings.Contains(string(bs), "SERVING"))
 			}),
-			wait.ForHTTP("/playground").WithPort("3000/tcp").WithStatusCodeMatcher(func(status int) bool {
```
mdelapenya, line 65: Question: If a user passes a different version of the image, would it still need the playground? If so, isn't this a breaking change?

selvakannanjr: It wouldn't need the playground to be operational. However if the user needs to access the playground in the code, they need to specify the wait strategy. I have captured this in the examples_test.go and in the new commit (docs update)

selvakannanjr: In the scope of code that already uses Playground, this is a breaking change when used with older images. Should I mention this in the issue and redo the commit message?

selvakannanjr: The current Run() needs workarounds in form of custom wait strategy to work for newer images, which could be considered broken already. This fixes that and is a breaking change only if Playground was accessed in the testcontainer.

mdelapenya (approved):
LGTM, thanks!

selvakannanjr:
@mdelapenya will you take care of this hereafter and merge it when ready? do I have to do anything else?

mdelapenya:
Merged, thanks for your support to Testcontainers! 🙇

PR #3764, opened 2026-07-14
docs: fix typo in SkipIfProviderIsNotHealthy docstring

salandered:
## What does this PR do?

Fixes a typo. The last sentence contradicts what docstring says before it.

Alternative: "In this way tests that depend on Testcontainers will only run if the provider is provisioned correctly."
I like this version better because the current sentence simply repeats the function name. But I chose the simpler fix

## Why is it important?

The current version sounds confusing


mdelapenya (approved):
LGTM! Good catch!

PR #3765, opened 2026-07-15
chore(compose): use require instead of assert for service name checks

### LINKED ISSUE #2808: [Enhancement]: use testify's require instead of assert.

sunghyun1999:
## Summary
- Replace `assert.Contains` with `require.Contains` for service name validations in compose module tests (`compose_api_test.go`, `compose_test.go`)
- Replace `assert.ElementsMatch` with `require.ElementsMatch` in profile test
- Ensures consistency with other tests in the same files that already use `require.Contains` for the same pattern

## Rationale
These assertions validate expected service names after `compose.Up()`. If a service is missing, subsequent test logic (e.g., `ServiceContainer()` calls) would fail with misleading errors. Using `require` stops the test immediately, making the root cause clear.

Assertions inside loops (label checks, env var checks) are intentionally kept as `assert` for comprehensive feedback on all iterations.

## Related Issue
Closes #2808

## Test plan
- [ ] Existing compose module tests pass (`go test ./modules/compose/...`)
- [ ] No behavioral change — only fail-fast semantics improved

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/modules/compose/compose_api_test.go
+++ b/modules/compose/compose_api_test.go
@@ -60,7 +60,7 @@ func TestDockerComposeAPIStrategyForInvalidService(t *testing.T) {
 	serviceNames := compose.Services()
 
 	require.Len(t, serviceNames, 1)
-	assert.Contains(t, serviceNames, "api-nginx")
+	require.Contains(t, serviceNames, "api-nginx")
 }
 
 func TestDockerComposeAPIWithWaitLogStrategy(t *testing.T) {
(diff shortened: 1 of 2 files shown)


mdelapenya (approved):
LGTM, thanks!

PR #3762, opened 2026-07-10
feat: add new modules porting them from other TC libs

mdelapenya:
> 🤖 **Automated implementer agent** — *this comment was posted by the implementer bot from Docker Agentic Platform, not by a human developer*

## Summary

Adds 21 new modules to testcontainers-go covering databases, messaging, search, storage, and web infrastructure. Each module was implemented with a `Run()` function, typed container methods, options, tests, examples, and documentation.

## Modules added

| Module | Image | Key methods |
|---|---|---|
| `activemq` | `apache/activemq-classic:5.18` | `BrokerURL()`, `WebConsoleURL()` |
| `azure/sqledge` | `mcr.microsoft.com/azure-sql-edge:1.0.7` | `ConnectionString()` |
| `couchdb` | `couchdb:3` | `ConnectionString()` |
| `cratedb` | `crate:5.7` | `HTTPEndpoint()`, `PGConnectionString()` |
| `fakegcsserver` | `fsouza/fake-gcs-server:1.47` | `StorageURL()` |
| `firebird` | `ghcr.io/jacobalberty/firebird:v3.0` | `ConnectionString()` |
| `kurrentdb` | `kurrentplatform/kurrentdb:latest` | `ConnectionString()` |
| `mailpit` | `axllent/mailpit:v1.20` | `SMTPEndpoint()`, `HTTPURL()` |
| `mosquitto` | `eclipse-mosquitto:2` | `BrokerURL()` |
| `nginx` | `nginx:1.25` | `HTTPEndpoint()`, `HTTPSEndpoint()` |
| `orientdb` | `orientdb:3.2` | `ServerURL()`, `StudioURL()` |
| `papercutsmtp` | `changemakerstudiosus/papercut-smtp:latest` | `SMTPEndpoint()`, `HTTPURL()` |
| `presto` | `prestodb/presto:0.286` | `ConnectionString()` |
| `questdb` | `questdb/questdb:7.4` | `HTTPEndpoint()`, `PGEndpoint()`, `InfluxDBEndpoint()` |
| `ravendb` | `ravendb/ravendb:6.0-ubuntu-latest` | `ManagementURL()` |
| `s3mock` | `adobe/s3mock:3.9` | `EndpointURL()`, `HTTPSEndpointURL()` |
| `sftp` | `atmoz/sftp:latest` | `Address()` |
| `solr` | `solr:9` | `Address()`, `CollectionURL()` |
[...]

PR #3778, opened 2026-07-16
feat: add Presto module

mdelapenya:
> 🤖 **Automated implementer agent** — *this comment was posted by the implementer bot from Docker Agentic Platform, not by a human developer*

Adds the `presto` testcontainers-go module.

Part of the broader effort to add missing modules — split from #3762.


mdelapenya:
> 🤖 **Automated implementer agent** — *this comment was posted by the implementer bot from Docker Agentic Platform, not by a human developer*

Addressed both P2 findings in commit 461b470 on `feat/module-presto`:

**P2 #1 (line 57 — worker readiness):** Wait strategy now uses `wait.ForAll` chaining two HTTP checks: `/v1/info` (starting=false) **and** `/v1/cluster` (activeWorkers>0). The second check ensures at least one worker has registered before `Run` returns, eliminating the 'No nodes available to run query' startup race.

**P2 #2 (line 38 — IPv6):** `ConnectionString` now delegates to `c.PortEndpoint(ctx, httpPort, "http")` which correctly brackets IPv6 hosts (e.g. `http://[::1]:12345`) instead of the broken manual `fmt.Sprintf`.

PR #3816, opened 2026-07-16
fix(postgres): make BasicWaitStrategies safe with container reuse

aayushcodex17:
`postgres.BasicWaitStrategies()` gives a false-positive ready signal when containers are reused via `testcontainers.WithReuseByName`, causing tests to fail non-deterministically with `FATAL: the database system is not yet accepting connections`.

Docker container logs survive across restarts of the same container. When a reused container starts after an unclean shutdown, the log-based wait is satisfied immediately by readiness messages from the previous run -- before the current Postgres process has finished crash recovery -- so `Run()` returns a container that is not yet accepting connections. The race is most reliably triggered by stopping a container with a zero-grace timeout while a client session is open, which forces crash recovery on the next start.

This PR adds a live-state probe after the log gate that blocks until Postgres is genuinely accepting connections, regardless of what the logs say. The probe requires no credentials, so custom `WithUsername`, `WithPassword`, and `WithDatabase` options keep working. It is skipped gracefully on images that do not ship the necessary binary, preserving existing behaviour for those images. Fresh-start behaviour is unchanged -- the log gate still passes only once Postgres is up, so the probe adds no meaningful overhead on a first boot.

| Scenario | Before | After |
|---|---|---|
| Fresh container | Pass | Pass |
| Reused container, clean shutdown | Pass (usually) | Pass |
| Reused container, unclean shutdown | Flaky / fail | Pass |
| Image without probe binary | Pass | Pass (probe skipped) |
| Custom user / password / database | Pass | Pass |
| Custom PGPORT | Pass | Pass |

The added regression test starts a container with `WithReuseByName`, inserts enough data to generate WAL, kills the container with a zero-grace timeout while a session is open to guarantee crash recovery on the next start, then asserts that a single retry-free connection attempt succeeds immediately after `Run()` returns.

```bash
cd modules/postgres
go test -race -count=1 -run TestBasicWaitStrategies_reusedContainer .
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,13 @@
+FROM public.ecr.aws/d3j8x8q7/olympus-base-go:latest
+
+WORKDIR /app
+
+COPY . .
+
+# Download dependencies for the root module and the postgres module.
+# modules/postgres/go.mod has a replace directive pointing to the root (../../..),
+# so the root module must be present and downloaded first for offline compilation.
+RUN go mod download && \
+    cd modules/postgres && go mod download
+
+CMD ["/bin/bash"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,12 +2,20 @@ FROM public.ecr.aws/d3j8x8q7/olympus-base-go:latest
 
 WORKDIR /app
 
-COPY . .
+# Copy dependency manifests first to cache the download layer separately from source.
+COPY go.mod go.sum ./
+COPY modules/postgres/go.mod modules/postgres/go.sum modules/postgres/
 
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM public.ecr.aws/d3j8x8q7/olympus-base-go:latest
```
mdelapenya, line 1: question: do we need this Dockerfile and test.sh files?

aayushcodex17: No, Actually We dont need it. Was just trying to prepare a testing situation, pushed it all by mistake.

PR #3833, opened 2026-08-04
fix: update lowkey-vault image to 7.3.48 for azsecrets SDK v1.5.0 compatibility

mdelapenya:
Fix for the lowkeyvault test failures in #3806.

- Update lowkey-vault image from 7.0.9 to 7.3.48-ubi10-minimal
- Switch pkcs12 import from `golang.org/x/crypto/pkcs12` to `software.sslmate.com/src/go-pkcs12` (already a dependency) to support SHA-256 based PKCS#12 encoding used by newer lowkey-vault versions

All lowkeyvault tests verified passing locally.

### DIFF
--- a/docs/modules/azure.md
+++ b/docs/modules/azure.md
@@ -420,7 +420,7 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
 #### Image
 
 Use the second argument in the `Run` function to set a valid Docker image.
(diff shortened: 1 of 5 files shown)


mdelapenya:
Addressed the docs feedback: updated both lowkey-vault image references in `docs/modules/azure.md` (lines 423 and 446) from `7.0.9-ubi10-minimal` to `7.3.48-ubi10-minimal`.

PR #3837, opened 2026-08-05
fix: escape the container name in the Docker name filter

jcmfernandes:
## What does this PR do?

Quotes the `name` filter of the Docker container list API is a regular expression, so metacharacters in a container name are interpreted instead of matched literally.

## Why is it important?

A lookup for "cache.1" also matches "cacheX1", and `ReuseOrCreateContainer` can reuse the wrong container.

Since container names may only include `^[a-zA-Z0-9][a-zA-Z0-9_.-]+$` char `.` (dot) is the only metacharacter reachable through a legal container name.

## Related issues

Not related to, by raised in #3051 by an [LLM code review](https://github.com/testcontainers/testcontainers-go/pull/3051#pullrequestreview-4856522740).

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -1382,7 +1382,7 @@ func (p *DockerProvider) findContainerByName(ctx context.Context, name string) (
 	// Note that, 'name' filter will use regex to find the containers
 	containers, err := p.client.ContainerList(ctx, client.ContainerListOptions{
 		All:     true,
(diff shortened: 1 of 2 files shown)


jcmfernandes:
Rebased.

mdelapenya (approved):
LGTM, thanks

PR #3842, opened 2026-08-08
chore: skip test jobs when module requires a newer Go version

mdelapenya:
Adds a Go version compatibility check to `ci-test-go.yml` that runs after "Set up Go". It compares the module's `go.mod` minimum Go version against the installed version and skips build/test/sonar steps when incompatible, emitting a warning annotation instead of failing.

**Why**: dependency bumps (e.g. weaviate v1.38.0 requiring `go 1.26`) cause hard failures on the Go 1.25.x matrix entry with `go: go.mod requires go >= 1.26`. This is expected behavior — the module simply can't be built with an older Go, so the CI should skip gracefully.

Unblocks #3829 and any future dependency bump that raises the minimum Go version for a module.

### DIFF
--- a/.github/workflows/ci-test-go.yml
+++ b/.github/workflows/ci-test-go.yml
@@ -72,11 +72,29 @@ jobs:
             cache-dependency-path: '${{ inputs.project-directory }}/go.sum'
         id: go
 
+      - name: Check Go version compatibility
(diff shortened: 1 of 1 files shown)


mdelapenya:
Closing as superseded by #3829

Issue #203, opened 2020-06-01
Way to define HostIP (interface)


dimuls:
Need a way to define on which host interface will be used. Now `HostIP=0.0.0.0`. For testing purposes we do not want to expose app on public interfaces. Need a way to define `HostIP=127.0.0.1`.

gianarb:
Hello! Do you mind to detail a bit more what you are trying to do? Is this a docker daemon configuration or what?

dimuls:
Haven't used `testcontainers-go` for months. As far as i remember it is docker container configuration. We were need to bind testing containers' ports on `127.0.0.1` not on `0.0.0.0`.

For example, we have web server which uses port 80. To test it we need to run docker with this web server and somehow expose this port 80. But we do not want to expose it on public net interface (using `0.0.0.0`) but only `127.0.0.1`.

If we draw an analogy with `docker run` command's `-p` flag: we need `-p 127.0.0.1:80:80` not just `80:80`.

So it will be nice if using `testcontainers-go` one may able to configure interface on which exposed ports will be binded.

mdelapenya:
Hi @dimuls I found this issue interesting and kind of related to this discussion: https://github.com/testcontainers/testcontainers-go/discussions/590

With the 0.18.0 release, you are able to define 3 different modifiers at the ContainerRequest layer:
- Docker config: https://github.com/moby/moby/blob/master/api/types/container/config.go#L70
- Docker HostConfig: https://github.com/moby/moby/blob/master/api/types/container/hostconfig.go#L378
- Docker EndpointSettings: https://github.com/moby/moby/blob/master/api/types/network/network.go#L49

If your setup involves updating any of those types, then you'll be able to achieve with the modifiers: https://golang.testcontainers.org/features/creating_container/#advanced-settings

mdelapenya:
I'm going to close this issue as stale, although please reopen it if you need anything from us 🙏 

Thanks!

PR #3844, opened 2026-08-10
feat: add WithHostIP option to bind exposed ports to a host IP

shyim:
## What does this PR do?

> Disclaimer AI used

Adds a first-class `WithHostIP` option (and a matching `HostIP` field on `ContainerRequest`) to control the IP address to which a container's exposed ports are bound on the host.

By default, testcontainers binds exposed ports to all host interfaces (`0.0.0.0`). With this PR:

```go
ctr, err := testcontainers.Run(ctx, "nginx:alpine",
    testcontainers.WithExposedPorts("80/tcp"),
    testcontainers.WithHostIP("127.0.0.1"),
)
```

restricts the binding to `127.0.0.1` only (the `WithHostConfigModifier` dance is no longer needed).

Implementation notes:

- `WithHostIP(ip string)` validates the address with `netip.ParseAddr` at request-customization time and stores it as `req.HostIP`.
- `preCreateContainerHook` applies the IP to **all** port bindings *after* `mergePortBindings`, so it covers both the default ephemeral bindings and per-port bindings set via `WithHostConfigModifier` (the `HostIP` is overridden while the `HostPort` is preserved).
- Direct `ContainerRequest` users get the same validation, returning an error for invalid IPs.

## Why is it important?

Exposing containers on `0.0.0.0` is a security concern on shared/multi-tenant machines: any process on the host (or, depending on the network setup, other hosts) can reach the container. Binding to loopback was previously only possible by reaching into the internal `HostConfig` via a modifier, with no validation and no documentation.

My usecase is running agents on a remote host, and while testcontainer runs my postgres / and so on are publicly reachable in the Internet, without proper Firewall

## Related issues

- Relates #203

```bash
# unit tests
go test -run 'TestWithHostIP|TestPreCreateContainerHookAppliesHostIP' .

# integration test (requires Docker): verifies the container actually
# binds to 127.0.0.1 via Inspect().NetworkSettings.Ports
go test -run 'TestContainerWithHostIP' -v .
```


lifecycle.go
```diff
@@ -544,6 +545,22 @@ func (p *DockerProvider) preCreateContainerHook(ctx context.Context, req Contain
 
 	dockerInput.ExposedPorts = exposedPortSet
 	hostConfig.PortBindings = mergePortBindings(hostConfig.PortBindings, exposedPortSet)
+
+	// Bind all exposed ports to the requested host IP address.
+	// The binding happens after the port bindings are merged, so it applies to both
+	// the default ephemeral bindings and the ones set by the HostConfigModifier.
+	if req.HostIP != "" {
+		ip, err := netip.ParseAddr(req.HostIP)
```
mdelapenya, line 553: suggestion: can we move this validation to the functional option? I think we can then add the field to the container request as private, so that nobody can actually populate it apart from the option. Thoughts?

PR #3877, opened 2026-09-04
fix: dial the sshd container on the Docker daemon host

timofurrer:
## What does this PR do?

Host port access (`WithHostPortAccess`) dialed its sshd sidecar at `localhost:<mapped port>`. It now resolves the host through `Container.Host()`, so the dial goes to the Docker daemon host like every other exposed port.

## Why is it important?

With a remote daemon, for example a `docker:dind` service in GitLab CI reached via `DOCKER_HOST=tcp://docker:2376`, the mapped port is published on the daemon host and not on the test process's loopback. Host port access then always fails with:

```
new port forwarder: ssh dial: dial tcp [::1]:32837: connect: connection refused
```

`TESTCONTAINERS_HOST_OVERRIDE` does not help because this code path never consulted it. testcontainers-java already uses `container.getHost()` for the same connection.

## Related issues

None found.

## How to test this PR

`TestExposeHostPorts` still passes against a local daemon. For the remote case, run `docker:dind` on a user-defined network with the alias `docker`, then run a program that uses `WithHostPortAccess` in a sibling container on that network with `DOCKER_HOST=tcp://docker:2376`. Before this change the sshd dial fails as above; after it the container reaches `host.testcontainers.internal:<port>`.

### DIFF
--- a/port_forwarding.go
+++ b/port_forwarding.go
@@ -205,7 +205,7 @@ func newSshdContainer(ctx context.Context, opts ...ContainerCustomizer) (*sshdCo
 // It's an internal type that extends the DockerContainer type, to add the SSH tunnelling capabilities.
 type sshdContainer struct {
 	Container
-	port           string
+	addr           string
 	sshConfig      *ssh.ClientConfig
 	portForwarders []*portForwarder
 }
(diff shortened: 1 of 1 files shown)


mdelapenya (approved):
LGTM, thanks! Waiting for the CI to pass

timofurrer:
@mdelapenya awesome, thanks for the speedy review, I appreciate it. 

Will this automatically be released? If not, can you kick one off? 🙏 This would avoid a nasty "replace" directive on my end :)

mdelapenya (approved):
LGTM

mdelapenya:
> @mdelapenya awesome, thanks for the speedy review, I appreciate it.
> 
> Will this automatically be released? If not, can you kick one off? 🙏 This would avoid a nasty "replace" directive on my end :)

Hi @timofurrer, I'm in a conference until today, with short periods of time for OSS work, but I'll try to find time the next week or the following one 🙏

PR #3881, opened 2026-09-07
fix(rabbitmq): pass the node config from memory instead of a shared temp file

### LINKED ISSUE #3880: [Bug]: RabbitMQ testcontainer leaves /tmp/rabbitmq-testcontainers.conf after destruction
### Testcontainers version

0.43.0

Yes

### Host OS

Linux

### Host arch

x86

### Go version

1.26

### Docker version

```shell
 Version:           29.5.0
 API version:       1.54
 Go version:        go1.26.3
 Git commit:        98f1464
 Built:             Thu May 14 14:40:04 2026
 Context:           default

 Engine:
  Version:          29.5.0
  API version:      1.54 (minimum version 1.40)
  Go version:       go1.26.3
  Git commit:       ff8d90a
  Built:            Thu May 14 14:40:04 2026
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v2.2.3
  GitCommit:        77c84241c7cbdd9b4eca2591793e3d4f4317c590
 runc:
  Version:          1.3.5
  GitCommit:        v1.3.5-0-g488fc13e
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### Docker info

```shell
 Version:    29.5.0
 Context:    default
 Debug Mode: false
 Plugins:
    Version:  v0.34.0
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
    Version:  v5.1.3
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 41
  Running: 3
  Paused: 0
  Stopped: 38
 Images: 37
 Server Version: 29.5.0
 Storage Driver: overlayfs
  driver-type: io.containerd.snapshotter.v1
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
 CDI spec directories:
  /etc/cdi
  /var/run/cdi
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc runsc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 77c84241c7cbdd9b4eca2591793e3d4f4317c590
 runc version: v1.3.5-0-g488fc13e
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 6.8.0-139-generic
 Operating System: Ubuntu 24.04.4 LTS
 OSType: linux
 Architecture: x86_64
 CPUs: 12
 Total Memory: 15.36GiB
 Name: pancakepc
 ID: 8ca2eba9-2ebc-4719-b04c-95c5eaae24bf
 Debug Mode: false
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
  ::1/128
 Live Restore Enabled: false
 Firewall Backend: iptables
  EnableUserlandProxy: true
  UserlandProxyPath: /usr/bin/docker-proxy
```

### What happened?

When the RabbitMQ testcontainer is used under either a root or non-root user for a test, and a subsequent test runs the container under the opposite user, a permission error occurs.

This happens because `/tmp/rabbitmq-testcontainers.conf` is not removed after the previous test completes. The file retains the ownership and permissions of the user that created it, causing a permission conflict when the next test runs under a different user.

### Steps to Reproduce

1. Run a test using the RabbitMQ testcontainer as a root user.
2. Let the test complete and destroy the container.
3. Run another test using the RabbitMQ testcontainer as a non-root user.
4. Observe the resulting permission error.

The same issue occurs when switching from a non-root user to a root user.

### Expected Behavior

The temporary `/tmp/rabbitmq-testcontainers.conf` file should be removed after the testcontainer is destroyed, so that subsequent tests can run under a different user without encountering permission issues.

### Actual Behavior

The `/tmp/rabbitmq-testcontainers.conf` file is not removed after the testcontainer is destroyed.
Because the file retains the ownership and permissions of the user from the previous test, running the RabbitMQ testcontainer under a different user in a subsequent test results in a permission error.

### Relevant log output

```shell
failed to start container: open /tmp/rabbitmq-testcontainers.conf: permission denied
exit status 1
```

### Additional information

_No response_

manduinca:
## What does this PR do?

The RabbitMQ module no longer writes its rendered node configuration to a file on the host. `ContainerFile` already takes a `Reader` instead of a `HostFilePath`, and the file is copied into the container in the `PostCreates` hook either way, so the config can be handed over from memory and the temporary file disappears entirely.

## Why is it important?

The path was fixed: `filepath.Join(os.TempDir(), "rabbitmq-testcontainers.conf")`, written with mode `0600` and never removed. On a shared machine that path belongs to whoever created it first, so a run under a different user fails with `open /tmp/rabbitmq-testcontainers.conf: permission denied` before the container ever starts. Removing the file after each run would fix the symptom; not writing it at all also removes the leftover, the fixed name, and the two failure modes that come with writing to a shared directory.

## Related issues

- Closes #3880

## How to test this PR

`go test ./...` in `modules/rabbitmq` (needs a Docker daemon). The existing suite covers this: `ExampleRun_withSSL` and the plugin examples assert the config actually lands at `/etc/rabbitmq/rabbitmq-testcontainers.conf` inside the container, which is the path that matters and is unchanged.

To see the original bug, run any test in the module as root, then again as a normal user.

### DIFF
--- a/modules/rabbitmq/rabbitmq.go
+++ b/modules/rabbitmq/rabbitmq.go
@@ -5,8 +5,6 @@ import (
 	"context"
 	_ "embed"
 	"fmt"
-	"os"
-	"path/filepath"
 	"text/template"
 	"time"
 
@@ -93,12 +91,6 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
 		return nil, err
 	}
 
-	tmpConfigFile := filepath.Join(os.TempDir(), "rabbitmq-testcontainers.conf")
-	err = os.WriteFile(tmpConfigFile, nodeConfig, 0o600)
-	if err != nil {
-		return nil, err
-	}
-
 	moduleOpts := []testcontainers.ContainerCustomizer{
 		testcontainers.WithEnv(map[string]string{
 			"RABBITMQ_DEFAULT_USER": settings.AdminUsername,
@@ -111,7 +103,7 @@ func Run(ctx context.Context, img string, opts ...testcontainers.ContainerCustom
 			DefaultHTTPPort,
 		),
 		testcontainers.WithWaitStrategy(wait.ForLog(".*Server startup complete.*").AsRegexp().WithStartupTimeout(60 * time.Second)),
-		withConfig(tmpConfigFile),
+		withConfig(nodeConfig),
 	}
 
 	if settings.SSLSettings != nil {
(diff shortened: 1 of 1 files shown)


mdelapenya (approved):
LGTM, thanks!

PR #3264, opened 2025-08-19
fix(postgres): snapshot restore

ascheman:
Label: Bug

## What does this PR do?

It fixes #3263 by providing two changes
- Close any existing connection on the snapshot DB
- Drop the database only if it exists

## Why is it important?

We had very flaky tests without this change.

## Related issues

Seldomly, we still run into #3233 and would be happy about a fix as well.

- Closes #3263
- Relates #3233

## How to test this PR

As the error(s) only occur in some unknown circumstances it is very hard to enforce their occurrence and also a test situation where they fire.

Issue #3263, opened 2025-08-19
[Bug]: Multiple errors by PostgreSQL snapshot restore


ascheman:
### Testcontainers version

v0.38.0

### Using the latest Testcontainers version?

Yes

### Host OS

Linux, macOS

### Host arch

x86_64, arm64

### Go version

1.24.4

### Docker version

```shell
Client:
 Version:           28.3.2
 API version:       1.51
 Go version:        go1.24.5
 Git commit:        578ccf6
 Built:             Wed Jul  9 16:12:57 2025
 OS/Arch:           darwin/arm64
 Context:           desktop-linux

Server: Docker Desktop 4.43.2 (199162)
 Engine:
  Version:          28.3.2
  API version:      1.51 (minimum version 1.24)
  Go version:       go1.24.5
  Git commit:       e77ff99
  Built:            Wed Jul  9 16:13:56 2025
  OS/Arch:          linux/arm64
  Experimental:     false
 containerd:
  Version:          1.7.27
  GitCommit:        05044ec0a9a75232cad458027ca83437aae3f4da
 runc:
  Version:          1.2.5
  GitCommit:        v1.2.5-0-g59923ef
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### Docker info

```shell
Client:
 Version:    28.3.2
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
  ai: Docker AI Agent - Ask Gordon (Docker Inc.)
    Version:  v1.9.11
    Path:     $HOME/.docker/cli-plugins/docker-ai
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.26.1-desktop.1
    Path:     $HOME/.docker/cli-plugins/docker-buildx
  cloud: Docker Cloud (Docker Inc.)
    Version:  v0.4.18
    Path:     $HOME/.docker/cli-plugins/docker-cloud
  compose: Docker Compose (Docker Inc.)
    Version:  v2.39.1-desktop.1
    Path:     $HOME/.docker/cli-plugins/docker-compose
  debug: Get a shell into any image or container (Docker Inc.)
    Version:  0.0.42
    Path:     $HOME/.docker/cli-plugins/docker-debug
  desktop: Docker Desktop commands (Docker Inc.)
    Version:  v0.2.0
    Path:     $HOME/.docker/cli-plugins/docker-desktop
  extension: Manages Docker extensions (Docker Inc.)
    Version:  v0.2.29
[...]

Issue #3233, opened 2025-07-17
[Bug]: Error on postgres restore


aBMania:
### Testcontainers version

0.38.0

### Using the latest Testcontainers version?

Yes

### Host OS

Linux

### Host arch

x86

### Go version

1.24.0

### Docker version

```shell
Docker version 28.1.1, build 4eba377
```

### Docker info

```shell
Client: Docker Engine - Community
 Version:    28.1.1
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.23.0
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.35.1
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 16
  Running: 5
  Paused: 0
  Stopped: 11
 Images: 220
 Server Version: 28.1.1
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 Swarm: inactive
 Runtimes: runc io.containerd.runc.v2
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 05044ec0a9a75232cad458027ca83437aae3f4da
 runc version: v1.2.5-0-g59923ef
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 6.8.0-60-generic
 Operating System: Ubuntu 22.04.5 LTS
 OSType: linux
 Architecture: x86_64
 CPUs: 8
 Total Memory: 38.92GiB
 Name: mania-exaion
 ID: 4f2e98b2-ae93-4512-85b3-a2276c32d22e
 Docker Root Dir: /var/lib/docker
 Debug Mode: false
 Experimental: false
 Insecure Registries:
  ::1/128
  127.0.0.0/8
 Live Restore Enabled: false
```

### What happened?

When calling `PostgresContainer`'s `Restore` method, i expect never receive an error. But sometimes (randomly), i get a `database "[DBNAME]" already exists` error

[...]

PR #3907, opened 2026-09-20
fix(postgres): terminate connections to target database before restore

cipherprofessor:
Relates #3233

## Problem

`Restore()` sometimes fails with `database "X" already exists` when `DROP DATABASE "X" WITH (FORCE)` is immediately followed by `CREATE DATABASE "X" ...`, reported in #3233.

## Context — why this isn't just re-litigating #3264

A related, earlier fix (#3264) added an explicit `pg_terminate_backend` call targeting the template/snapshot database before the `CREATE ... WITH TEMPLATE` step, fixing a different bug (#3263). That call was load-bearing because `CREATE DATABASE ... TEMPLATE` has no `FORCE`-equivalent at all.

During review of #3264, @stevenh asked whether `WITH (FORCE)` already handles connection termination on its own. It does — per [Postgres's own docs](https://www.postgresql.org/docs/current/sql-dropdatabase.html), `FORCE` terminates connections via the same `pg_terminate_backend` mechanism, against the same target, with the same documented exceptions (prepared transactions, active logical replication slots/subscriptions). So an explicit terminate call against the *same* database `DROP` is about to drop doesn't reach any connection `FORCE` couldn't already reach on its own — I want to be upfront about that rather than imply otherwise.

**Its value is timing, not capability:** `pg_terminate_backend` is fire-and-forget — it signals termination but doesn't block until the backend actually exits. Issuing it here, as its own round trip before `DROP`, buys those backends additional wall-clock time to fully exit before `FORCE`'s own connection check runs immediately afterward. This empirically narrows, but does not eliminate, the race — consistent with the maintainers' own note on #3264 that these errors "only occur in some unknown circumstances" and are "very hard to enforce... in a test."

## Testing

Given the above, I didn't try to force a red/green test of the actual race (matching the precedent already set on #3264). Instead:

- Extracted the SQL commands `Restore()` issues into a new unexported `restoreCommands()` method (byte-for-byte identical to the original three, plus the one new line), so the exact commands and their ordering can be asserted directly without a live database. Added `restore_test.go` (`package postgres`, matching this monorepo's existing white-box-test convention used elsewhere, e.g. `modules/redis/options_test.go`) verifying: exactly 4 commands are produced, the new terminate call targets the target database specifically, and it comes before the `DROP`.
- Ran the full existing Docker-backed `Snapshot`/`Restore` integration suite against a real Postgres container, repeated 5x (`-count=5`), all green.

`go build`, `go vet`, `gofmt`, and the full `postgres` module test suite are all clean.


cipherprofessor:
Thanks for the catch @coderabbitai — verified and fixed in 9d5c8edc.

`snapshotName`/`c.dbName` were interpolated directly into the `datname = '%s'` literal, so a name containing a single quote would break out of it. Added `quoteLiteral()`, which mirrors Postgres's own `quote_literal()`: doubles embedded quotes, and if the value contains a backslash wraps it in an `E'...'` escape string with backslashes doubled too, so it's correct regardless of the server's `standard_conforming_strings` setting (a plain doubled-quote-only escape is unsafe when that's off, since a trailing backslash would otherwise escape the closing quote).

I intentionally didn't reach for `lib/pq`'s `QuoteLiteral` (equivalent function, already a module dependency) — it's currently only pulled in via a test file, and importing it from `postgres.go` itself would force `lib/pq`'s driver-registration side effect onto every consumer of this module regardless of the `SQLDriverName` they configure (e.g. `pgx`-only users).

Left the double-quoted identifiers in the `DROP`/`CREATE DATABASE` statements as-is, per your note — that's a pre-existing pattern (also present in `Snapshot()`, which this PR doesn't touch) and out of scope here; happy to open a separate issue for it if useful.

Added two regression tests (apostrophe, backslash) to `restore_test.go`, confirmed red against the prior code, green after the fix. `go build`/`go vet`/`gofmt`/full module test suite all clean.

modules/postgres/postgres.go
```diff
@@ -300,15 +300,47 @@ func (c *PostgresContainer) Restore(ctx context.Context, opts ...SnapshotOption)
...
+
+// quoteLiteral returns s as a safely-quoted SQL string literal (as opposed
+// to a double-quoted identifier, which restoreCommands quotes separately),
+// including the surrounding quotes. It mirrors Postgres's own
+// quote_literal(): single quotes are doubled, and if s contains a
+// backslash the literal is escaped with the E prefix and backslashes are
+// doubled too, so the value round-trips correctly whether or not the
+// server has standard_conforming_strings enabled -- see
+// https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-STRINGS-ESCAPE.
+func quoteLiteral(s string) string {
```
mdelapenya, line 338: suggestion: even though the added tests already execute this function, I'd like to also include full coverage for it with dedicated unit tests. Could you add them? 🙏

cipherprofessor: Good call, done in 2156155 — added `TestQuoteLiteral` with dedicated table-driven coverage (empty string, plain identifier, single/multiple/leading-and-trailing apostrophes, single/multiple backslashes, and both combined orderings). Confirmed the cases are load-bearing, not tautological, by temporarily reverting `quoteLiteral` to a naive no-escaping implementation and checking that 7 of the 9 cases fail as expected before restoring it.

